PR #33, opened 2019-07-25
Intrinsic function BigIntAsInt

msoeken:
Based on a discussion with @cgranade in https://github.com/microsoft/QuantumLibraries/pull/140, we were wondering whether it can be useful to have a function `BigIntAsInt` in the runtime. There is a use case in the PR, where I create a sequence from a range of `BigInteger`s specified given a `from` and `to` parameter. The difference is a `BigInteger`, which should be used as the length of a `BigInteger[]`, which must be an `Int`.

However, there are some open issues.

1. The conversion may fail, if the value held by the `BigInteger` is too large. My suggestion is to return a tuple `(Int, Bool)` instead of just an `Int`. The Boolean return value indicates whether the conversion was successful.
1. Further, I have not found a place where these conversion methods in the runtime are tested.

### DIFF
--- a/src/Simulation/Intrinsic/Convert/Convert.qs
+++ b/src/Simulation/Intrinsic/Convert/Convert.qs
@@ -37,7 +37,7 @@ namespace Microsoft.Quantum.Convert {
     /// which is true, if and only if the conversion was possible.
     /// # Remarks
     /// See [C# BigInteger constructor](https://docs.microsoft.com/dotnet/api/system.numerics.biginteger.-ctor?view=netframework-4.7.2#System_Numerics_BigInteger__ctor_System_Int64_) for more details.
-    function BigIntAsInt(a : BigInt) : (Int, Bool) {
+    function MaybeBigIntAsInt(a : BigInt) : (Int, Bool) {
         body intrinsic;
     }
     
--- a/src/Simulation/Intrinsic/Convert/Convert.qs
+++ b/src/Simulation/Intrinsic/Convert/Convert.qs
@@ -32,6 +32,17 @@ namespace Microsoft.Quantum.Convert {
     
     
     /// # Summary
+    /// Converts a given big integer to an equivalent integer, if possible.
+    /// The function returns a pair of the resulting integer and a Boolean flag
+    /// which is true, if and only if the conversion was possible.
+    /// # Remarks
+    /// See [C# BigInteger constructor](https://docs.microsoft.com/dotnet/api/system.numerics.biginteger.-ctor?view=netframework-4.7.2#System_Numerics_BigInteger__ctor_System_Int64_) for more details.
+    function BigIntAsInt(a : BigInt) : (Int, Bool) {
+        body intrinsic;
+    }
+    
+    
+    /// # Summary
     /// Converts a given big integer to an array of Booleans.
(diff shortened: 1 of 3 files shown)


src/Simulation/Intrinsic/Convert/Convert.qs
```diff
@@ -31,6 +31,17 @@ namespace Microsoft.Quantum.Convert {
     }
     
     
+    /// # Summary
+    /// Converts a given big integer to an equivalent integer, if possible.
+    /// The function returns a pair of the resulting integer and a Boolean flag
+    /// which is true, if and only if the conversion was possible.
+    /// # Remarks
+    /// See [C# BigInteger constructor](https://docs.microsoft.com/dotnet/api/system.numerics.biginteger.-ctor?view=netframework-4.7.2#System_Numerics_BigInteger__ctor_System_Int64_) for more details.
+    function BigIntAsInt(a : BigInt) : (Int, Bool) {
```
cgranade: I really like this as a signature, as per our discussion on the other thread. I think that this signature is distinct enough from other `XAsY` functions, however, that we should call that out in the name. Perhaps `MaybeBigIntAsInt`, thinking of `(Int, Bool)` as a kind of "Int option" or "Int?" pattern? If so, I'm happy to add in M.Q.Convert new functions `Definitely<'T>(maybeValue : ('T, Bool)) : 'T` and `BigIntAsInt(a : BigInt) : Int` that `fail` if the conversion fails.

msoeken:
Thanks for your comments @anpaz-msft and @cgranade. I think that `MaybeBigIntAsInt` is a good name to better distinguish the different signature. But how would the `BigIntAsInt` function fail? With an assertion? But that code would anyway be in microsoft/QuantumLibraries.

msoeken:
@anpaz-msft I have added some test code. However, when trying to convert the smallest possible int value, I run into a problem with the Q# compiler. I am currently looking into that one in qsharp-compiler.

@cgranade I have updated the code with the new suggestion for the name.

cgranade:
> Thanks for your comments @anpaz-msft and @cgranade. I think that `MaybeBigIntAsInt` is a good name to better distinguish the different signature. But how would the `BigIntAsInt` function fail? With an assertion? But that code would anyway be in microsoft/QuantumLibraries.

@msoeken: My thought was that `BigIntAsInt` could call `Definitely(MaybeBigIntAsInt(value))`, and that `Definitely` could include a `fail` statement. I definitely agree, though, that would work better in libraries. I'll make a PR for that shortly.

anpaz-msft:
once approved by @cgranade, I'll merge this. Thanks!

msoeken:
Thanks for your comments, @anpaz-msft and @cgranade. Meanwhile, the problem with the test case due to the handling of the smallest possible `Int` value has been fixed in qsharp-compiler (see https://github.com/microsoft/qsharp-compiler/pull/87). Now I am not sure about the dependencies among the repositories. Can I already fix the test case or do I have to wait until the qsharp-compiler reference was updated in this repository to include the PR.

anpaz-msft:
@msoeken, we're planning to build `-beta` packages of the compiler soon. For the time being, I suggest just merging these changes so they can be picked up.

PR #39, opened 2019-08-19
Increment version number and add prerelease feed to README.

cgranade:
This PR updates the version number used to build the Q# runtime in CI and updates the README file to point to the public prerelease feed.

### DIFF
--- a/NuGet.Config
+++ b/NuGet.Config
@@ -1,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright (c) Microsoft Corporation. All rights reserved.
+     Licensed under the MIT License. -->
+
+<!--
+    This configuration tells the .NET Core SDK to build the Q# runtime using the
+    Quantum Development Kit prerelease feed in addition to any other currently
+    configured NuGet feeds.
+    This allows us to use prerelease packages from other parts of the Quantum
+    Development Kit in between releases.
+
+    See https://docs.microsoft.com/en-us/nuget/consume-packages/configuring-nuget-behavior#how-settings-are-applied
+    for more information about how NuGet and the .NET Core SDK find package
+    sources from NuGet.Config files.
+-->
+<configuration>
+  <packageSources>
+    <add key="qdk-alpha" value="https://pkgs.dev.azure.com/ms-quantum-public/Microsoft Quantum (public)/_packaging/alpha/nuget/v3/index.json" protocolVersion="3" />
+  </packageSources>
+</configuration>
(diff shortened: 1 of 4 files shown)


cgranade:
Sounds good, I'll get on that now, then.

build/ci.yml
```diff
@@ -4,7 +4,7 @@ trigger:
 
 variables:
   Build.Major: 0
-  Build.Minor: 0
+  Build.Minor: 9
```
anpaz-msft: I wonder how we are going to keep these in-sync.

README.md
```diff
@@ -107,3 +107,32 @@ For more information see the [Code of Conduct FAQ](https://opensource.microsoft.
 contact [opencode@microsoft.com](mailto:opencode@microsoft.com) with any additional questions or comments.
 
 For more details, please see [CONTRIBUTING.md](./CONTRIBUTING.md), or the [contribution guide](https://docs.microsoft.com/quantum/contributing/).
+
+## [Optional] Using Prerelease Versions ##
```
anpaz-msft: In hindsight, I'm wondering if this adds more noise than being helpful.
Reading it, it appears people have to do something to enable the prerelease feed and if they don't they will be using released bytes, when the matter of fact is that they are already using it if building from master. 

Maybe instead of instructions, we just need to give notice that when building from `master` you will be using pre-release packages picked up from our public pre-release feed configured on the nuget.config in the repo?

NuGet.Config
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright (c) Microsoft Corporation. All rights reserved.
+     Licensed under the MIT License. -->
+
+<!--
+    This configuration tells the .NET Core SDK to build the Q# runtime using the
```
anpaz-msft: This really just tells .NET Core how to find prereleased packages. 
Each project then tells .NET Core which version of the package they want to use, and most of the projects in `master` use prereleased versions, which is why it's needed. I would tailor the comment to reflect this.

README.md
bettinaheim: That's not even true - the config only makes it possible to reference packages from the prerelease feed, but looking at the PR, I think all packages still point to the released ones. So what it really should say is that the prerelease feed is added for all code in this repo and you can use prerelease packages if you want to.

build/ci.yml
bettinaheim: Came here to look what the consensus on this one was - I don't think they need to be in sync, but intended to keep it the same as other QDK repos nonetheless. What do we want to do here?

PR #52, opened 2019-09-10
Add type annotation to empty array

msoeken:
Without this type annotation the code causes a runtime exception on MacOS 10.14.6 with .NET Command Line Tools (2.2.401).

Before this fix:

```
Unhandled Exception: System.ArgumentException: Object of type 'System.Object[]' cannot be converted to type 'System.Collections.Generic.IEnumerable`1[System.String]'.
   at System.RuntimeType.TryChangeType(Object value, Binder binder, CultureInfo culture, Boolean needsSpecialCast)
   at System.Reflection.MethodBase.CheckArguments(Object[] parameters, Binder binder, BindingFlags invokeAttr, CultureInfo culture, Signature sig)
   at System.Reflection.RuntimeConstructorInfo.Invoke(BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   at CommandLine.Core.InstanceBuilder.BuildImmutable[T](Type typeInfo, Maybe`1 factory, IEnumerable`1 specProps, IEnumerable`1 specPropsWithValue, List`1 setPropertyErrors)
   at CommandLine.Core.InstanceBuilder.<>c__DisplayClass0_0`1.<Build>b__5()
   at CommandLine.Parser.ParseArguments[T](IEnumerable`1 args)
   at Microsoft.Quantum.QsCompiler.CsharpGeneration.Program.main(String[] args) in Program.fs:line 64
```


msoeken:
It only happens, when the references parameter is left empty. In most usages, it's not empty, so the problem is not triggered.

PR #55, opened 2019-09-15
Fixing empty array slices

### LINKED ISSUE #54: Incorrect array slicing
In https://github.com/microsoft/qsharp-runtime/blob/master/src/Simulation/Core/QArray.cs
in the slicing method, the line 
"long rangeCount = 1 + (range.End - range.Start) / range.Step;"
is not correct; it mistakenly computes that e.g. the range 1 .. 2 .. 0 has one element, which subsequently causes an ArgumentOutOfRangeException exception.

bettinaheim:
Fixes #54

### DIFF
--- a/src/Simulation/Core/QArray.cs
+++ b/src/Simulation/Core/QArray.cs
@@ -387,11 +387,11 @@ namespace Microsoft.Quantum.Simulation.Core
                 return new QArray<T>();
             }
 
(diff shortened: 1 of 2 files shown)


anpaz-msft:
Thought about it after I merged it, but adding tests would have been ideal.

bettinaheim:
> Thought about it after I merged it, but adding tests would have been ideal.

Ah, yes, of course. Sorry, it was late...

Issue #56, opened 2019-09-25
Improve runtime exception handling


vadym-kl:
## Please describe what you would like the feature to accomplish. 
Currently, when `fail` statement is used in Q# only Q# entry-point is shown in .NET call-stack. 
This is caused by the way how `try`,`catch` and `finally` are used in the generated C# code. 
The exception is rethrown by using `throw` keyword which cuts off the call stack . 

## Describe the solution you'd like 
The call stack cut-off during rethrowing of the exception is a known issue and the solutions is also [known](https://stackoverflow.com/questions/57383/how-to-rethrow-innerexception-without-losing-stack-trace-in-c). It would be nice to incorporate the solution into the the runtime.

## Describe alternatives you've considered 
I am not aware of any alternatives to recover full call stack from the point where exception was actually thrown from. 

## Additional context 
This makes it very hard to debug Q# code that relies on library functions such as [`Microsoft.Quantum.Diagnostics.Fact`](https://docs.microsoft.com/qsharp/api/qsharp/microsoft.quantum.diagnostics.fact).  This negatively affects debugging experience across Visual Studio, Visual Studio Code, IQ# and command line. 

For example, try creating Q# application with the following code:
```qsharp
namespace stackTrace
{
    open Microsoft.Quantum.Canon;
    open Microsoft.Quantum.Intrinsic;
    open Microsoft.Quantum.Diagnostics;

    operation U( q : Qubit ) : Unit is Adj {
        Fact(false,"Always fails");
    } 

    operation HelloQ () : Unit {
        using ( q = Qubit() ) {
            ApplyWith(H,U,q);
        }
        
    }
}
```
with the following driver:
```csharp
using System;
using Microsoft.Quantum.Simulation.Core;
using Microsoft.Quantum.Simulation.Simulators;

namespace stackTrace
{
    class Driver
    {
        static void Main(string[] args)
        {
            using (var qsim = new QuantumSimulator())
            {
                HelloQ inst = qsim.Get<HelloQ>(typeof(HelloQ));
                inst.Apply(QVoid.Instance);
            }
        }
    }
}
```
and try to run it with `dotnet run`.  The error you see  is:
```
Unhandled Exception: Microsoft.Quantum.Simulation.Core.ExecutionFailException: Always fails
   at stackTrace.HelloQ.<get_Body>b__26_0(QVoid __in__) in  Operations.qs:line 15
   at Microsoft.Quantum.Simulation.Core.Operation`2.Apply(I a)
[...]

PR #57, opened 2019-09-26
Updating code-generation for exception handling in Q# runtime

vadym-kl:
Using [`ExceptionDispatchInfo.Capture`](https://docs.microsoft.com/dotnet/api/system.runtime.exceptionservices.exceptiondispatchinfo.capture?view=netframework-4.8#System_Runtime_ExceptionServices_ExceptionDispatchInfo_Capture_System_Exception_) to maintain call-stack when exceptions are thrown in Q#.

See also this [post])https://stackoverflow.com/questions/57383/how-to-rethrow-innerexception-without-losing-stack-trace-in-c.) on StackOverflow.

This PR addresses #56.


src/Simulation/CsharpGeneration.Tests/SimulationCodeTests.fs
```diff
@@ -1093,7 +1093,7 @@ namespace N1
             {
                 var qubits = Allocate.Apply(3L);
 #line hidden
-                Exception __arg1__ = null;
```
vadym-kl: All the tests that I updated check that generated code is equal to the string provided in the text.
I had to update all the tests like this, because new exception handling generates new code.

Issue #58, opened 2019-09-26
Notify simulators when runtime exception has occurred during Q# execution


vadym-kl:
## Please describe what you would like the feature to accomplish.
It would be nice if simulators will get notified when runtime exception has occurred during Q# execution. This way, the simulator can preserve additional information about the state of the program to help debug the problem. 

## Describe the solution you'd like 
I would like a method `Fail` to be added to [`IOperationFactory`](https://github.com/microsoft/qsharp-runtime/blob/ebb4b2a9de0110ab45a6e38b38cfbd664759483f/src/Simulation/Core/IOperationFactory.cs) interface that should be called by an operation when an exception occurs, similar to `StartOperation` and `EndOperation` methods.
I would also like to have `OnFail` event in [`SimulatorBase`](https://github.com/microsoft/qsharp-runtime/blob/c8d2c658e937e729d1b843a406a33e5e27b0f219/src/Simulation/Common/SimulatorBase.cs) class, similar to `OnOperationStart`, `OnOperationEnd` events invoked through `StartOperation`, `EndOperation`.  

## Describe alternatives you've considered 
I am not aware of alternatives achieving the same result.

## Additional context 
There is a potential for state to change due to code in `finally` blocks in the C# generated code.
For this reason, the state of the simulator might change by the time the exception has been caught in the C# driver and this in turn makes debugging more difficult especially in absence of interactive debuggers, for example, when using from command line or IQ#.

PR #59, opened 2019-09-26
Enabling simulators OnFail event when exception is thrown

vadym-kl:
This implements #58.


src/Simulation/Simulators.Tests/CoreTests.cs
```diff
@@ -259,5 +259,24 @@ public void BigInts()
...
+        {
+            int exceptionCount = 0;
+            System.Action<System.Runtime.ExceptionServices.ExceptionDispatchInfo> inc = (System.Runtime.ExceptionServices.ExceptionDispatchInfo e) => exceptionCount++;
+            var sim = new TrivialSimulator();
+            sim.OnFail += inc; // increment exception counter when exception is caught
+            var inst = sim.Get<Microsoft.Quantum.Simulation.Simulators.Tests.Circuits.AlwaysFail>();
+            try
+            {
+                inst.Apply(QVoid.Instance);
+            }
```
anpaz-msft:
```suggestion
                Assert.True(false); // make sure that exception actually happened
            }
```

src/Simulation/Simulators.Tests/CoreTests.cs
```diff
@@ -259,5 +259,24 @@ public void BigInts()
...
+            catch(System.Exception e)
+            {
+                Assert.True(true); // make sure that exeption actually happened
```
anpaz-msft:
```suggestion
                Assert.True(true); // make sure that exception actually happened
```

src/Simulation/Common/SimulatorBase.cs
```diff
@@ -21,6 +21,7 @@ public abstract class SimulatorBase : AbstractFactory<AbstractCallable>, IOperat
     {
         public event Action<ICallable, IApplyData> OnOperationStart = null;
         public event Action<ICallable, IApplyData> OnOperationEnd = null;
+        public event Action<System.Runtime.ExceptionServices.ExceptionDispatchInfo> OnFail = null;
```
anpaz-msft: I hate to ask, but what is the `ExceptionDispatchInfo`, and do we really need to provide it?

src/Simulation/Core/Operations/Operation.cs
```diff
@@ -77,6 +77,11 @@ public O Apply(I a)
                 this.Factory?.StartOperation(this, __dataIn(a));
                 __result__ = this.Body(a);
             }
+            catch( Exception e)
```
anpaz-msft: Will the `fail` statement throw an Exception right now? Can we use that same exception instead of the `ExceptionDispatchInfo`?

vadym-kl: `ExceptionDispatchInfo` containst exception itself and relevant debugging information from the moment exception was thrown. So if you say wanted to say access the call-stack of the exception at the moment when it was thrown, you could do it. Original exceptions is available through [`ExceptionDispatchInfo.SourceException`](https://docs.microsoft.com/dotnet/api/system.runtime.exceptionservices.exceptiondispatchinfo.sourceexception) property. 

Fail statement throws exceptions the same way as before.

src/Simulation/Common/SimulatorBase.cs
vadym-kl: [`ExceptionDispatchInfo`](https://docs.microsoft.com/en-gb/dotnet/api/system.runtime.exceptionservices.exceptiondispatchinfo) contains exception itself and relevant debugging information from the moment exception was thrown. So if you say wanted to say access the call-stack of the exception at the moment when it was thrown, you could do it. Original exceptions is available through `ExceptionDispatchInfo.SourceException` property. 

From the docs:
> The ExceptionDispatchInfo object stores the stack trace information and Watson information that the exception contains at the point where it is captured. The exception can be thrown at another time and possibly on another thread by calling the ExceptionDispatchInfo.Throw method. The exception is thrown as if it had flowed from the point where it was captured to the point where the Throw method is called.

I thought it might be nice sometimes to know about the call-stack at the moment when exception was thrown.
But I am fine with just passing an exception itself.

vadym-kl: If you think that `ExceptionDispatchInfo` is an overkill, I could change the argument back to just `Exception`.

PR #64, opened 2019-10-07
Removing project templates

bettinaheim:
Migrating project templates to the compiler repo.


anpaz-msft:
These were migrated into the `qsharp-compiler` repo with [PR 191](https://github.com/microsoft/qsharp-compiler/pull/191)

PR #73, opened 2019-10-22
Functionality to disable qubits in QubitManager.

avasch01:
The QubitManager::Disable(Qubit qubit) interface can be called by a client on a qubit that is currently allocated in order to prevent this qubit from being reused (with the same id) after it will be released. It also immediately prevents this qubit from being available for borrowing or to have any quantum operation perform on it. This might be necessary if the qubit physically went bad or is otherwise locked and cannot be used anymore. A disabled qubit cannot be re-enabled.

The QubitManager::Disable(IQArray<Qubit> qubitsToDisable) allows to disable several qubits at once.

The bool IsDisabled(Qubit qubit) interface can be called to find out if a specific qubit has been disabled.


anpaz-msft:
I'm wondering if there is a better/simpler way to do this...

If I understand correctly, the goal is to prevent that Qubit to be reused. The reason it is re-use is because the QubitManager tries to reuse Qubits after they are released.  What if instead of tracking this at the Qubit level there is a global flag on the QubitManager to indicate it should not be reusing Qubits once released? When this flag is on, if the user wants to allocate re-usable Qubits they use Borrow instead (as opposed to Allocate)?
If we do want to do this at the Qubit level, I still think that having a ReUsable property on the Qubit might be better approach than a method in the QubitManager.
If we truly need a method in the QubitManager, Disable is not intuitive. What about something like `Freeze` or `Hold`?

anpaz-msft:
Actually the opposite of reuse is discard, so we instead of `Disable` we could use `Discard`... again, that is my third option, though.

Issue #68, opened 2019-10-15
Additional handles in generated code for debug support


bettinaheim:
Right now, auto-generated specializations do not contain any positional information. We could provide better debugging support if we pointed to the original specialization declaration (implicit or explicit) upon code gen and have properties with the full name of the callable and it's range within the Q# file.

bettinaheim:
FYI @vadym-kl.

PR #75, opened 2019-10-23
Additional handles in generated code

bettinaheim:
This PR addresses #68. 
It sets the line numbers for autogenerated statements to the line number of the corresponding implicit or explicit specialization declaration in the Q# code, and adds an attribute for each specialization to the callable class that contains the source file where it has been specified as well as the start and end line of the specialization declaration.
The line numbers are 0-based and the end line number is set to -1 if it is the last specialization declaration in the file.

### DIFF
--- a/src/Simulation/Core/Attributes.cs
+++ b/src/Simulation/Core/Attributes.cs
@@ -7,14 +7,14 @@ using System;
 namespace Microsoft.Quantum.Simulation.Core
 {
     [AttributeUsage(AttributeTargets.Class, AllowMultiple = true)]
-    public class SpecializationRangeAttribute : Attribute
+    public class SourceLocationAttribute : Attribute
     { 
         public string SourceFile { get; }
-        public string SpecializationKind { get; }
+        public OperationFunctor SpecializationKind { get; }
         public int StartLine { get; }
         public int EndLine { get; }
 
-        public SpecializationRangeAttribute(string sourceFile, string kind, int startLine, int endLine)
+        public SourceLocationAttribute(string sourceFile, OperationFunctor kind, int startLine, int endLine)
         {
             this.SourceFile = sourceFile;
             this.SpecializationKind = kind;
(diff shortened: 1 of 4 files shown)


src/Simulation/CsharpGeneration.Tests/SimulationCodeTests.fs
```diff
@@ -2303,6 +2304,10 @@ namespace N1
         |> testOneClass randomAbstractOperation
 
         """
+    [SpecializationRangeAttribute("%%%", "Body", 107, 112)]
```
anpaz-msft: nit: we could simple use [SepecializationRange()]

src/Simulation/Core/Attributes.cs
```diff
@@ -6,6 +6,23 @@
 
 namespace Microsoft.Quantum.Simulation.Core
 {
+    [AttributeUsage(AttributeTargets.Class, AllowMultiple = true)]
+    public class SpecializationRangeAttribute : Attribute
```
anpaz-msft: nit: would `SpecializationSourceLocationAttribute` or just `SourceLocation` be a better name?

src/Simulation/Core/Attributes.cs
```diff
@@ -6,6 +6,23 @@
...
+    { 
+        public string SourceFile { get; }
+        public string SpecializationKind { get; }
```
vadym-kl: Sorry for not pointing out this in earlier discussion, but there is `enum`
https://github.com/microsoft/qsharp-runtime/blob/ebb4b2a9de0110ab45a6e38b38cfbd664759483f/src/Simulation/Core/OperationFunctor.cs#L9 
that might be useful here.

src/Simulation/Core/Attributes.cs
bettinaheim: Yes, renamed it to SourceLocationAttribute.

src/Simulation/Core/Attributes.cs
bettinaheim: changed.

Issue #65, opened 2019-10-08
Error messaging in AssertProb


bettinaheim:
The probability given as the actual probability in the error message of AssertProb is not the same as the one given as argument, which may be slightly confusing. This is the offending line: https://github.com/microsoft/qsharp-runtime/blob/ebb4b2a9de0110ab45a6e38b38cfbd664759483f/src/Simulation/Simulators/QuantumSimulator/AssertProb.cs#L49

adrianwyatt:
looking at this now

adrianwyatt:
Fix submitted. #78

bettinaheim:
Thanks, @adrianwyatt!

PR #78, opened 2019-10-29
Updated error message in AssertProb.Body to be clearer when modifying  the expected vs the actual.

adrianwyatt:
In the AssertProb.Body function, to account for the modification of expectedPr to (1 - expectedPr) when (result == Result.Zero), we must also update the ensemblePr to (1 - ensemblePr) when reporting the failure.


adrianwyatt:
Fix for issue #65

src/Simulation/Simulators/QuantumSimulator/AssertProb.cs
```diff
@@ -46,7 +49,18 @@ public QSimAssertProb(QuantumSimulator m) : base(m)
 
                 if (Abs(ensemblePr - expectedPr) > tol)
                 {
-                    var extendedMsg = $"{msg}\n\tExpected:\t{expectedPr}\n\tActual:\t{ensemblePr}";
+                    string extendedMsg;
+                    if (result == Result.Zero)
+                    {
+                        // To account for the modification of expectedPr to (1 - expectedPr) when (result == Result.Zero), 
+                        // we must also update the ensemblePr to (1 - ensemblePr) when reporting the failure.
+                        extendedMsg = $"{msg}\n\tExpected:\t{expectedPr}\n\tActual:\t{(1 - ensemblePr)}";
```
bettinaheim: Now I may be confused, but shouldn't this be originalExpectedPr for the expected value here as well?
```suggestion
                        extendedMsg = $"{msg}\n\tExpected:\t{originalExpectedPr}\n\tActual:\t{(1 - ensemblePr)}";
```

Issue #18, opened 2019-07-11
Comments in src/Simulation/Simulators/QCTraceSimulator/Circuits/*.qs in XML format


anpaz-msft:
These files have comments in XML-doc format. They should be moved to use Q# doc comments format (Markdown).

codingupastorm:
Hey there,

I am interested in working on this. Do you have any requirements in terms of all of the inputs being documented, even if they're not in the old docs? 

I have started here:

https://github.com/microsoft/qsharp-runtime/compare/master...codingupastorm:master

PR #89, opened 2019-11-05
Converting API Docs for Simulation/Simulators/QCTraceSimulator/Circuits/*.qs

codingupastorm:
Contributes to #18 

I commented on the issue but didn't get a response so thought I'd open this PR.

I'll do all of the files, this is just the first of many that need to be converted, to check that it's in line with your expectations.


bettinaheim:
@cgranade Could you take a quick look at this? Looks good to me.

anpaz-msft:
Thanks! This was a great contribution.
Build 0.10.1911.1710 failed, but looks like a random error and only in macOS. Tried again on build 0.10.1911.2002 and it worked like a charm. Merging this changes.

PR #92, opened 2019-11-11
Restrict gate fusion to at most 15 gates

damiansteiger:
@ Alex Borachov:

This change restricts gate fusion to at most 15 gates. Does this improve your timings?


anpaz-msft:
Based on internal testing, we noticed that gate fusion was making simulation pretty slow as the accumulation of gates took longer than just applying them. With this change we now restrict gate fusion to at most 15 gates improving performance by potentially 2x or 3x.

PR #95, opened 2019-11-13
Invoking code gen as rewrite step

bettinaheim:
Right now this has marginal benefits, but makes sense as an intermediate step for eliminating the code gen app and packaging an sdk leveraging the command line compiler.


src/Simulation/CsharpGeneration/RewriteStep.fs
```diff
@@ -0,0 +1,49 @@
...
+            let step = this :> IRewriteStep
+            let dir = step.AssemblyConstants.TryGetValue AssemblyConstants.OutputPath |> function
+                | true, outputFolder when outputFolder <> null -> outputFolder
+                | _ -> step.Name
+            let allSources = 
+                GetSourceFiles.Apply compilation.Namespaces 
+                |> Seq.filter (fun fileName -> (fileName.Value |> Path.GetFileName).StartsWith "Microsoft.Quantum" |> not)
+            for source in allSources do
+                let content = compilation.Namespaces |> SimulationCode.generate source
+                CompilationLoader.GeneratedFile(source, dir, ".g.cs", content) |> ignore
```
anpaz-msft: Remind me, why is this necessary? Shouldn't all that is needed is to write the file to disk?

src/Simulation/CsharpGeneration/RewriteStep.fs
```diff
@@ -0,0 +1,49 @@
...
+        member this.ImplementsPreconditionVerification = false
+        member this.ImplementsPostconditionVerification = false
+        member this.ImplementsTransformation = true
+
+        member this.PreconditionVerification _ = NotImplementedException() |> raise
+        member this.PostconditionVerification _ = NotImplementedException() |> raise
+        
+        member this.Transformation (compilation, transformed) = 
+            let step = this :> IRewriteStep
+            let dir = step.AssemblyConstants.TryGetValue AssemblyConstants.OutputPath |> function
```
anpaz-msft: I'm not convinced that the OutputPath from the compiler should be used here. Maybe by default, but probably this should be a parameter for the constructor of this Emitter.

src/Simulation/CsharpGeneration.App/Program.fs
```diff
@@ -49,26 +48,25 @@ type Logger() =
 let generateFiles (options : Options) = 
     let logger = new Logger()
     let outputFolder = if String.IsNullOrWhiteSpace options.QSTFileName then null else options.OutputFolder
+    let codeGenDll = typeof<Emitter>.Assembly.Location
+    let assemblyConstants = new Dictionary<string, string>()
+    assemblyConstants.[Emitter.SupportUnitTests] <- "true" // always permit unit tests
```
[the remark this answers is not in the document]
bettinaheim: We could remove it, but the idea was to have a handle that would permit to e.g. check if the project references the necessary xUnit packages and only support unit tests if it does (rather than failing). Just failing the C# compilation in this case is the alternative.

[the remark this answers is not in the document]
bettinaheim: removed

src/Simulation/CsharpGeneration.App/Program.fs
```diff
@@ -49,26 +48,25 @@ type Logger() =
...
-            CompilationLoader.GeneratedFile(source, options.OutputFolder, ".g.cs", content) |> ignore
-        with | ex -> logger.Log(ex)
+    if loaded.Success then 0 else 2
 
 
 let [<EntryPoint>] main args = 
     match Parser.Default.ParseArguments<Options> args with 
-    | :? Parsed<Options> as options -> generateFiles options.Value; 0
+    | :? Parsed<Options> as options -> generateFiles options.Value
     | _ -> 1
```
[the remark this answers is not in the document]
bettinaheim: I added it, though the intend is to eliminate the CsharpGeneration.App in the near future.

src/Simulation/CsharpGeneration/AssemblyInfo.fs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace QsCompiler.AssemblyInfo
+
+open System.Runtime.CompilerServices
+
+[<assembly: InternalsVisibleTo("Microsoft.Quantum.CsharpGeneration.App" + SigningConstants.PUBLIC_KEY)>]
```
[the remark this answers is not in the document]
bettinaheim: removed

src/Simulation/CsharpGeneration/RewriteStep.fs
bettinaheim: Waht do you mean? That is what it's doing?

src/Simulation/CsharpGeneration/RewriteStep.fs
bettinaheim: There needs to be a default constructor. The way to pass information is to use the AssemblyConstants dictionary (string, string dict). This is exactly how it worked before.

src/Simulation/CsharpGeneration/RewriteStep.fs
anpaz-msft: yeah, i was asking about `GeneratedFile`. It was unclear why that logic to write the file to disk was wrapped in that method.

PR #98, opened 2019-11-18
Unit testing

ScottCarda-MS:
Added the ability to use unit tests from within Q#.
This implements #104.


src/Simulation/CsharpGeneration.Tests/Circuits/UnitTests.qs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.Quantum.Core {
+    @Attribute ()
+    newtype Attribute = Unit;
+
+    @Attribute ()
+    newtype TestOperation = String;
```
cgranade: May I suggest that this attribute live in M.Q.Diagnostics?

src/Simulation/CsharpGeneration.Tests/Circuits/UnitTests.qs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.Quantum.Core {
+    @Attribute ()
```
cgranade: I thought the suggested style was to not have a space between the attribute name and the inputs?

src/Simulation/CsharpGeneration.Tests/Circuits/UnitTests.qs
ScottCarda-MS: This attribute normally lives in the core library. Since we don't have access to core here, this is just trying to replicate that behavior.

bettinaheim: @cgranade 
Yes, I think having it live in the Diagnostics namespace makes sense, too. I wasn't sure whether we would want it auto-opened so I added it to Core, but I am happy either way. We should consider having a file template for unit tests that contains that namespace an a stump to make it easier to get going.

src/Simulation/CsharpGeneration.App/Program.fs
```diff
@@ -29,6 +28,10 @@ type Options = {
       HelpText = "Destination folder where the output of the compilation will be generated.")>]
     OutputFolder : string
 
+    [<Option("test", Required = false, Default = false,
```
anpaz-msft: Why is this an option? If it *has* to be an option, then why is there no way to turn it off from a user project?

src/Simulation/CsharpGeneration.Tests/SimulationCodeTests.fs
```diff
@@ -3204,3 +3203,112 @@ namespace Microsoft.Quantum.Tests.LineNumbers
         |> 
         testOneFile (Path.Combine("Circuits","LineNumbers.qs"))
 
+        
+    [<Fact>]
+    let ``one file - UnitTests`` () =
+        """
```
anpaz-msft: In hindsight, I think it would be a lot better if these strings, instead of a big string inside the .fs file would be all saved in their own file and during the test we just read the content of the file.

src/Simulation/CsharpGeneration.Tests/SimulationCodeTests.fs
```diff
@@ -3204,3 +3203,112 @@ namespace Microsoft.Quantum.Tests.LineNumbers
...
+        public override IApplyData __dataOut(QVoid data) => data;
+        public static System.Threading.Tasks.Task<QVoid> Run(IOperationFactory __m__)
+        {
+            return __m__.Run<UnitTest1, QVoid, QVoid>(QVoid.Instance);
+        }
+    }
+
+    public partial class QuantumSimulatorTests
+    {
+        [Fact(DisplayName = "UnitTest1")]
```
anpaz-msft: do we want to keep using `xUnit` as the main framework?
If we're doing this change I would suggest on taking the opportunity to move to `mstest`.

src/Simulation/CsharpGeneration.Tests/SimulationCodeTests.fs
```diff
@@ -3204,3 +3203,112 @@ namespace Microsoft.Quantum.Tests.LineNumbers
...
+        
+    [<Fact>]
+    let ``one file - UnitTests`` () =
+        """
+#pragma warning disable 1591
+using System;
+using Microsoft.Quantum.Core;
+using Microsoft.Quantum.Intrinsic;
+using Microsoft.Quantum.Simulation.Core;
+using Microsoft.Quantum.Simulation.Simulators;
```
anpaz-msft: From experience, it is better if instead of `using` statements, you actually use full-qualified-names for the types that you need from this namespaces.

src/Simulation/CsharpGeneration.Tests/SimulationCodeTests.fs
```diff
@@ -3204,3 +3203,112 @@ namespace Microsoft.Quantum.Tests.LineNumbers
...
+
+        public override IApplyData __dataIn(QVoid data) => data;
+        public override IApplyData __dataOut(QVoid data) => data;
+        public static System.Threading.Tasks.Task<QVoid> Run(IOperationFactory __m__)
+        {
+            return __m__.Run<UnitTest1, QVoid, QVoid>(QVoid.Instance);
+        }
+    }
+
+    public partial class QuantumSimulatorTests
```
anpaz-msft: I wouldn't name this `QuantumSimulatorTests`. It can clash, but also, these are not tests for the quantum simulator, so the name is missleading.
What about something like simply `__AllTests__`.

src/Simulation/CsharpGeneration.Tests/SimulationCodeTests.fs
```diff
@@ -3204,3 +3203,112 @@ namespace Microsoft.Quantum.Tests.LineNumbers
...
+    {
+        [Fact(DisplayName = "UnitTest1")]
+        public void __UnitTest1Test__()
+#line 20 "%%%"
+        {
+            var sim = new QuantumSimulator();
```
anpaz-msft: I would provide a mechanism such that the simulator to use can be programmatically specified.
Maybe you could pass it during constructor? Or maybe creating and initializing is a protected method that can be overriden?

src/Simulation/CsharpGeneration/AssemblyInfo.fs
```diff
@@ -0,0 +1,11 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace QsCompiler.AssemblyInfo
+
+open System.Runtime.CompilerServices
+
+[<assembly: InternalsVisibleTo("Microsoft.Quantum.CsharpGeneration.App" + SigningConstants.PUBLIC_KEY)>]
```
anpaz-msft: why are these necessary? In particular, I would make sure that anything that `CsharpGeneration.App` needs is public as it will probably be needed by iq#.

src/Simulation/Simulators.Tests/Circuits/UnitTests.qs
```diff
@@ -0,0 +1,16 @@
...
+// Licensed under the MIT License.
+
+namespace Microsoft.Quantum.Simulation.Tests.QsUnitTests {
+
+    open Microsoft.Quantum.Intrinsic;
+    open Microsoft.Quantum.Diagnostics;
+    
+    @TestOperation("QuantumSimulator")
+    operation QsharpUnitTest() : Unit {
+		Message("Worked!");
```
anpaz-msft: How do you know if this unittest was actually executed? It could not be executed and then nothing bad would happen.

src/Simulation/CsharpGeneration/SimulationCode.fs
```diff
@@ -30,77 +30,26 @@ open Microsoft.Quantum.QsCompiler.Transformations.BasicTransformations
...
-        }
               
-    let autoNamespaces = 
-        [
-            "System"
-            "Microsoft.Quantum.Core"
-            "Microsoft.Quantum.Intrinsic"
-            "Microsoft.Quantum.Simulation.Core" 
-        ]
+    let unitTestingNamespaces = [
```
anpaz-msft: As mentioned before, I would not add these to the top, but simply use FQN for whatevery types it needs.

src/Simulation/CsharpGeneration/SimulationCode.fs
```diff
@@ -385,7 +334,7 @@ module SimulationCode =
             match ex.Expression with
 // TODO: Diagnostics
             | InvalidExpr                   -> failwith "Can't generate code for error expression"
-            | UnitValue                     -> (``ident`` "QVoid") <|.|> (``ident`` "Instance")
```
anpaz-msft: I like keeping the ticks because it's easier to identify what are calls that are generating C# code.

src/Simulation/CsharpGeneration/SimulationCode.fs
```diff
@@ -1032,7 +981,7 @@ module SimulationCode =
                 | false, _ -> startLine
             ``attribute`` None (ident "SourceLocation") [ 
                 ``literal`` sp.SourceFile.Value 
-                ``ident`` "OperationFunctor" <|.|> ``ident`` bodyName
```
anpaz-msft: Again, add these ticks back so its consistent that any method that generates C# uses the backticks.

src/Simulation/CsharpGeneration/Context.fs
```diff
@@ -0,0 +1,101 @@
...
+    allCallables            : ImmutableDictionary<QsQualifiedName,QsCallable>
+    declarationPositions    : ImmutableDictionary<NonNullable<string>, ImmutableSortedSet<int * int>>
+    byName                  : ImmutableDictionary<NonNullable<string>,(NonNullable<string>*QsCallable) list>
+    current                 : QsQualifiedName option
+    signature               : ResolvedSignature option
+    fileName                : string option
+    unitTests               : ILookup<NonNullable<string>, string> // for each namespace contains the targets on which unit tests need to be executed
+} 
+    with
+    static member internal Create (syntaxTree, assemblyConstants, mayContainTests) =
```
anpaz-msft: I dont' think this should be a parameter. If tests are defined, it should just generate the code for them.

src/Simulation/CsharpGeneration.Tests/SimulationCodeTests.fs
bettinaheim: After a round a syncing with everyone, we should briefly discuss the switch to MsTest as a team, including Viktor and Cathy.

src/Simulation/Simulators.Tests/Circuits/UnitTests.qs
ScottCarda-MS: If the test was not run, then whatever tool is being used to run tests would show that the test was not run. Maybe I'm missing the point of your comment?

src/Simulation/CsharpGeneration.Tests/SimulationCodeTests.fs
ScottCarda-MS: Would this involve a rename of the partial class definitions for QuantumSimulatorTests found in BasicTests.cs, CircuitsTests.cs, and VerifyGates.cs under Simulators.Tests\QuantumSimulatorTests?

src/Simulation/CsharpGeneration.Tests/SimulationCodeTests.fs
ScottCarda-MS: When decorate a Q# operation with the TestsOperation attribute, you will be expected to provide the name of the target as an argument to the attribute. This name will appear here.

src/Simulation/CsharpGeneration.Tests/SimulationCodeTests.fs
bettinaheim: I think the bundling according to what target they run on is both useful and actually also needed. How about something like "__ExecutedOnQuantumSimulator__" instead?

src/Simulation/CsharpGeneration.Tests/SimulationCodeTests.fs
bettinaheim: That's a good idea. I made a task for it: https://dev.azure.com/ms-quantum/Quantum%20Program/_workitems/edit/9144

ScottCarda-MS: This would be a fairly involved processes. Let's have that be in a separate PR.

src/Simulation/CsharpGeneration/SimulationCode.fs
```diff
@@ -1412,11 +1352,81 @@ module SimulationCode =
...
+                // namespace AssemblyName // we omit this nested namespace if the assembly name is not specified
+                // {
+                //   public partial class __ExecuteOnTargetName__
+                //   #line nr source
+                //   {
+                //     [Fact(DisplayName = "TestOperationName")]
+                //     public void __TestOperationNameTest__()
+                //     {
+                //         var sim = new TargetName(); 
+                //         sim.OnLog += this.Output.WriteLine;
```
anpaz-msft: And we should check that `sim` is an instance of `SimulatorBase` before attaching to the event.

src/Simulation/Intrinsic/Diagnostics/UnitTests.qs
```diff
@@ -0,0 +1,18 @@
...
+
+namespace Microsoft.Quantum.Diagnostics {
+
+    /// # Summary
+    /// Compiler recognized attribute used to mark a unit test.
+    /// 
+    /// # Input
+    /// ## ExecutionTarget
+    /// The name of the target to execute the test on. 
+    /// Possible values are: QuantumSimulator, TraceSimulator, ToffoliSimulator
```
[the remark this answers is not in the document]
anpaz-msft: Optionally, use this namespace if a FQN is not specified, but don't add it if it is specified.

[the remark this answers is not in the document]
ScottCarda-MS: Unfortunately, it doesn't look like that is supported at this time.

build/Props/QSharp.targets
```diff
@@ -7,7 +7,7 @@
   -->
   <Target Name="QsharpCompile"
       Inputs="@(QsharpFiles)"
-      Outputs="$(QsharpSrcOutDir)\$(QsharpTree).bson"
+      Outputs="$(QsharpSrcOutDir)\$(QsharpTree)"
```
anpaz-msft: This is reverting #100, please undo.

src/Simulation/CsharpGeneration.Tests/SimulationCodeTests.fs
```diff
@@ -3207,3 +3206,229 @@ namespace Microsoft.Quantum.Tests.LineNumbers
...
+namespace Microsoft.Quantum.Tests.UnitTests
+{
+    [SourceLocation("%%%", OperationFunctor.Body, 22, 26)]
+    public partial class UnitTest1 : Operation<QVoid, QVoid>, ICallable
+    {
+        public UnitTest1(IOperationFactory m) : base(m)
+        {
+        }
+
+        public class QuantumSimulator
```
anpaz-msft: A different approach we could have used is to instead of creating a subclass per simulator, we could have created one single `Tests` class and use `[Xunit.Theory]` accepting the `Simulator` in the constructor and creating the instances using the `[Xunit.InlineData(new QuantumSimulator())]`. This might be a little cleaner and group the tests better.

anpaz-msft: BTW, I'm ok keeping the generation as-is, just putting the idea as that might be cleaner.

bettinaheim: I like the idea, but it looks like IndlineData attribute (like any attribute) is not happy about the non-constant expression "new QuantumSimulator()"

PR #99, opened 2019-11-21
Tagging the generated C# code as auto-generated

bettinaheim:
We should make the C# files that we generate recognizable as auto-generated files to analyzers and such. Auto-generated files are and should be treated differently from user code in terms of e.g. what info and tools are provided to the user. 
There doesn't seem to be one unified way to do so that works for everything. I added the three ways of which I am aware of to make the code recognizable as generated: 
1.) There is a handle in the project system that tags included files as auto-generated. Unfortunately, that information is not available to a lot of tools. 
2.) A lot of tools therefore just check for files in the obj folder that starts with a comment // \<auto-generated\>, and assume that these were indeed auto-generated. 
3.) There also seems to be a class attribute (https://docs.microsoft.com/en-us/dotnet/api/system.codedom.compiler.generatedcodeattribute?view=netcore-3.0); 
we are *not* generating that one unless a need for it arises.


ScottCarda-MS:
Could you please provide a description for this pull request? I can tell the effect is to add a comment to the top of generated files, "// <auto-generated>", as well as add this codegen attribute to many of the classes, but I am not sure why we should do this. What is the benefit?

bettinaheim:
> Could you please provide a description for this pull request? I can tell the effect is to add a comment to the top of generated files, "// <auto-generated>", as well as add this codegen attribute to many of the classes, but I am not sure why we should do this. What is the benefit?

Sorry, forgot to add one. Will do.

PR #117, opened 2019-12-27
Added new hex- and bit-format dump for ToffoliSimulator.

cgranade:
This PR implements an old feature request for the Toffoli simulator, providing a new hex-dump based output for `DumpMachine` and `DumpRegister`.


src/Simulation/Simulators/ToffoliSimulator/ToffoliSimulator.cs
```diff
@@ -483,4 +493,67 @@ public override void Apply(IQArray<Qubit> qubits)
             }
         }
     }
+    
+    internal static class Extensions
```
anpaz-msft: do we want to do any null checks?

cgranade: Good point, yeah. I can use `#nullable enable` as well to help make sure I've caught relevant null-checks.

Issue #116, opened 2019-12-26
AccessViolationException after updating to .NET Core 3.1.100 Sdk


bettinaheim:
## Describe the bug
When using the latest version of the .NET Core Sdk (3.1.100), the Borrow test in Microsoft.Quantum.Simulation.Simulators under CoreTests starts to not execute silently. The reason is a AccessViolationException. See the attached log from the build. 

## To Reproduce
Download and install the latest .NET Core Sdk from here: https://dotnet.microsoft.com/download. 
Try to run the Borrowing test. 

## Expected behavior
The test executes as it has with .NET Core Sdk 3.0.100. 

## Actual behavior
The test host process crashes due to a System.AccessViolationException (see attached log). 

## System information
- OS: Windows (not sure if the same problem exists on Mac/Linux)
- .NET Core Version:  3.1.100

For more details, see [Test log](https://github.com/microsoft/qsharp-runtime/files/4001654/8_Testing.Q.runtime.txt)

bettinaheim:
@anpaz-msft @avasch01 For your awareness, since I am oof.

bettinaheim:
Actually, come to think of it - our latest Binder issues might be related.

avasch01:
The issue happens when we are trying to extract qubits from operation parameters:

2019-12-26T06:34:48.3736519Z The active test run was aborted. Reason: Test host process crashed : Fatal error. System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
2019-12-26T06:34:48.3737508Z    at Microsoft.Quantum.Tests.CoreOperations.BGen`1+In[[System.__Canon, System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKey***=7cec85d7bea7798e]].Microsoft.Quantum.Simulation.Core.IApplyData.get_Qubits()
2019-12-26T06:34:48.3737760Z    at

anpaz-msft:
There is some bug in .NET Core, probably in the compiler.

This causes an AccessViolation:
            System.Collections.Generic.IEnumerable<Qubit> IApplyData.Qubits
            {
                get
                {
                    var q2 = Data.Item2?.GetQubits();
                    return q2;
                }
            }

but this works file:


            System.Collections.Generic.IEnumerable<Qubit> IApplyData.Qubits
            {
                get
                {
                    var i2 = Data.Item2;
                    var q2 = i2?.GetQubits();
                    return q2;
                }
            }

[...]

PR #120, opened 2019-12-31
Workaround for .NET Core 3.1.100 issue

bettinaheim:
Workaround for https://github.com/microsoft/qsharp-runtime/issues/116, as identified by @anpaz-msft. Thanks, Andres!

### DIFF
--- a/src/Simulation/CsharpGeneration/SimulationCode.fs
+++ b/src/Simulation/CsharpGeneration/SimulationCode.fs
@@ -1118,6 +1118,8 @@ module SimulationCode =
                 ``property-get`` "System.Collections.Generic.IEnumerable<Qubit>" "IApplyData.Qubits" [] 
                     ``get`` (fieldPaths |> List.map buildOne)
         else
+            // this implementation is a workaround for the .NET Core issue discussed here: 
+            // https://github.com/microsoft/qsharp-runtime/issues/116
             let mutable count = 0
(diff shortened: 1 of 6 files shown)


src/Simulation/CsharpGeneration/SimulationCode.fs
```diff
@@ -1117,25 +1117,45 @@ module SimulationCode =
...
                 | QsTypeKind.ArrayType _
                 | QsTypeKind.UserDefinedType _
-                | QsTypeKind.Qubit -> ``((`` ( ``cast`` "IApplyData" (field |> snd) ) ``))`` <|?.|> ``ident`` "Qubits"
-                | _       -> (field |> snd)  <?.>  ( ``ident`` "GetQubits", [] )
-            let body = 
+                | QsTypeKind.Qubit -> 
+                    let qs = ``((`` ( ``cast`` "IApplyData" token) ``))`` <|?.|> ``ident`` "Qubits"
+                    items <- (null, qs) :: items
+                | _       -> 
+                    let id = nextName()
```
cgranade: If I was reading this without knowing about the .NET Core bug, this would seem pretty awkward to me. Perhaps would it be reasonable to add a comment here indicating the workaround?

bettinaheim: Sure, done.

PR #130, opened 2020-01-12
Revert "Revert "Added new hex- and bit-format dump for ToffoliSimulat…

bettinaheim:
We should now be ready to take this in again. 
Please wait with merging until we've confirmed the end-to-end build.


cgranade:
Approved, but holding off merging as per your advice.

PR #133, opened 2020-01-28
Fix mismatch in line break format in ToffoliDumpState test

irinayat-MS:
The test was failing locally on my machine due to mismatch between the oracle strings (\r) and the text dumped into the file (\r\n). I've changed the string to use Environment.NewLine. It passes on my machine now. Hopefully, it can work with other system defaults as well.


cgranade:
Thanks for fixing this!

PR #140, opened 2020-02-04
Better Exceptions for QuantumProcessorBase

avasch01:
Better Exceptions for QuantumProcessorBase

### DIFF
--- a/src/Simulation/Common/QuantumProcessorBase.cs
+++ b/src/Simulation/Common/QuantumProcessorBase.cs
@@ -4,178 +4,179 @@
 using System;
 using Microsoft.Quantum.Simulation.Core;
 using System.Diagnostics;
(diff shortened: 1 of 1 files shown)


src/Simulation/Common/QuantumProcessorBase.cs
```diff
@@ -282,4 +283,18 @@ public virtual void OnMessage(string msg)
         }
 
     }
+
+    /// <summary>
+    /// A class that implements exception to be thrown when Operation is not supported by a QuantumProcessor.
+    /// </summary>
+    public class UnsupportedOperationException : NotImplementedException
```
irinayat-MS: Why inherit from NotImplementedException? Do we expect code that catches NYI to catch these too? Also, could we use either of the built-in NotSupportedException/PlatformNotSuppotedException?

PR #145, opened 2020-02-11
Added OperationInfo static property to the generated C# class

cesarzc:
Added a property to the generated C# class to be able to retrieve information about the operation.


cesarzc:
Currently working on addressing the build breaks on the simulation projects as well as expanding the unit test. I will update the PR soon.

PR #146, opened 2020-02-11
QuantumProcessor interface calls with empty controls

avasch01:
QuantumProcessor interface will not call controlled versions of APIs when the list of controls is empty or null. It will call non-controlled versions instead.

### DIFF
--- a/src/Simulation/Simulators/QuantumProcessor/ClassicalControl.cs
+++ b/src/Simulation/Simulators/QuantumProcessor/ClassicalControl.cs
@@ -257,7 +257,7 @@ namespace Microsoft.Quantum.Simulation.QuantumProcessor
 
             public override Func<(IQArray<Result>, IQArray<Result>, IUnitary, IUnitary), QVoid> Body => (q) =>
             {
-                var (measurementResults, resultsValues, onEqualOp, onNonEqualOp) = q;
+                (IQArray<Result> measurementResults, IQArray<Result> resultsValues, ICallable onEqualOp, ICallable onNonEqualOp) = q;
                 return ExecuteConditionalStatement(Simulator, measurementResults, resultsValues, onEqualOp, onNonEqualOp, OperationFunctor.Body, null);
             };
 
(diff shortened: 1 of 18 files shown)


src/Simulation/Simulators/QuantumProcessor/ClassicalControl.cs
```diff
@@ -123,14 +123,21 @@ public class QuantumProcessorApplyIfElseC : Extensions.ApplyIfElseIntrinsicC
...
             };
 
             public override Func<(IQArray<Qubit>, (Result, IControllable, IControllable)), QVoid> ControlledBody => (q) =>
             {
-                var (ctrls, (measurementResult, onZero, onOne)) = q;
-                return ExecuteConditionalStatement(Simulator, measurementResult, Result.Zero, onZero, onOne, OperationFunctor.Controlled, ctrls);
+                (IQArray<Qubit> ctrls, (Result measurementResult, ICallable onZero, ICallable onOne)) = q;
+                if ((ctrls == null) || (ctrls.Count == 0))
+                {
+                    return ExecuteConditionalStatement(Simulator, measurementResult, Result.Zero, onZero, onOne, OperationFunctor.Body, null);
```
irinayat-MS: > return ExecuteConditionalStatement(Simulator, measurementResult, Result.Zero, onZero, onOne, OperationFunctor.Body, null); [](start = 19, length = 123)

Why is it useful to replace empty ctrls with null? It will be passed through to the primitive ops in RunClause so both should be handled by them, no?

avasch01: The contract here is to pass null when OperationFunctor.Body is passed. It would work either way, but a little cleaner to maintain the invariant.

---
In reply to: [377963209](https://github.com/microsoft/qsharp-runtime/pull/146#discussion_r377963209) [](ancestors = 377963209)

irinayat-MS: Oh, I see! I didn't notice that OperationFunctor was also changed.

---
In reply to: [377963994](https://github.com/microsoft/qsharp-runtime/pull/146#discussion_r377963994) [](ancestors = 377963994,377963209)

PR #147, opened 2020-02-13
Updated exception handling code generation

DmitryVasilevsky:
Before this fix the stack trace was captured incorrectly on exceptions. Several locations of the same C# function were present on the stack due to incorrect re-throwing and it looked like a recursion. This change modifies exception handling to use "throw" without arguments to re-throw exceptions without changing stack trace. This is needed for correct matching between C# and Q# stacks, which is needed to report correct locations in Q# stack (note that this change alone won't fix missing locations in Q# stack as other fixes are needed).
The main change is in F# code that generates error handing for Q#. Therefore all generated code is now different. This required a lot of changes in expected generated code in tests. New code uses fewer temporary variables so temporary variable names have also changed causing changes in places not related to exception handling.
This pull request includes new test AllocateQubit2Test which will eventually test proper exception information propagation in case of an assert. Currently this test only test basic functionality as error propagation is still incorrect at the moment.

### DIFF
--- a/src/Simulation/CsharpGeneration/SimulationCode.fs
+++ b/src/Simulation/CsharpGeneration/SimulationCode.fs
@@ -787,9 +787,8 @@ module SimulationCode =
                 releases    
                 
             let symbols = removeDiscarded using.Binding.Lhs
-            let exDispatchInfoName = nextArgName() 
-            let exDispatchInfoHandle = ``ident`` exDispatchInfoName
-            let caughtEx = nextArgName()
+            let deallocationFlagName = nextArgName()
+            let deallocationFlagIdentifier = ``ident`` deallocationFlagName
 
             // allocations and deallocations
             let lhs = symbols |> buildSymbolNames id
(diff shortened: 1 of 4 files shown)


src/Simulation/CsharpGeneration/SimulationCode.fs
```diff
@@ -800,19 +799,15 @@ module SimulationCode =
             // To force that exceptions thrown during the execution of the allocation scope take precedence over the ones thrown upon release
             // we catch all exceptions in a variable and throw after releaseing if necessary. 
 
-            // System.Runtime.ExceptionServices.ExceptionDispatchInfo is used to keep all call stack information when rethrowing
-            let exceptionHandle = ``typed var`` "System.Runtime.ExceptionServices.ExceptionDispatchInfo" exDispatchInfoName (``:=`` ``null`` |> Some) |> ``#line hidden`` :> StatementSyntax
+            // Indicates if deallocation is needed. It is not needed when excaption is thrown.
```
avasch01: >excaption [](start = 74, length = 9)

exception

DmitryVasilevsky: Fixed,

src/Simulation/Simulators.Tests/StackTraceTests.cs
```diff
@@ -26,6 +26,23 @@ public StackTraceTests(ITestOutputHelper output)
...
+            using (var sim = new QuantumSimulator())
+            {
+                try
+                {
+                    QVoid res = AllocateQubit2.Run(sim).Result;
+                }
+                catch (AggregateException ex)
+                {
+                    Assert.True(ex.InnerException is ExecutionFailException);
+                    // TODO: Need more checks here - stack frames and line numbers
```
bettinaheim: What was the reason to not add the additional checks for the stack frames?

DmitryVasilevsky: As I mentioned in the description, more fixes are needed for the stack frames to be correct. If I add additional checks now they will fail.
I'm currently working on better C#-Q# frame matching. When this is done I'll add more checks.

PR #148, opened 2020-02-19
Changed stack matching algorithm not to match some Q# frames

DmitryVasilevsky:
Updated Q# - C# stack matching algorithm. Skipping Q# frames that have no chance of matching - i.e. those that have no source information.
This helps continue matching even if Q# intrinsic function is on stack, which has no source information.

### DIFF
--- a/src/Simulation/Common/StackTrace.cs
+++ b/src/Simulation/Common/StackTrace.cs
@@ -123,6 +123,7 @@ namespace Microsoft.Quantum.Simulation.Common
         /// </summary>
         public static StackFrame[] PopulateSourceLocations(Stack<StackFrame> qsharpCallStack, System.Diagnostics.StackFrame[] csharpCallStack)
         {
+            // Populate source location information in CSharp stack from attributes
             foreach (StackFrame currentFrame in qsharpCallStack)
             {
                 ICallable op = currentFrame.Callable.UnwrapCallable();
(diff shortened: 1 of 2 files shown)


src/Simulation/Common/StackTrace.cs
```diff
@@ -135,36 +136,94 @@ public static StackFrame[] PopulateSourceLocations(Stack<StackFrame> qsharpCallS
...
+                    return currentFrame;
+                }
+            }
+            return null;
+        }
+
+        /// <summary>
+        /// Return next C# stack frame that has enough information to match
+        /// </summary>
+        private static System.Diagnostics.StackFrame GetNextCSharpStackFrame(System.Diagnostics.StackFrame[] csharpStackFrames, ref int currentFrameIndex)
```
avasch01: >GetNextCSharpStackFrame [](start = 53, length = 23)

Let's simplify to currentFrameIndex++; without skipping empties.

src/Simulation/Common/StackTrace.cs
```diff
@@ -135,36 +136,94 @@ public static StackFrame[] PopulateSourceLocations(Stack<StackFrame> qsharpCallS
...
+
+        /// <summary>
+        /// Return next Q# stack frame that has enough information to match
+        /// </summary>
+        private static StackFrame GetNextQSharpStackFrame(StackFrame[] qsharpStackFrames, ref int currentFrameIndex)
+        {
+            while (currentFrameIndex < qsharpStackFrames.Length)
+            {
+                StackFrame currentFrame = qsharpStackFrames[currentFrameIndex];
+                currentFrameIndex++;
```
avasch01: >currentFrameIndex [](start = 16, length = 17)

currentFrame does not correspond to currentFrameIndex. It would be much nicer if they matched.

src/Simulation/Common/StackTrace.cs
DmitryVasilevsky: Updated code not to skip C# frames.

src/Simulation/Common/StackTrace.cs
DmitryVasilevsky: Updated "current frame" indexes to point to current frame (and not to current+1 frame as before).

src/Simulation/Common/StackTrace.cs
```diff
@@ -135,36 +136,106 @@ public static StackFrame[] PopulateSourceLocations(Stack<StackFrame> qsharpCallS
...
+            }
+            return null;
+        }
+
+        /// <summary>
+        /// Return next C# stack frame
+        /// </summary>
+        private static System.Diagnostics.StackFrame GetNextCSharpStackFrame(System.Diagnostics.StackFrame[] csharpStackFrames, ref int currentFrameIndex)
+        {
+            if (currentFrameIndex >= csharpStackFrames.Length)
```
avasch01: >if [](start = 12, length = 2)

Do we really need this if twice?

PR #153, opened 2020-02-25
Remove link to project templates and fix typo in README

marshallsa:
Project templates were moved to the compiler repository, so the link to the project templates folder doesn't work anymore.


marshallsa:
Thanks, I forgot about that.

PR #157, opened 2020-03-03
Separate stack trace handling into new OnException event.

cgranade:
This PR separates stack trace handling into a new event, paralleling the design of `OnLog`. This PR also separates our source file identification logic into a method that can be called separately in anticipation of picking that feature up from IQ# and providing rich stack traces there.

(NB: as I've been making PRs, I try to add `#nullable enable`. In this case, that has identified some potential issues with qubit managers being null that might be worth addressing in future PRs.)

### DIFF
--- a/src/Simulation/Common/SimulatorBase.cs
+++ b/src/Simulation/Common/SimulatorBase.cs
@@ -174,7 +174,7 @@ namespace Microsoft.Quantum.Simulation.Common
         }
 
         /// <summary>
-        ///     Adds an event to the OnStackTrace event that logs stack traces
+        ///     Adds an event to the OnException event that logs stack traces
         ///     as plain text via the OnLog event (e.g.: for console output).
         /// </summary>
         public void EnableExceptionPrinting()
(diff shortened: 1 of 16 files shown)


src/Simulation/Common/SimulatorBase.cs
```diff
@@ -165,6 +173,25 @@ public void DisableLogToConsole()
             OnLog -= Console.WriteLine;
         }
 
+        /// <summary>
+        ///     Adds an event to the OnStackTrace event that logs stack traces
```
anpaz-msft:
```suggestion
        ///     Adds an event to the OnException event that logs stack traces
```

src/Simulation/Common/SimulatorBase.cs
```diff
@@ -165,6 +173,25 @@ public void DisableLogToConsole()
...
+        ///     as plain text via the OnLog event (e.g.: for console output).
+        /// </summary>
+        public void EnableDefaultStackTraceHandling()
```
anpaz-msft: Maybe name this `EnablePrintException` is not handling StackTrace but how exceptions are handled...

PR #161, opened 2020-03-16
Adding a EnableTestingViaName attribute

bettinaheim:
For testing purposes, it is convenient if it is possible to compare different implementation for the same callables. We can facilitiate that with some built-in support. See also https://github.com/microsoft/qsharp-compiler/pull/373


bettinaheim:
included in https://github.com/microsoft/qsharp-runtime/pull/163

PR #163, opened 2020-03-25
Additional support for targeting

bettinaheim:


src/Simulation/CsharpGeneration/SimulationCode.fs
```diff
@@ -1336,7 +1341,14 @@ module SimulationCode =
         let outType  = op.Signature.ReturnType   |> roslynTypeName context
 
         let constructors = [ (buildConstructor context name) ]
-        let properties = buildName name :: buildFullName context.current.Value :: buildOperationInfoProperty inType outType nonGenericName :: buildOpsProperties context opNames
+        let properties = 
+            let opProperties = buildOpsProperties context opNames
+            buildName name :: 
+            buildFullName context.current.Value :: 
+            if globalContext.entryPoints |> Seq.contains op.FullName then
```
cesarzc: A question to understand a bit better what is going on here. It seems like the EntryPointInfo property and the properties that are generated based on the targeted architecture are only going to be generated for Operations that are marked as entry points. However, I don't get how this is done. What is the purpose of "Seq.contains op.FullName"?

src/Simulation/Core/EntryPointInfo.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        public EntryPointInfo(Type operation) => 
+            this.Operation = operation;
+    }
+
+    /// <summary>
+    /// Base class containing information about an entry point 
+    /// for a Q# executable targeted for a Alfred quantum processor.
+    /// </summary>
+    public class AlfredExecutable<I, O>
```
cesarzc: Regarding the names of the classes, I think it would be more consistent to have AlfredInfo, BrunoInfo and ClementineInfo as names.

The current names are ok with me if you feel strongly about keeping them.

src/Simulation/CsharpGeneration/SimulationCode.fs
bettinaheim: The global context contains an ImmutableArray of entry points. We do currently enforce that no more than one entry point exists, but for the sake of keeping data structures flexible, the compilation actually contains an ImmutableArray. Seq.contains just checks whether that array lists op as entry point.

src/Simulation/Core/EntryPointInfo.cs
bettinaheim: AlfredInfo seems slightly misleading, since it doesn't actually contain infos about Alfred. How about renaming EntryPointInfo to Executable, or renaming AlfredExecutable to AlfreadEntryPointInfo etc?

bettinaheim: Going with AlfredEntryPointInfo etc for now - I'm happy to change it later, when make the pass to adapt these names, since the names are only temporary.

PR #165, opened 2020-03-27
Generate private-protected C# properties for internal Q# callables

### LINKED ISSUE #164: Calling an internal type constructor in a public operation generates invalid C#
## Describe the bug
Calling an internal type constructor in a public operation generates a C# class for the operation that uses the internal type as the type of a protected field, which is not legal.

## To Reproduce
```
internal newtype MyInt = Int;

operation HelloQ() : Unit {
    let x = MyInt(3);
}
```

## Expected behavior
The generated C# code should compile.

## System information
- OS: Windows
- .NET Core 3.1.200
- Q# SDK 0.11.2003.2505-beta

**Additional context**
The problem property is:

```
public partial class HelloQ : Operation<QVoid, QVoid>, ICallable
{
    // ...

    protected ICallable<Int64, MyInt> MyInt
    {
        get;
        set;
    }

    // ...
}
```

marshallsa:
Fixes #164.

### DIFF
--- a/src/Simulation/CsharpGeneration.Tests/Circuits/CodegenTests.qs
+++ b/src/Simulation/CsharpGeneration.Tests/Circuits/CodegenTests.qs
@@ -1309,11 +1309,20 @@ namespace Microsoft.Quantum.Compiler.Generics {
 
     // Access Modifiers
 
-    internal function InternalFunction () : Unit {
-    }
+    internal function EmptyInternalFunction () : Unit { }
 
-    internal operation InternalOperation () : Unit {
-    }
+    internal operation EmptyInternalOperation () : Unit { }
 
     internal newtype InternalType = Unit;
+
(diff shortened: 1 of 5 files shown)


bettinaheim:
@marshallsa It would be good to have an execution test. Since you need to test this across compilation units, I'd suggest to create a Q# test project, reference that from the simulator tests project, and add a couple of tests there to see if everything executes as expected.

marshallsa:
@bettinaheim I can add an execution test to ensure that the generated C# compiles, but it seems like  using a project reference to test that C#'s `private protected` works correctly is unnecessary?

bettinaheim:
> @bettinaheim I can add an execution test to ensure that the generated C# compiles, but it seems like  using a project reference to test that C#'s private protected works correctly is unnecessary?

Yes, that's fair. But and execution test would be good.

PR #168, opened 2020-04-03
IQuantumMachine

anpaz-msft:
As part of Azure Quantum and Q# executables, we are introducing the IQuantumMachine as the interface that defines an machine you can submit quantum jobs to.

### DIFF
--- a/src/Simulation/Core/IQuantumMachine.cs
+++ b/src/Simulation/Core/IQuantumMachine.cs
@@ -9,13 +9,9 @@ namespace Microsoft.Quantum.Runtime
 {
     /// <summary>
     /// Interface that a quantum machine must implement.
-    ///
-    /// Type parameters:
-    ///     - Job: Type of job that handles the execution of a program in the quantum machine.
-    ///     - QResult: Type of result the quantum machine returns.
     /// </summary>
-    // TODO: This interface should be migrated to the qsharp-runtime repository.
-
+    /// <typeparam name="Job">Type of job that handles the execution of a program in the quantum machine.</typeparam>
+    /// <typeparam name="QRawResult">Type of result the quantum machine returns.</typeparam>
     public interface IQuantumMachine<Job, QRawResult>
     {
         /// <summary>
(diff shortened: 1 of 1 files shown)


src/Simulation/Core/IQuantumMachine.cs
```diff
@@ -0,0 +1,60 @@
...
+namespace Microsoft.Quantum.Runtime
+{
+    /// <summary>
+    /// Interface that a quantum machine must implement.
+    ///
+    /// Type parameters:
+    ///     - Job: Type of job that handles the execution of a program in the quantum machine.
+    ///     - QResult: Type of result the quantum machine returns.
+    /// </summary>
+    // TODO: This interface should be migrated to the qsharp-runtime repository.
```
cgranade, line 17: It looks like this PR resolves this TODO?

src/Simulation/Core/IQuantumMachine.cs
```diff
@@ -0,0 +1,60 @@
...
+using System;
+using System.Threading.Tasks;
+using Microsoft.Quantum.Simulation.Core;
+
+namespace Microsoft.Quantum.Runtime
+{
+    /// <summary>
+    /// Interface that a quantum machine must implement.
+    ///
+    /// Type parameters:
```
cgranade, line 13: Would it make sense to convert these to `<typeparam>` elements?

src/Simulation/Core/IQuantumMachine.cs
```diff
@@ -0,0 +1,60 @@
...
+        /// </summary>
+        string Target { get; }
+
+        /// <summary>
+        /// Executes a Q# program.
+        /// Submits a job to execute it and continuously checks whether it has been completed.
+        /// Once its execution completes, returns its output.
+        ///
+        /// Type parameters:
+        ///     - QInput: Type of input the quantum program receives.
```
cgranade, line 39: Similarly here, typeparam elements may be helpful.

src/Simulation/Core/IQuantumMachine.cs
```diff
@@ -0,0 +1,52 @@
...
+using Microsoft.Quantum.Simulation.Core;
+
+namespace Microsoft.Quantum.Runtime
+{
+    /// <summary>
+    /// Interface that a quantum machine must implement.
+    /// </summary>
+    /// <typeparam name="Job">Type of job that handles the execution of a program in the quantum machine.</typeparam>
+    /// <typeparam name="QRawResult">Type of result the quantum machine returns.</typeparam>
+    public interface IQuantumMachine<Job, QRawResult>
```
cesarzc, line 15: It might be worth it to conform to the type parameters name convention in C# and change the type parameter names to start with T instead of Q.

PR #169, opened 2020-04-08
Generate command-line program for Q# projects with an entry point

samarsha:
This is the "standalone executables" feature which allows Q# applications to run without a C# driver in the project.

TODO:

- [x] Use entry point documentation for help text
- [x] Support non-standard Q# types: `Range`, `BigInt`, `Pauli`, `Result`, and `Unit`
- [x] Combine `simulate` and `resources` commands
- [x] Handle parameters with the same name as a standard option
- [x] Support custom simulators
- [x] Support a default simulator in the project configuration
- [x] Use option name constants added by microsoft/qsharp-compiler#399
- [x] Add unit tests
- [x] Add project test


samarsha:
All of the planned features are implemented. I still need to write the unit tests, and some details of the command-line interface are still being finalized. So I'll leave it as a draft for now, but early reviews are welcome. :)

src/Simulation/CsharpGeneration/RewriteStep.fs
```diff
@@ -54,7 +54,12 @@ type Emitter() =
                 let content = SimulationCode.loadedViaTestNames source context
                 if content <> null then CompilationLoader.GeneratedFile(source, dir, ".dll.g.cs", content) |> ignore
 
+            match Seq.tryExactlyOne compilation.EntryPoints with
+            | Some entryPoint ->
+                let callable = context.allCallables.[entryPoint]
+                let content = EntryPoint.generate context callable
+                CompilationLoader.GeneratedFile(callable.SourceFile, dir, ".EntryPoint.g.cs", content) |> ignore
+            | None -> ()
```
bettinaheim, line 62: Do you want to generate a diagnostic if more than one entry point exists? You should at least return false if several entry point exist.

src/Simulation/CsharpGeneration/Resources/EntryPointDriver.cs
```diff
@@ -0,0 +1,355 @@
...
+        /// <param name="name">The name of the custom simulator.</param>
+        private static void DisplayCustomSimulatorError(string name)
+        {
+            var originalForeground = Console.ForegroundColor;
+            Console.ForegroundColor = ConsoleColor.Red;
+            Console.Error.WriteLine($"The simulator '{name}' could not be found.");
+            Console.ForegroundColor = originalForeground;
+            Console.Error.WriteLine();
+            Console.Error.WriteLine(
+                $"If '{name}' is a custom simulator, it must be set in the DefaultSimulator project property:");
```
bettinaheim, line 252: I take it you went with disallowing defining execution on a custom simulator via command line? (works for me, just double checking)

src/Simulation/CsharpGeneration/Resources/EntryPointDriver.cs
```diff
@@ -0,0 +1,355 @@
...
+        /// <param name="entryPoint">The entry point adapter.</param>
+        /// <param name="createSimulator">A function that creates an instance of the simulator to use.</param>
+        private static async Task DisplayEntryPointResult(
+            @EntryPointAdapter entryPoint, Func<IOperationFactory> createSimulator)
+        {
+            var simulator = createSimulator();
+            try
+            {
+                var value = await entryPoint.Run(simulator);
+#pragma warning disable CS0184
```
bettinaheim, line 147: What warning is that?

src/Simulation/CsharpGeneration/Resources/EntryPointDriver.cs
```diff
@@ -0,0 +1,355 @@
...
+        /// Simulates the entry point.
+        /// </summary>
+        /// <param name="entryPoint">The entry point adapter.</param>
+        /// <param name="simulator">The simulator to use.</param>
+        /// <returns>The exit code.</returns>
+        private static async Task<int> Simulate(@EntryPointAdapter entryPoint, string simulator)
+        {
+            switch (simulator)
+            {
+                case "ResourcesEstimator":
```
bettinaheim, line 112: Instead of strings, I suggest to use the AssemblyConstants.ResourcesEstimator etc.

src/Simulation/CsharpGeneration/Resources/EntryPointDriver.cs
```diff
@@ -0,0 +1,355 @@
...
+
+        /// <summary>
+        /// Runs the entry point.
+        /// </summary>
+        /// <param name="args">The command-line arguments.</param>
+        /// <returns>The exit code.</returns>
+        private static async Task<int> Main(string[] args)
+        {
+            var simulate = new Command("simulate", "(default) Run the program using a local simulator.");
+            // TODO: Replace the option aliases with constants defined in Microsoft.Quantum.QsCompiler.ReservedKeywords.
```
bettinaheim, line 85: Coming. :)

src/Simulation/CsharpGeneration/Resources/EntryPointDriver.cs
samarsha: Yes. I did not want to go to the trouble of updating the reflection loading to use a separate option for the assembly name. :)

src/Simulation/CsharpGeneration/Resources/EntryPointDriver.cs
samarsha: [CS0184](https://docs.microsoft.com/en-us/dotnet/csharp/misc/cs0184):
> The expression can never be true because the variable you are testing is neither declared as type nor derived from type.

The return type of the entry point is known at compile time, so the C# compiler knows when `value is QVoid` is always false. But it's convenient to write this if statement so I don't need to generate different code depending on the return type.

src/Simulation/CsharpGeneration/RewriteStep.fs
samarsha: I can do that. Is the thinking that even though the Q# compiler enforces only one entry point, it is still technically possible to give the transformation a syntax tree with more than one?

samarsha: This is now a precondition for the transformation. I did not change the return value of `Transformation` because it assumes that the precondition holds.

src/Simulation/CsharpGeneration/Resources/EntryPointDriver.cs
bettinaheim: Thanks for the explanation!

src/Simulation/CsharpGeneration/EntryPoint.fs
```diff
@@ -0,0 +1,205 @@
...
+                            QsharpType = variable.Type
+                            CsharpTypeName = SimulationCode.roslynTypeName context variable.Type
+                            Description = ParameterDescription doc name.Value }
+        | InvalidName, _ -> Seq.empty
+    | QsTuple items -> items |> Seq.map (parameters context doc) |> Seq.concat
+
+/// The custom argument handler for the given Q# type.
+let private argumentHandler =
+    function
+    | UnitType -> Some "UnitArgumentHandler"
```
bettinaheim, line 71: I saw that for entry points that take Unit no argument is required. Nice!

src/Simulation/CsharpGeneration.Tests/Circuits/EntryPointTests.qs
```diff
@@ -0,0 +1,247 @@
...
+    operation AcceptString(s : String) : String {
+        return s;
+    }
+}
+
+// ---
+
+namespace Microsoft.Quantum.CsharpGeneration.Testing.EntryPoint {
+    @EntryPoint()
+    operation AcceptStringArray(xs : String[]) : String[] {
```
bettinaheim, line 111: Does something like a Range[] as argument work fine? Looking at the code I think it should, but wouldn't hurt to check.

samarsha: Thanks for bringing this up, I didn't consider it. No, it doesn't work, for the same reason that nested arrays don't work - since range arguments can be space-separated, it doesn't know when one range stops and the other one begins.

We could either require ranges to not have spaces (only allow `1..2..10`, not `1 2 10`), or disallow range arrays. I like the first option (since it's Q# syntax anyway).

src/Simulation/CsharpGeneration/EntryPoint.fs
samarsha: Yes, although that is not related to the `UnitArgumentHandler`. Command line options are created for each named argument in the entry point's argument tuple. Since `()` has no named items, there are no options. `UnitArgumentHandler` is for when a named item is of type `Unit`, such as `(foo : Unit)`, and in that case the argument is required. :)

src/Simulation/CsharpGeneration.Tests/Circuits/EntryPointTests.qs
samarsha: There is actually a larger issue that is preventing this from being fixed. Custom argument handlers are currently not used for arrays, so arrays of `BigInt`s and `Result`s are not working either.

src/Simulation/CsharpGeneration/EntryPoint.fs
bettinaheim: I know, I just randomly thought of it when I saw this. :)

src/Simulation/CsharpGeneration.Tests/Circuits/EntryPointTests.qs
bettinaheim: As discussed, that is fine. Sooner or later we should look into supporting this, but this is a good first version. Thanks!

PR #172, opened 2020-04-17
Alexva/qp custom gate

avasch01:
This change adds extensibility to IQuantumProcessor interface. Users can now add custom gates to the processors.


src/Simulation/Common/IQuantumProcessor.cs
```diff
@@ -391,6 +391,41 @@ public interface IQuantumProcessor
         /// <param name="qubits">Register to apply the exponent to.</param>
         void ControlledExpFrac(IQArray<Qubit> controls, IQArray<Pauli> paulis, long numerator, long power, IQArray<Qubit> qubits);
 
+        /// <summary>
+        /// Used as extensibility mechanism to add custom gates to IQuantumProcessor.
```
cgranade, line 395: The term "gate" can be ambiguous, as it is commonly used to refer to quantum operations in general, specifically intrinsic quantum operations, and to unitary operators that represent the action of quantum operations. Given that ambiguity, I would strongly suggest a name that communicates the intent and purpose of this extensibility mechanism; e.g.: `ApplyCustomIntrinsicOperation`.

src/Simulation/Common/IQuantumProcessor.cs
```diff
@@ -391,6 +391,41 @@ public interface IQuantumProcessor
...
+        /// <param name="controls">Array of control qubits for this operation.</param>
+        /// <param name="qubits">Array of qubits to apply the gate to. Their meaning is specific to the custom gate.</param>
+        /// <param name="paulis">Array of single-qubit Pauli values. Their meaning is specific to the custom gate.</param>
+        /// <param name="longs">Array of long parameters. Their meaning is specific to the custom gate.</param>
+        /// <param name="doubles">Array of double parameters. Their meaning is specific to the custom gate.</param>
+        /// <param name="bools">Array of bool parameters. Their meaning is specific to the custom gate.</param>
+        /// <param name="results">Array of Result parameters. Their meaning is specific to the custom gate.</param>
+        /// <param name="strings">Array of string parameters. Their meaning is specific to the custom gate.</param>
+        void ControlledGate(long gateId, IQArray<Qubit> controls, IQArray<Qubit> qubits, IQArray<Pauli> paulis, IQArray<long> longs, IQArray<double> doubles, IQArray<bool> bools, IQArray<Result> results, IQArray<string> strings);
```
cgranade, line 427: Why are there methods for applying the body and controlled body, but not the adjoint and controlled adjoint? Also, the same comment applies here about providing a default implementation to avoid breaking changes to the interface.

src/Simulation/Simulators/QuantumProcessor/Gate.cs
```diff
@@ -0,0 +1,43 @@
...
+
+using System;
+using Microsoft.Quantum.Simulation.Core;
+
+namespace Microsoft.Quantum.Simulation.QuantumProcessor
+{
+    public partial class QuantumProcessorDispatcher
+    {
+
+        public class QuantumProcessorApplyGate : Extensions.Gate
```
cgranade, line 12: As a public type, this should have an API documentation comment, even if that comment is just an inheritdoc.

src/Simulation/Simulators/QuantumProcessor/Gate.qs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.Quantum.Simulation.QuantumProcessor.Extensions
+{
+    open Microsoft.Quantum.Intrinsic;
+
+    operation Gate(gateId : Int, qubits : Qubit[], paulis : Pauli[], longs : Int[], doubles : Double[], bools : Bool[], results : Result[], strings : String[]) : Unit is Ctl {
```
cgranade, line 8: Is this intended to be a public Q# API? If so, then it needs to adhere to Q# API design principles and to the Q# style guide.

Design principle questions and comments:
- What is the usecase for this new API, and why is this the right API to address that usecase? In particular, I am concerned that, as written, this presents internal details of the simulation runtime as a public Q# API rather than addressing a usecase external to the runtime.
- Q# namespaces should follow the `Publisher.Product.DomainArea` format, optionally with a subnamespace for focused tasks. Following that guidance, the domain area here is "simulation," but that contradicts the existing meaning of the `Microsoft.Quantum.Simulation` namespace, where the term "simulation" refers to operations and functions that support quantum simulation, not classical simulation of quantum programs.
- Inputs should be ordered following likely use of partial application, defaulting to qubits occuring last in input order.

Style guide comments:
- Since this is an operation, it should be named as a verb (e.g.: `ApplyCustomIntrinsicOperation`)
- This operation should have an API documentation comment explaining its usage, inputs, and ideally providing examples.

swernli:
@avasch01 I think I'd like to better understand the intended usage of this extension before I'm able to evaluate it. Would this be used for simulators, physical processors, or both? Since someone looking to take advantage of this would have to implement the processor backend for each of the gates supported via this extension, couldn't they also achieve this kind of extensibility by adding methods to their specific processor implementation and matching Q# and C# to a decomposition package for that implementation? That approach would certainly have a very different developer experience during authoring the extension, but it would allow them to make the extension operation signature match exactly the gate they wanted to expose, which seems like it could provide a better experience for developers using that operation in their code.

src/Simulation/Common/IQuantumProcessor.cs
avasch01: Valid concern! I deliberated on it for quite some time and decided for controlled but against adjoint. We do not provide adjoint for all gates, so this is not expected. There are two ways to implement adjoint of such custom gate: One is to add Adjoint as another interface function (actually two: Adjoint and AdjointControled). Another way is to treat it as a gate with a different id. The second one looked appropriate and easier. The drawback of the first one is addition of two functions to the API, which could be avoided. Controlled is provided separately because it adds another parameter - an array of control qubits. 
So, I could go either way, but I think current one is better.

---
In reply to: [410524847](https://github.com/microsoft/qsharp-runtime/pull/172#discussion_r410524847) [](ancestors = 410524847)

src/Simulation/Common/IQuantumProcessor.cs
avasch01: Makes sense. I was thinking to make it more limited, but switching gate to operation is also reasonable. I was thinking about it myself, so if you don't like "Gate" I can certainly understand.

---
In reply to: [410524692](https://github.com/microsoft/qsharp-runtime/pull/172#discussion_r410524692) [](ancestors = 410524692)

src/Simulation/Common/IQuantumProcessor.cs
```diff
@@ -391,6 +391,41 @@ public interface IQuantumProcessor
...
+        /// </remarks>
+        /// <param name="gateId">The id of the custom gate. It is up to the custom gate implementor to define the meaning.</param>
+        /// <param name="qubits">Array of qubits to apply the gate to. Their meaning is specific to the custom gate.</param>
+        /// <param name="paulis">Array of single-qubit Pauli values. Their meaning is specific to the custom gate.</param>
+        /// <param name="longs">Array of long parameters. Their meaning is specific to the custom gate.</param>
+        /// <param name="doubles">Array of double parameters. Their meaning is specific to the custom gate.</param>
+        /// <param name="bools">Array of bool parameters. Their meaning is specific to the custom gate.</param>
+        /// <param name="results">Array of Result parameters. Their meaning is specific to the custom gate.</param>
+        /// <param name="strings">Array of string parameters. Their meaning is specific to the custom gate.</param>
+        void Gate(long gateId, IQArray<Qubit> qubits, IQArray<Pauli> paulis, IQArray<long> longs, IQArray<double> doubles, IQArray<bool> bools, IQArray<Result> results, IQArray<string> strings);
```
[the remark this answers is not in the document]
avasch01: A default implementation is provided in a QuantumProcessorBase class. While formally this is a breaking change, none of known implementations will be broken on this, since all of them derive from the QuantumProcessorBase class. In general, I would strongly recommend against implementations in interfaces. And in this case it can be easily avoided. I bet nobody would notice (be broken) on this, and we are not yet at version 1.0, so a breaking change like this is acceptable,

---
In reply to: [410516089](https://github.com/microsoft/qsharp-runtime/pull/172#discussion_r410516089) [](ancestors = 410516089)

src/Simulation/Simulators/QuantumProcessor/Gate.cs
avasch01: Sure.

---
In reply to: [410524961](https://github.com/microsoft/qsharp-runtime/pull/172#discussion_r410524961) [](ancestors = 410524961)

avasch01:
This provides an easy extension mechanism for people who implement IQuantumProcessor. The entire IQuantumProcessor has been designed to ease the development experience, our customers were scared at the old mechanism and asked for this easier one. It can be used by anybody who is using IQuantumProcessor, whether for hardware or simulator. The immediate intent is to enable custom gates for simulators, layout, printers and other software components. But hardware is also welcome to use it (or not to use it) if desired. it is optional and enables research and some more efficient implementations. One could for example implement adding an integer to a quantum register and let a specific simulator to do it as efficient as possible, depending of the internals of the simulator. You make a good point about creating Q# operation with an exact signature - that would be much nicer. Well, we did think about it. It is easy to add a Q# wrapper around this CustomOperation call with a nice name and only the required parameters and expose it on top of CustomOperation. Flexibility, and ease of development.

---
In reply to: [616209141](https://github.com/microsoft/qsharp-runtime/pull/172#issuecomment-616209141) [](ancestors = 616209141)

avasch01:
Please see Stephan's comment below.

---
In reply to: [395826946](https://github.com/microsoft/qsharp-runtime/pull/172#issuecomment-395826946) [](ancestors = 395826946)

swernli:
I see what you mean... this design allows for experimentation and extensibility in parallel or in conjunction with operations that have gate specific signatures. That does maximize the flexibility for developers!

cgranade:
> This provides an easy extension mechanism for people who implement IQuantumProcessor. The entire IQuantumProcessor has been designed to ease the development experience, our customers were scared at the old mechanism and asked for this easier one.

Could you please forward that feedback to me?
It would be good to make sure our existing features meet customer needs before adding a new extensibility mechanism that duplicates and overlaps with existing functionality.



> It can be used by anybody who is using IQuantumProcessor, whether for hardware or simulator.



That argues against this being a feature in the simulation runtime, then.
In particular, `IQuantumProcessor` is very much a simulation runtime feature, such that I wouldn't want it to also take on responsibilities for hardware.



> The immediate intent is to enable custom gates for simulators, layout, printers and other software components. But hardware is also welcome to use it (or not to use it) if desired. it is optional and enables research and some more efficient implementations. One could for example implement adding an integer to a quantum register and let a specific simulator to do it as efficient as possible, depending of the internals of the simulator.
> You make a good point about creating Q# operation with an exact signature - that would be much nicer. Well, we did think about it. It is easy to add a Q# wrapper around this CustomOperation call with a nice name and only the required parameters and expose it on top of CustomOperation. Flexibility, and ease of development.



This is exactly the scenario and the workflow enabled by writing a Q# operation with either a `body intrinsic;` statement or a pure-Q# fallback, and then subclassing with `Native`. That's worked well in the past in cases such as `EstimateFrequencyA`. We may want to document that process better, of course, so that end-users can make use of that feature as is appropriate, but what that gives is a _type-safe_ and portable means of providing custom extensibility for the simulation runtime.

bettinaheim:
> I see what you mean... this design allows for experimentation and extensibility in parallel or in conjunction with operations that have gate specific signatures. That does maximize the flexibility for developers! #Resolved

So to follow up on this: 
@avasch01, Do I understand correctly that the idea is to have as flexible a definition of a general purpose gate in the C# interface that each simulator is expected to support with the idea that that gate will never directly be exposed as intrinsic in Q#, but rather that people would define intrinsic operations in Q# with more specific signatures? But then people would also need to provide C# implementations for those? I still don't understand what people don't have to do if we add this gate to the interface that they otherwise would, given that also the base class has no implementation for it.

cgranade:
As a quick follow-up, I talked with @vadym-kl a bit offline to understand the context for this PR a bit better. Given that, I'd like to suggest making this a formal feature request so that we can iterate on designs that avoid exposing public Q# APIs that aren't type-safe. One idea that came to mind, for instance, would be for `QuantumProcessorDispatcher` to look for attributes on its `IQuantumProcessor` argument for methods that can be used to implement custom Q# intrinsics without needing to use inner classes.

Concretely, something like the following may work:
```C#
// CustomSimulator.cs
namespace Customer.Quantum.Simulator
{
    public class CustomSimulator : QuantumProcessorBase
    {
        // implement IQuantumProcessor methods

        [CustomIntrinsic(typeof(Customer.Quantum.Intrinsic.ApplyCustomRotation), Specializations.Body)]
        public QVoid ApplyCustomRotation(double angle, IQArray<Qubit> targets)
        {
            // ...
            return QVoid.instance;
        }

        [CustomIntrinsic(typeof(Customer.Quantum.Intrinsic.ApplyCustomRotation), Specializations.Controlled)]
        public QVoid ApplyCustomRotation(double angle, IQArray<Qubit> targets)
        {
            // ...
            return QVoid.instance;
        }
    }
}
```

```Q#
// Intrinsics.qs
namespace Customer.Quantum.Intrinsic {
    operation ApplyCustomRotation(angle : Double, targets : Qubit[]) : Unit {
        body (...) {
            // Pure Q# decomposition here as a fall-back for portability.
        }
        adjoint self;
        controlled (ctrls, (...)) {
             // Pure Q# decomposition here.
        }
        controlled adjoint auto;
    }
}
```

src/Simulation/Common/IQuantumProcessor.cs
```diff
@@ -391,6 +391,41 @@ public interface IQuantumProcessor
...
+        /// <param name="qubits">Array of qubits to apply the operation to. Their meaning is specific to the custom operation.</param>
+        /// <param name="paulis">Array of single-qubit Pauli values. Their meaning is specific to the custom operation.</param>
+        /// <param name="longs">Array of long parameters. Their meaning is specific to the custom operation.</param>
+        /// <param name="doubles">Array of double parameters. Their meaning is specific to the custom operation.</param>
+        /// <param name="bools">Array of bool parameters. Their meaning is specific to the custom operation.</param>
+        /// <param name="results">Array of Result parameters. Their meaning is specific to the custom operation.</param>
+        /// <param name="strings">Array of string parameters. Their meaning is specific to the custom operation.</param>
+        void CustomOperation(long operationId, IQArray<Qubit> qubits, IQArray<Pauli> paulis, IQArray<long> longs, IQArray<double> doubles, IQArray<bool> bools, IQArray<Result> results, IQArray<string> strings);
```
viktorveis, line 409: >CustomOperation [](start = 13, length = 15)

What are the first use cases for this?

viktorveis, line 409: >CustomOperation [](start = 13, length = 15)

How will we decide which gates are explicitly spelled in the interface and which go through the custom operation?

src/Simulation/Common/IQuantumProcessor.cs
```diff
@@ -391,6 +391,41 @@ public interface IQuantumProcessor
...
         void ControlledExpFrac(IQArray<Qubit> controls, IQArray<Pauli> paulis, long numerator, long power, IQArray<Qubit> qubits);
 
+        /// <summary>
+        /// Used as extensibility mechanism to add custom operations to IQuantumProcessor. 
+        /// </summary>
+        /// <remarks>
+        /// The custom operation may use only some of the parameters. Depending on the nature of the operation, all of them may be optional. 
+        /// A variety of parameter types are provided for flexibility.
+        /// </remarks>
+        /// <param name="operationId">The id of the custom operation. It is up to the custom operation implementor to define the meaning.</param>
```
viktorveis, line 401: >up to the custom operation implementor [](start = 76, length = 38)

Different implementations may use different IDs with the same meaning. This means that client code will only be able to run on a single quantum processor or a set of quantum processors with standardized the IDs. Have we evaluated long term implications on Q# ecosystem?

src/Simulation/Common/IQuantumProcessor.cs
avasch01: The immediate use cases are for table lookup for state preparation which is used in benchmarks for one of our projects, but will be used in any state preparation also. Another current use case is for testing reversible circuits for arithmetic functions, which are, for example, used in the same benchmarks.

---
In reply to: [413474648](https://github.com/microsoft/qsharp-runtime/pull/172#discussion_r413474648) [](ancestors = 413474648)

avasch01: The set of gates that are explicitly spelled out is fixed in the definition of Q# intrinsics. I believe this set was decided based on being a minimally reasonable set supporting a closure of quantum operations and supporting functionality. But anyway, it is decided and users cannot change these. Custom operation allows users to add any gates they deem necessary for them. Usually for more complex gates for more efficient execution, and also for custom supporting functionality that is supported only by their custom quantum processor. 

---
In reply to: [413475308](https://github.com/microsoft/qsharp-runtime/pull/172#discussion_r413475308) [](ancestors = 413475308)

src/Simulation/Common/IQuantumProcessor.cs
avasch01: Correct. Custom operation functionality requires user to be aware of which custom processor it is running against. it is users responsibility to ensure correct quantum processor. But user is empowered to do this.

---
In reply to: [413476401](https://github.com/microsoft/qsharp-runtime/pull/172#discussion_r413476401) [](ancestors = 413476401)

viktorveis: This is a very generic API. Unfortunately, it can fragment the ecosystem and promote writing code that is hard to read and maintain. An analogy in classical programming would be something like void *DoAnything(int id, void*). What other options have we considered to address this scenario?

---
In reply to: [414257458](https://github.com/microsoft/qsharp-runtime/pull/172#discussion_r414257458) [](ancestors = 414257458,413476401)

cgranade: @viktorveis: As detailed above, the existing emulation feature has been used well in the libraries to extend the simulation framework with new accelerated functionality (e.g.: `EstimateFrequencyA`). That works especially well when a pure-Q# fallback can be offered for portability across target machines. I definitely take the point that this extension mechanism could be documented better, or that alternative mechanisms could be provided (e.g.: the attribute-based design discussed above) in order to better enable advanced users looking to extend the simulation runtime, but I think that would be best discussed in the context of a formal feature request issue where the various alternatives can be compared and evaluated against our Q# API Design Principles, and against the need to maintain portability across simulators.

src/Simulation/Common/IQuantumProcessor.cs
bettinaheim: To make it more concrete, could you please give the Q# signature for these cases, and why these operations should be intrinsic for benchmarking opposed to being decompositions?

bettinaheim: Any custom operation can always be defined as intrinsic. The problem to solve is how to make it more user friendly to do that. There are several possibilities on how to achieve that, and we should discuss these.

bettinaheim:
Closing as per our simulator extensibility plans.

PR #174, opened 2020-04-22
adding an execution test

bettinaheim:


src/Simulation/Simulators.Tests/Tests.Microsoft.Quantum.Simulation.Simulators.csproj
```diff
@@ -24,7 +33,18 @@
...
+  <Target Name="PostBuild" AfterTargets="PostBuildEvent">
+    <PropertyGroup>
+      <_ExeDir>$(MSBuildThisFileDirectory)TestProjects\QsharpExe\built\</_ExeDir>
+    </PropertyGroup>
+    <ItemGroup>
+      <_ExeFiles Include="$(_ExeDir)*" />
+    </ItemGroup>
+    <MakeDir Directories="$(OutputPath)TestExe" />    
+    <Copy SourceFiles="@(_ExeFiles)" DestinationFolder="$(OutputPath)TestExe" ContinueOnError="true" />
+  </Target>
```
samarsha, line 47: So the Q# executable project copies its binaries from the platform-specific build folder to a known folder, and the test project copies those binaries from the known folder to its own build folder? I see what you mean about this solution not being the nicest. I don't know a better way, either, though. :)

src/Simulation/Simulators.Tests/CoreTests.cs
```diff
@@ -25,10 +26,27 @@ public CoreTests(ITestOutputHelper output)
...
+        [Fact]
+        public void BasicExecution()
+        {
+            var asmPath = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
+            var exe = Path.Combine(asmPath, "TestExe", "QsharpExe.exe");
+
+            ProcessRunner.Run(exe, "", out StringBuilder output, out var _, out int exitCode, out Exception ex);
+            Assert.Null(ex);
+            Assert.Equal(1, exitCode);
+            Assert.Contains("NotImplementedException", output.ToString());
```
samarsha, line 38: I'm surprised that the exception message goes to standard out. Looking into this more, it seems like there are two exception messages, one that goes to standard out and one that goes to standard error. The one in standard out comes from `SimulatorBase.WriteStackTraceToLog` - all log messages seem to go to standard out.

This is not a big deal and you don't need to change anything here (although I think standard error would work here also, since it goes to both), but just wanted to mention it.

src/Simulation/Simulators.Tests/CoreTests.cs
```diff
@@ -44,18 +62,18 @@ public void Borrowing()
...
-                var testOneAdjoint = new System.Action<int, (int, Qubit)>((callsCount, info) => testOne(callsCount, OperationFunctor.Adjoint, info));
-                var testOneCtrl = new System.Action<int, (int, Qubit)>((callsCount, info) => testOne(callsCount, OperationFunctor.Controlled, info));
-                var testOneCtrlAdj = new System.Action<int, (int, Qubit)>((callsCount, info) => testOne(callsCount, OperationFunctor.ControlledAdjoint, info));
+                var testOneBody = new System.Action<int, (int, Qubit)>((int callsCount, (int, Qubit) info) => testOne(callsCount, OperationFunctor.Body, info));
+                var testOneAdjoint = new System.Action<int, (int, Qubit)>((int callsCount, (int, Qubit) info) => testOne(callsCount, OperationFunctor.Adjoint, info));
+                var testOneCtrl = new System.Action<int, (int, Qubit)>((int callsCount, (int, Qubit) info) => testOne(callsCount, OperationFunctor.Controlled, info));
+                var testOneCtrlAdj = new System.Action<int, (int, Qubit)>((int callsCount, (int, Qubit) info) => testOne(callsCount, OperationFunctor.ControlledAdjoint, info));
```
samarsha, line 76: Why are the explicit lambda parameter types needed now?

src/Simulation/Simulators.Tests/TestProjects/QsharpExe/Main.qs
```diff
@@ -0,0 +1,16 @@
...
+
+    @EntryPoint()
+    operation Main () : Result {
+        using (q = Qubit()) {
+            H(q);
+            let res = M(q);
+            if (res == One) {
+                X(q);
+            }
+            return res;
```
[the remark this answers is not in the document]
bettinaheim: I don't have access to the library here. MResetZ is in the standard library I believe.

src/Simulation/Simulators.Tests/TestProjects/QsharpExe/QsharpExe.csproj
```diff
@@ -0,0 +1,40 @@
...
+    </PropertyGroup>
+    <ItemGroup>
+      <FilesToRemove Include="$(_ExeDir)**" />
+    </ItemGroup>
+    <Delete Files="@(_ExeDir)" />
+    <MakeDir Directories="$(_ExeDir)" />
+    <ItemGroup>
+      <_ExeFiles Include="$(OutputPath)*" />
+    </ItemGroup>
+    <Copy SourceFiles="@(_ExeFiles)" DestinationFolder="$(_ExeDir)" ContinueOnError="true" />
```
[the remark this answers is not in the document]
bettinaheim: No, just that the copying is for the test, and it seemed that the build should succeed in either case. But I am also happy to set it to false.

src/Simulation/Simulators.Tests/Tests.Microsoft.Quantum.Simulation.Simulators.csproj
bettinaheim: Yeah... It's such a silly problem - I just need to somehow know where the executable is that I'm calling from the test, and this actually seemed like the best option...

src/Simulation/Simulators.Tests/CoreTests.cs
bettinaheim: Interesting. Either one works for me - I can also check the standard error.

src/Simulation/Simulators.Tests/CoreTests.cs
bettinaheim: Hm, I didn't add them. Maybe I picked them up from another file version or one of the tools did? I'll remove them again.

PR #175, opened 2020-04-23
Support arrays of BigInt, Range, Result, and Unit in entry point options

samarsha:
* Refactor the argument parsers so they support both single-valued and many-valued arguments (arrays). This was done by primarily supporting many-valued arguments (the more general case), and then calling `Single()` when the argument is single-valued.
* Pass the argument-parsing functions directly to the `Option<T>` constructors instead of creating `Argument<T>` instances and setting the `Option<T>.Argument` property. These should be functionally equivalent, it is just recommended to do it this way in System.CommandLine.
* Various cleanup:
   - Split `EntryPointDriver.cs` into three files.
   - Rename `Result<T>` to `Validation<T>` to avoid confusion with Q# `Result`.


src/Simulation/CsharpGeneration.Tests/Circuits/EntryPointTests.qs
```diff
@@ -117,8 +126,35 @@ namespace Microsoft.Quantum.CsharpGeneration.Testing.EntryPoint {
...
+    operation AcceptRangeArray(rs : Range[]) : Range[] {
+        return rs;
+    }
+}
+
+// ---
+
+namespace Microsoft.Quantum.CsharpGeneration.Testing.EntryPoint {
+    @EntryPoint()
+    operation AcceptResultArray(rs : Result[]) : Result[] {
```
bettinaheim, line 147: Pauli[] is another one to add I believe.

src/Simulation/CsharpGeneration/Resources/EntryPoint/Driver.cs
```diff
@@ -0,0 +1,206 @@
+﻿namespace @Namespace
```
bettinaheim, line 1: I take it this file hasn't changed, has it?

src/Simulation/CsharpGeneration/EntryPoint.fs
```diff
@@ -45,36 +45,52 @@ let private constantsClass =
...
-    | BigInt -> Some "BigIntArgumentHandler"
-    | Range -> Some "RangeArgumentHandler"
-    | _ -> None
-    >> Option.map (fun handler -> ``ident`` "Driver" <|.|> ``ident`` handler)
+/// The argument parser for the Q# type.
+let private argumentParser qsType =
+    let rec valueParser = function
+        | ArrayType (itemType : ResolvedType) -> valueParser itemType.Resolution
+        | BigInt -> Some "TryParseBigInteger"
+        | Range -> Some "TryParseQRange"
```
bettinaheim, line 72: Are you missing Pauli here?

src/Simulation/CsharpGeneration/Resources/EntryPoint/Driver.cs
samarsha: Yes, no changes to Driver.cs besides moving things to the other files. Also, no change to Validation.cs besides renaming `Result<T>` to `Validation<T>`.

src/Simulation/CsharpGeneration/EntryPoint.fs
samarsha: `Pauli` is an enum, so it is natively supported by System.CommandLine.

PR #176, opened 2020-04-24
Replacing some hardcoded strings

bettinaheim:


src/Simulation/CsharpGeneration/EntryPoint.fs
```diff
@@ -40,9 +40,9 @@ let private constantsClass =
             [readonlyProperty "SimulatorOptions" "System.Collections.Generic.IEnumerable<string>"
                 (``new array`` (Some "") [``literal`` ("--" + fst CommandLineArguments.SimulatorOption)
                                           ``literal`` ("-" + snd CommandLineArguments.SimulatorOption)])
-             constant "QuantumSimulator" "string" (``literal`` AssemblyConstants.QuantumSimulator)
-             constant "ToffoliSimulator" "string" (``literal`` AssemblyConstants.ToffoliSimulator)
-             constant "ResourcesEstimator" "string" (``literal`` AssemblyConstants.ResourcesEstimator)]
+             constant AssemblyConstants.QuantumSimulator "string" (``literal`` AssemblyConstants.QuantumSimulator)
+             constant AssemblyConstants.ToffoliSimulator "string" (``literal`` AssemblyConstants.ToffoliSimulator)
+             constant AssemblyConstants.ResourcesEstimator "string" (``literal`` AssemblyConstants.ResourcesEstimator)]
```
samarsha, line 45: This is not a valid substitution. The strings here refer to the field names used at [EntryPointDriver.cs:94](https://github.com/microsoft/qsharp-runtime/blob/master/src/Simulation/CsharpGeneration/Resources/EntryPointDriver.cs#L94). If the value of the assembly constants are changed, the C# driver will not compile.

PR #180, opened 2020-04-30
Create a new project for the standalone driver

samarsha:
* The new driver project is a library. The generated code includes a Main method that calls the non-generated driver.
* Moved a lot of code that used to be generated into the non-generated code, which required a lot of refactoring.
* @cgranade, I refactored FindNuspecReferences.ps1 for the CsharpGeneration project so it was easier to add manual dependencies, but didn't unify it with the other FindNuspecReferences.ps1 in the Simulators project.

### DIFF
--- a/src/Simulation/CsharpGeneration.Tests/EntryPointTests.fs
+++ b/src/Simulation/CsharpGeneration.Tests/EntryPointTests.fs
@@ -67,9 +67,9 @@ let private generateCsharp defaultSimulator (syntaxTree : QsNamespace seq, entry
         | None -> ImmutableDictionary.Empty
     let context = CodegenContext.Create (syntaxTree, assemblyConstants)
     let entryPoint = context.allCallables.[Seq.exactlyOne entryPoints]
-    List.concat [
-        [SimulationCode.generate (NonNullable<_>.New testFile) context]
-        EntryPoint.generate context entryPoint |> List.map snd
+    [
+        SimulationCode.generate (NonNullable<_>.New testFile) context
+        EntryPoint.generate context entryPoint
     ]
 
 /// The full path to a referenced assembly given its short name.
--- a/src/Simulation/CsharpGeneration.Tests/EntryPointTests.fs
+++ b/src/Simulation/CsharpGeneration.Tests/EntryPointTests.fs
@@ -94,7 +94,7 @@ let private compileCsharp (sources : string seq) =
             "System.Runtime"
             "System.Runtime.Extensions"
             "System.Runtime.Numerics"
-            "EntryPointDriver"
+            "Microsoft.Quantum.CsharpGeneration.EntryPointDriver"
             "Microsoft.Quantum.QSharp.Core"
             "Microsoft.Quantum.QsDataStructures"
             "Microsoft.Quantum.Runtime.Core"
--- a/src/Simulation/CsharpGeneration.Tests/EntryPointTests.fs
+++ b/src/Simulation/CsharpGeneration.Tests/EntryPointTests.fs
@@ -122,7 +122,7 @@ let private testAssembly testNum defaultSimulator =
 /// Runs the entry point in the assembly with the given command-line arguments, and returns the output, errors, and exit
 /// code.
 let private run (assembly : Assembly) (args : string[]) =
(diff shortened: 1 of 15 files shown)


src/Simulation/EntryPointDriver/Driver.cs
```diff
@@ -0,0 +1,230 @@
...
+        {
+            if (IsAliasAvailable(entryPoint, alias))
+            {
+                return value;
+            }
+            else
+            {
+                var originalForeground = Console.ForegroundColor;
+                Console.ForegroundColor = ConsoleColor.Yellow;
+                Console.Error.WriteLine($"Warning: Option {alias} is overriden by an entry point parameter name.");
```
bettinaheim, line 180: Nice! I like the warning.
```suggestion
                Console.Error.WriteLine($"Warning: Option {alias} is overridden by an entry point parameter name.");
```

src/Simulation/EntryPointDriver/Driver.cs
```diff
@@ -0,0 +1,230 @@
...
+                Console.Error.WriteLine($"         Using default value {defaultValue}.");
```
bettinaheim, line 181: Hm, whitespace in my experience tends to never work out. Maybe it would make more sense to have it formatted as 

Warning:
Option … is overridden...
Using default value ...

src/Simulation/CsharpGeneration/FindNuspecReferences.ps1
```diff
@@ -19,53 +19,52 @@
 # nuget is tracking this problem at: https://github.com/NuGet/Home/issues/4491
 ########################################
```
bettinaheim, line 20: Did you double check the content of the CsharpGeneration NuGet package and compare it against the released one to make sure everything is where it's supposed to be?

src/Simulation/CsharpGeneration.Tests/EntryPointTests.fs
```diff
@@ -212,10 +214,10 @@ let ``Accepts Double`` () =
 [<Fact>]
 let ``Accepts Bool`` () =
     let given = test 7
-    given ["-b"] |> yields "True"
     given ["-b"; "false"] |> yields "False"
     given ["-b"; "true"] |> yields "True"
     given ["-b"; "one"] |> fails
+    given ["-b"] |> fails
```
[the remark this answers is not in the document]
samarsha: Yes - to clarify, it's always been required to include the option `-b`, which meant the same thing as `-b true`, but now you have to type either `-b true` or `-b false`. While I was refactoring the parsers it ended up being simpler to require all options to have at least one argument, and `bool` was the only exception.

src/Simulation/EntryPointDriver/Driver.cs
samarsha: Could you clarify what the problems with whitespace are? Console output should reliably support whitespace formatting - e.g., it's needed for command-line programs that display tables, etc.

src/Simulation/CsharpGeneration/FindNuspecReferences.ps1
samarsha: Seems like everything is the same except for the DLLs and dependencies I added.

src/Simulation/CsharpGeneration.Tests/EntryPointTests.fs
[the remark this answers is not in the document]
bettinaheim: I like that Bool is now the same as everything else.

src/Simulation/EntryPointDriver/Driver.cs
bettinaheim: I suppose I was thinking more along the lines when the output is redirected. But then maybe that's not something to worry about.

samarsha: I'm leaning now towards putting it all on one line, what do you think? (The console window will wrap it if it's too long.)

```
Warning: Option {alias} is overriden by an entry point parameter name. Using default value {defaultValue}.
```

samarsha: We chatted offline, going with all on one line now.

PR #184, opened 2020-04-30
Add "submit" command to the standalone executable using a mock IQuantumMachine

samarsha:
This is a bare-bones implementation of the `submit` command. It doesn't support any options, and the `IQuantumMachine` that I am using just runs the entry point on the `QuantumSimulator`. But now everything should be ready to start adding support for actual `IQuantumMachine`s.

This depends on #180, so targeting that branch for an accurate diff, but after that's merged I'll un-draft this and retarget to the feature branch.

### DIFF
--- a/src/Simulation/CsharpGeneration.Tests/Circuits/EntryPointTests.qs
+++ b/src/Simulation/CsharpGeneration.Tests/Circuits/EntryPointTests.qs
@@ -192,6 +192,46 @@ namespace Microsoft.Quantum.CsharpGeneration.Testing.EntryPoint {
 // ---
 
 //
+// Tuples
+//
+
+namespace Microsoft.Quantum.CsharpGeneration.Testing.EntryPoint {
+    @EntryPoint()
+    operation RedundantOneTuple((x : Int)) : String {
+        return $"{x}";
+    }
+}
+
+// ---
+
+namespace Microsoft.Quantum.CsharpGeneration.Testing.EntryPoint {
+    @EntryPoint()
+    operation RedundantTwoTuple((x : Int, y : Int)) : String {
+        return $"{x} {y}";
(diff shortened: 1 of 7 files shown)


src/Simulation/CsharpGeneration.Tests/Circuits/EntryPointTests.qs
```diff
@@ -191,6 +191,46 @@ namespace Microsoft.Quantum.CsharpGeneration.Testing.EntryPoint {
 
 // ---
 
+//
+// Tuples
```
bettinaheim, line 195: Good thinking to add these!

src/Simulation/EntryPointDriver/Driver.cs
```diff
@@ -112,25 +101,37 @@ private static async Task<int> Simulate<T>(IEntryPoint<T> entryPoint, ParseResul
...
+        /// </summary>
+        /// <param name="parseResult">The command-line parsing result.</param>
+        private async Task Submit(ParseResult parseResult)
+        {
+            // TODO: Use an actual quantum machine.
+            var machine = new SimulatorMachine();
+            var output = await machine.ExecuteAsync(entryPoint.Info, entryPoint.CreateArgument(parseResult));
+            foreach (var (result, frequency) in output.Histogram)
+            {
+                Console.WriteLine($"{result} (frequency = {frequency})");
```
bettinaheim, line 120: I assume that this piece will still change (I think we had said the default would be to not show the entire histogram but instead the most frequent value?).

samarsha: Yes, this was just temporary since the method that gets the most frequent output is not available yet. Added a TODO.

PR #185, opened 2020-05-02
Removed type parameters from IQuantumMachine interface.

cesarzc:
This change removes type parameters from the IQuantumMachine interface and replaces them with the IQuantumMachineJob and the IQuantumMachineOutput interfaces.

### DIFF
--- a/src/Simulation/Core/IQuantumMachineJob.cs
+++ b/src/Simulation/Core/IQuantumMachineJob.cs
@@ -1,0 +1,53 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Quantum.Runtime
+{
+    /// <summary>
+    /// Interface to track a job submitted to a quantum machine.
+    /// </summary>
+    public interface IQuantumMachineJob
+    {
+        /// <summary>
+        /// Gets the ID of submitted job.
+        /// </summary>
+        string Id { get; }
+
+        /// <summary>
+        /// Gets whether job execution is in progress.
+        /// </summary>
+        bool InProgress { get; }
+
+        /// <summary>
+        /// Gets the status of the submitted job.
+        /// </summary>
+        string Status { get; }
+
+        /// <summary>
+        /// Gets whether the job execution completed successfully.
+        /// </summary>
+        bool Succeeded { get;  }
+
+        /// <summary>
+        /// Gets an URI to access the job.
+        /// </summary>
+        Uri Uri { get; }
+
+        /// <summary>
+        /// Cancels the job.
+        /// </summary>
+        /// <param name="cancellationToken">The cancellation token.</param>
+        Task CancelAsync(CancellationToken cancellationToken = default);
+
+        /// <summary>
+        /// Refreshes the state of the job.
+        /// </summary>
+        /// <param name="cancellationToken">The cancellation token.</param>
+        Task RefreshAsync(CancellationToken cancellationToken = default);
+
+    }
+}
--- a/src/Simulation/Core/IQuantumMachineJob.cs
+++ b/src/Simulation/Core/IQuantumMachineJob.cs
@@ -13,6 +13,11 @@ namespace Microsoft.Quantum.Runtime
     public interface IQuantumMachineJob
     {
         /// <summary>
+        /// Gets whether job execution failed.
+        /// </summary>
+        bool Failed { get; }
+
+        /// <summary>
         /// Gets the ID of submitted job.
         /// </summary>
         string Id { get; }
@@ -40,14 +45,13 @@ namespace Microsoft.Quantum.Runtime
         /// <summary>
         /// Cancels the job.
         /// </summary>
-        /// <param name="cancellationToken">The cancellation token.</param>
+        /// <param name="cancellationToken">The cancellation token for the cancel operation.</param>
         Task CancelAsync(CancellationToken cancellationToken = default);
 
         /// <summary>
         /// Refreshes the state of the job.
         /// </summary>
-        /// <param name="cancellationToken">The cancellation token.</param>
+        /// <param name="cancellationToken">The cancellation token for the refresh operation.</param>
         Task RefreshAsync(CancellationToken cancellationToken = default);
-
     }
 }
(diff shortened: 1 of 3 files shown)


src/Simulation/Core/IQuantumMachineOutput.cs
```diff
@@ -0,0 +1,42 @@
...
+    /// Interface to access the results of a program executed in a quantum machine.
+    /// </summary>
+    public interface IQuantumMachineOutput
+    {
+
+        /// <summary>
+        /// Gets an output probabilistically.
+        /// Calls to this method are not deterministic, each call can return a different output.
+        /// </summary>
+        object GetOutput();
```
samarsha, line 18: It seems like this could be accomplished by randomly sampling `Histogram`. Does it need to be a separate method in the interface?

src/Simulation/Core/IQuantumMachineOutput.cs
```diff
@@ -0,0 +1,42 @@
...
+
+        /// <summary>
+        /// Gets the most frequent output.
+        /// If multiple outputs have the same frequency, it gets the first one in ascending order.
+        /// </summary>
+        object GetMostFrequentOutput();
```
samarsha, line 24: Similarly, this could also be obtained by using the `Histogram` property, so I am not sure that it should be a separate method.

src/Simulation/Core/IQuantumMachineOutput.cs
```diff
@@ -0,0 +1,42 @@
...
+
+        /// <summary>
+        /// Gets a histogram of outputs.
+        /// The key is the output and the value is its frequency.
+        /// </summary>
+        IDictionary<object, double> Histogram { get; }
```
samarsha, line 30: To be more precise and avoid floating-point precision issues, could `double` be an `int` that represents the exact number of occurrences of each output? If needed, the frequency could still be calculated by dividing that by the number of shots.

src/Simulation/Core/IQuantumMachineJob.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        /// <summary>
+        /// Gets the status of the submitted job.
+        /// </summary>
+        string Status { get; }
+
+        /// <summary>
+        /// Gets whether the job execution completed successfully.
+        /// </summary>
+        bool Succeeded { get;  }
```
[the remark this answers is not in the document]
cesarzc: As you mentioned, all of them are related. However, depending on the specific implementation of the Job, the status can be more than just InProgress, Succeeded and Failed. The InProgress and Succeeded properties are wrappers around Status to get information used for control flow.

src/Simulation/Core/IQuantumMachineJob.cs
```diff
@@ -0,0 +1,53 @@
...
+        /// <summary>
+        /// Gets an URI to access the job.
+        /// </summary>
+        Uri Uri { get; }
+
+        /// <summary>
+        /// Cancels the job.
+        /// </summary>
+        /// <param name="cancellationToken">The cancellation token.</param>
+        Task CancelAsync(CancellationToken cancellationToken = default);
```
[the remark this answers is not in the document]
cesarzc: It is cancelling the operation, I will expand the parameter description :)

src/Simulation/Core/IQuantumMachineJob.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        /// <summary>
+        /// Refreshes the state of the job.
+        /// </summary>
+        /// <param name="cancellationToken">The cancellation token.</param>
+        Task RefreshAsync(CancellationToken cancellationToken = default);
```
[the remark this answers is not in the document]
cesarzc: I will add the clarification.

src/Simulation/Core/IQuantumMachineOutput.cs
cesarzc: It could be obtained by doing that, but it needs to be proportional to the frequency of the specific output. That requires a little bit of logic which is provided by this convenience so that the developer using the interface does not have to write that logic.

src/Simulation/Core/IQuantumMachineOutput.cs
cesarzc: As with the GetOutput method, this is a convenience method. Someone can achieve the same result from the Histogram property but it requires a bit more code than just calling a method. It also allows the class that implements the IQuantumMachine interface to provide a consistent return value when there is ambiguity (e.g. when multiple outputs have the same frequency).

src/Simulation/Core/IQuantumMachineOutput.cs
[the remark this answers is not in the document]
cesarzc: In theory, ideally the whole interface should be parametrized, but I think that would remove the flexibility we need for standalone executables.

src/Simulation/Core/IQuantumMachineJob.cs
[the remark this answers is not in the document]
cesarzc: I added the `Failed` property to make it less weird to check for that case.

src/Simulation/Core/IQuantumMachineOutput.cs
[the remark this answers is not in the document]
cesarzc: Per our conversation, I made it parametrized.

src/Simulation/Core/IQuantumMachineOutput.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// Interface to access the results of a program executed in a quantum machine.
+    /// <typeparam name="TOutput">Type of output the quantum program returns.</typeparam>
+    /// </summary>
+    public interface IQuantumMachineOutput<TOutput>
+    {
+        /// <summary>
+        /// Gets a histogram of outputs.
+        /// The key is the output and the value is its frequency.
+        /// </summary>
+        IReadOnlyDictionary<object, TOutput> Histogram { get; }
```
samarsha, line 18:
```suggestion
        IReadOnlyDictionary<TOutput, double> Histogram { get; }
```

src/Simulation/Core/IQuantumMachineOutput.cs
samarsha: After chatting with @cesarzc, it seems like the Azure Quantum service itself is returning a histogram with doubles, so this would need to be a bigger change than just updating the interface.

src/Simulation/Core/IQuantumMachineOutput.cs
samarsha: We chatted offline and agreed that `GetOutput` and `GetMostFrequentOutput` could be moved to extension methods.

PR #190, opened 2020-05-04
Performance improvement to QubitManager for Borrowing.

avasch01:
Performance improvement to QubitManager for Borrowing. This performance issue has been noted by me before and documented as TODO, but its resolution has been delayed. We have hit it recently while benchmarking, so it is time for a fix.


bettinaheim:
Lgtm. 
We have had microbenchmarks at some point where I believe this has popped up previously. Unfortunately we have never gotten around to integrate these microbenchmarks into our Solid build. It would still be a good opportunity to revive that effort and see if we can add some perf tests for this.

avasch01:
Yes, it would be great to revive that effort find the benchmarks and integrate them! This checkin improves assymptotic complexity, so it should improve the results of those benchmarks for large numbers of qubits.

---
In reply to: [623739869](https://github.com/microsoft/qsharp-runtime/pull/190#issuecomment-623739869) [](ancestors = 623739869)

PR #191, opened 2020-05-05
Make sure tests can run without  build.

anpaz-msft:
Adding the step to explicitly build the native binaries before their unittests as it doesn't happen automatically.

### DIFF
--- a/build/test.ps1
+++ b/build/test.ps1
@@ -6,6 +6,7 @@ $all_ok = $True
 
 Write-Host "##[info]Test Native simulator"
 pushd (Join-Path $PSScriptRoot "../src/Simulation/Native/build")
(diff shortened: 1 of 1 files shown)


build/test.ps1
```diff
@@ -6,6 +6,7 @@ $all_ok = $True
 
 Write-Host "##[info]Test Native simulator"
 pushd (Join-Path $PSScriptRoot "../src/Simulation/Native/build")
+cmake --build . --config $Env:BUILD_CONFIGURATION
```
cgranade, line 9: Out of curiosity, does cmake rebuild in this case if the build has already run?

anpaz-msft: No, cmake checks timestamps and will not build again.

PR #194, opened 2020-05-07
Add support for submitting standalone entry points to Azure Quantum

samarsha:


src/Simulation/CsharpGeneration.Tests/EntryPointTests.fs
```diff
@@ -481,6 +481,9 @@ let ``Supports default custom simulator`` () =
     given ["--simulator"; typeof<QuantumSimulator>.FullName; "--use-h"; "false"] |> fails
 
 
+// TODO: Add tests for the "submit" command.
```
bettinaheim, line 484: :) 
What's the plan for tests? Are there any test planned on the runtime repo, or will this be tested as part of our service tests?

src/Simulation/EntryPointDriver/Azure.cs
```diff
@@ -0,0 +1,181 @@
...
+                return 1;
+            }
+
+            var job = await machine.SubmitAsync(entryPoint.Info, entryPoint.CreateArgument(parseResult));
+            switch (settings.Output)
+            {
+                case OutputFormat.FriendlyUri:
+                    Console.WriteLine("Job submitted. To track your job status and see the results use:");
+                    Console.WriteLine();
+                    // TODO: Show the friendly URI.
```
bettinaheim, line 40: What's this is a todo rather than done?

src/Simulation/EntryPointDriver/Azure.cs
```diff
@@ -0,0 +1,181 @@
...
+        /// <summary>
+        /// Creates a quantum machine based on the Azure Quantum submission settings.
+        /// </summary>
+        /// <param name="settings">The Azure Quantum submission settings.</param>
+        /// <returns>A quantum machine.</returns>
+        private static IQuantumMachine? CreateMachine(AzureSettings settings)
+        {
+            if (!(settings.Target is null) && settings.Target.StartsWith("ionq."))
+            {
+                // TODO: Give the number of shots to the quantum machine.
```
bettinaheim, line 61: same question here

src/Simulation/EntryPointDriver/Simulation.cs
```diff
@@ -0,0 +1,101 @@
...
+        /// <param name="entryPoint">The entry point.</param>
+        /// <param name="parseResult">The command-line parsing result.</param>
+        /// <param name="createSimulator">A function that creates an instance of the simulator to use.</param>
+        private static async Task RunSimulator(
+            IEntryPoint<TIn, TOut> entryPoint, ParseResult parseResult, Func<IOperationFactory> createSimulator)
+        {
+            var simulator = createSimulator();
+            try
+            {
+                var value = await simulator.Run<TCallable, TIn, TOut>(entryPoint.CreateArgument(parseResult));
```
bettinaheim, line 67: @anpaz-msft Is invoking .Run the recommended way for executing? I vaguely remember some other method having been added related to getting a better stack trace handling?

src/Simulation/CsharpGeneration.Tests/EntryPointTests.fs
samarsha: My plan is to test all of the command-line option handling in the runtime repo using the `NothingMachine` as the target. For other quantum machines like IonQ, I think those will have to go in another repository where those machines are available by a direct reference.

src/Simulation/EntryPointDriver/Azure.cs
samarsha: There is not a way to get the friendly URI from the job at the moment. I think @cesarzc might have something planned for that?

src/Simulation/EntryPointDriver/Azure.cs
samarsha: The `IQuantumMachine` interface does not support specifying the number of shots when a job is submitted. @cesarzc was planning to update it.

src/Simulation/EntryPointDriver/Simulation.cs
samarsha: When C# generation generates a `Run` method for operations, it calls `IOperationFactory.Run`, so I was just following that.

src/Simulation/CsharpGeneration.Tests/EntryPointTests.fs
samarsha: Tests are in #204.

bettinaheim: sounds good.

src/Simulation/EntryPointDriver/Azure.cs
bettinaheim: Ah, I see. Just curious.

src/Simulation/EntryPointDriver/Simulation.cs
bettinaheim: Ok, let's keep with that then - it's a quick thing to change in any case.

PR #203, opened 2020-05-14
Add Microsoft.Azure.Quantum.Client sources and QuantumMachineFactory

rmshaffer:
This PR targeting `feature/azure-quantum-preview` contains the sources of the `Microsoft.Azure.Quantum.Client` package.

It also contains an `IQuantumMachine` factory in `Machine/QuantumMachineFactory.cs`, which is mostly copied from [a similar function in this PR](https://github.com/microsoft/qsharp-runtime/blob/256ebbd8cffc7d3a9e9dde2f19e573a565f9459d/src/Simulation/EntryPointDriver/Azure.cs#L58) from @samarsha.

Before merging:
- [x] Confirm successful end-to-end CI build with signing enabled.


src/Azure/Azure.Quantum.Client/generated/IProvidersOperations.cs
```diff
@@ -0,0 +1,68 @@
...
+    using Models;
+    using System.Collections;
+    using System.Collections.Generic;
+    using System.Threading;
+    using System.Threading.Tasks;
+
+    /// <summary>
+    /// ProvidersOperations operations.
+    /// </summary>
+    public partial interface IProvidersOperations
```
george-moussa, line 24: should this be IProviderOperations instead? we might need to update the swagger for this.
Andres's comment also makes sense we can use IProviderStatusOperations.Get

rmshaffer: I think the plural `Providers` does make some sense because it's operating on the list of providers, rather than a single provider. But I don't have a strong opinion.

PR #204, opened 2020-05-14
Add tests for the standalone executable "submit" command

samarsha:


src/Simulation/EntryPointDriver.Tests/Tests.fs
```diff
@@ -510,3 +582,28 @@ Commands:
     given ["--help"] |> yields message
     given ["-h"] |> yields message
     given ["-?"] |> yields message
+
+[<Fact>]
+let ``Shows help text for submit command`` () =
```
bettinaheim, line 587: The submit command itself won't be listed in the help function for the executable or in tab completion, but one can query the help for the submit explicitly - is that correct?

src/Simulation/EntryPointDriver.Tests/Tests.fs
```diff
@@ -481,7 +481,79 @@ let ``Supports default custom simulator`` () =
     given ["--simulator"; typeof<QuantumSimulator>.FullName; "--use-h"; "false"] |> fails
 
 
-// TODO: Add tests for the "submit" command.
+// Azure Quantum Submission
```
anpaz-msft, line 484: do you have tests for when missing required arguments? like missing target/workspace?

src/Simulation/EntryPointDriver.Tests/Tests.fs
```diff
@@ -481,7 +481,79 @@ let ``Supports default custom simulator`` () =
...
+     "myStorage"
+     "--subscription"
+     "mySubscription"
+     "--resource-group"
+     "myResourceGroup"
+     "--workspace"
+     "myWorkspace"]
+
+[<Fact>]
+let ``Submit can submit a job`` () =
```
cesarzc, line 501: It might be worth it to add a test case in which an unexpected exception occurs when calling the submit method. Catching the exception and showing an output that lets the user know something unexpected happened would be valuable.

src/Simulation/EntryPointDriver.Tests/Tests.fs
```diff
@@ -481,7 +481,79 @@ let ``Supports default custom simulator`` () =
...
+Shots: 500
+Output: FriendlyUri
+Verbose: True
+
+Job submitted. To track your job status and see the results use:
+
+00000000-0000-0000-0000-0000000000000"
+
+[<Fact>]
+let ``Submit allows overriding default values`` () =
```
cesarzc, line 531: Maybe not specifically related to the submit command, but having a test in which one of the Entry-Point arguments has the same name as one of the options would also be useful.

src/Simulation/EntryPointDriver.Tests/Tests.fs
samarsha: Yep!

src/Simulation/EntryPointDriver.Tests/Tests.fs
samarsha: I think @bettinaheim and I were wanting to wait until next release to improve the exception handling. However, I can add a test where an exception is thrown and check that the first line or so of the exception stack trace is correct. Would that work for now?

samarsha: I see that you improved the exception handling in #218, so I can add tests for those instead now.

samarsha:
After talking with @bettinaheim, I'm going to merge this PR as-is so we can merge the feature branch to master with some tests, but I'll add more tests to address the remaining comments in a separate PR.

src/Simulation/EntryPointDriver.Tests/Tests.fs
samarsha: No, those would be good to add - I'll include them in a separate PR (see comment a the bottom).

src/Simulation/EntryPointDriver.Tests/Tests.fs
samarsha: I'll include this in the next tests PR (see comment at the bottom).

src/Simulation/EntryPointDriver.Tests/Tests.fs
samarsha: Sounds good, I'll include those tests in a follow-up PR.

bettinaheim:
@cesarzc @anpaz-msft. I've asked Sarah to make a separate PR to add the additional tests that you suggested and merge this one.  I'd much rather have a good set of tests than no tests, and I want to make sure that we have sufficient time for the release testing.

PR #205, opened 2020-05-18
Allow reflection to find type in loaded assembly

rmshaffer:
Two changes to `QuantumMachineFactory` here, to address issues I found while consuming this from IQ#:
- Take an `IWorkspace` parameter instead of a `Workspace` parameter.
- Load the type directly from the already-loaded assembly with the expected name, rather than using `Type.GetType()`, which seems to sometimes require an `AssemblyQualifiedName` including version number and public key token in order to load the type successfully. (Open to suggestions here, in case there's a simpler way of doing this, but this seems to work.)

### DIFF
--- a/src/Azure/Azure.Quantum.Client/Machine/QuantumMachineFactory.cs
+++ b/src/Azure/Azure.Quantum.Client/Machine/QuantumMachineFactory.cs
@@ -26,12 +26,21 @@ namespace Microsoft.Azure.Quantum
                 ? "Microsoft.Quantum.Providers.Honeywell.Targets.HoneywellQuantumMachine, Microsoft.Quantum.Providers.Honeywell"
                 : null;
 
-            // First try to load the signed assembly with the correct version, then try the unsigned one.
-            var machineType =
-                machineName is null
-                ? null
-                : Type.GetType($"{machineName}, Version={typeof(IWorkspace).Assembly.GetName().Version}, Culture=neutral, PublicKeyToken=40866b40fd95c7f5")
-                ?? Type.GetType(machineName, throwOnError: true);
+            Type machineType = null;
+            if (machineName != null)
+            {
+                // First try to load the signed assembly with the correct version, then try the unsigned one.
+                try
+                {
+                    machineType = Type.GetType($"{machineName}, Version={typeof(IWorkspace).Assembly.GetName().Version}, Culture=neutral, PublicKeyToken=40866b40fd95c7f5");
+                }
+                catch
+                {
+                    machineType = null;
+                }
+
+                machineType ??= Type.GetType(machineName, throwOnError: true);
+            }
 
             return machineType is null
                 ? null
--- a/src/Azure/Azure.Quantum.Client/Machine/QuantumMachineFactory.cs
+++ b/src/Azure/Azure.Quantum.Client/Machine/QuantumMachineFactory.cs
@@ -17,21 +17,26 @@ namespace Microsoft.Azure.Quantum
         /// <returns>A quantum machine for job submission targeting <c>targetName</c>.</returns>
         public static IQuantumMachine? CreateMachine(Workspace workspace, string targetName, string storageAccountConnectionString)
         {
-            if (string.IsNullOrEmpty(targetName))
-            {
-                return null;
-            }
-
             var machineName =
-                targetName.StartsWith("ionq.")
+                targetName is null
+                ? null
+                : targetName.StartsWith("ionq.")
                 ? "Microsoft.Quantum.Providers.IonQ.Targets.IonQQuantumMachine, Microsoft.Quantum.Providers.IonQ"
                 : targetName.StartsWith("honeywell.")
                 ? "Microsoft.Quantum.Providers.Honeywell.Targets.HoneywellQuantumMachine, Microsoft.Quantum.Providers.Honeywell"
                 : null;
-            return machineName is null
+
+            // First try to load the signed assembly with the correct version, then try the unsigned one.
+            var machineType =
+                machineName is null
+                ? null
+                : Type.GetType($"{machineName}, Version={typeof(Workspace).Assembly.GetName().Version}, Culture=neutral, PublicKeyToken=40866b40fd95c7f5")
+                ?? Type.GetType(machineName, throwOnError: true);
+
+            return machineType is null
                 ? null
                 : (IQuantumMachine)Activator.CreateInstance(
-                    Type.GetType(machineName, throwOnError: true),
+                    machineType,
                     targetName,
                     storageAccountConnectionString,
                     workspace);
--- a/src/Azure/Azure.Quantum.Client/Machine/QuantumMachineFactory.cs
+++ b/src/Azure/Azure.Quantum.Client/Machine/QuantumMachineFactory.cs
@@ -1,8 +1,10 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+#nullable enable
+
 using System;
-using System.Threading.Tasks;
+using System.Linq;
 using Microsoft.Quantum.Runtime;
 
 namespace Microsoft.Azure.Quantum
@@ -16,7 +18,7 @@ namespace Microsoft.Azure.Quantum
         /// <param name="targetName">The execution target for job submission.</param>
         /// <param name="storageAccountConnectionString">The connection string for the Azure storage account.</param>
         /// <returns>A quantum machine for job submission targeting <c>targetName</c>.</returns>
-        public static IQuantumMachine? CreateMachine(Workspace workspace, string targetName, string storageAccountConnectionString)
+        public static IQuantumMachine? CreateMachine(IWorkspace workspace, string targetName, string storageAccountConnectionString)
         {
             if (string.IsNullOrEmpty(targetName))
             {
@@ -25,13 +27,13 @@ namespace Microsoft.Azure.Quantum
 
             if (targetName.StartsWith("ionq."))
             {
-                var ionQType = Type.GetType(
-                    "Microsoft.Quantum.Providers.IonQ.Targets.IonQQuantumMachine, Microsoft.Quantum.Providers.IonQ",
-                    throwOnError: true);
+                var ionQType = AppDomain.CurrentDomain.GetAssemblies()
+                    .First(a => a.FullName.StartsWith("Microsoft.Quantum.Providers.IonQ,"))
+                    .GetType("Microsoft.Quantum.Providers.IonQ.Targets.IonQQuantumMachine", throwOnError: true);
                 return (IQuantumMachine)Activator.CreateInstance(
                     ionQType, targetName, storageAccountConnectionString, workspace);
             }
-            
+
             return null;
         }
     }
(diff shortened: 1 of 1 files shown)


rmshaffer:
Thanks for trying this out! Did you do anything special such that the existing `Type.GetType()` call was succeeding? When I tried to call the code using `Type.GetType()`, it couldn't find the assembly (even if it had already been loaded) unless I also included the version and the PublicKeyToken in the type string, i.e., in a format like this:
`System.Web.Extensions, Version=1.0.61025.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35`
I suppose we could always expose both versions of this, if we need to obtain the `Type` in different ways. IQ# does its own package loading and management at runtime, so I guess it's not surprising that this behaves differently.

---
In reply to: [414007852](https://github.com/microsoft/qsharp-runtime/pull/205#issuecomment-414007852) [](ancestors = 414007852)

samarsha:
No, I don't think I needed to do anything to get `Type.GetType` to work. Could you post the error message you get from it?

samarsha:
Could it have something to do with IQ# using signed versions of the IonQ assembly? That could mean that `Type.GetType` would fail for the standalone executable as well if it uses a signed version. (I'm not sure if the versions we have been using for testing have been signed or not.) We could try specifying just the assembly name and public key token (but not the version), but I am not sure if the public key token stays the same or changes when the version number is changed.

rmshaffer:
Below is the error I get. I'll try just specifying the PublicKeyToken and not the version; I think that should be consistent for all signed versions, but I'd need to confirm this.
```
System.IO.FileNotFoundException
  HResult=0x80070002
  Message=Could not load file or assembly 'Microsoft.Quantum.Providers.IonQ, Culture=neutral, PublicKeyToken=null'. The system cannot find the file specified.
  Source=<Cannot evaluate the exception source>
  StackTrace:
  System.Private.CoreLib.dll!System.RuntimeTypeHandle.GetTypeByName(string name, bool throwOnError, bool ignoreCase, System.Runtime.CompilerServices.StackCrawlMarkHandle stackMark, System.Runtime.CompilerServices.ObjectHandleOnStack assemblyLoadContext, bool loadTypeFromPartialName, System.Runtime.CompilerServices.ObjectHandleOnStack type, System.Runtime.CompilerServices.ObjectHandleOnStack keepalive) Unknown
  System.Private.CoreLib.dll!System.RuntimeTypeHandle.GetTypeByName(string name, bool throwOnError, bool ignoreCase, ref System.Threading.StackCrawlMark stackMark, System.Runtime.Loader.AssemblyLoadContext assemblyLoadContext, bool loadTypeFromPartialName) Unknown
  System.Private.CoreLib.dll!System.RuntimeType.GetType(string typeName, bool throwOnError, bool ignoreCase, ref System.Threading.StackCrawlMark stackMark) Unknown
  System.Private.CoreLib.dll!System.Type.GetType(string typeName, bool throwOnError) Unknown
  Microsoft.Azure.Quantum.Client.dll!Microsoft.Azure.Quantum.QuantumMachineFactory.CreateMachine(Microsoft.Azure.Quantum.Workspace workspace, string targetName, string storageAccountConnectionString) Unknown
  Microsoft.Quantum.IQSharp.AzureClient.dll!Microsoft.Quantum.IQSharp.AzureClient.AzureClient.SubmitJobAsync(Microsoft.Jupyter.Core.IChannel channel, Microsoft.Quantum.IQSharp.IOperationResolver operationResolver, string operationName) Line 162 C#
> Microsoft.Quantum.IQSharp.AzureClient.dll!Microsoft.Quantum.IQSharp.AzureClient.SubmitMagic.RunAsync(string input, Microsoft.Jupyter.Core.IChannel channel) Line 63 C#
  Microsoft.Quantum.IQSharp.AzureClient.dll!Microsoft.Quantum.IQSharp.AzureClient.AzureClientMagicBase.Run(string input, Microsoft.Jupyter.Core.IChannel channel) Line 35 C#

```

---
In reply to: [630504039](https://github.com/microsoft/qsharp-runtime/pull/205#issuecomment-630504039) [](ancestors = 630504039)

src/Azure/Azure.Quantum.Client/Machine/QuantumMachineFactory.cs
```diff
@@ -17,21 +17,26 @@ public static class QuantumMachineFactory
...
                 : targetName.StartsWith("honeywell.")
                 ? "Microsoft.Quantum.Providers.Honeywell.Targets.HoneywellQuantumMachine, Microsoft.Quantum.Providers.Honeywell"
                 : null;
-            return machineName is null
+
+            // First try to load the signed assembly with the correct version, then try the unsigned one.
+            var machineType =
+                machineName is null
+                ? null
+                : Type.GetType($"{machineName}, Version={typeof(Workspace).Assembly.GetName().Version}, Culture=neutral, PublicKeyToken=40866b40fd95c7f5")
```
rmshaffer, line 33: I checked several different signed versions from our CI build of both the IonQ and Honeywell assemblies and they all had this same `PublicKeyToken` value.

rmshaffer: The version seems to be required, but it seems safe to just enforce that it matches the currently-loaded version of `Microsoft.Azure.Quantum.Client`.

samarsha: I get a different error message now:

```
Unhandled exception: System.IO.FileLoadException: Could not load file or assembly 'Microsoft.Quantum.Providers.IonQ, Version=1.0.0.0, Culture=neutral, PublicKeyToken=40866b40fd95c7f5'. The located assembly's manifest definition does not match the assembly reference. (0x80131040)
File name: 'Microsoft.Quantum.Providers.IonQ, Version=1.0.0.0, Culture=neutral, PublicKeyToken=40866b40fd95c7f5'
   at System.RuntimeTypeHandle.GetTypeByName(String name, Boolean throwOnError, Boolean ignoreCase, StackCrawlMarkHandle stackMark, ObjectHandleOnStack assemblyLoadContext, Boolean loadTypeFromPartialName, ObjectHandleOnStack type, ObjectHandleOnStack keepalive)
   at System.RuntimeTypeHandle.GetTypeByName(String name, Boolean throwOnError, Boolean ignoreCase, StackCrawlMark& stackMark, AssemblyLoadContext assemblyLoadContext, Boolean loadTypeFromPartialName)
   at System.RuntimeType.GetType(String typeName, Boolean throwOnError, Boolean ignoreCase, StackCrawlMark& stackMark)
   at System.Type.GetType(String typeName)
   at Microsoft.Azure.Quantum.QuantumMachineFactory.CreateMachine(IWorkspace workspace, String targetName, String storageAccountConnectionString) in C:\Users\samarsha\Source\Repos\microsoft\qsharp-runtime\src\Azure\Azure.Quantum.Client\Machine\QuantumMachineFactory.cs:line 30
   at Microsoft.Quantum.CsharpGeneration.EntryPointDriver.Azure.CreateMachine(AzureSettings settings) in C:\Users\samarsha\Source\Repos\microsoft\qsharp-runtime\src\Simulation\EntryPointDriver\Azure.cs:line 60
   at Microsoft.Quantum.CsharpGeneration.EntryPointDriver.Azure.Submit[TIn,TOut](IEntryPoint`2 entryPoint, ParseResult parseResult, AzureSettings settings) in C:\Users\samarsha\Source\Repos\microsoft\qsharp-runtime\src\Simulation\EntryPointDriver\Azure.cs:line 28
   at Microsoft.Quantum.CsharpGeneration.EntryPointDriver.Driver`3.Submit(ParseResult parseResult, AzureSettings settings) in C:\Users\samarsha\Source\Repos\microsoft\qsharp-runtime\src\Simulation\EntryPointDriver\Driver.cs:line 124
   at System.CommandLine.Invocation.CommandHandler.GetResultCodeAsync(Object value, InvocationContext context)
   at System.CommandLine.Invocation.ModelBindingCommandHandler.InvokeAsync(InvocationContext context)
   at System.CommandLine.Invocation.InvocationPipeline.<>c__DisplayClass4_0.<<BuildInvocationChain>b__0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.CommandLine.Builder.CommandLineBuilderExtensions.<>c.<<UseParseErrorReporting>b__20_0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.CommandLine.Builder.CommandLineBuilderExtensions.<>c__DisplayClass15_0.<<UseHelp>b__0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.CommandLine.Builder.CommandLineBuilderExtensions.<>c__DisplayClass24_0.<<UseVersionOption>b__0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.CommandLine.Builder.CommandLineBuilderExtensions.<>c__DisplayClass22_0.<<UseTypoCorrections>b__0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.CommandLine.Builder.CommandLineBuilderExtensions.<>c.<<UseSuggestDirective>b__21_0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.CommandLine.Builder.CommandLineBuilderExtensions.<>c.<<UseParseDirective>b__19_0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.CommandLine.Builder.CommandLineBuilderExtensions.<>c.<<UseDebugDirective>b__11_0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.CommandLine.Builder.CommandLineBuilderExtensions.<>c.<<RegisterWithDotnetSuggest>b__10_0>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.CommandLine.Builder.CommandLineBuilderExtensions.<>c__DisplayClass13_0.<<UseExceptionHandler>b__0>d.MoveNext()
```

I think this is because I'm using a locally built version of Microsoft.Azure.Quantum.Client (version 1.0.0.0), but a a packaged version of IonQ with a different version number. My local setup is hacky, but it was the only way I was able to get something working where I could test local changes to the entry point driver with the quantum machines (without making a full Solid build to synchronize the version numbers).

It looks like `Type.GetType` is throwing this error even though the `throwOnError` parameter is not set to true. It's not ideal, but maybe the exception needs to be caught so the next call to `Type.GetType` can run?

rmshaffer: Yes, that sounds like the right approach. I shouldn't have assumed that `throwOnError=false` means that it won't throw on error.

rmshaffer: Or, wait, actually, since you're using a project reference to Microsoft.Azure.Quantum.Client, you'll probably always be using a locally-built version and will never have the correct version number.

Could you use a package reference to Microsoft.Azure.Quantum.Client instead? I'm using version 0.11.2005.1420-beta and things are working.

samarsha: Well, I seem to have broken my local project after changing it a bit and can't get it back to the state it was in before. I'll approve the PR since the new change fixes the assembly loading problem - I think the other problem I'm running into is unrelated. :)

About using a package reference, that might work, but since Microsoft.Azure.Quantum.Client is in the same repository as EntryPointDriver I'm not sure if it makes sense - I think all of the projects I have seen use project references to other projects in the same repository, and package references to projects in different repositories.

My idea now to avoid version conflicts when testing locally was to just build the entire Solid repository locally, so everything would have the same version, but at the moment I'm having some issues with that. :) Do you want to remove the try/catch and require that the versions match?

rmshaffer: I think I'll leave it this way for now, since at least it should allow you to continue with your existing workflow (if you can get that working again). If/when you figure out a way to get around the version issue, you could always remove the try/catch at that time. I don't think it hurts anything to have it there, though.

Thanks for reviewing and working through this with me!

PR #207, opened 2020-05-19
Add Honeywell to QuantumMachineFactory

samarsha:
Keeping `Type.GetType` since that works for the standalone executable, but this will have a merge conflict with #205.


rmshaffer:
Looks good! No worries about the merge conflict with #205, we need to first decide what the right approach is there anyway.

anpaz-msft:
This is going to conflict with #205 , you might want to pick that up?

anpaz-msft:
doh. sorry, i looked at the comment afterwards.

PR #209, opened 2020-05-20
Create separate project for entry point driver tests and clean up references

samarsha:
The entry point driver's references to the Q# runtime are now in one place (test projects get them transitively). This makes it easier to update the references to test changes to the entry point driver locally with the Azure Quantum providers without version conflicts (since both the entry point driver and the providers depend on the runtime).


samarsha:
This breaks the Solid build since the package versions are updated to the current build version, but those versions haven't been built yet (since they are in the same repository).

So I think I will change them back to project references but keep the rest of the changes that made EntryPointDriver more modular from the rest of the repository. That way it would be easier for me to change the project references back to package references when I need to test the providers locally.

PR #211, opened 2020-05-26
Add support for submitting standalone executables to Azure Quantum

samarsha:
TODO:

- [x] Add tests for the standalone executable "submit" command (#204)
- [x] Handle Azure quantum known exceptions (#218)


src/Azure/Azure.Quantum.Client/Authentication/IAccessTokenProvider.cs
```diff
@@ -0,0 +1,22 @@
+﻿// <copyright file="IAccessTokenProvider.cs" company="Microsoft Corporation">
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
bettinaheim, line 2: I just noticed this here; We license everything under MIT license, so we should make sure to add that to all files. Do you know why that was missing in the first place?

samarsha: That's probably because the Azure client came from a private repository. I'll update the headers.

src/Azure/Azure.Quantum.Client/Utility/Constants.cs
```diff
@@ -0,0 +1,35 @@
...
+{
+    internal static class Constants
+    {
+        internal const string DefaultBaseUri = "https://app-jobscheduler-prod.azurewebsites.net/";
+
+        internal static class Aad
+        {
+            internal const string ApplicationId = "84ba0947-6c53-4dd2-9ca9-b3694761521b";
+
+            // TODO: Confirm audience
```
bettinaheim, line 15: Was this confirmed?

samarsha: I'm not sure what it's referring to. @rmshaffer @anpaz-msft @cesarzc @swernli ?

src/Azure/Azure.Quantum.Client/Authentication/IAccessTokenProvider.cs
rmshaffer: Sorry, that's my fault, I didn't add the headers when I moved these files here.

src/Azure/Azure.Quantum.Client/Utility/Constants.cs
rmshaffer: I'm not sure how old the comment is, but the `Audience` string here looks correct to me.

FWIW, all of this code under Azure.Quantum.Client has already been in master of the private repo for some time. A detailed review here shouldn't be necessary, since it should have already been reviewed when it was first written and committed.

rmshaffer: (The exception is `Azure.Quantum.Client/Machine/QuantumMachineFactory.cs`, which did not come from the private repo.)

bettinaheim: All of the code here has already been reviewed since this is just the merge of the feature branch. I am also just skimming, but even so, merging is a good opportunity to look through the code to exactly spot things here and there that god forgotten (like the todo or the licenses).

src/Azure/Azure.Quantum.Client/Authentication/IAccessTokenProvider.cs
samarsha: No worries!

src/Azure/Azure.Quantum.Client/Utility/Constants.cs
samarsha: TODO has been removed.

PR #212, opened 2020-05-26
Created QuantumProcessorException

cesarzc:
This change creates the base exception for errors that occur when invoking a quantum processor.

### DIFF
--- a/src/Simulation/Common/Exceptions/QuantumProcessorException.cs
+++ b/src/Simulation/Common/Exceptions/QuantumProcessorException.cs
@@ -1,0 +1,15 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+
+namespace Microsoft.Quantum.Simulation.Common.Exceptions
+{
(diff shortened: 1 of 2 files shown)


src/Simulation/Common/Exceptions/QuantumProcessorException.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+
+namespace Microsoft.Quantum.Simulation.Common.Exceptions
+{
+    public class QuantumProcessorException : Exception
```
swernli, line 8: Maybe a nitpick, but do we want this to refer to translation somehow? Something like `QuantumProcessorTranslationException` perhaps? As it is it might imply that the exception came from the actual hardware processor. Same thought with the default string below, where "invoking" seems to imply that it already tried to run on hardware.

cesarzc: I agree `QuantumProcessorTranslationException` is a better name. I changed it :)

PR #214, opened 2020-05-27
Add --dry-run option and implement --shots option

samarsha:
I decided to use `submit --dry-run` since:

* With `submit --validate`, it looks like it is validating *and* submitting, rather than validating *instead of* submitting.
* I considered making a separate `validate` command, but it would have all the same options as the `submit` command, so that seemed like unnecessarily duplication.

I'll add the tests to #204.

### DIFF
--- a/src/Simulation/EntryPointDriver/Azure.cs
+++ b/src/Simulation/EntryPointDriver/Azure.cs
@@ -6,6 +6,7 @@ using System.CommandLine.Parsing;
 using System.Threading.Tasks;
 using Microsoft.Azure.Quantum;
 using Microsoft.Quantum.Runtime;
+using static Microsoft.Quantum.CsharpGeneration.EntryPointDriver.Driver;
 
 namespace Microsoft.Quantum.CsharpGeneration.EntryPointDriver
 {
@@ -28,7 +29,8 @@ namespace Microsoft.Quantum.CsharpGeneration.EntryPointDriver
             var machine = CreateMachine(settings);
             if (machine is null)
             {
-                DisplayUnknownTargetError(settings.Target);
+                DisplayWithColor(ConsoleColor.Red, Console.Error,
+                    $"The target '{settings.Target}' was not recognized.");
                 return 1;
             }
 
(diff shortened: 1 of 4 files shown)


src/Simulation/EntryPointDriver/Azure.cs
```diff
@@ -32,23 +32,45 @@ internal static class Azure
...
-            // TODO: Specify the number of shots. The IQuantumMachine interface should be updated.
-            var job = await machine.SubmitAsync(entryPoint.Info, entryPoint.CreateArgument(parseResult));
-            switch (settings.Output)
+            var input = entryPoint.CreateArgument(parseResult);
+            if (settings.DryRun)
             {
-                case OutputFormat.FriendlyUri:
-                    Console.WriteLine("Job submitted. To track your job status and see the results use:");
+                var (isValid, message) = machine.Validate(entryPoint.Info, input);
+                Console.WriteLine(isValid ? "✔️  The program is valid!" : "❌  The program is invalid.");
```
cesarzc, line 39: I really like the ✔️ and ❌!

src/Simulation/EntryPointDriver/Azure.cs
```diff
@@ -32,23 +32,45 @@ internal static class Azure
...
+        /// <param name="job">The job.</param>
+        /// <param name="format">The output format.</param>
+        /// <exception cref="ArgumentOutOfRangeException">Thrown if the output format is invalid.</exception>
+        private static void DisplayJob(IQuantumMachineJob job, OutputFormat format)
+        {
+            switch (format)
+            {
+                case OutputFormat.FriendlyUri:
                     // TODO: Show the friendly URI. The friendly URI is not yet available from the job.
-                    Console.WriteLine(job.Id);
```
cesarzc, line 43: In this case maybe we could display something like "Not yet available" or "Available in a future release". This way we can let users know that this format option is not available in this release but it will be available in a future one. Otherwise, users that try this option early might not get why both the ID and the URI options display the same output.

PR #218, opened 2020-05-27
Handle Azure quantum known exceptions

cesarzc:
This change handles two known exceptions that can happen when submitting/executing a program in Azure quantum:
-  `AzureQuantumException`
- `QuantumProcessorTranslationException`


src/Simulation/EntryPointDriver/Azure.cs
```diff
@@ -48,9 +50,35 @@ internal static class Azure
             }
             else
             {
-                var job = await machine.SubmitAsync(
+                try
+                {
+                    var job = await machine.SubmitAsync(
                     entryPoint.Info, input, new SubmissionContext { Shots = settings.Shots });
```
samarsha, line 56:
```suggestion
                        entryPoint.Info, input, new SubmissionContext { Shots = settings.Shots });
```

src/Simulation/EntryPointDriver/Azure.cs
```diff
@@ -48,9 +50,35 @@ internal static class Azure
...
+                }
+                catch (AzureQuantumException azureQuantumEx)
+                {
+                    DisplayWithColor(
+                        ConsoleColor.Red,
+                        Console.Error,
+                        "Something went wrong related to Azure quantum.");
+
+                    Console.WriteLine();
+                    Console.WriteLine(azureQuantumEx.Message);
```
samarsha, line 67:
```suggestion
                    Console.Error.WriteLine();
                    Console.Error.WriteLine(azureQuantumEx.Message);
```

src/Simulation/EntryPointDriver/Azure.cs
```diff
@@ -48,9 +50,35 @@ internal static class Azure
...
+                }
+                catch (QuantumProcessorTranslationException translationEx)
+                {
+                    DisplayWithColor(
+                        ConsoleColor.Red,
+                        Console.Error,
+                        "Something went wrong when performing translation to the intermediate representation used by the target quantum machine.");
+
+                    Console.WriteLine();
+                    Console.WriteLine(translationEx.Message);
```
samarsha, line 78:
```suggestion
                    Console.Error.WriteLine();
                    Console.Error.WriteLine(translationEx.Message);
```

src/Simulation/EntryPointDriver/Azure.cs
```diff
@@ -48,9 +50,35 @@ internal static class Azure
...
                     entryPoint.Info, input, new SubmissionContext { Shots = settings.Shots });
-                DisplayJob(job, settings.Output);
+                    DisplayJob(job, settings.Output);
+                }
+                catch (AzureQuantumException azureQuantumEx)
+                {
+                    DisplayWithColor(
+                        ConsoleColor.Red,
+                        Console.Error,
+                        "Something went wrong related to Azure quantum.");
```
samarsha, line 64: IMO, this message could describe a translation error as well, since for the user, the entire process is related to Azure Quantum. This is a little more specific:

```suggestion
                        "Something went wrong when submitting the program to the Azure Quantum service.");
```

cesarzc: Yes, you are right, I will change it such that both have the same message.

cesarzc: Thinking a little bit more about it, I think it is ok to have to separate messages since one indicates a problem with service related things which can be credentials, unavailability and the other one is a problem with the program itself.

samarsha: Yes, having different messages for translation and service errors is fine - just wanted to make the service error message more specific.

PR #223, opened 2020-05-28
Remove bond.

anpaz-msft:
We had a reference to bond on the Azure.Quantum.Client just for a Prototype parameter which was never used.
This is an unnecessary and heavy reference.
I'm changing the signature and removing the reference to the package.


bettinaheim:
@anpaz-msft Can this be merged?

PR #224, opened 2020-05-28
Add Microsoft.Azure.Quantum.Client.Test to test.ps1

samarsha:
I think that without this, the Azure client tests weren't running on the CI and Solid builds.


rmshaffer:
Thanks for fixing this! Just curious, is there a reason each .csproj is listed there individually, instead of testing the whole .sln? e.g., what is done here: https://github.com/microsoft/qsharp-compiler/blob/master/build/test.ps1. Wondering if that would help to avoid something like this from happening again.

samarsha:
Oh, that seems like a better way. :) I created #227.

PR #225, opened 2020-05-28
Create a separate package for the entry point driver

samarsha:
* The package is Microsoft.Quantum.EntryPointDriver, but I'm not 100% sure what to call it so I'm open to other suggestions.
* I renamed the project Microsoft.Quantum.Simulation.Simulators.csproj to Microsoft.Quantum.Simulators.csproj, so that the project name matches the package name. This was needed since EntryPointDriver depends on it and `dotnet pack` replaces project references with package references.
* There will also need to be changes to the Quantum SDK in microsoft/qsharp-compiler to reference the new package in Q# projects.


src/Simulation/EntryPointDriver/Microsoft.Quantum.EntryPointDriver.csproj
```diff
@@ -4,9 +4,8 @@
   
   <PropertyGroup>
     <TargetFramework>netstandard2.1</TargetFramework>
-    <RootNamespace>Microsoft.Quantum.CsharpGeneration.EntryPointDriver</RootNamespace>
-    <AssemblyName>Microsoft.Quantum.CsharpGeneration.EntryPointDriver</AssemblyName>
```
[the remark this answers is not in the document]
samarsha: It seems to be based on the name of the project file.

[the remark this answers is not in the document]
samarsha: You're right, it actually is based on `AssemblyName`. By default that is the same name as the project file, but changing the assembly name will also change the package reference.

src/Simulation/EntryPointDriver/Microsoft.Quantum.EntryPointDriver.csproj
```diff
@@ -16,7 +15,11 @@
```
samarsha, line 15:
```
<PackageReference Include="Microsoft.Quantum.Compiler" Version="0.10.2004.1717-alpha" />
```

I think that this is bringing in a lot of DLLs from the compiler that weren't there before when you run `dotnet publish` on a standalone application (see microsoft/qsharp-compiler#432). This is probably not what we want.

The only reason EntryPointDriver has this dependency is to get the constants defined in ReservedKeywords for the names of simulators and command line arguments. A few alternatives:

1. Read the constants at compile time (CsharpGeneration naturally has a dependency on the compiler) and embed them in the generated C# entry point file, and pass those into the driver so it can use them instead. This requires explicitly listing out all the constants the driver uses and making a new class to hold them, but their values will still come from the compiler.
2. Create a separate package in the compiler repository for ReservedKeywords and reference that instead of the full compiler.
3. Hard-code the strings instead of using the defined constants.

Thoughts? @bettinaheim @anpaz-msft

src/Simulation/CsharpGeneration/FindNuspecReferences.ps1
```diff
@@ -59,13 +59,6 @@ function Add-PackageReferenceDependencies($ProjectFileName) {
 
 # Add dependencies for the projects included in this NuGet package.
 Add-PackageReferenceDependencies 'Microsoft.Quantum.CsharpGeneration.fsproj'
-Add-PackageReferenceDependencies '..\EntryPointDriver\EntryPointDriver.csproj'
```
cesarzc, line 62: I am curious about why adding these references is no longer needed?

src/Simulation/CsharpGeneration/Microsoft.Quantum.CsharpGeneration.nuspec.template
```diff
@@ -23,12 +23,5 @@
   <files>
     <file src="bin\$Configuration$\**\Microsoft.Quantum.*.pdb" target="lib" />
     <file src="bin\$Configuration$\**\Microsoft.Quantum.*.xml" target="lib" />
-    
-    <file src="..\EntryPointDriver\bin\$Configuration$\**\Microsoft.Quantum.CsharpGeneration.EntryPointDriver.dll"
```
cesarzc, line 27: Out of curiosity, why are these file references no longer needed?

src/Simulation/CsharpGeneration/FindNuspecReferences.ps1
samarsha: This was added when I made EntryPointDriver part of the CsharpGeneration package, so that the package included dependencies for both projects. Since EntryPointDriver is no longer part of this package, the dependencies aren't needed anymore.

src/Simulation/CsharpGeneration/Microsoft.Quantum.CsharpGeneration.nuspec.template
samarsha: This PR is removing EntryPointDriver from the CsharpGeneration package, so the DLL for it should no longer be included here (it has its own package now).

src/Simulation/EntryPointDriver/Microsoft.Quantum.EntryPointDriver.csproj
samarsha: I think it is slightly different than EntryPointInfo, since that holds operation-specific data, while this class would have global constants that only have one value throughout the program, which makes a class feel weird. But yes I agree that it's still probably the best option.

src/Simulation/EntryPointDriver/Microsoft.Quantum.EntryPointDriver.csproj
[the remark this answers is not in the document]
bettinaheim: That's pretty cool! I'd suggest to keep the explicit AssemblyName property then (same as for all other projects).

src/Simulation/EntryPointDriver/Microsoft.Quantum.EntryPointDriver.csproj
bettinaheim: +1 for option 1, too.

src/Simulation/EntryPointDriver/Microsoft.Quantum.EntryPointDriver.csproj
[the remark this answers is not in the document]
samarsha: > (same as for all other projects)

Do you mean that I should add an `AssemblyName` so that EntryPointDriver is consistent with the other projects? I searched the repository, and out of 20 projects, I found only 4 that use an `AssemblyName` property, and only 2 of them where the `AssemblyName` is the same as the `.csproj`/`.fsproj` file name.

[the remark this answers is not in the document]
samarsha: Looks like there is also a `PackageId` property, which can override the `AssemblyName` property.

PR #227, opened 2020-05-29
Replace individual test projects with the solution in test.ps1

samarsha:
Now we'll never forget to add a new test project to the build script. :) Thanks for the suggestion @rmshaffer.


samarsha:
Seems like it is attempting to test Microsoft.Quantum.Xunit.csproj, even though that is not a test project.

PR #228, opened 2020-06-01
Simulator Classical Control Support

ScottCarda-MS:
This adds support for the simulators to process the ApplyIf methods of the classical control conversion logic. Previously, running Q# with if-statements would cause an error because the ApplyIf method the if-statement is converted into during compilation was not defined for the simulator runtime.


src/Simulation/Simulators/QuantumSimulator/ClassicalControl.cs
```diff
@@ -0,0 +1,290 @@
...
+            for (int i = 0; i < measurementResults.Count; i++)
+            {
+                if (measurementResults[i] != comparisonResults[i])
+                {
+                    equal = false;
+                    break;
+                }
+            }
+
+            return equal;
```
swernli, line 68: Is there any reason this can't just return the result directly instead of setting the variable and breaking out of the loop?
```suggestion
            for (int i = 0; i < measurementResults.Count; i++)
            {
                if (measurementResults[i] != comparisonResults[i])
                {
                    return false;
                }
            }

            return true;
```

ScottCarda-MS: done

src/Simulation/Simulators/QuantumProcessor/ClassicalControl.cs
```diff
@@ -83,7 +83,7 @@ private static void RunClause(ICallable op, OperationFunctor type, IQArray<Qubit
         }
 
 
-        public class QuantumProcessorApplyIfElse : Extensions.ApplyIfElseIntrinsic
+        public class QuantumProcessorApplyIfElse : Quantum.Simulation.QuantumProcessor.Extensions.ApplyIfElseIntrinsic
```
samarsha, line 86: Since this is in the `Microsoft.Quantum.Simulation.QuantumProcessor` namespace, is the long name actually needed?

src/Simulation/Simulators/QuantumProcessor/ClassicalControl.cs
```diff
@@ -83,7 +83,7 @@ private static void RunClause(ICallable op, OperationFunctor type, IQArray<Qubit
         }
```
samarsha, line 83: In ClassicalControl.qs, I saw that you changed the namespace back to `Microsoft.Quantum.Simulation.QuantumProcessor.Extensions`. Did it break things to use `Microsoft.Quantum.ClassicalControl` instead?

src/Simulation/Simulators/QuantumSimulator/ClassicalControl.cs
```diff
@@ -0,0 +1,288 @@
...
+using System;
+using Microsoft.Quantum.Simulation.Core;
+
+namespace Microsoft.Quantum.Simulation.Simulators
+{
+    public partial class QuantumSimulator
+    {
+        #region Static Methods
+
+        private static void RunClause(ICallable op, OperationFunctor type, IQArray<Qubit>? ctrls)
```
samarsha, line 13: I was looking around and it seems like most things in QuantumSimulator are missing doc comments, but they would be nice to have for new code. :)

src/Simulation/Simulators/QuantumProcessor/ClassicalControl.cs
ScottCarda-MS: Yah, that change requires cross-repo changes. This is the namespace change I brought up in the daily meeting, to which Bettina had suggested I make this a separate task. Once this PR is completed, the functionality will be provided and we can better address what the name should be changed to, and how to go about doing it.

samarsha: That makes sense, maybe there can be a TODO at the top of ClassicalControl.qs mentioning this? (Since the namespace looks inconsistent from the other files right now.)

src/Simulation/Simulators/QuantumProcessor/ClassicalControl.cs
ScottCarda-MS: Looks like it doesn't make a difference. I'll remove the extra qualifiers.

src/Simulation/Simulators/QuantumSimulator/ClassicalControl.cs
ScottCarda-MS: done

src/Simulation/Simulators/QuantumProcessor/ClassicalControl.cs
ScottCarda-MS: done

src/Simulation/Simulators/QuantumProcessor/ClassicalControl.cs
ScottCarda-MS: done

src/Simulation/Simulators/QuantumSimulator/ClassicalControl.cs
samarsha: Thanks!

src/Simulation/Simulators/QCTraceSimulator/Circuits/Interface.qs
```diff
@@ -78,6 +78,36 @@ namespace Microsoft.Quantum.Simulation.Simulators.QCTraceSimulators.Implementati
...
+
+    operation Interface_ApplyIfElseC (ctrls : Qubit[], measurementResult : Result, onResultZeroOp : (Unit => Unit is Ctl) , onResultOneOp : (Unit => Unit is Ctl)) : Unit {
+        body intrinsic;
+    }
+
+    operation Interface_ApplyIfElseCA (ctrls : Qubit[], measurementResult : Result, onResultZeroOp : (Unit => Unit is Ctl + Adj) , onResultOneOp : (Unit => Unit is Ctl + Adj)) : Unit {
+        body intrinsic;
+    }
 
+    operation Interface_ApplyConditionally (measurementResults : Result[], resultsValues : Result[], onEqualOp : (Unit => Unit) , onNonEqualOp : (Unit => Unit)) : Unit {
```
swernli, line 97: Could these be `internal`? Not necessary, but might clarify their intended usage.

ScottCarda-MS: I can make them `internal`, but I am wondering if that makes sense for something in the Interface.qs. I am trying to find out more about this file, if it is supposed to be a reflection of some pre-defined interface for the product, then maybe these should be located elsewhere.

ScottCarda-MS: The classical control handles are now internal.

ScottCarda-MS: Oops looks like making these internal invalidates the same test I set out to satisfy. There being internal makes them invisible to the test project. I've made them public again to satisfy the test.

swernli: Ah, ok. That’s good to know! Sorry for the detour.

PR #229, opened 2020-06-01
Update QubitManagerTests and avoid unnecessary IgnorableAssert

swernli:
The use of `Debug.Assert` was causing our tests to fail with an unhandled exception. Switching to `IgnorableAssert.Assert` keeps the debug assert behavior but also allow tests to disable them and test corner cases.

### DIFF
--- a/src/Simulation/Common/QubitManager.cs
+++ b/src/Simulation/Common/QubitManager.cs
@@ -259,7 +259,6 @@ namespace Microsoft.Quantum.Simulation.Common
         /// </summary>
         public virtual IQArray<Qubit> Allocate(long numToAllocate)
         {
-            IgnorableAssert.Assert(numToAllocate >= 0, "Attempt to allocate negative number of qubits.");
             if (numToAllocate < 0)
             {
                 throw new ArgumentException("Attempt to allocate negative number of qubits.");
@@ -297,7 +296,6 @@ namespace Microsoft.Quantum.Simulation.Common
             else
             {
                 long Occupied = (usedOnlyForBorrowing ? AllocatedForBorrowing : Allocated);
-                IgnorableAssert.Assert(qubits[qubit.Id] == Occupied, "Attempt to free qubit that has not been allocated.");
                 if (qubits[qubit.Id] != Occupied)
                 {
                     throw new ArgumentException("Attempt to free qubit that has not been allocated.");
@@ -337,7 +335,7 @@ namespace Microsoft.Quantum.Simulation.Common
         protected virtual void DisableOneQubit(Qubit qubit)
         {
             // Note: Borrowed qubits cannot be disabled.
-            IgnorableAssert.Assert(qubits[qubit.Id] == Allocated, "Attempt to disable qubit that has not been allocated.");
+            Debug.Assert(qubits[qubit.Id] == Allocated);
             qubits[qubit.Id] = Disabled;
             numDisabledQubits++;
 
@@ -416,7 +414,6 @@ namespace Microsoft.Quantum.Simulation.Common
         /// </summary>
         public virtual IQArray<Qubit> Borrow(long numToBorrow, IEnumerable<Qubit> excludedQubitsSortedById) // Note, excluded could be an array of Ids for efficiency, if it is convenient for compiler.
         {
-            IgnorableAssert.Assert(numToBorrow >= 0, "Attempt to borrow negative number of qubits.");
             if (numToBorrow < 0)
             {
                 throw new ArgumentException("Attempt to borrow negative number of qubits.");
--- a/src/Simulation/Simulators.Tests/QubitManagerTests.cs
+++ b/src/Simulation/Simulators.Tests/QubitManagerTests.cs
@@ -123,17 +123,33 @@ namespace Microsoft.Quantum.Simulation.Simulators.Tests
                 Assert.True(n_q.Length == 0);
             }
 
-            // Test allocating qubits over capacity
-            OperationsTestHelper.IgnoreDebugAssert(() =>
+            // NOTE: The below tests trigger exceptions, which but the QubitManager into a bad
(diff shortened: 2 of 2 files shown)


swernli:
The test that has been failing is over in [QubitManagerTests](https://github.com/microsoft/qsharp-runtime/blob/master/src/Simulation/Simulators.Tests/QubitManagerTests.cs#L127), and tried to verify behavior when qubits are allocated over capacity. The debug assert meant the test actually crashed, which because of how our pipeline works meant it was being treated as a success without looking at the actual result. The test already used `IgnorableAssert` mechanics, so just updated the manager was sufficient to resolve the failure.

src/Simulation/Common/QubitManager.cs
```diff
@@ -116,14 +116,14 @@ private void ExtendQubitArray()
             {
                 this.qubits[i] = i + 1;
             }
-            Debug.Assert(this.qubits[NumQubits - 1] == None);
+            IgnorableAssert.Assert(this.qubits[NumQubits - 1] == None, "Qubitis should be initialized to None.");
```
bettinaheim, line 119: typo

swernli: Fixed!

avasch01:
This does not look like the right fix. The asserts in private functions are not ignorable. They are hard asserts. Let's talk on Wednesday so that I can understand exactly how the test has been failing (it worked OK for a couple of years, right?) and try to come up with the right fix, most probably in the public entry points.

---
In reply to: [637018971](https://github.com/microsoft/qsharp-runtime/pull/229#issuecomment-637018971) [](ancestors = 637018971)

swernli:
@avasch01 Sure thing! Let's follow up to find the right solution here. For what it's worth, in the effort to gather more info on how and why the test was failing, I discovered a bug in our build scripts, including test.ps1, that explains why this doesn't hit during normal pipeline execution. It turns out that the tests are executed using this powershell:
```
    dotnet test $(Join-Path $PSScriptRoot $project) `
        -c $Env:BUILD_CONFIGURATION `
        -v $Env:BUILD_VERBOSITY `
        --logger trx `
        /property:DefineConstants=$Env:ASSEMBLY_CONSTANTS `
        /property:Version=$Env:ASSEMBLY_VERSION
```
Note where it uses the environment variable `ASSEMBLY_CONSTANTS` as input to the `DefineConstants` property of the build. `DefineConstants` does not perform an append of any constants, but rather acts as an override. Because we don't actually define `ASSEMBLY_CONSTANTS` as part of our build scripts (I assume it was just put there for later extensibility, or perhaps leftover from an early build setup), the line that actually gets executed by the test script has these two properties: `/property:DefineConstants= /property:Version=0.0.1.0`. This ends up undefining all build constants during compilation and test execution, including the build constants that identify it as a debug build. As a result, all debug asserts are disabled, and the existing tests seem to run fine. Removing this constant override shows that our tests are actually failing/crashing due to debug asserts.

I'll file an issue about this larger build issue so we can track and fix that separate from this specific assert, but it's likely that we will need to fix several tests that fail due to debug asserts in order for the update to the test script to build clean.

@bettinaheim @anpaz-msft for FYI.

swernli:
@avasch01 After our discussion, I've updated the approach to this fix. For others, the `IgnorableAssert` patterns isn't really proper here, and instead I've left the internal asserts as `Debug.Assert` and adjusted the tests to avoid the bad state. The previous test tried to test multiple exceptional states, but once a `QubitManager` hits an exception, it is no longer usable and its internal state is unreliable, so these cases should not be in a single test.

In addition, we had a few `IgnorableAssert` uses that were immediately before a matching exception case, which doesn't really add value. I removed these in favor of relying on the exception, which the tests were already set up to verify.

src/Simulation/Simulators.Tests/QubitManagerTests.cs
```diff
@@ -122,18 +122,50 @@ public void TestQubitManager()
...
-            // Test allocating qubits over capacity
-            OperationsTestHelper.IgnoreDebugAssert(() =>
-            {
-                IQArray<Qubit> n_q;
+        /// <summary>
+        /// Tests for exceptional cases or over allocation.
+        /// </summary>
+        [Fact]
+        public void TestAllocateTooManyQubits()
+        {
```
avasch01, line 132: It is Ok, but I would not take this (and below) out as a complete separate test. I would keep it as a part of the above test function, just make it a separate scope {} with its own Qubit manager. Reason: not to add 4 new tests to the "set" of tests in the TestExplorer window. Kind of polutes it. If it grows by 4 with every little test, the list becomes too long. Up to you, though.

avasch01:
Great fix! Now it nails the problem exactly where it happens! Thank you.

---
In reply to: [638572741](https://github.com/microsoft/qsharp-runtime/pull/229#issuecomment-638572741) [](ancestors = 638572741)

src/Simulation/Common/QubitManager.cs
```diff
@@ -259,7 +259,6 @@ public virtual Qubit Allocate()
         /// </summary>
         public virtual IQArray<Qubit> Allocate(long numToAllocate)
         {
-            IgnorableAssert.Assert(numToAllocate >= 0, "Attempt to allocate negative number of qubits.");
             if (numToAllocate < 0)
```
avasch01, line 262: I would strongly suggest against removing asserts like this. In fact, just revert changes to this file. The fix you did to the test file is proper and resolves all issues. As for this, I do not see any harm in the presence of assert, and leaving it in helps debugging (without adding any drawbacks).

src/Simulation/Simulators.Tests/QubitManagerTests.cs
swernli: I've always tended toward more tests with smaller scopes since it allows targeted investigation by running a single test case either from the editor or the cmdline, though I can see the argument the other way. I'll move these back into the above function.

PR #231, opened 2020-06-01
Change to allow a qubit to be released if it is measured

AniketDalvi:
This PR is for a change I made to allow a qubit to be released if it has been measured and no operation has been made on it after the measurement. This involved adding an isMeasured field to Qubit, and setting its values appropriately. I also added 2 unit tests to confirm the desired behavior.


src/Simulation/Simulators/QuantumSimulator/QubitManager.cs
```diff
@@ -64,12 +64,14 @@ protected override void ReleaseOneQubit(Qubit qubit, bool usedOnlyForBorrowing)
                 if (qubit != null)
                 {
                     bool areAllReleasedQubitsZero = ReleaseOne(this.SimulatorId, (uint)qubit.Id);
-                    if (!areAllReleasedQubitsZero && throwOnReleasingQubitsNotInZeroState)
+                    bool isQubitMeasured = qubit.IsMeasured;
```
bettinaheim, line 67: This seems like a rather unnecessary definition of a variable to me, given that qubit.IsMeasured is as readable and also not longer than the variable name.

src/Simulation/Simulators.Tests/QuantumSimulatorTests/QubitReleaseTest.cs
```diff
@@ -0,0 +1,36 @@
...
+        }
+
+        //test to check that qubit can be released if measured
+        [Fact]
+        public async Task MeasuredQubitReleaseTest()
+        {
+            var sim = new QuantumSimulator();
+
+            //should not throw an exception
+            await UsingMeasuredQubitCheck.Run(sim);
```
bettinaheim, line 33: It would be good to somehow also check that the state vector is processed appropriately. Let me take a quick look at what options we have.

src/Simulation/Simulators/QuantumSimulator/QubitManager.cs
```diff
@@ -64,12 +64,14 @@ protected override void ReleaseOneQubit(Qubit qubit, bool usedOnlyForBorrowing)
                 if (qubit != null)
                 {
                     bool areAllReleasedQubitsZero = ReleaseOne(this.SimulatorId, (uint)qubit.Id);
```
anpaz-msft, line 66: why is this named `areAllReleasedQubitsZero` if it's only checking the status of the last qubit released? :)
it threw me off a little.

cgranade:
> 2. Releasing a Qubit measured in a different base, the qubit is still in super-position on the Z bases but it doesn't throw:
>    ```
>         using (q = Qubit()) {
>             H(q);
>             let _ = Measure([PauliX], [q]);
>         }
>    ```
Quick comment, but this should be entirely OK, as the qubit is still in a deterministically known state at the point that it is released, such that appropriate instructions can be automatically added to reset it to the |0⟩ state.

swernli:
For a bit of extra context, Aniket already found that the simulator handles release of qubits properly. See this code from the [simulator.hpp](https://github.com/microsoft/qsharp-runtime/blob/master/src/Simulation/Native/src/simulator/simulator.hpp#L99) elsewhere the repo:
```
    bool release(unsigned q)
    {
      recursive_lock_type l(mutex());
      bool allok = isclassical(q);
      if (allok)
      allok = (psi.getvalue(q)==false);
      else
      M(q);
      psi.release(q);
      return allok;
    }
```
When the qubit is released, the simulator will check if it's in a classical value (defined as |0> or |1> in the Z-basis), and if not, it performs a measure. That measure has code to explicitly collapses the state. Then it returns the boolean indicating whether the qubit was originally in a classical state and whether that classical state was |0>, then the higher level code in QubitManager would throw. So the simulator doesn't actually care and will handle the qubit no matter what state it is in, it's just a matter of what the QubitManager will allow, hence the fix that Aniket is building.

anpaz-msft:
@swernli, yes, the simulator works correctly, and there is an option to disable this behavior (is just *on* by default).
The reason is on is because this is a common cause for invalid algorithms. You could question if there is a real value on doing this, since at the end users just get around this by calling `Reset`; I believe not having to call `Reset` but still throwing un-measured Qubits makes it more robust and makes the code cleaner.

cgranade:
> The reason is on is because this is a common cause for invalid algorithms.

Agreed, but there's also value in that avoid reset by coherently unpreparing qubits is a common and useful optimization. For example, the target qubit in Deutsch–Jozsa is deterministically in |−⟩ after the query, such that an `H` and `X` operation call are sufficient to return the target qubit to |0⟩ for reuse, avoiding a potentially expensive measurement. Similarly, the same phase kickback trick is often used to perform common reflections, such that the state of a scratch qubit used in reflections can be efficiently unprepared without measurement or resetting.

swernli:
Agreed on all points. I was just clarifying the simulator behavior because of some phrasing; as we are discussing this issue, we keep talking about what is "ok" or what "works" but any and all of these approaches can work, it's about what patterns we want to allow and encourage. So the question we should be asking ourselves is: what do we want the simulator to teach people about how to manage their qubits? What usage patterns do we want the simulator to reject so that it forces developers to adjust their programs?

The question of whether or not to clear the `IsMeasured` property when a qubit is used as a control is a very interesting one to me. Is using something as a control a "usage" in the sense that it invalidates the collapsing that occurred during the previous measure? Is that something that can vary between different kinds of qubits or hardware implementations? It's easy to ignore control qubits in the code, so either way can be quickly implemented. But I honestly don't know which way is correct given our goals of using this exception as instructive for algorithm correctness.

cgranade:
> The question of whether or not to clear the `IsMeasured` property when a qubit is used as a control is a very interesting one to me. Is using something as a control a "usage" in the sense that it invalidates the collapsing that occurred during the previous measure? Is that something that can vary between different kinds of qubits or hardware implementations? It's easy to ignore control qubits in the code, so either way can be quickly implemented. But I honestly don't know which way is correct given our goals of using this exception as instructive for algorithm correctness.

Suppose after a measurement, the qubit `left` is left deterministically in the |+⟩ state, and then used as a control for a `CNOT` operation `CNOT(left, right)`, where the initial state of `right : Qubit` is |0⟩. Immediately following that, the joint state of `[left, right]` is (|00⟩ + |11⟩) / √2.
Even worse, if you instead do a `CNOT(left, right)` with `right` starting in |−⟩, then the final state of the register is |−−⟩ not |+−⟩.

Put differently, the word "control" in the context of a controlled quantum operation is at most a mnemonic. A two-qubit operation can in general change the joint state of both qubits, even if one is only a "control." You can even reverse which qubit is a control and which is a target using `H` operations:

```Q#
within {
    H(left); H(right);
} apply {
    CNOT(right, left);
}

// is equivalent to
CNOT(left, right);
```

From that perspective, I would strongly argue that `IsMeasured` has to be cleared when a qubit is used as a control in a quantum operation.

swernli:
That's an excellent explanation, thanks @cgranade !

cgranade:
> That's an excellent explanation, thanks @cgranade !

Thanks, happy to help!

src/Simulation/Simulators.Tests/QuantumSimulatorTests/QubitReleaseTest.cs
bettinaheim: Per @swernli's comment this is handled in the C++ piece. It should hence be sufficient to just properly document when the behavior here.

cgranade:
I have to say, this is an exciting change from the libraries perspective! We've gotten feedback about the variety of different ways that measurement is used in the libraries, and in particular, that it's not clear when a reset is needed and not without looking at detailed API docs. This change would allow us to adopt the convention that measurements other than MReset* do not reset ever.

src/Simulation/Simulators/QuantumSimulator/QuantumSimulator.cs
```diff
@@ -93,6 +93,8 @@ static void CheckAngle(double angle)
         void CheckQubit(Qubit q1)
```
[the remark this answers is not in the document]
AniketDalvi: I am unsure of why it is in the current state. However, before making a change around this, I would let the original author comment on this, as this was as is when I looked at it.

src/Simulation/Simulators/QuantumSimulator/QubitManager.cs
AniketDalvi: Yes, will incorporate this!

src/Simulation/Simulators.Tests/QuantumSimulatorTests/QubitReleaseTest.cs
AniketDalvi: Will add some documentation around these tests

src/Simulation/Simulators/QuantumSimulator/QubitManager.cs
AniketDalvi: Will rename appropriately

AniketDalvi:
After looking at this thread, I am attempting to summarize the required changes:
1. The behavior for control qubits can stay as is, and thus the isMeasured should stay as false for the controls after a multi qubit operation
2. As far as properly releasing goes, the C++ side of the code handles it correctly
3. The only changes in behavior that would be required is around the DumpManager behavior which calls CheckQubit, but should be allowed to be Release if it has been previously Measured
@anpaz-msft @swernli @cgranade

swernli:
That summary looks good to me (plus the renaming feedback which you already mentioned you'd be addressing).

cgranade:
> After looking at this thread, I am attempting to summarize the required changes:
> 
> 1. The behavior for control qubits can stay as is, and thus the isMeasured should stay as false for the controls after a multi qubit operation

That's correct; control qubits aren't special, such that the postcondition represented by `isMeasured` can be violated by passing qubits as controls. In particular, if I understand correctly, `isMeasured` represents the postcondition that the qubit is: (0) not entangled with any other qubits, and is (1) in a deterministically known state, such that it can be coherently returned to the |0⟩ state (that is, without further measurement).

src/Simulation/Simulators/QuantumSimulator/QuantumSimulator.cs
ScottCarda-MS, line 93: For the methods in this file that set `IsMeasured` to `false`, could you add to their summaries that they do this?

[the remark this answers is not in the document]
bettinaheim: If you don't here back, please add that change - I will be to blame for any resulting issues. :)

bettinaheim:
@anpaz-msft Could you please take another look whether the changes you requested have been addressed?

PR #233, opened 2020-06-04
Throw an AzureQuantumException (or an exception that derives from it) when there is a failure related to Azure

cesarzc:
This change creates two exceptions that derive from `AzureQuantumExecption`:
- `WorkspaceClientException`: used when there is a problem related to job management (e.g. submit, get).
- `StorageClientException`: used when there is a problem related to Azure storage (e.g. upload/download blobs).

These exceptions are used in the workspace and storage clients respectively when a failure occurs.

### DIFF
--- a/src/Azure/Azure.Quantum.Client/JobManagement/Workspace.cs
+++ b/src/Azure/Azure.Quantum.Client/JobManagement/Workspace.cs
@@ -195,7 +195,7 @@ namespace Microsoft.Azure.Quantum
             catch (Exception ex)
             {
                 throw new WorkspaceClientException(
-                    $"Could not cancel job",
+                    "Could not cancel job",
                     SubscriptionId,
                     ResourceGroupName,
                     WorkspaceName,
--- a/src/Azure/Azure.Quantum.Client/JobManagement/Workspace.cs
+++ b/src/Azure/Azure.Quantum.Client/JobManagement/Workspace.cs
@@ -94,7 +94,7 @@ namespace Microsoft.Azure.Quantum
             catch (Exception ex)
             {
                 throw new WorkspaceClientException(
-                    $"Could not create an access token provider",
+                    "Could not create an access token provider",
                     SubscriptionId,
                     ResourceGroupName,
                     WorkspaceName,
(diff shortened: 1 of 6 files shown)


src/Azure/Azure.Quantum.Client/Storage/StorageHelper.cs
```diff
@@ -81,14 +120,27 @@ public StorageHelper(string connectionString)
...
+                    .CreateCloudBlobClient()
+                    .GetContainerReference(containerName)
+                    .GetBlobReference(blobName);
+
+                return blob.Uri + blob.GetSharedAccessSignature(adHocSAS);
+            }
+            catch (Exception ex)
+            {
+                throw new StorageClientException(
+                    "Could not get BLOB sas URI",
```
samarsha, line 137:
```suggestion
                    "Could not get BLOB shared access signature URI",
```

(Alternatively, "SAS" should be capitalized, but that is a lot of acronyms in a row.)

src/Azure/Azure.Quantum.Client/Storage/StorageHelper.cs
```diff
@@ -103,11 +155,23 @@ public StorageHelper(string connectionString)
...
-            CloudBlobContainer container = this.storageAccount.CreateCloudBlobClient().GetContainerReference(containerName);
-            return container.Uri + container.GetSharedAccessSignature(adHocPolicy, null);
+                // Generate the shared access signature on the container, setting the constraints directly on the signature.
+                CloudBlobContainer container = this.storageAccount.CreateCloudBlobClient().GetContainerReference(containerName);
+                return container.Uri + container.GetSharedAccessSignature(adHocPolicy, null);
+            }
+            catch (Exception ex)
+            {
+                throw new StorageClientException(
+                    "Could not get BLOB container sas URI",
```
samarsha, line 169:
```suggestion
                    "Could not get BLOB container shared access signature URI",
```

src/Azure/Azure.Quantum.Client/JobManagement/Workspace.cs
```diff
@@ -76,21 +79,53 @@ public class Workspace : IWorkspace
...
+            catch (Exception ex)
+            {
+                throw new WorkspaceClientException(
+                    $"Could not create an access token provider",
+                    SubscriptionId,
+                    ResourceGroupName,
+                    WorkspaceName,
+                    BaseUri,
+                    string.Empty,
+                    ex);
```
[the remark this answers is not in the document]
cesarzc: Yes! I think that is a good idea, I will make the change.

Issue #232, opened 2020-06-02
build.ps1 and test.ps1 scripts incorrectly override all build constants, hiding failures


swernli:
## Describe the bug
Our build and test scripts use `/property:DefineConstants=$Env:ASSEMBLY_CONSTANTS` in an attempt to provide extensibility, but `DefineConstants` acts as an override by default (not an append). So when the environment variable is empty, as it is during our pipeline runs, this will undefine all build constants, which disables any debug asserts, conditionally compiled code or logging, etc. If this part of the build and test script is removed, some of the resulting tests hit debug asserts and fail in ways that weren't caught before.

## To Reproduce
Compare the result of running these two commands from the root:
```
dotnet test .\src\Simulation\Simulators.Tests\Tests.Microsoft.Quantum.Simulation.Simulators.csproj /property:DefineConstants= 
```
vs
```
dotnet test .\src\Simulation\Simulators.Tests\Tests.Microsoft.Quantum.Simulation.Simulators.csproj
```
The first command will execute all tests and report success, while the second command aborts the test run due to a crash from a debug assert.

## Expected behavior
Build constants should not be unconditionally undefined as a result of the build/test scripts, and any tests that fail due to asserts should be investigated and fixed either in the test or the product code.

**Additional context**
NOTE: In my experiments, I only actually saw one test failure due to an assert, which was how I first ended up trying to investigate previous test performance (see https://github.com/microsoft/qsharp-runtime/pull/229). Hopefully that means we can fix the script and one test to get back to clean tests and builds.

swernli:
See https://github.com/dotnet/runtime/commit/613f9b270deebd35656637399ac1d77627368962 from the dotnet repo for an example of how others have hit and resolved this bug. Because our use of `DefineConstants` is not in the project file itself, this exact fix wont work for us, but perhaps we could have a new parameter, `AdditionalDefineConstants`, that we append using this mechanism in the project file and then define at the commandline from the environment via the powershell.

PR #234, opened 2020-06-04
Fixing scripts to not undefine all build constants

swernli:
This updates the build.ps1, test.ps1, and pack.ps1 scripts to only provide the `DefineConstants` property to dotnet cli when there is an actual value in the environment variable. This ensures that it won't end up clearing all of the defined build constants, possibly hiding build or test failures due to the missing DEBUG constant.

This fixes https://github.com/microsoft/qsharp-runtime/issues/232

### DIFF
--- a/build/build.ps1
+++ b/build/build.ps1
@@ -21,12 +21,17 @@ function Build-One {
     );
 
     Write-Host "##[info]Building $project ($action)..."
-    Invoke-Expression ("dotnet $action $(Join-Path $PSScriptRoot $project) " +
-        "-c $Env:BUILD_CONFIGURATION " +
-        "-v $Env:BUILD_VERBOSITY " +
-        "$(if ($Env:ASSEMBLY_CONSTANTS) { "/property:DefineConstants=$Env:ASSEMBLY_CONSTANTS" }) " +
-        "/property:Version=$Env:ASSEMBLY_VERSION " +
-        "/property:QsharpDocsOutputPath=$Env:DOCS_OUTDIR")
+    if ("" -ne "$Env:ASSEMBLY_CONSTANTS") {
+        $args = @("/property:DefineConstants=$Env:ASSEMBLY_CONSTANTS");
+    }  else {
+        $args = @();
+    }
+    dotnet $action $(Join-Path $PSScriptRoot $project) `
+        -c $Env:BUILD_CONFIGURATION `
+        -v $Env:BUILD_VERBOSITY  `
+        @args `
+        /property:Version=$Env:ASSEMBLY_VERSION `
+        /property:QsharpDocsOutputPath=$Env:DOCS_OUTDIR;
 
     if ($LastExitCode -ne 0) {
         Write-Host "##vso[task.logissue type=error;]Failed to build $project."
@@ -41,4 +46,3 @@ Build-One 'build' '../Simulation.sln'
 if (-not $all_ok) {
     throw "At least one project failed to compile. Check the logs."
 }
-
--- a/build/build.ps1
+++ b/build/build.ps1
@@ -21,12 +21,12 @@ function Build-One {
     );
 
     Write-Host "##[info]Building $project ($action)..."
-    dotnet $action (Join-Path $PSScriptRoot $project) `
-        -c $Env:BUILD_CONFIGURATION `
-        -v $Env:BUILD_VERBOSITY `
-        /property:DefineConstants=$Env:ASSEMBLY_CONSTANTS `
-        /property:Version=$Env:ASSEMBLY_VERSION `
-        /property:QsharpDocsOutputPath=$Env:DOCS_OUTDIR
+    Invoke-Expression ("dotnet $action $(Join-Path $PSScriptRoot $project) " +
+        "-c $Env:BUILD_CONFIGURATION " +
+        "-v $Env:BUILD_VERBOSITY " +
+        "$(if ($Env:ASSEMBLY_CONSTANTS){"/property:DefineConstants=$Env:ASSEMBLY_CONSTANTS"}) " +
+        "/property:Version=$Env:ASSEMBLY_VERSION " +
+        "/property:QsharpDocsOutputPath=$Env:DOCS_OUTDIR")
 
     if  ($LastExitCode -ne 0) {
         Write-Host "##vso[task.logissue type=error;]Failed to build $project."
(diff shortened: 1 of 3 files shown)


build/build.ps1
```diff
@@ -21,12 +21,12 @@ function Build-One {
...
-        -v $Env:BUILD_VERBOSITY `
-        /property:DefineConstants=$Env:ASSEMBLY_CONSTANTS `
-        /property:Version=$Env:ASSEMBLY_VERSION `
-        /property:QsharpDocsOutputPath=$Env:DOCS_OUTDIR
+    Invoke-Expression ("dotnet $action $(Join-Path $PSScriptRoot $project) " +
+        "-c $Env:BUILD_CONFIGURATION " +
+        "-v $Env:BUILD_VERBOSITY " +
+        "$(if ($Env:ASSEMBLY_CONSTANTS){"/property:DefineConstants=$Env:ASSEMBLY_CONSTANTS"}) " +
+        "/property:Version=$Env:ASSEMBLY_VERSION " +
+        "/property:QsharpDocsOutputPath=$Env:DOCS_OUTDIR")
```
samarsha, line 29: I have been trying to look for a way to do this in PowerShell that doesn't involve string concatenation, but I have failed. :) I guess it might be the best way here.

build/build.ps1
```diff
@@ -21,12 +21,12 @@ function Build-One {
...
-    dotnet $action (Join-Path $PSScriptRoot $project) `
-        -c $Env:BUILD_CONFIGURATION `
-        -v $Env:BUILD_VERBOSITY `
-        /property:DefineConstants=$Env:ASSEMBLY_CONSTANTS `
-        /property:Version=$Env:ASSEMBLY_VERSION `
-        /property:QsharpDocsOutputPath=$Env:DOCS_OUTDIR
+    Invoke-Expression ("dotnet $action $(Join-Path $PSScriptRoot $project) " +
+        "-c $Env:BUILD_CONFIGURATION " +
+        "-v $Env:BUILD_VERBOSITY " +
+        "$(if ($Env:ASSEMBLY_CONSTANTS){"/property:DefineConstants=$Env:ASSEMBLY_CONSTANTS"}) " +
```
samarsha, line 27:
```suggestion
        "$(if ($Env:ASSEMBLY_CONSTANTS) { "/property:DefineConstants=$Env:ASSEMBLY_CONSTANTS" }) " +
```

build/pack.ps1
```diff
@@ -36,15 +36,15 @@ function Pack-One() {
...
-        /property:DefineConstants=$Env:ASSEMBLY_CONSTANTS `
-        /property:Version=$Env:NUGET_VERSION `
-        $option1 `
-        $option2 `
-        $option3
+    Invoke-Expression ("dotnet pack $project " +
+        "-o $Env:NUGET_OUTDIR " +
+        "-c $Env:BUILD_CONFIGURATION " +
+        "-v detailed " +
+        "$(if ($Env:ASSEMBLY_CONSTANTS){"/property:DefineConstants=$Env:ASSEMBLY_CONSTANTS"}) " +
```
samarsha, line 43:
```suggestion
        "$(if ($Env:ASSEMBLY_CONSTANTS) { "/property:DefineConstants=$Env:ASSEMBLY_CONSTANTS" }) " +
```

build/test.ps1
```diff
@@ -19,12 +19,12 @@ function Test-One {
...
-        -c $Env:BUILD_CONFIGURATION `
-        -v $Env:BUILD_VERBOSITY `
-        --logger trx `
-        /property:DefineConstants=$Env:ASSEMBLY_CONSTANTS `
-        /property:Version=$Env:ASSEMBLY_VERSION
+    Invoke-Expression ("dotnet test (Join-Path $PSScriptRoot $project) " +
+        "-c $Env:BUILD_CONFIGURATION " +
+        "-v $Env:BUILD_VERBOSITY " +
+        "--logger trx " +
+        "$(if ($Env:ASSEMBLY_CONSTANTS){"/property:DefineConstants=$Env:ASSEMBLY_CONSTANTS"}) " +
```
samarsha, line 26:
```suggestion
        "$(if ($Env:ASSEMBLY_CONSTANTS) { "/property:DefineConstants=$Env:ASSEMBLY_CONSTANTS" }) " +
```

build/build.ps1
swernli: Sadly, it is also the best I could come up with too...

build/build.ps1
swernli: I went ahead and made this an the other formatting changes, as well as ran the autoformatter in VSCode that caught a few more spaces in if-statements that were missing. Thanks!

build/build.ps1
```diff
@@ -21,14 +21,14 @@ function Build-One {
...
-        /property:Version=$Env:ASSEMBLY_VERSION `
-        /property:QsharpDocsOutputPath=$Env:DOCS_OUTDIR
-
-    if  ($LastExitCode -ne 0) {
+    Invoke-Expression ("dotnet $action $(Join-Path $PSScriptRoot $project) " +
+        "-c $Env:BUILD_CONFIGURATION " +
+        "-v $Env:BUILD_VERBOSITY " +
+        "$(if ($Env:ASSEMBLY_CONSTANTS) { "/property:DefineConstants=$Env:ASSEMBLY_CONSTANTS" }) " +
+        "/property:Version=$Env:ASSEMBLY_VERSION " +
+        "/property:QsharpDocsOutputPath=$Env:DOCS_OUTDIR")
```
cgranade, line 29: It is often easier to use PowerShell's argument splat operators to expand a variable list of arguments than to use `Invoke-Expression`. There's some subtle bugs that can be introduced by `Invoke-Expression` swallowing `LASTEXITCODE` that often make spalatting a simpler solution.
```suggestion
    if ("" -ne "$Env:ASSEMBLY_CONSTANTS") {
        $args = @("/property:DefineConstants=$Env:ASSEMBLY_CONSTANTS");
    }  else {
        $args = @();
    }
    dotnet $action $(Join-Path $PSScriptRoot $project) `
        -c $Env:BUILD_CONFIGURATION `
        -v $Env:BUILD_VERBOSITY  `
        @args `
        /property:Version=$Env:ASSEMBLY_VERSION `
        /property:QsharpDocsOutputPath=$Env:DOCS_OUTDIR;
```
Using `@args` instead of `$args` there expands the contents of `$args` as a list of arguments; if `$args` is empty, then `@args` is a no-op.

swernli: Oh, neat! I'll have to try that instead.

PR #244, opened 2020-06-08
Tests for Classical Control Support on Simulator and Resource Estimator

ScottCarda-MS:
Added two execution tests that test the simulator and resource estimator have support for the classical control API.

Work Item: [AB#15598](https://ms-quantum.visualstudio.com/cd78fdaf-a553-43e7-b11f-fb8d51709d64/_workitems/edit/15598)


src/Simulation/Simulators.Tests/Tests.Microsoft.Quantum.Simulation.Simulators.csproj
```diff
@@ -38,12 +41,17 @@
   <Target Name="PostBuild" AfterTargets="PostBuildEvent">
     <PropertyGroup>
       <_ExeDir>$(MSBuildThisFileDirectory)TestProjects\QsharpExe\built\</_ExeDir>
+      <_ClassicControlDir>$(MSBuildThisFileDirectory)TestProjects\ClassicControlSupportTest\built\</_ClassicControlDir>
+      <ExecutionTarget>$(MSBuildThisFileDirectory)..\TestTargets\Simulation\Example\bin\$(Configuration)\netcoreapp3.1\Example.dll</ExecutionTarget>
```
bettinaheim, line 45: What were you planning to use that for?

src/Simulation/Simulators.Tests/TestProjects/ClassicControlSupportTest/Main.qs
```diff
@@ -0,0 +1,13 @@
+﻿namespace Quantum.ClassicControlSupportTest {
+    open Microsoft.Quantum.Intrinsic;
+    open Microsoft.Quantum.Simulation.QuantumProcessor.Extensions;
+    @EntryPoint()
+    operation Main () : Unit {
+        using (q = Qubit())
+        {
+            H(q);
+            let result = M(q);
+            ApplyIfOne(result, (X, q));
```
bettinaheim, line 10: Instead of just a single test with an explicit call to ApplyIfOne, you should instead target the ClassicControlSupportTest.csproj project to the Honeywell processor, and test all the cases that you also test on the compiler ([here](https://github.com/microsoft/qsharp-compiler/blob/master/src/QsCompiler/Tests.Compiler/TestCases/ClassicalControl.qs)).

ScottCarda-MS: Am I allowed to put something specific to Honeywell in this repo? Are we allowed to even talk about that here on GitHub?

src/Simulation/Simulators.Tests/CoreTests.cs
```diff
@@ -43,6 +43,30 @@ public void BasicExecution()
             Assert.Empty(error.ToString().Trim());
         }
 
+        [Fact]
+        public void ClassicControlSimulatorSupport()
```
bettinaheim, line 47: Also a minor thing: Everywhere else the code talks about ClassicalControl opposed to ClassicControl, so I'd suggest to adapt the name for both tests to match that.

src/Simulation/Simulators.Tests/Tests.Microsoft.Quantum.Simulation.Simulators.csproj
ScottCarda-MS: I copied what was done for the QsharpExe project. Should I have something else here?

Simulation.sln
```diff
@@ -1,423 +1,442 @@
...
+		{E2F30496-19D8-46A8-9BC0-26936FFE70D2} = {A567C185-A429-418B-AFDE-6F1785BA4A77}
+		{8AD4386C-89CE-4DD7-B919-15E35D32324A} = {09C842CB-930C-4C7D-AD5F-E30DE4A55820}
+		{7256B986-6705-42FC-9F57-485D72D9DE51} = {09C842CB-930C-4C7D-AD5F-E30DE4A55820}
+		{A85277B3-4E07-4E15-8F0C-07CC855A3BCB} = {09C842CB-930C-4C7D-AD5F-E30DE4A55820}
+		{46278108-D247-4EFC-AC34-23D4A676F62F} = {09C842CB-930C-4C7D-AD5F-E30DE4A55820}
+	EndGlobalSection
+	GlobalSection(ExtensibilityGlobals) = postSolution
+		SolutionGuid = {929C0464-86D8-4F70-8835-0A5EAF930821}
+	EndGlobalSection
+EndGlobal
```
samarsha, line 442: I'm worried that the line endings change for this file will cause merge conflicts with other PRs. It looks like it changed from LF to CRLF - I think generally things in the repo should be LF, can this be changed back?

src/Simulation/Simulators.Tests/TestProjects/ClassicControlSupportTest/Main.qs
bettinaheim: Yes. :)

ScottCarda-MS: Ah okay, I thought everything related to specific partners was confidential, but it seems that is in the past now.

src/Simulation/Simulators.Tests/Tests.Microsoft.Quantum.Simulation.Simulators.csproj
bettinaheim: Looking at the diff, the ExecutionTarget property seems to have been added and is not used by anything.

Simulation.sln
ScottCarda-MS: It being in LF caused that exact issue for me. I will see about changing it to CRLF.

src/Simulation/Simulators.Tests/TestProjects/ClassicControlSupportTest/Main.qs
ScottCarda-MS: I've updated the tests to reference each of the API operations. I don't think targeting a specific processor is going to have an effect here as the API referenced directly.

src/Simulation/Simulators.Tests/CoreTests.cs
ScottCarda-MS: done

Simulation.sln
samarsha: See [here](https://help.github.com/en/github/using-git/configuring-git-to-handle-line-endings) which recommends configuring Git to check out as CRLF but commit as LF. I think it would be better to keep it as LF in the repo to avoid merge conflicts, but it looks like Git didn't automatically translate the line endings when it was committed in this PR.

ScottCarda-MS: The line endings have been changed to LF.

src/Simulation/Simulators.Tests/Tests.Microsoft.Quantum.Simulation.Simulators.csproj
ScottCarda-MS: I've removed the ExecutionTarget tag.

src/Simulation/Simulators.Tests/TestProjects/ClassicControlSupportTest/Main.qs
bettinaheim: That still doesn't test whether a project that has been compiled for Honeywell will actually execute on the QuantumSimulator; if e.g. the compiler changes what operations it relies on when translating if-statements, these tests will still pass fine.

swernli: Hmm... Scott and I had discussed the possibility of doing a more general test of the conditional rewrite logic and thought that might be out of scope for this change, but it sounds like that was more along the lines of what you were imagining here. It felt a little awkward to use Honeywell specifically for testing this. Is there any merit in creating a test target that just invokes all rewrite steps without being specific to a partner or has any connections to submission logic? That was the other idea we discussed, and while that addition would have to go into the compiler repo first it would give us a nice way to invoke everything needed for testing without relying on a specific partner target.

bettinaheim: Some of the steps may be mutually exclusive. 
What we want to test is always the functionality, and not the implementation - and certainly not the implementation in another repository...
The functionality here we want to provide that should be tested is that for all execution target, the built executable will execute on the QuantumSimulator and the Resources estimator as well. Testing that exhaustively is a slightly ambitious endeavor (there are general frameworks for other languages to generate "sample" programs for similar purposes, but I haven't looked into what sort of tooling exist for us). Hence my suggestion is to for now to be a bit less ambitious and test a handful of programs that we know will be modified during compilation and test those for all execution targets. Setting up a project compiled for Honeywell, one compiled for QCI and on for IonQ actually seem like to right thing to do to me. As a compromise to keep the testing effort reasonable (though strictly speaking that also deviates from testing exactly the functionality we want) and not literally generate a separate compilation unit for each program and target to test, you could leverage the NoEntryPoint property we've introduced specifically for testing purposes. The other option would be to do an in-memory compilation, but that would bypass the Sdk part of the setup then, so also not fully reflect the functionality.

bettinaheim: As a side note: The reason that the test for the QsharpExe actually does an out of process invocation is because we specifically wanted to test the command line functionality. For the tests here, it would be fine to go with Q# unit tests (making the above suggested compromise of relying on the guarantee of NoEntryPoint that all code will still be fully targeted).

ScottCarda-MS:
We won't be needing these test.

PR #251, opened 2020-06-13
Fix NUGET_VERSION issues

cgranade:
This PR fixes two places where NUGET_VERSION was accidentally being ignored, causing outdated packages to be used during CI builds via NuGet's nearest version matching logic.


src/Simulation/Simulators.Tests/TestProjects/UnitTests/UnitTests.csproj
```diff
@@ -15,10 +15,10 @@
     <ProjectReference Include="..\..\..\Simulators\Microsoft.Quantum.Simulators.csproj" />
     <ProjectReference Include="..\Library1\Library1.csproj" />
     <ProjectReference Include="..\Library2\Library2.csproj" />
+    <ProjectReference Include="..\..\..\..\Xunit\Microsoft.Quantum.Xunit.csproj" />
```
bettinaheim, line 18: Actually I believe we don't need that reference - if you could remove it.

cgranade: Sounds good, will do!

PR #252, opened 2020-06-13
[Workspace] Support linked storage accounts

george-moussa:


src/Azure/Azure.Quantum.Client/generated/IJobsOperations.cs
```diff
@@ -8,7 +8,7 @@
 // regenerated.
 // </auto-generated>
 
-namespace Microsoft.Azure.Quantum.Client
+namespace Microsoft.Quantum
```
rmshaffer, line 11: Was it intentional to change the namespace of the generated code? My feeling is that "Azure" should still be in there somewhere.

george-moussa: i used the swagger we currently have. I believe @anpaz-msft updated it to Microsoft.Quantum

rmshaffer: The namespace can be specified via the `--namespace` argument to `autorest`. I think the reason for previously using `--namespace=Microsoft.Azure.Quantum.Client` was to be consistent with the package name, which is `Microsoft.Azure.Quantum.Client`.

george-moussa: i synced with Andres and reverted it.

PR #253, opened 2020-06-13
Experimental documentation generation tool

cgranade:
This PR opens for early feedback a possible path towards solving microsoft/qsharp-compiler#277, namely outputting documentation Markdown directly rather than as YAML. This PR packages that effort as an IRewriteStep that can be controlled by the `IRewriteStep.AssemblyConstants` property, and that can be included as an `IsQscReference`. The step also attaches documentation information as attributes that can be used by other parts of the Quantum Development Kit, such as IQ#.

Improvements over the existing approach:
- Can display Q# syntax for UDT named items.
- Extensible with new Q# features.
- Can be used more easily by third-party libraries.
- Attaches documentation data to the AST for reuse by other tools.

Remaining work:
- [ ] Currently does not filter by the current compilation unit, such that all symbols included from references are documented as well.
- [ ] Does not yet provide special support for `@Deprecated()`; this is currently supported by the non-experimental documentation generator.
- [ ] Does not yet provide special support for `@Attribute()`, making it hard to learn how attributes should be used.
- [ ] Missing unit tests.
- [ ] Missing explanatory comments.
- [ ] Named items don't yet have a dedicated Markdown section; this will require additions to the DocumentationParser in the qsharp-compiler repo.
- [ ] Need to add comments to output Markdown files indicating that they are auto-generated.
- [ ] Does not yet support making or merging toc.yml and index.md files.
- [ ] Does not yet resolve unqualified names in `/// # See Also` blocks.

### DIFF
--- a/src/Simulation/DocumentationGenerationStep/Extensions.cs
+++ b/src/Simulation/DocumentationGenerationStep/Extensions.cs
@@ -1,0 +1,246 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+using Microsoft.Quantum.QsCompiler;
+using Microsoft.Quantum.QsCompiler.DataTypes;
+using Microsoft.Quantum.QsCompiler.SyntaxTokens;
+using Microsoft.Quantum.QsCompiler.SyntaxTree;
+using Microsoft.Quantum.QsCompiler.Transformations.Core;
+using Microsoft.Quantum.QsCompiler.Transformations.QsCodeOutput;
+using YamlDotNet.Serialization;
+
+#nullable enable
+
+namespace Microsoft.Quantum.Documentation
+{
+    internal interface IAttributeBuilder<T>
+    {
+        public IAttributeBuilder<T> AddAttribute(QsDeclarationAttribute attribute);
+        public QsNullable<QsLocation> Location { get; }
+
+        public T Build();
+    }
+
+    internal class Callable : IAttributeBuilder<QsCallable>
+    {
+        private QsCallable callable;
+        internal Callable(QsCallable callable)
+        {
+            this.callable = callable;
+        }
+
+        public QsNullable<QsLocation> Location => callable.Location;
+
+        public IAttributeBuilder<QsCallable> AddAttribute(QsDeclarationAttribute attribute) =>
+            new Callable(callable.AddAttribute(attribute));
+
+        public QsCallable Build() => callable;
+    }
+
+    internal class Udt : IAttributeBuilder<QsCustomType>
+    {
+        private QsCustomType type;
+        internal Udt(QsCustomType type)
+        {
+            this.type = type;
+        }
+
+        public QsNullable<QsLocation> Location => type.Location;
+
+        public IAttributeBuilder<QsCustomType> AddAttribute(QsDeclarationAttribute attribute) =>
+            new Udt(type.AddAttribute(attribute));
+
+        public QsCustomType Build() => type;
+    }
+
+    internal static class Extensions
+    {
+        internal static IAttributeBuilder<QsCallable> AttributeBuilder(
+            this QsCallable callable
+        ) => new Callable(callable);
+        internal static IAttributeBuilder<QsCustomType> AttributeBuilder(
+            this QsCustomType type
+        ) => new Udt(type);
+
+        internal static IAttributeBuilder<T> WithAttribute<T>(
+            this IAttributeBuilder<T> builder, string @namespace, string name,
+            TypedExpression input
+        ) =>
+            builder.AddAttribute(
+                new QsDeclarationAttribute(
+                    QsNullable<UserDefinedType>.NewValue(
+                        new UserDefinedType(
+                            NonNullable<string>.New(@namespace),
+                            NonNullable<string>.New(name),
+                            QsNullable<Tuple<QsPositionInfo, QsPositionInfo>>.Null
+                        )
+                    ),
+                    input,
+                    builder.Location.Item.Offset,
+                    QsComments.Empty
+                )
+            );
+
+        private static IAttributeBuilder<T> WithDocumentationAttribute<T>(
+            this IAttributeBuilder<T> builder, string attributeName,
+            TypedExpression input
+        ) => builder.WithAttribute("Microsoft.Quantum.Documentation", attributeName, input);
+
+        private static TypedExpression AsLiteralExpression(this string literal) =>
+            SyntaxGenerator.StringLiteral(
+                NonNullable<string>.New(literal),
+                ImmutableArray<TypedExpression>.Empty
+            );
+
+        internal static IAttributeBuilder<T> MaybeWithSimpleDocumentationAttribute<T>(
+            this IAttributeBuilder<T> builder, string attributeName, string? value
+        ) =>
+            value == null || value.Trim().Length == 0
+            ? builder
+            : builder.WithDocumentationAttribute(
+                attributeName, value.AsLiteralExpression()
+            );
+
+        internal static IAttributeBuilder<T> WithListOfDocumentationAttributes<T>(
+            this IAttributeBuilder<T> builder, string attributeName, IEnumerable<string> items
+        ) =>
+            items
+            .Aggregate(
+                builder,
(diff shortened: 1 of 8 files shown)


cgranade:
Adding @tcNickolas and @KittyYeungQ for reference documentation impact, @anpaz-msft for build pipeline impact. If you have any time to offer feedback on the early prototype / experiment and the broader approach, I'd really appreciate it. Thank you!

src/Simulation/DocumentationGenerationStep/Extensions.cs
```diff
@@ -0,0 +1,246 @@
...
+        ) =>
+            items
+            .Aggregate(
+                builder,
+                (acc, item) => acc.WithDocumentationAttribute(
+                    attributeName, item.AsLiteralExpression()
+                )
+            );
+
+        internal static IAttributeBuilder<T> WithDocumentationAttributesFromDictionary<T>(
```
bettinaheim, line 119: I believe you might want to use the tools in this PR: https://github.com/microsoft/qsharp-compiler/pull/463
Let me know what attribute related things are missing, and we can see if it makes sense to add it to that static class.

src/Simulation/DocumentationGenerationStep/ProcessDocComments.cs
```diff
@@ -0,0 +1,228 @@
...
+using Microsoft.Quantum.QsCompiler.SyntaxTree;
+using Microsoft.Quantum.QsCompiler.Transformations.Core;
+using Newtonsoft.Json.Linq;
+
+#nullable enable
+
+namespace Microsoft.Quantum.Documentation
+{
+    public class ProcessDocComments
+    : SyntaxTreeTransformation<ProcessDocComments.TransformationState>
```
bettinaheim, line 23: If you don't use a transformation state, you can just use the SyntaxTreeTransformation class (no type argument).

src/Simulation/DocumentationGenerationStep/ProcessDocComments.cs
```diff
@@ -0,0 +1,228 @@
...
+        {
+            OutputPath = outputPath;
+            // If the output path is not null, make sure the directory exists.
+            if (outputPath != null && !Directory.Exists(outputPath))
+            {
+                Directory.CreateDirectory(outputPath);
+            }
+
+            // We provide our own custom namespace transformation, and expression kind transformation.
+            this.Namespaces = new ProcessDocComments.NamespaceTransformation(this, outputPath);
```
bettinaheim, line 43: I'd suggest to disable transformations on nodes that are not needed. This can be done by setting 
`this.Statements = new StatementTransformation(this, TransformationOptions.Disabled);
this.Expressions = ...; 
this.Types = ...`
Since you only need the top-level transformation, you alternatively just pass TransformationOptions.Disabled to the base constructor, and setting this.Namespaces here makes sure that the transformation on the namespace executed (though this.Apply would be disabled as well).

src/Simulation/DocumentationGenerationStep/ProcessDocComments.cs
```diff
@@ -0,0 +1,228 @@
...
+        }
+
+        public QsCompilation OnCompilation(QsCompilation compilation) =>
```
bettinaheim, line 46: Apply (inherited from the base class) would do that too.

src/Simulation/DocumentationGenerationStep/ProcessDocComments.cs
```diff
@@ -0,0 +1,228 @@
...
+        : NamespaceTransformation<TransformationState>
+        {
+            private string? outputPath;
+            internal NamespaceTransformation(ProcessDocComments parent, string? outputPath)
+            : base(parent)
+            { this.outputPath = outputPath; }
+            
+            private static QsCallable AddSummaryAttribute(QsCallable callable, string summary) =>
+                callable.AddAttribute(
+                    new QsDeclarationAttribute(
```
bettinaheim, line 64: See the tools in the PR I mentioned in another comment.

src/Simulation/DocumentationGenerationStep/ProcessDocComments.cs
```diff
@@ -0,0 +1,228 @@
...
+                                QsNullable<Tuple<QsPositionInfo, QsPositionInfo>>.Null
+                            )
+                        ),
+                        SyntaxGenerator.StringLiteral(NonNullable<string>.New(summary), ImmutableArray<TypedExpression>.Empty),
+                        callable.Location.Item.Offset,
+                        QsComments.Empty
+                    )
+                );
+
+            private async Task MaybeWriteOutput(QsCustomType type, DocComment docComment)
```
bettinaheim, line 78: I would put these writers etc into a separate file.

cgranade:
Thanks for your feedback, @bettinaheim!

> What was the reason to put this on the runtime repo rather than the compiler repo? I believe you could just have a reference to the source code version of the compiler to build the rewrite step, couldn't you?

My thinking was that this is similar in principle to how the C# generation app works, and thus to put it there, but I'm happy to move this to the compiler instead if that makes more sense.

src/Simulation/DocumentationGenerationStep/Extensions.cs
cgranade: Sounds good, I'll take a look at that, then. One thing I ran into was that a lot of the documentation attributes take a similar form on UDTs and callables, but even though QsCallable and QsCustomType both have an `AddAttribute` method, there's no common interface that allows calling `AddAttribute` on both. The `AttributeBuilder` class here was a hack around that to reduce code duplication.

src/Simulation/DocumentationGenerationStep/ProcessDocComments.cs
cgranade: Sounds good. I wasn't sure if I needed a transformation state or not yet; I will likely need one for the TOC, but I can definitely remove it until I get there. Thanks!

src/Simulation/DocumentationGenerationStep/ProcessDocComments.cs
cgranade: Sounds good, I agree; this was convenient when quickly prototyping, but is not the most maintainable.

src/Simulation/DocumentationGenerationStep/ProcessDocComments.cs
bettinaheim: renamed to OnCompilation

anpaz-msft:
I think it makes sense on the compiler repo, as Bettina suggested.

Could you provide more information about usage? If I understand correctly these are Q# attributes that get added to the code directly and that would replace the current markdown docs, is that correct, are you planning to replace comments with attributes, or are these attributes go inside the comments?
I personally like much better using comments and markdown to write the documentation as we do today. I rather make enhancement to the YAML parser we use that to embed the documentation directly into the code.
Maybe you can give a couple of samples on how the new documentation would look?

Also, will these changes be backwards compatible? Switching all the documentation will be a huge PITA.

cgranade:
Thanks for your comments, @anpaz-msft! Please see responses inline below.

> I think it makes sense on the compiler repo, as Bettina suggested.

Sounds good, I'll move it there then and close this PR.

> Could you provide more information about usage? If I understand correctly these are Q# attributes that get added to the code directly and that would replace the current markdown docs, is that correct, are you planning to replace comments with attributes, or are these attributes go inside the comments?

The new attributes are added automatically by the IRewriteStep in this PR, such that no change to Markdown documentation comments is required.

> I personally like much better using comments and markdown to write the documentation as we do today. I rather make enhancement to the YAML parser we use that to embed the documentation directly into the code.

Agreed, hence this PR doesn't target changing that workflow at all, but makes the result of parsing those comments more widely available to other parts of the Quantum Development Kit (e.g.: IQ#) via attributes.

> Maybe you can give a couple of samples on how the new documentation would look?
> 
> Also, will these changes be backwards compatible? Switching all the documentation will be a huge PITA.

So far the only backwards-incompatible change posed is to finally fix the casing convention for API reference UIDs to not require all-lowercase.

src/Simulation/DocumentationGenerationStep/ProcessDocComments.cs
cgranade: Does that imply that this method can be safely deleted?

cgranade:
Moved logic to compiler repo, closing this PR in favor of new approach.

Issue #249, opened 2020-06-11
Q# should include decompositions to alternate quantum gate sets


swernli:
Q# defines the set of basic gate operations in the Microsoft.Quantum.Intrinsic namespace, where the fundamental operations are implemented as `body intrinsic` and other operations are expressed as decompositions onto these body-intrinsic operations. These effectively represent a specific quantum gate set that the current version of the simulator implements. To facilitate experimentation, testing, and eventually targeting of simulators and hardware with a different quantum gate set, the runtime should have infrastructure and examples of alternate implementations of operations in the Microsoft.Quantum.Intrinsic namespace as decompositions on top of the body-intrinsic operations for that alternate gate set.

Introducing this functionality will require several tasks.

- [x] Refactor [Qsharp.Core](https://github.com/microsoft/qsharp-runtime/blob/master/src/Simulation/QsharpCore/Microsoft.Quantum.QSharp.Core.csproj) into the foundational elements shared by all decompositions and the replaceable elements that are override by alternate decompositions. (Resolved by #254)

- [x] Introduce the first new decomposition package on top of a different quantum gate set with alternate implementations for Microsoft.Quantum.Intrinsic operations. (Resolved by #363)

- [x] Allow alternate decompositions to be tested against the default QuantumSimulator to verify behavior. (Resolved by #363)

- [x] ~~Refactor the EntryPointDriver to allow for target specific simulators.~~ (Moved out to issue #388)

- [x] Fix e2e build to allow integration of refactored QsharpCore into main (#397, #414)

- [x] Add decomposition packages for other common quantum gate sets. (Resolved by #367)

- [x] Add documentation on individual decomposition packages and general guidance on authoring new packages.

- [x] ~~Rework use of `internal` in defining callabes that can be implemented outside of their original binary (see discussion in #367).~~ (Moved out to issue #389)

samarsha:
@swernli Can this issue be closed or is there still work in progress?

swernli:
@samarsha You beat me by one minute :). The new PR I opened (#476) is the last one needed to finally resolve this issue. It's the big merge of the feature branch into main, and will likely take a while, but should auto-close this once it is done.

PR #254, opened 2020-06-15
Refactor QSharp.Core into Foundation + Core

swernli:
This change refactors the Microsoft.Quantum.QSharp.Core project to introduce a new Microsoft.Quantum.QSharp.Foundation project, where Core is build on top of Foundation. This is the first step toward allowing for alternative decomposition packages (see https://github.com/microsoft/qsharp-runtime/issues/249). The Foundation project includes those fundamental elements of the language that all decomposition packages rely on, while the implementation that remains in Core is expected to be overridden/replaced with alternative implementation defined in a specific decomposition package. These decomposition packages will be introduced as part of a later change.


swernli:
To make the larger effort of introducing decomposition packages easier to review and discuss, I'm going to have separate PRs that should roughly line up with the tasks in #249. This set of changes incorporates feedback from my earlier draft PR, #240.

src/Simulation/QsharpFoundation/Utils.qs
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.Quantum.Intrinsic {
```
cgranade, line 4: It doesn't have any user-facing impact, but I'm not sure the name "Utils.qs" communicates to me that this source file contains definitions for two intrinsic callables that are classical, by contrast with the quantum intrinsics in Intrinsic.qs.

swernli: I struggled to come up with a name for this, actually... the only clearer option I could think of would be separating them into Message.qs and Random.qs and having only one operation per file, which feels silly to do for body-intrinsic methods where there isn't any complicated implementation. Anyone have any suggestions here?

swernli: You know, the more I think about it the more I like splitting them anyway, even if it does make very small files. This will kind of line up with what I'm imagining later parts of this decomposition work will look like anyway.

swernli:
I just noticed that ClassicalControl.qs opens the Microsoft.Quantum.Intrinsic namespace but doesn't use it. Between that and the fact that there are rewrite dependencies on it, it should get moved out of the Core and into the Foundation.

src/Simulation/QsharpCore/Intrinsic.qs
```diff
@@ -118,10 +16,7 @@ namespace Microsoft.Quantum.Intrinsic {
     /// This is a no-op. It is provided for completeness and because
     /// sometimes it is useful to call the identity in an algorithm or to pass it as a parameter.
     operation I (target : Qubit) : Unit
-    is Adj + Ctl {
-        body (...) { }        
-        adjoint self;
-    }
+    is Adj + Ctl { }
```
bettinaheim, line 19: We actually will want to keep the explicitly "adjoint self" - it is the only way currently to declare that an operation is self-adjoint (which is information that can be used for optimization and that the compiler propagates).

swernli: Ah, I didn't know that. I will put this back.

swernli:
@bettinaheim I was able to confirm that there are no file duplications. I think the reason it seems like there are is because I moved several files that have both a .qs and .cs version, such as
src/Simulation/QsharpCore/Bitwise/Bitwise.cs → src/Simulation/QsharpFoundation/Bitwise/Bitwise.cs
src/Simulation/QsharpCore/Bitwise/Bitwise.qs → src/Simulation/QsharpFoundation/Bitwise/Bitwise.qs

PR #258, opened 2020-06-17
Fixed Documentation Errors in S and T gate

jainvasu631:
The S gate which applies a PI/2 phase shift and T gate which applies a PI/4 phase shift gate were wrongly documented to term them as PI/4 and PI/8 gate respectively. The PI/8 gate is a valid and used name for T gate but it is quite confusing for beginners since the matrix only adds in a phase of PI/4 if qubit is 1.

The S gate error was pointed out in [Issue #370](https://github.com/microsoft/Quantum/issues/370) of the Microsoft Quantum GitHub page. I have redone the documentation in the hopes of fixing that. This change in turn needs to be reflected in the Microsoft/quantum-docs-pr so that the confusing documentation is fixed.


bettinaheim:
@cgranade Could you please take a look?

cgranade:
Thanks for the contribution, @jainvasu631! In this case, the 𝑇 matrix used to simulate the [`Microsoft.Quantum.Intrinsic.T` operation](https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.intrinsic.t) has historically also been known as the "π / 8 gate" despite that it can be thought of as a π / 4 rotation about the 𝑍-axis. This comes back to that, up to a global phase, 𝑇 can be written as diag(𝑒^{−𝑖π / 8}, 𝑒^{𝑖π / 8}); that is, that both the |0⟩ and |1⟩ computational basis states rotate by a phase of π / 8, but in opposite directions. When thinking about quantum physics concepts like spin, this is often a much easier way to think of rotations, such that the name "π / 8" has stuck around for the 𝑇 matrix despite the apparent contradiction.

Given that, I think it would be more correct to leave the API documentation as is. That said, to resolve the ambiguity inherent in the name "π / 8", it might perhaps make sense to add an explicit matrix form as well.

jainvasu631:
@cgranade I agree with the historical context for calling the T gate the "π / 8 gate". However calling the S gate the "π / 4 gate" doesn't make that much sense even from a historical point of view. Moreover I think it may be confusing for beginners as we saw in [Issue #370](https://github.com/microsoft/Quantum/issues/370).

cgranade:
@jainvasu631: We appreciate the feedback! In #287, I've gone on and removed the confusing terminology entirely, relying on the explicit matrix representations of each operation. I'm going to go on and close this, as your feedback has been included in that new PR. Thank you again!

PR #263, opened 2020-06-18
Update C# generation and tests for file path format change

samarsha:
This goes with microsoft/qsharp-compiler#468. Updated the C# generation to remove the assumption that the file ID is `URI.AbsolutePath`, and added a test for a project with spaces in the path that uses `EmbedAllSources`.

### DIFF
--- a/src/Simulation/CsharpGeneration/Microsoft.Quantum.CsharpGeneration.fsproj
+++ b/src/Simulation/CsharpGeneration/Microsoft.Quantum.CsharpGeneration.fsproj
@@ -21,7 +21,7 @@
 
   <ItemGroup>
     <PackageReference Update="FSharp.Core" Version="4.7.0" />
-    <PackageReference Include="Microsoft.Quantum.Compiler" Version="0.11.2006.602-beta" />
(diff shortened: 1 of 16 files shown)


src/Simulation/CsharpGeneration/Microsoft.Quantum.CsharpGeneration.fsproj
```diff
@@ -21,7 +21,7 @@
 
   <ItemGroup>
     <PackageReference Update="FSharp.Core" Version="4.7.0" />
-    <PackageReference Include="Microsoft.Quantum.Compiler" Version="0.11.2006.602-beta" />
+    <PackageReference Include="Microsoft.Quantum.Compiler" Version="0.11.2006.1806-alpha" />
```
ScottCarda-MS, line 24: Changing from a beta version to an alpha version means your operating off of not released nugets right? Isn't that going to be an issue?

samarsha: Both -alpha and -beta packages are published to the qdk-alpha NuGet feed. The -alpha packages come from individual repo builds and the -beta ones come from full cross-repo Solid builds.

PR #264, opened 2020-06-18
Aggregate results in resource estimator across multiple runs

### LINKED ISSUE #189: ResourcesEstimator instance shows wrong estimates when used more than once
## Describe the bug
`ResourcesEstimator` does not reset internal state and a single instance cannot be used to simulate two operations.

## To Reproduce
*Operation.qs*
```q#
namespace Namespace {

  operation Operation1() : Unit {
    using ((a, b) = (Qubit(), Qubit())) {
      CNOT(a, b);
      CNOT(a, b);
      CNOT(a, b);
    }
  }

  operation Operation2() : Unit {
    using ((a, b) = (Qubit(), Qubit())) {
      CNOT(a, b);
      CNOT(a, b);
      CNOT(a, b);
      CNOT(a, b);
      CNOT(a, b);
    }
  }
}
```

*Driver.cs*
```c#
using System;
using Microsoft.Quantum.Simulation.Simulators;

namespace Namespace {
  public class Program {
    public static void Main(string[] args) {
      var sim = new ResourcesEstimator();

      Operation1.Run(sim).Wait();
      Console.WriteLine(sim.ToTSV());

      Operation2.Run(sim).Wait();
      Console.WriteLine(sim.ToTSV());
    }
  }
}
```

## Expected behavior
Output should be to accumulate the estimates:
```
Metric          Sum            
CNOT            3
QubitClifford   0
R               0
Measure         0
T               0
Depth           0
Width           2
BorrowedWidth   0
Metric          Sum            
CNOT            8
QubitClifford   0
R               0
Measure         0
T               0
Depth           0
Width           2
BorrowedWidth   0
```
or to reset the estimates:
```
Metric          Sum            
CNOT            3
QubitClifford   0
R               0
Measure         0
T               0
Depth           0
Width           2
BorrowedWidth   0
Metric          Sum            
CNOT            5
QubitClifford   0
R               0
Measure         0
T               0
Depth           0
Width           2
BorrowedWidth   0
```

## Actual behavior
Output currently shows the estimates for the first operation:
```
Metric          Sum            
CNOT            3
QubitClifford   0
R               0
Measure         0
T               0
Depth           0
Width           2
BorrowedWidth   0
Metric          Sum            
CNOT            3
QubitClifford   0
R               0
Measure         0
T               0
Depth           0
Width           2
BorrowedWidth   0
```

## System information
- Q# Version: 0.11.2004.2825
- .NET Core Version: 3.1.201

irinayat-MS:
Fixes #189 

I'm not sure what the expected behavior is for the "Width" metric. Personally, I'd expect it to report the max of widths across the runs, not adding the per-run widths. However, I'd rather not have special logic for width in the processing loop because it would have to rely on the label and will be fragile, so I've added the max column to the results for all metrics. Is it useful to know the highest count across operations for gates or measurements? The data can be accessed programmatically, as the new test demonstrates, but I didn't update the ToTSV method to include it -- should I? If we go this route the documentation (https://docs.microsoft.com/en-us/quantum/machines/resources-estimator) also will need to be updated.

### DIFF
--- a/src/Simulation/Simulators.Tests/ResourcesEstimatorTests.cs
+++ b/src/Simulation/Simulators.Tests/ResourcesEstimatorTests.cs
@@ -94,10 +94,10 @@ namespace Microsoft.Quantum.Simulation.Simulators.Tests
 
             var cols = rows[0].Split('\t');
             Assert.Equal("Metric", cols[0].Trim());
-            Assert.Equal(2, cols.Length);
+            Assert.Equal(3, cols.Length);
 
             var cliffords = rows.First(r => r.StartsWith("QubitClifford")).Split('\t');
-            Assert.Equal(2, cliffords.Length);
+            Assert.Equal(3, cliffords.Length);
             Assert.Equal("2", cliffords[1]);
         }
 
--- a/src/Simulation/Simulators.Tests/ResourcesEstimatorTests.cs
+++ b/src/Simulation/Simulators.Tests/ResourcesEstimatorTests.cs
@@ -2,9 +2,8 @@
 // Licensed under the MIT License.
 
 using System;
-using System.Collections.Generic;
 using System.Linq;
-using System.Text;
+using System.Data;
 using Microsoft.Quantum.Simulation.QCTraceSimulatorRuntime;
 using Xunit;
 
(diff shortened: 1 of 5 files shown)


src/Simulation/Simulators.Tests/ResourcesEstimatorTests.cs
```diff
@@ -101,5 +100,43 @@ public void ToTSVTest()
...
+
+            Operation_1_of_2.Run(sim).Wait();
+            DataTable data1 = sim.Data;
+
+            Assert.Equal(1.0, data1.Rows.Find("CNOT")["Sum"]);
+            Assert.Equal(1.0, data1.Rows.Find("QubitClifford")["Sum"]);
+            Assert.Equal(1.0, data1.Rows.Find("T")["Sum"]);
+            Assert.Equal(0.0, data1.Rows.Find("R")["Sum"]);
+            Assert.Equal(0.0, data1.Rows.Find("Measure")["Sum"]);
+            Assert.Equal(2.0, data1.Rows.Find("Width")["Sum"]);
```
msoeken, line 121: This label changes after merging #267

src/Simulation/Simulators.Tests/ResourcesEstimatorTests.cs
```diff
@@ -101,5 +100,43 @@ public void ToTSVTest()
...
+            DataTable data2 = sim.Data;
+
+            // Aggregated stats for both operations.
+            Assert.Equal(1.0 + 2.0, data2.Rows.Find("CNOT")["Sum"]);
+            Assert.Equal(1.0 + 1.0, data2.Rows.Find("QubitClifford")["Sum"]);
+            Assert.Equal(1.0 + 0.0, data2.Rows.Find("T")["Sum"]);
+            Assert.Equal(0.0 + 1.0, data2.Rows.Find("R")["Sum"]);
+            Assert.Equal(0.0 + 1.0, data2.Rows.Find("Measure")["Sum"]);
+            Assert.Equal(2.0 + 3.0, data2.Rows.Find("Width")["Sum"]);
+            Assert.Equal(System.Math.Max(2.0, 3.0), data2.Rows.Find("Width")["Max"]);
```
msoeken, line 133: This label changes after merging #267

src/Simulation/Simulators.Tests/ResourcesEstimatorTests.cs
```diff
@@ -101,5 +100,43 @@ public void ToTSVTest()
...
+
+            // Run again to confirm two operations isn't the limit!
+            VerySimpleEstimate.Run(sim).Wait();
+            DataTable data3 = sim.Data;
+            Assert.Equal(2.0 + 3.0 + 3.0, data3.Rows.Find("Width")["Sum"]);
+            Assert.Equal(3.0, data3.Rows.Find("Width")["Max"]);
```
msoeken, line 139: These labels change after merging #267

src/Simulation/Simulators/ResourcesEstimator/ResourcesEstimator.cs
```diff
@@ -121,24 +121,32 @@ public virtual DataTable Data
                     if (l is ICallGraphStatistics collector)
                     {
                         var results = collector.Results.ToTable();
-                        Debug.Assert(results.rows.Count() == 1);
+                        //Debug.Assert(results.rows.Count() == 1);
```
msoeken, line 124: Is it safe to delete this line completely. Might be confusing if it stays commented.

bettinaheim: I believe the assert is removed since for each tracer run we would add a row here - so far it seems like it wasn't intended to be used more than once.

msoeken: Thanks. Shouldn't the line better be removed completely then?

irinayat-MS: Removed the assert

src/Simulation/Simulators.Tests/ResourcesEstimatorTests.cs
irinayat-MS: n/a anymore as abandoned column name changes in #267

src/Simulation/Simulators.Tests/ResourcesEstimatorTests.cs
irinayat-MS: n/a anymore as abandoned column name changes in #267

src/Simulation/Simulators.Tests/ResourcesEstimatorTests.cs
irinayat-MS: n/a anymore as abandoned column name changes in #267

irinayat-MS:
> I just left some minor comments. I think it would be convenient to have a `Clear()` or `Reset()` method on `ResourcesEstimator` having this new behaviour.

How having Clear/Reset is different from creating a new simulator? Does providing this additional functionality (that needs to be documented and tested) make user experience better? I'm not saying we shouldn't do it, but clearing methods are notorious for being confusing to the users and causing subtle bugs. I believe we should do it only if the perf benefits of reusing the simulator to get non-aggregated results provably exist.

PR #266, opened 2020-06-18
Make --storage option optional for Azure Quantum submissions with the standalone executable

samarsha:


src/Azure/Azure.Quantum.Client/Machine/QuantumMachineFactory.cs
```diff
@@ -1,6 +1,8 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+#nullable enable
```
cesarzc, line 4: Should we enable nullable in the whole project?

samarsha: We could do that, but maybe in a separate PR. :) I haven't tried it, but I suspect it would add a lot of warnings to the build. I kept it to this file because it was relatively small and I could fix all of the nullable warnings.

cesarzc: Sounds good!

Issue #192, opened 2020-05-05
Width and depth metrics from ResourcesEstimator can seem inconsistent


msoeken:
**Please describe what you would like the feature to accomplish.**
The ResourcesEstimator does not take potential dependencies between width and depth into account, which may result in resource estimates that seem inconsistent. This issue is to discuss alternatives on how to make the output more clear.

The following example illustrates the problem:

*Q# code:*
```q#
namespace Namespace {
  open Microsoft.Quantum.Canon;

  operation Operation() : Unit {
    using (qs = Qubit[6]) {
      ApplyLowDepthAnd(qs[0], qs[1], qs[2]);
      ApplyLowDepthAnd(qs[3], qs[4], qs[5]);
    }
  }
}
```

*C# driver:*
```c#
namespace Namespace {
  public class Program {
    public static void Main(string[] args) {
      var sim = new ResourcesEstimator();

      Operation.Run(sim).Wait();
      Console.WriteLine(sim.ToTSV());
    }
  }
}
```

*Output:*
```
Metric          Sum            
CNOT            16
QubitClifford   6
R               0
Measure         0
T               8
Depth           1
Width           7
BorrowedWidth   0
```

Here, width and depth are conflicting, because we cannot distribute 8 T gates over 7 qubits in depth 1.

**Describe the solution you'd like**
At least change the output to ToTSV() and add clarifications to the [documentation](https://docs.microsoft.com/en-us/quantum/machines/resources-estimator?view=qsharp-preview#metrics-reported). The preferred solution would be to output consistent width and depth estimates.

bettinaheim:
Being explicit about that the depth and width output are minimal counts that are not necessarily mutually satisfiable is a good first step. @msoeken, do you want to pick this up and make a PR to clarify that?

KittyYeungQ:
@geduardo to add clarification in docs

cryptosidh:
It would be great if this issue with the estimator could be fixed. The result in https://github.com/microsoft/qsharp-runtime/pull/267 is disappointing. Renaming metrics to WidthLowerBound and DepthLowerBound would have been a welcome first step. 

[...]

PR #267, opened 2020-06-18
Confusing output for width and depth from resource estimator

irinayat-MS:
Related to #192 

The resource estimator currently calculates independent lower bounds for width and depth metrics. When reported next to each other they might look confusing because it might not be possible to satisfy both bounds at the same time.

While we consider whether it's worth implementing the full optimization logic in the resource estimator (so it could calculate corelated width/depth metrics), this change renames the current metrics to WidthLowerBound and DepthLowerBound to communicate the intent more clearly.

Have decided that this breaking change isn't justified and will only update the documentation:
https://github.com/MicrosoftDocs/quantum-docs-pr/pull/930


src/Simulation/Simulators/ResourcesEstimator/ResourcesEstimator.cs
```diff
@@ -78,7 +78,11 @@ public ResourcesEstimator(QCTraceSimulatorConfiguration config) : base(config)
             }
             else if (name == MetricsNames.WidthCounter.ExtraWidth)
             {
-                return "Width";
+                return "WidthLowerBound";
```
cgranade, line 81: As a brief note, this could break user code. At the moment, users may extract the `"Width"`  key from the returned dictionary. I think this even happens in samples, such that I'd suggest firing off a cross-repo to validate this API change.

irinayat-MS: Yes, this _is_ a breaking change. How do we usually stage changes like this? Shall we provide both properties in parallel for a while and document that "Width/Depth" are deprecated?

cgranade: @bettinaheim could speak much better to .NET API deprecation policies; I'm mainly responsible for Q# API deprecation policies, and for maintaining samples / libraries. From that stand point, a cross-repo build could tell us if any current samples are broken and give us the heads-up we need to fix that. Otherwise, the cross-repo `master` would break when merging this in.

irinayat-MS:
Have decided that this breaking change isn't justified and will only update the documentation:
MicrosoftDocs/quantum-docs-pr#930

irinayat-MS:
The additional tests would be nice to have, though... I'll update them to use the old statistics names.

irinayat-MS:
Warning messages like this become annoying very quickly because the user learns the fact after seeing the warning once or twice but the message would keep showing up, complicating the analysis of the output and demanding attention. I am also worried to proliferate hard-coded text messages through the codebase as that is against the globalization guidelines. Arguably, if a user is confused by what they are seeing in the statistics, that would be the perfect time for them to check the documentation. Basically, I believe we should keep code and documentation separate.

I am not resolving #192, please feel free to add more comments there if you think additional fixes could improve the situation. For this change I'd like to go ahead and add the tests -- would you mind unblocking the PR?

Issue #40, opened 2019-08-19
C# code generation generates invalid properties for UDTs with tuple-valued named items


cgranade:
## Describe the bug
When defining a UDT with a named item whose type is a non-singleton tuple, the C# code generation step produces invalid properties that then fail to compile when running `dotnet build`.

## To Reproduce
```
namespace runtime_bug {
    open Microsoft.Quantum.Arrays;
    open Microsoft.Quantum.Canon;
    open Microsoft.Quantum.Intrinsic;

    newtype GeneratorIndex = (IndexData: (Int[], Double[]), Subsystems: Int[]);

    function Example() : GeneratorIndex[] {
        return Mapped(
            GeneratorIndex(_, [0, 1, 2]),
            [
                ([0], [0.0]),
                ([1, 1], [1.1, 1.1]),
                ([2, 2, 2], [2.2, 2.2, 2.2])
            ]
        );
    }

    function Boom() : Int {
        let foos = Example();
        Message($"{foos}");
        return (Fst(foos[0]::IndexData))[0];
    }
}
```

## Expected behavior
`Boom` should return `0`.

## Actual behavior
```
obj\qsharp\src\Operations.g.cs(27,62): error CS0029: Cannot implicitly convert type 'Microsoft.Quantum.Simulation.Core.IQArray<long>' to '(Microsoft.Quantum.Simulation.Core.IQArray<long>, Microsoft.Quantum.Simulation.Core.IQArray<double>)' [C:\Users\chgranad.REDMOND\tmp\runtime-bug\runtime-bug.csproj]
obj\qsharp\src\Operations.g.cs(28,45): error CS0266: Cannot implicitly convert typ  'Microsoft.Quantum.Simulation.Core.IQArray<double>' to 'Microsoft.Quantum.Simulation.Core.IQArray<long>'. An explicit conversion exists (are you missing a cast?) [C:\Users\chgranad.REDMOND\tmp\runtime-bug\runtime-bug.csproj]
```
The generated code referred to by the above error:
```C#
public (IQArray<Int64>,IQArray<Double>) IndexData => Data.Item1.Item1;
public IQArray<Int64> Subsystems => Data.Item1.Item2;
public (IQArray<Int64>,IQArray<Double>) Item1 => Data.Item1;
public IQArray<Int64> Item2 => Data.Item2;
```

## System information
- OS: Windows 10

- .NET Core Version: 3.0.100-preview8-013656

```xml
  <ItemGroup>
[...]

PR #270, opened 2020-06-19
Fix C# Generation for UDT Tuple Field

ScottCarda-MS:
Fixes issue #40.

Fixed an issue with the C# generation for UDT classes where tuple fields of the UDT where not being correctly parsed.


src/Simulation/CsharpGeneration/SimulationCode.fs
```diff
@@ -1454,17 +1454,24 @@ module SimulationCode =
                     []
```
ScottCarda-MS, line 1454: From @bettinaheim post: Could you please add tests for this?

ScottCarda-MS: Sure. I've been looking at SimulationCodeTests.fs to get a feel for these kinds of tests, but that file is huge. Is there any tools or tests in particular you could point out to me to help me write a test for this?

samarsha: We should probably split SimulationCodeTests.fs into multiple files. I agree it's way too big.

bettinaheim: Looking at the code gen test for the piece that you modified makes sense. It would also be good to additionally have some execution tests - there is e.g. a nice case in the bug report, and you should add some more complex cases as well.

ScottCarda-MS: @bettinaheim I've added a test for the bug fix. As for the execution test, I don't think we need to go that far. This PR is a simple bug fix, after all.

ScottCarda-MS: @samarsha showed me that writing execution tests is way easier in this case than I thought, so I have added an execution test for this as well.

src/Simulation/Simulators.Tests/Circuits/UserDefinedTypes.qs
```diff
@@ -7,10 +7,11 @@ namespace Microsoft.Quantum.Simulation.Simulators.Tests.Circuits
 
     newtype P1 = (Int, Int);
     newtype P2 = ((Int, Int), Int);
+    newtype NamedTuple = (FirstItem: (Int, Double), SecondItem: Int);
```
bettinaheim, line 10: Looks good. If you could please add one more test for an inner named tuple item (it's all working fine, but never hurts to cover it with a test): 
```suggestion
    newtype NamedTuple = (FirstItem: (Int, Double), SecondItem: Int);
    newtype P3 = ((Fst : Result, Snd : (Int, Int)), Trd : String);

    function UdtInnerNamedTupleFieldTest () : Unit {
        let t = P3((Zero, (0,1)), "");
        AssertEqual(Zero, t::Fst);
        let snd = t::Snd;
        let (s1, s2) = snd;
        AssertEqual(0, s1);
        AssertEqual(1, s2);
        AssertEqual("", t::Trd);
    }
```

ScottCarda-MS: Added. Thanks!

PR #271, opened 2020-06-19
Adding props file to C# generation 

bettinaheim:
Going forward, compiler extensions need to declare which dll needs to be loaded. This PR makes the C# generation compliant with that.


samarsha:
Is there a PR for the update to the compiler that reads this property?

bettinaheim:
> Is there a PR for the update to the compiler/SDK that reads this property?

Yes, here is the corresponding compiler PR: https://github.com/microsoft/qsharp-compiler/pull/472.

PR #276, opened 2020-06-22
Porting ToffoliSimulator to IQuantumProcessor

msoeken:
This PR reimplements the current `ToffoliSimulator` to use the `IQuantumProcessor` interface. All supported features should be ported.


src/Simulation/Simulators/ToffoliSimulator/ToffoliSimulator.cs
```diff
@@ -20,478 +17,29 @@ public enum ToffoliDumpFormat
...
-        /// </summary>
-        /// <param name="q">The qubit being returned.</param>
-        internal void ReturnQubit(Qubit q)
-        {
-            var stack = borrowedQubitStates[q.Id];
-            if (stack.Pop() != State[q.Id])
-            {
-                throw new ExecutionFailException("Borrowed qubit returned not in the original state");
-            }
+    public class ToffoliSimulator : QuantumProcessorDispatcher {
```
irinayat-MS, line 20: >{ [](start = 63, length = 1)

From other *.cs files in this repo it seems that the coding style for C# is to put the opening brace on the new line.

src/Simulation/Simulators/ToffoliSimulator/ToffoliSimulator.cs
```diff
@@ -20,478 +17,29 @@ public enum ToffoliDumpFormat
...
+        public const uint DEFAULT_QUBIT_COUNT = 64;
```
irinayat-MS, line 21: >64 [](start = 48, length = 2)

The new default qubit count is much lower than it used to be. Why change?

src/Simulation/Simulators/ToffoliSimulator/ToffoliSimulator.cs
```diff
@@ -20,478 +17,29 @@ public enum ToffoliDumpFormat
...
+        private ToffoliSimulatorProcessor processor;
 
-            if (stack.Count == 0)
-            {
-                borrowedQubitStates.Remove(q.Id);
-            }
-        }
+        // Whether an exception is thrown when releasing qubits that are in a
+        // non-zero state is configured as an argument to the constructor of
+        // ReversibleSimulator, defaulted to `true`.
```
irinayat-MS, line 26: I'm failing to parse the grammatical structure of this comment. Maybe reword it?

src/Simulation/Simulators/ToffoliSimulator/Processor.cs
```diff
@@ -0,0 +1,429 @@
...
+using Microsoft.Quantum.Simulation.Simulators.Exceptions;
+
+namespace Microsoft.Quantum.Simulation.Simulators
+{
+
+    // ToffoliSimulator is implemented based on `QuantumProcessorDispatcher`,
+    // which is constructed using a specialization `QuantumProcessorBase`.  The
+    // specialization overrides methods to specify actions on intrinsic
+    // operations in the Q# code.
+    class ToffoliSimulatorProcessor : QuantumProcessorBase {
```
irinayat-MS, line 22: > { [](start = 58, length = 2)

Please follow the coding style of the project and place opening braces on new lines.

src/Simulation/Simulators/ToffoliSimulator/Processor.cs
```diff
@@ -0,0 +1,429 @@
...
+
+        // When allocating a qubit whose index equals or exceeds the current
+        // number of simulation values stored in the bit array, the bit array is
+        // resized to twice its current size.
+        public override void OnAllocateQubits(IQArray<Qubit> qubits) {
+            if (qubits.Count == 0) return;
+
+            var maxId = qubits.Max(q => q.Id);
+            while (maxId >= SimulationValues.Length) {
+                SimulationValues.Length *= 2;
```
irinayat-MS, line 74: >SimulationValues.Length *= 2; [](start = 16, length = 29)

What if the doubled size is still less than maxId?

src/Simulation/Simulators/ToffoliSimulator/Processor.cs
```diff
@@ -0,0 +1,429 @@
...
+        /// Make sure that a rotation is either the identity or an X, up to global phase.
+        /// Assumes that the rotation is exp(i*theta*sigma(axis)), where sigma(axis) is
+        /// the appropriate Pauli matrix normalized to +/-1 eigenvalues.
+        /// Throws an exception if the rotation is invalid.
+        /// </summary>
+        /// <param name="axis">The rotation axis</param>
+        /// <param name="theta">The rotation angle, properly normalized</param>
+        /// <returns>A pair of flags. The first indicates whether the rotation is
+        /// equivalent to X; the second whether or not the rotation is exactly equal
+        /// to the identity (no phase), and thus may be controlled.</returns>
```
irinayat-MS, line 151: Why is it valuable to support in Toffoli simulator rotations that are equivalent to X or I? Do we expect the users to write their code for running on a Toffoli simulator in terms of such rotations?

src/Simulation/Simulators/ToffoliSimulator/Processor.cs
```diff
@@ -0,0 +1,429 @@
...
+        // If the simulation value of qubit is true, when it's being reset, we
+        // restore it to false.
+        public override void Reset(Qubit qubit) {
+            SetValue(qubit, false);
+        }
+
+        // Measuring the qubit corresponds to translating the current simulation
+        // value into a Result value.
+        public override Result M(Qubit qubit) =>
+            GetValue(qubit).ToResult();
```
irinayat-MS, line 323: For consistency reasons with other simple methods in this file, I'd suggest to use the standard function syntax rather than a lambda.

src/Simulation/Simulators/ToffoliSimulator/Processor.cs
```diff
@@ -0,0 +1,429 @@
...
+
+
+        private Func<Pauli, Qubit, Qubit?> PauliFilter(string operation) =>
+            (Pauli p, Qubit q) =>
+                p switch {
+                    Pauli.PauliI => null,
+                    Pauli.PauliZ => q,
+                    _ => throw new InvalidOperationException($"{operation} on bases other than PauliZ not supported")
```
irinayat-MS, line 331: >PauliZ [](start = 95, length = 6)

and PauliI

src/Simulation/Simulators/ToffoliSimulator/ToffoliSimulator.cs
msoeken: Thanks for pointing this out. I will fix this in the next commit.

src/Simulation/Simulators/ToffoliSimulator/ToffoliSimulator.cs
msoeken: The previous simulator did not support resizing and therefore had a very high default count. The new implementation does support resizing by doubling.

src/Simulation/Simulators/ToffoliSimulator/ToffoliSimulator.cs
msoeken: I will fix that one. It also still refers to the old name from samples. I'll scan the files for that.

src/Simulation/Simulators/ToffoliSimulator/Processor.cs
msoeken: Am going to fix it in one of the next commits.

src/Simulation/Simulators/ToffoliSimulator/Processor.cs
msoeken: It'll double until it's larger than `maxId` using the condition in the `while` loop.

src/Simulation/Simulators/ToffoliSimulator/Processor.cs
msoeken: I don't know, and doubt that this is a common scenario. This is supported by the current implementation and I wanted to make sure to take over all currently supported features.

src/Simulation/Simulators/ToffoliSimulator/Processor.cs
irinayat-MS: Duh! Of course :)

src/Simulation/Simulators/ToffoliSimulator/Processor.cs
msoeken: I think it's consistent for single-line non-void functions to use expression syntax, see e.g., https://github.com/microsoft/qsharp-runtime/blob/master/src/Simulation/EntryPointDriver/Parsers.cs

src/Simulation/Simulators/ToffoliSimulator/Processor.cs
msoeken: `PauliI` is supported, but ignored. Only `PauliX` and `PauliY` raise an error.

msoeken:
@irinayat-MS Thanks for your comments. I addressed them and left some responses to your comments.

src/Simulation/Simulators/ToffoliSimulator/Processor.cs
irinayat-MS: Supporting this case does complicate the code and sends a somewhat confusing message to the users: "Rotation gates are supported _sometimes_", so I was wondering whether we have any data on this functionality actually being valuable...

---
In reply to: [444429979](https://github.com/microsoft/qsharp-runtime/pull/276#discussion_r444429979) [](ancestors = 444429979)

src/Simulation/Simulators/ToffoliSimulator/Processor.cs
irinayat-MS: I am not sure about the recommended style in those cases but I find it weird (and harder to read) when the syntax keeps switching between function and expression styles like it does here for Reset and M, both of which are single line (Parsers.cs doesn't contain any void single line functions).

---
In reply to: [444431113](https://github.com/microsoft/qsharp-runtime/pull/276#discussion_r444431113) [](ancestors = 444431113)

src/Simulation/Simulators/ToffoliSimulator/Processor.cs
```diff
@@ -0,0 +1,508 @@
...
+            }
+
+            // If there are any X- or Y-basis measurements, then the probability is 50%
+            if (bases.Any(p => p == Pauli.PauliX || p == Pauli.PauliY))
+            {
+                if (Abs(0.5 - probabilityOfZero) > tolerance)
+                {
+                    throw new ExecutionFailException(msg);
+                }
+            }
```
irinayat-MS, line 450: I am confused. The processor only supports measurements in Z basis, right? So what does it mean to assert probability for X and Y?

src/Simulation/Simulators/ToffoliSimulator/Processor.cs
msoeken: I wouldn't want to remove any functionality in this conversion PR. Maybe this can be another discussion and another PR. I think there are sufficiently many warnings to report to the user which kind of rotations are not supported.

src/Simulation/Simulators/ToffoliSimulator/Processor.cs
msoeken: It's also not completely clear to me, but carried over from the conversion.

src/Simulation/Simulators/ToffoliSimulator/Processor.cs
```diff
@@ -0,0 +1,518 @@
...
+        public ToffoliDumpFormat DumpFormat { get; set; } = ToffoliDumpFormat.Automatic;
+
+        // A property that must be assigned from the dispatcher to return the allocated
+        // qubit ids from the qubit manager
+        public Func<IEnumerable<long>>? GetAllocatedIds { get; set; }
+
+        // A property that must be assigned from the dispatcher to print a Q# message
+        public Action<String>? Message { get; set; }
+
+        private Dictionary<Qubit, Stack<bool>> borrowedQubitStates { get; } = new Dictionary<Qubit, Stack<bool>>();
```
avasch01, line 43: >Stack<bool> [](start = 34, length = 11)

Is there a way to implement it more efficiently? Maybe BitArray and StackPointer?

src/Simulation/Simulators/ToffoliSimulator/Processor.cs
```diff
@@ -0,0 +1,518 @@
...
+                var stack = borrowedQubitStates[qubit];
+                if (stack.Pop() != GetValue(qubit))
+                {
+                    throw new ExecutionFailException("Borrowed qubit returned not in the original state");
+                }
+
+                if (stack.Count == 0)
+                {
+                    borrowedQubitStates.Remove(qubit);
+                }
```
avasch01, line 132: Does it make sense to remove empty stacks, or is it better to keep them around once created to avoid churn in memory?

src/Simulation/Simulators/ToffoliSimulator/Processor.cs
```diff
@@ -0,0 +1,518 @@
...
+            if (controls.All(control => GetValue(control)))
+            {
+                InvertValue(qubit);
+            }
+        }
+
+        // SWAP interchanges the simulation values at two locations.
+        public override void SWAP(Qubit qubit1, Qubit qubit2)
+        {
+            var tmp = GetValue(qubit1);
```
avasch01, line 156: >var [](start = 12, length = 3)

Please avoid using var, as per our Quantum Systems style guide.

src/Simulation/Simulators/ToffoliSimulator/Processor.cs
avasch01: Please do not use lambdas unless absolutely necessary (which is not the case here).  Please use simple concepts instead of complex ones in such simple cases.

---
In reply to: [444449794](https://github.com/microsoft/qsharp-runtime/pull/276#discussion_r444449794) [](ancestors = 444449794,444431113)

src/Simulation/Simulators/ToffoliSimulator/Processor.cs
avasch01: I agree with Mathias that backwards compatibility is important here, regardless of whether the functionality us more or less useful.

---
In reply to: [444679013](https://github.com/microsoft/qsharp-runtime/pull/276#discussion_r444679013) [](ancestors = 444679013)

src/Simulation/Simulators/ToffoliSimulator/Processor.cs
```diff
@@ -0,0 +1,518 @@
...
+                        safeToControl = false;
+                        break;
+                    case Pauli.PauliY:
+                    case Pauli.PauliZ:
+                        throw new InvalidOperationException($"The Toffoli simulator can only perform non-identity rotations around the X axis.");
+                }
+            }
+            // If power == 0, then we're a multiple of pi, possible an even multiple.
+            else
+            {
```
avasch01, line 274: > { [](start = 11, length = 2)

assert(power == 0)
Code does tend to change, and these asserts come very handy.

src/Simulation/Simulators/ToffoliSimulator/Processor.cs
```diff
@@ -0,0 +1,518 @@
...
+
+            var piOverTwoRatio = Abs(theta) / (PI / 2.0);
+            if (Abs(piOverTwoRatio - Round(piOverTwoRatio)) > Tolerance)
+            {
+                throw new InvalidOperationException($"The Toffoli simulator can only perform rotations of multiples of pi/2.");
+            }
+
+            var piOverTwoCount = (int)Round(piOverTwoRatio);
+
+            var equivalentToX = (axis == Pauli.PauliX) && (piOverTwoCount % 2 == 1);
```
avasch01, line 197: >var [](start = 12, length = 3)

This is just a bool, right? please help me understand the code faster by stating it. Here and everywhere.

src/Simulation/Simulators/ToffoliSimulator/Processor.cs
```diff
@@ -0,0 +1,518 @@
...
+        // resized to twice its current size.
+        public override void OnAllocateQubits(IQArray<Qubit> qubits)
+        {
+            if (qubits.Count == 0) return;
+
+            var maxId = qubits.Max(q => q.Id);
+            while (maxId >= SimulationValues.Length)
+            {
+                SimulationValues.Length *= 2;
+            }
```
avasch01, line 81: This may cause the array to be reallocated multiple times. Alternatively, first calculate the newLength variable, then set Length property to this new value once, not in a loop.

src/Simulation/Simulators/ToffoliSimulator/Processor.cs
msoeken: Are you referring to `Dictionary` or `Stack` or both?

alan-geller:
What is the benefit of this change? What are the benefits of this implementation?

src/Simulation/Simulators/ToffoliSimulator/ToffoliSimulator.cs
```diff
@@ -1,15 +1,14 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+﻿// Copyright (c) Microsoft Corporation.
```
alan-geller, line 1: What is the benefit of this change? What are the benefits of this implementation over the existing one? It seems to me that this implementation will be slightly slower and is less extensible.

msoeken: The simulator uses the `IQuantumProcessor` interface. It requires fewer code and is therefore easier to maintain. Further, we expect that the change in data structures, e.g., how to store simulation values, will increase the performance. Also `QuantumProcessorDispatcher` can be extended in the same way as `SimulatorBase`, so we won't loose in extensibility.

msoeken: I compared the proposed implementation to the current implementation using our ToffoliChain performance benchmarks and found no performance degradation, but also no noticeable speed-up. However, we still think to switch to the proposed implementation for more simplicity and easier maintainability.

bettinaheim:
Closing this as per our simulator extensibility plans.

PR #278, opened 2020-06-22
v-yusudz/tracer-minor-fixes

yusudz:


src/Simulation/QCTraceSimulator/DepthCounter.cs
```diff
@@ -87,8 +87,8 @@ public void OnOperationEnd(object[] returnedQubitsTraceData)
             OperationCallRecord caller = operationCallStack.Peek();
             HashedString callerName = caller.OperationName;
 
-            caller.ReleasedQubitsAvailableTime = Max(opRec.ReleasedQubitsAvailableTime, caller.ReleasedQubitsAvailableTime );
-            caller.ReleasedQubitsAvailableTime = Max(opRec.ReleasedQubitsAvailableTime, caller.ReleasedQubitsAvailableTime );
```
yusudz, line 91: Is this a bug?

irinayat-MS: Could you please provide a description for the PR? Something along the lines of "Removed duplicated code and unnecessary namespace includes. Fixed bug in calculation of depth for borrowed qubits."

Is it possible to add a test that fails prior to this fix and passes after?

vadym-kl: Yep. Good catch.

PR #286, opened 2020-06-25
Improve core Q# API consistency

cgranade:
This PR makes several changes as part of making the Q# API for the core library more consistent and usable:

- Move NoOp and IndexRange from QuantumLibraries.
- Remove year-old deprecation stubs.
- Move diagnostic functionality out of Microsoft.Quantum.Intrinsic and into Microsoft.Quantum.Diagnostics.

Adding @guenp and @msoeken for Q# API discussion, @bettinaheim for runtime discussion.


src/Simulation/QsharpCore/Canon/NoOp.qs
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.Quantum.Canon {
```
bettinaheim, line 4: Is this the namespace to keep them under, or would we even want them in Core?

src/Simulation/Simulators.Tests/Circuits/Bug3758.qs
```diff
@@ -22,43 +20,31 @@ namespace Bug3758
...
-namespace Microsoft.Quantum.Simulation.Simulators.Tests.Circuits
-{
+namespace Microsoft.Quantum.Simulation.Simulators.Tests.Circuits {
     open Bug3758;
-    open Microsoft.Quantum.Extensions.Testing;
+    open Microsoft.Quantum.Diagnostics;
 
     operation TransformedOperationTest() : Unit {
-        AssertOperationsEqualReferenced(
+        AssertOperationsEqualReferenced(3,
```
ScottCarda-MS, line 45: The order of the arguments changed here, but I didn't see this callable's signature being changed in this PR. The  doc page for this operation [here](https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.diagnostics.assertoperationsequalreferenced) suggests that the number of qubits should be the first argument.

ScottCarda-MS: I read this backwards. I see now that the new ordering is in line with the documentation. My bad!

src/Simulation/Simulators.Tests/Circuits/Bug3758.qs
```diff
@@ -1,12 +1,10 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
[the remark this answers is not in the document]
cgranade: Updated guidance is to not include the "All rights reserved" for new code, but that it's not a high priority to remove it from existing code. Following that, I've just tried to clean that up where I notice it.

src/Simulation/QsharpCore/Canon/NoOp.qs
cgranade: That would be an additional breaking change (one we could manage with a deprecation stub), such that I'd like to defer that for now if possible. That said, I think it makes sense to keep that under Canon so that it's guarded with an explicit open statement; NoOp is really useful in some cases, but not generically enough in my mind to elevate to auto-open.

src/Simulation/Simulators.Tests/Circuits/Bug3758.qs
cgranade: No worries, happy for the double-check on that!

src/Simulation/Simulators.Tests/Circuits/Bug3758.qs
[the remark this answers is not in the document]
ScottCarda-MS: Interesting. I will keep that in mind for new files I add.

src/Simulation/Simulators.Tests/StackTraceTests.cs
```diff
@@ -29,32 +31,33 @@ public StackTraceTests(ITestOutputHelper output)
         [Fact]
         public void AllocateQubit2Test()
         {
-            using (var sim = new QuantumSimulator())
+            using var sim = new QuantumSimulator();
```
ricardo-espinoza, line 34: NIT: I see a mix of style between using and not parenthesis. Was this intentional?

cgranade: Parenthesis aren't allowed in the new C# 8 form that uses the current scope to trigger disposability. For more details: https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-8#using-declarations

src/Simulation/QsharpCore/Canon/NoOp.qs
bettinaheim: Ok.

PR #289, opened 2020-06-30
upgraded Roslyn to 3.6.0

filipw:
I noticed that Roslyn 3.3.1, which is almost 1 year old, is still in use. This PR updates to 3.6.0 which aligns with VS 2019 16.6

sibling PR to https://github.com/microsoft/qsharp-compiler/pull/499
cc @bettinaheim

### DIFF
--- a/src/Simulation/CsharpGeneration/Microsoft.Quantum.CsharpGeneration.fsproj
+++ b/src/Simulation/CsharpGeneration/Microsoft.Quantum.CsharpGeneration.fsproj
@@ -21,7 +21,7 @@
 
   <ItemGroup>
     <PackageReference Update="FSharp.Core" Version="4.7.0" />
-    <PackageReference Include="Microsoft.Quantum.Compiler" Version="0.11.2006.2309-alpha" />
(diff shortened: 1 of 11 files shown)


bettinaheim:
@filipw I pushed the package with the corresponding change on the compiler repo to our alpha channel. In order for this PR to work, you would need to pick up that change; i.e. the I believe 8 uses of the Microsoft.Quantum.Sdk need to point the version 0.11.2006.3017-alpha, as well as the one package reference to the Microsoft.Quantum.Compiler in the C# generation. 
Do you want to give this a go? Alternatively I am happy to do it if I may push to your branch?

filipw:
thanks - I pushed the change here. 
Overall I was genuinely trying to help but I now see that the required coordination between the two repos is only causing you extra headache instead of helping 😅

bettinaheim:
@filipw You are definitively helping! 
It's great to have an additional set of eyes on this to notice things like old packages! :) Sorry for the hassle regarding coordination between the compiler and runtime repo for certain things.

PR #290, opened 2020-07-01
Use a default target when submitting to Azure Quantum

samarsha:
If the `ExecutionTarget` assembly constant is blank, `--target` is required. Otherwise, `--target` is optional and uses the value of `ExecutionTarget` as its default value.

This is possible because microsoft/qsharp-compiler#503 changed the value of `ExecutionTarget` to be the original string specified in the project file, e.g. "ionq.qpu", instead of "IonQProcessor".

### DIFF
--- a/src/Simulation/EntryPointDriver/Azure.cs
+++ b/src/Simulation/EntryPointDriver/Azure.cs
@@ -150,12 +150,16 @@ namespace Microsoft.Quantum.EntryPointDriver
         /// Creates a quantum machine based on the Azure Quantum submission settings.
         /// </summary>
         /// <param name="settings">The Azure Quantum submission settings.</param>
+        /// <exception cref="ArgumentNullException">Thrown if <paramref name="settings"/>.Target is null.</exception>
         /// <returns>A quantum machine.</returns>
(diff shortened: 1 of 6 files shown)


src/Simulation/EntryPointDriver/Azure.cs
```diff
@@ -150,12 +150,16 @@ private static void DisplayError(string summary, string message)
         /// Creates a quantum machine based on the Azure Quantum submission settings.
         /// </summary>
         /// <param name="settings">The Azure Quantum submission settings.</param>
+        /// <exception cref="ArgumentNullException">Thrown if <paramref name="settings"/>.Target is null.</exception>
         /// <returns>A quantum machine.</returns>
         private static IQuantumMachine? CreateMachine(AzureSettings settings) => settings.Target switch
         {
             NothingMachine.TargetId => new NothingMachine(),
             ErrorMachine.TargetId => new ErrorMachine(),
```
anpaz-msft, line 158: Since you are doing the `switch` on `settings.Target` itself, I think you should match `null` instead of throwin when invoking `CreateMachine`

samarsha: Good catch, thanks!

PR #291, opened 2020-07-01
T anikda/experimental

AniketDalvi:
The code changes made in this PR primarily revolve around implementing the logic to reorder the wavefunction. It also includes some supporting functions to implement this change. This reordering routine has been added as an effort to optimize the full-state native simulator.


bettinaheim:
@AniketDalvi 
We should keep this on the feature branch until all the bits and pieces have been implemented. I retargeted the PR; could you please resolve the merge conflicts? Thanks!

irinayat-MS:
For performance improvements it's super useful to have tangible numbers that highlight the problem and show the fix is working (and the situations when the fix might not be applicable). Do we have benchmarks for the scenarios you are addressing? Can they be checked-in alongside with your change?

The PR doesn't link to any issues -- is there an issue tracking this perf improvements? Please add benchmark results to this PR's description and to the issue (if it exists).

src/Simulation/Native/src/simulator/capi_test.cpp
```diff
@@ -106,33 +106,54 @@ void test_gates()
 
     allocateQubit(sim_id, 0);
     allocateQubit(sim_id, 1);
+    /*allocateQubit(sim_id, 2);
```
irinayat-MS, line 109: Remove this new commented out code?

src/Simulation/Native/src/simulator/wavefunction.hpp
```diff
@@ -238,19 +290,55 @@ class Wavefunction
...
+        
+    }
     /// generic application of a gate
     template <class Gate>
     void apply(Gate const& g)
     {
-        fused_.apply(wfn_, g.matrix(), qubit(g));
+        //check flush condition
+        std::vector<qubit_t> cs;
+        if (fused_.shouldFlush(wfn_, g.matrix(), cs, g.qubit())) {
```
irinayat-MS, line 326: matrix() call creates the matrix, so now you are doing it twice -- maybe cache in a local instead?

src/Simulation/Native/src/simulator/wavefunction.hpp
```diff
@@ -238,19 +290,55 @@ class Wavefunction
...
+        }
+        fused_.apply(wfn_, g.matrix(), g.qubit());
     }
-
+    
     /// generic application of a multiply controlled gate
     template <class Gate>
     void apply_controlled(std::vector<qubit_t> cs, Gate const& g)
     {
         std::vector<qubit_t> pcs = qubits(cs);
```
irinayat-MS, line 336: looks unused now?

src/Simulation/Native/src/simulator/wavefunction.hpp
```diff
@@ -274,7 +362,8 @@ class Wavefunction
     template <class A>
     bool subsytemwavefunction(std::vector<unsigned> const& qs, std::vector<T, A>& qubitswfn, double tolerance)
     {
```
irinayat-MS, line 364: Why have you extracted flush from Fused::subsytemwavefunction? Is keeping that function still useful?

src/Simulation/Native/src/simulator/wavefunction.hpp
```diff
@@ -338,7 +427,7 @@ class Wavefunction
   private:
     unsigned num_qubits_;             // for convenience
     mutable WavefunctionStorage wfn_; // storing the wave function
-    std::vector<qubit_t> qubitmap_;   // mapping of logical to physical qubits
+    mutable std::vector<qubit_t> qubitmap_;   // mapping of logical to physical qubits
```
irinayat-MS, line 430: Does reorderWFN have to be const?

src/Simulation/Native/src/simulator/wavefunction.hpp
```diff
@@ -238,19 +290,55 @@ class Wavefunction
         rng_.seed(s);
     }
 
+    void reorderWFN(unsigned qubitLoc, unsigned newPos) const
+    {
+        //get id of qubit located at newPos and qubitLoc - getting index from the element
+        auto itr1 = std::find(qubitmap_.begin(), qubitmap_.end(), newPos);
```
irinayat-MS, line 296: Asseert that both iterators aren't equal to qubitmap_.end()?

src/Simulation/Native/src/simulator/wavefunction.hpp
```diff
@@ -238,19 +290,55 @@ class Wavefunction
...
+        auto newPosId = std::distance(qubitmap_.begin(), itr1);
+        auto itr2 = std::find(qubitmap_.begin(), qubitmap_.end(), qubitLoc);
+        auto qid = std::distance(qubitmap_.begin(), itr2);
+        //swap qubits in wfn between qubitLoc and newPos
+        if (newPos != qubitLoc)
+        {
+            for (unsigned i = 0; i < wfn_.size(); i++)
```
irinayat-MS, line 303: the size of wavefunction store might be quite large, correct? is it guaranteed to be less than max unsigned? (otherwise the loop counter will wrap before reaching wfn_.size() and the loop won't terminate)

src/Simulation/Native/src/simulator/wavefunction.hpp
```diff
@@ -238,19 +290,55 @@ class Wavefunction
...
+            {
+                unsigned int bit1 = (i >> qubitLoc) & 1;
```
irinayat-MS, line 305: Nit: as far as I know, `unsigned int` is equivalent to `unsigned` -- any reason to spell it out here but not in other places?

src/Simulation/Native/src/simulator/wavefunction.hpp
```diff
@@ -238,19 +290,55 @@ class Wavefunction
...
+                unsigned int x = (bit1 ^ bit2);
+                x = (x << qubitLoc) | (x << newPos);
+                unsigned new_i = i ^ x;
+                if (new_i > i)
+                {
+                    std::iter_swap(wfn_.begin() + i, wfn_.begin() + new_i);
+                }
+            }
+            //swap elemnts in qubitmap located at qid and newPosId
+            std::iter_swap(qubitmap_.begin() + qid, qubitmap_.begin() + newPosId);
```
irinayat-MS, line 316: rename itr1 and itr2 to something more descriptive and use them here directly?

src/Simulation/Native/src/simulator/wavefunction.hpp
```diff
@@ -238,19 +290,55 @@ class Wavefunction
...
+            {
+                unsigned int bit1 = (i >> qubitLoc) & 1;
+                unsigned int bit2 = (i >> newPos) & 1;
+                unsigned int x = (bit1 ^ bit2);
+                x = (x << qubitLoc) | (x << newPos);
+                unsigned new_i = i ^ x;
+                if (new_i > i)
+                {
+                    std::iter_swap(wfn_.begin() + i, wfn_.begin() + new_i);
+                }
```
irinayat-MS, line 313: This is a complicated piece of math that is easy to get wrong. Have you considered adding unit tests that check this logic directly?

src/Simulation/Native/src/simulator/wavefunction.hpp
```diff
@@ -90,6 +90,53 @@ class Wavefunction
 
     void flush() const
     {
+        //logic to reorder
+        Fusion fg = fused_.getFusedGates();
+        auto items = fg.getItems();
```
irinayat-MS, line 95: This is extremely cryptic code that is hard to read: no type information and names are generic and use inconsistent plurals. For example, what does fg.getSet() return and why it's set*s*?

src/Simulation/Native/src/simulator/wavefunction.hpp
```diff
@@ -90,6 +90,53 @@ class Wavefunction
...
+        auto sets = fg.getSet();
+        auto ctrlSets = fg.getCtrlSet();
+        //getting all qubits to move to lower end of the wfn
+        if (items.size() > 0) {
+            std::vector<unsigned> indices;
+            std::set<unsigned> indicesSet;
```
irinayat-MS, line 101: std::unordered_set might be a better choice here

src/Simulation/Native/src/simulator/wavefunction.hpp
```diff
@@ -90,6 +90,53 @@ class Wavefunction
...
+            for (int i = 0; i < items.size(); i++) {
+                auto tempIndices = items[i].get_indices();
+                for (unsigned j = 0; j < tempIndices.size(); j++) {
+                    if (indicesSet.count(tempIndices[j]) == 0) {
+                        indices.push_back(tempIndices[j]);
```
irinayat-MS, line 106: it would be more efficient to dump indicesSet into a vector at once after the union set is populated (push_back reallocates the vector as it has to expand it): `indices.assign(indicesSet.begin(), indicesSet.end());`

src/Simulation/Native/src/simulator/wavefunction.hpp
```diff
@@ -90,6 +90,53 @@ class Wavefunction
...
+        auto sets = fg.getSet();
+        auto ctrlSets = fg.getCtrlSet();
+        //getting all qubits to move to lower end of the wfn
+        if (items.size() > 0) {
+            std::vector<unsigned> indices;
+            std::set<unsigned> indicesSet;
+            for (int i = 0; i < items.size(); i++) {
+                auto tempIndices = items[i].get_indices();
+                for (unsigned j = 0; j < tempIndices.size(); j++) {
+                    if (indicesSet.count(tempIndices[j]) == 0) {
```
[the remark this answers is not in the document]
irinayat-MS: I'd also rename the locals to be more descriptive, something along the lines of `indicesUnionForAllItems`

src/Simulation/Native/src/simulator/wavefunction.hpp
```diff
@@ -90,6 +90,53 @@ class Wavefunction
 
     void flush() const
     {
+        //logic to reorder
+        Fusion fg = fused_.getFusedGates();
+        auto items = fg.getItems();
+        auto sets = fg.getSet();
```
irinayat-MS, line 96: unused?

src/Simulation/Native/src/simulator/wavefunction.hpp
```diff
@@ -90,6 +90,53 @@ class Wavefunction
...
+                }
+            }
+            for (auto it = ctrlSets.begin(); it != ctrlSets.end(); ++it) {
+                if (indicesSet.count(*it) == 0) {
+                    indices.push_back(*it);
+                    indicesSet.insert(*it);
+                }
+            }
+            //performing reorder
+            std::vector<qubit_t> indexLocs = qubits(indices);
```
irinayat-MS, line 118: This seems rather inefficient: the qubits vector gets copied as a whole after each transposition of two qubits. Maybe, refactor to deal with in-place transposition of qubitmap here and use (change it to private) reorderWFN to update wfn_'s state only?

src/Simulation/Native/src/simulator/wavefunction.hpp
```diff
@@ -238,19 +290,55 @@ class Wavefunction
         rng_.seed(s);
     }
 
+    void reorderWFN(unsigned qubitLoc, unsigned newPos) const
```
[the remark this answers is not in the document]
irinayat-MS: in kernels.hpp there is:
`
template <class T, class A>
void swap(std::vector<T, A>& wfn, unsigned q1, unsigned q2)
`
how is it different from what you are trying to do here?

src/Simulation/Native/src/simulator/capi_test.cpp
```diff
@@ -106,33 +106,54 @@ void test_gates()
...
 
-    assert(M(sim_id, 1)==false);
+    assert(M(sim_id, 1) == false);
 
     X(sim_id, 1);
 
-    assert(M(sim_id, 1)==true);
+    assert(M(sim_id, 1) == true);
 
+    /*X(sim_id, 2);
```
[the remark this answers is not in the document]
AniketDalvi: I have now gotten rid of these, as they were just simpler tests used for debugging.

src/Simulation/Native/src/external/fusion.hpp
```diff
@@ -58,7 +64,48 @@ class Fusion{
...
+		return ctrl_set_;
+	}
+
+	Complex getGlobalFactor() const {
+		return global_factor_;
+	}
 
+	void setSet(std::map<unsigned, unsigned> elemDict) {
+		std::set<Index> tempSet;
+		for (unsigned elem : set_) {
```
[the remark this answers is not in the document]
AniketDalvi: This has been changed. When I looked at he original name in retrospect,  got a really good laugh out of the silliness of it :)

src/Simulation/Native/src/external/fusion.hpp
```diff
@@ -20,6 +21,11 @@ class Item{
 	Item(Matrix mat, IndexVector idx) : mat_(mat), idx_(idx) {}
 	Matrix& get_matrix() { return mat_; }
 	IndexVector& get_indices() { return idx_; }
+	void setIdx(std::map<unsigned, unsigned> elemDict) {
```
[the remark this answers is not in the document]
AniketDalvi: using ordered map here to allow for faster look up (which happens in order)

src/Simulation/Native/src/external/fusion.hpp
```diff
@@ -58,7 +64,48 @@ class Fusion{
...
+		for (unsigned elem : set_) {
+			if (elemDict.find(elem) != elemDict.end()) {
+				tempSet.insert(elemDict[elem]);
+			}
+		}
+		set_.clear();
+		set_.insert(tempSet.begin(), tempSet.end()); //look into using swap instead
+	}
+
+	void setCtrlSet(std::map<unsigned, unsigned> elemDict) {
```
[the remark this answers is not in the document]
AniketDalvi: I was thinking of this, however on seeing that it was just a couple of lines, decided to leave as is. Would you recommended it being taken out into a helper method?

src/Simulation/Native/src/external/fused.hpp
```diff
@@ -41,7 +41,14 @@ class Fused
       fusedgates = Fusion();
     }
 
+    Fusion getFusedGates() const {
+        return fusedgates;
+    }
     
+    void setFusedGates(Fusion newFG) const {
```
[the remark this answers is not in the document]
AniketDalvi: Using the getter and setter  nomenclature, as we are trying to update the fusedgates data structure

src/Simulation/Native/src/simulator/wavefunction.hpp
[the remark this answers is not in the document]
AniketDalvi: I believe that is the swap gate operation. This is meant to be the reordering routine - which updates the mapping and the wavefunction. Also in it s final form, this reordering routine will not just 2 qubits to be swapped with each other but a collection of them.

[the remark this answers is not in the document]
AniketDalvi: Also the abbreviation is in accordance with the rest of the file also using WFN

src/Simulation/Native/src/simulator/wavefunction.hpp
AniketDalvi: Removed it to streamline flush() call through the repo

src/Simulation/Native/src/simulator/wavefunction.hpp
AniketDalvi: Yes as it is called inside flush() which is const

src/Simulation/Native/src/external/fusion.hpp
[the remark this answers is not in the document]
irinayat-MS: I'm sorry, I don't follow. The spec for std::map (https://en.cppreference.com/w/cpp/container/map/operator_at) states the lookup to be logarithmic in size of the container. The lookups inside the loop are still done one at a time even if they happen for consequtive elements in the map so, unless cache locality kicks in big time here, I wouldn't expect them to be faster than O(1) for the unordered_map (which is, effectively, a hash table). If you *are* seeing locality effects and that is driving your choice of the datastructure, please add a comment to document it.

src/Simulation/Native/src/simulator/wavefunction.hpp
AniketDalvi: Oh that is a great point! Let me look into this

src/Simulation/Native/src/external/fusion.hpp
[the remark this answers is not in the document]
irinayat-MS:
```
	static void remap_qubits(std::set<Index>& qubits, const std::map<unsigned, unsigned>& elemDict) {
		std::set<Index> tempSet;
		for (unsigned elem : qubits) {
			if (elemDict.find(elem) != elemDict.end()) {
				tempSet.insert(elemDict[elem]);
			}
		}
		qubits.swap(tempSet);
	}
    void setSet(std::map<unsigned, unsigned> elemDict) { remap_qubits(set_, elemDict); }
    void setCtrlSet(std::map<unsigned, unsigned> elemDict) { remap_qubits(ctrl_set_, elemDict); }
```
(haven't tried to compile the above, but the idea is that you should be able to extract _all_ of the logic into the helper)

src/Simulation/Native/src/external/fused.hpp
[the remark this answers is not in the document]
irinayat-MS: getFoo/setFoo is fine, I'm not fond of abbreviations in names of functions and arguments, especially when they don't match the type names.

src/Simulation/Native/src/simulator/wavefunction.hpp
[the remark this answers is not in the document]
irinayat-MS: Would it make sense to implement `swap` in terms of reorder_wavefunction?
Ss far as naming is concerned, having local variables named like "wfn" might be fine because the context is local and clear, but function names serve as documentation, especially at the callsites, and in talking about code we often refer to function names so it makes sense for them to be pronouncable...

src/Simulation/Native/src/simulator/wavefunction.hpp
irinayat-MS: Is keeping Fused::subsytemwavefunction still useful?

src/Simulation/Native/src/external/fused.hpp
```diff
@@ -41,7 +41,14 @@ class Fused
       fusedgates = Fusion();
     }
 
+    Fusion getFusedGates() const {
```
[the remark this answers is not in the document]
AniketDalvi: You bring out a very interesting point here, and I noticed that you brought this up in multiple places. However, making it a reference takes us away from the getFoo/setFoo design pattern used throughtout the repository. This change would require a significant overhaul in many aspects of the codebase. We could look into making use of std:move, however it would be appropriate to get back to this as part of a another PR linking it to an appropriate issue. Given that the data structures these work on are set of gates, I believe copying does not add much overhead. right now. Thank you for pointing this out!

src/Simulation/Native/src/external/fusion.hpp
```diff
@@ -20,6 +21,11 @@ class Item{
 	Item(Matrix mat, IndexVector idx) : mat_(mat), idx_(idx) {}
```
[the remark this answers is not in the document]
AniketDalvi: Thank you for pointing this out. However, this change takes us away a similar design pattern used throughtout the repository. This change would require a significant overhaul in many aspects of the codebase. We could look into making use of std:move, however it would be appropriate to get back to this as part of a another PR linking it to an appropriate issue.

src/Simulation/Native/src/external/fusion.hpp
```diff
@@ -58,7 +64,48 @@ class Fusion{
 		handle_controls(empty_matrix, empty_vec, {}); // remove all current control qubits (this is a GLOBAL factor)
 	}
 	
+	IndexSet getSet() const {
```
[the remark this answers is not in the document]
AniketDalvi: You bring out a very interesting point here, and I noticed that you brought this up in multiple places. However, making it a reference takes us away from the getFoo/setFoo design pattern used throughtout the repository. This change would require a significant overhaul in many aspects of the codebase. We could look into making use of std:move, however it would be appropriate to get back to this as part of a another PR linking it to an appropriate issue. Given that the data structures these work on are set of gates, I believe copying does not add much overhead. right now. Thank you for pointing this out!

src/Simulation/Native/src/external/fusion.hpp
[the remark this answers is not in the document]
irinayat-MS: You can avoid the extra copy by a one line fix inside the constructor which doesn't affect any of the callers and makes the code more efficient. Having multiple bugs of this type doesn't make any one of them less buggy. However, I do agree that this is a pre-existing code and _not_ fixing it as part of this PR is a valid choice.

src/Simulation/Native/src/external/fusion.hpp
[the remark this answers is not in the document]
irinayat-MS: If a design pattern leads to a lesser quality code, the design pattern shouldn't be used. You can keep the name as get_foo, if you so prefer (though I personally like to reflect the semantics in the name), the callsites then won't have to change but won't incure the copy costs _unless they need their own copy of the set_, which they can still do explicitly.

The getters you are adding are wasteful and can be easily fixed without the overhead of a follow-up PR that would have to cleanup the mess.

src/Simulation/Native/src/simulator/wavefunction.hpp
AniketDalvi: The reason this approach is being used is because we want to preserve the original order of the qubits in the target and control sets. Dumping the set into the vector will result in them being scrambled into an ascending order. Even if we switch to an unordered set it will not guarantee the preservation of the original order

src/Simulation/Native/src/external/fusion.hpp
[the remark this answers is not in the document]
swernli: I think Aniket's point here is that there are other constructors throughout the simulator that should also be updated to follow this. Rather than have one constructor use one pattern while the others use the old one, we should do a single change that updates all the constructors to use `std::move` that is not tied to this PR. That makes sure that the pattern is fixed everywhere at once and makes it easier to review and discuss that change without having get mixed in with this PR that is meant to introduce the new functionality.

src/Simulation/Native/src/external/fusion.hpp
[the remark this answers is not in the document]
swernli: Switching to a reference pattern would also mean removing the setters, wouldn't it? Each time get is called, any changes made to the returned structure would already be on the internal structure. I wonder if a move-based approach would work better, which would have the added benefit of only letting one part of the code own the data structure at a time (kind of like the Rust semantics).

That said, given how small these data structures are, this is a minor enough optimization that I think noting it and addressing it in a later change would make sense.

src/Simulation/Native/src/simulator/wavefunction.hpp
AniketDalvi: Just curious to understand the reason behind this recommendation?

src/Simulation/Native/src/simulator/wavefunction.hpp
irinayat-MS: Please document this requirement with a comment because it's not obvious from the code, and someone might later attempt refactors that would violate it. To avoid the mltiple re-allocations of the indices vector, consider reserving enough space upfront.

src/Simulation/Native/src/simulator/wavefunction.hpp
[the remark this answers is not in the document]
AniketDalvi: That could be an interesting proposition, but as it is not directly related to the purpose of this feature branch, I personally feel it could be more appropriate incorporated with a different PR appropriately linked to an issue.

src/Simulation/Native/src/external/fusion.hpp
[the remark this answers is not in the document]
irinayat-MS: returning const& prevents modifying the vectors by the clients and is usually the most efficient way to allow read-only access to the data. The setters that swap in a new data-structure as a whole are still valuable and have different semantics compared to allowing modifying the data in place. The structures aren't that small either (vectors of vectors plus other stuff) in some of the cases) -- why return by value if there is an established C++ pattern for such scenarios?

src/Simulation/Native/src/simulator/wavefunction.hpp
irinayat-MS: same as for map versus unordered_map -- it's likely to be more efficient

src/Simulation/Native/src/simulator/wavefunction.hpp
AniketDalvi: The tests currently present test the reordering routine, thereby confirming the correctness of this bit math. However, as far as testing this directly I am unsure of how that could be done, as this method is deeply embeded making it very difficult to access outside the class

src/Simulation/Native/src/external/fusion.hpp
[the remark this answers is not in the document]
irinayat-MS: As I said above, not fixing pre-existing code with this PR is a valid choice. I still believe that this pattern is sub-optimal and shouldn't be used to justify new code.

src/Simulation/Native/src/simulator/wavefunction.hpp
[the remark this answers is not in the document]
irinayat-MS: To your discretion. It might be confusing to readers of the code (as it was to me) to see two differently named functions that are implemented exactly the same, but it's not a big deal.

src/Simulation/Native/src/simulator/wavefunction.hpp
irinayat-MS: A common way to improve testability is to extract the functionality to be tested into a standalone function or a helper class with public methods to expose the code. The function/class then can be tested with given parameters in known state, while the production class will use them with its internal data. This way you can ensure much better test coverage and easier debugging of failures.

src/Simulation/Native/src/simulator/wavefunction.hpp
AniketDalvi: I will hold this of until the efficient reorder routine is implemented over the next week or so!

src/Simulation/Native/src/external/fusion.hpp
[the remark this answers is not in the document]
AniketDalvi: I have added const&, and it seemed that the callsites did not required any change as I anyway was just using them to create new structures which eventually get set later

src/Simulation/Native/src/simulator/wavefunction.hpp
AniketDalvi: Given the lack of any testing infrastructure in the native simulator, I want to punt this to after I have a discussion with Dave and Thomas. But will definitely keep this in mind!

src/Simulation/Native/src/external/nointrin/kernel1.hpp
```diff
@@ -1,4 +1,4 @@
-// (C) 2018 ETH Zurich, ITP, Thomas Häner and Damian Steiger
+// (C) 2018 ETH Zurich, ITP, Thomas HΣner and Damian Steiger
```
irinayat-MS, line 1: unintended change?

src/Simulation/Native/src/simulator/wavefunction.hpp
```diff
@@ -90,6 +91,52 @@ class Wavefunction
 
     void flush() const
     {
+        //logic to reorder
```
irinayat-MS, line 94: style: the common convention is to have a single space between `//` and the text that follows in text comments, and no space when commenting out code

src/Simulation/Native/src/external/nointrin/kernel1.hpp
swernli: Yes, this change should be reversed. Something in the git tools keeps doing this; I've had to reverse this change (and other occurrences of `Häner`) several times.

src/Simulation/Native/src/external/fused.hpp
```diff
@@ -41,7 +41,14 @@ class Fused
       fusedgates = Fusion();
     }
 
+    Fusion get_fusedgates() const& {
```
irinayat-MS, line 44: most people would be more familiar with signature like `const Fusion& get_fusedgates() const`, the first const referring to the returned type `Fusion&` and the second stating that the function doesn't change the object it's invoked on. I'm actually not sure what the signature with `const&` after function name means and I wouldn't be looking forward to reading the C++ standard on this (I'm surprised it compiles) :D

src/Simulation/Native/src/external/fused.hpp
```diff
@@ -41,7 +41,14 @@ class Fused
...
+        return fusedgates;
+    }
     
+    void set_fusedgates(Fusion newFusedGates) const {
```
irinayat-MS, line 48: declaring setters as `const` doesn't make sense: the purpose of the setter is to change the object, isn't it?

src/Simulation/Native/src/external/fused.hpp
```diff
@@ -133,16 +144,9 @@ class Fused
         }
 
         // New rules of when to stop fusing
-        Fusion::IndexVector qs      = std::vector<unsigned>(1, q);
-        if (fusedgates.predict(qs, cs) > maxFusedSpan || fusedgates.size() >= maxFusedDepth)  flush(wfn);
-        fusedgates.insert(convertMatrix(mat), qs, cs);
-    }
+        Fusion::IndexVector qs = std::vector<unsigned>(1, q);
```
irinayat-MS, line 147: Mix-and-match of custom type aliases and underlying built-in types makes the code harder to understand and implies there might be some non-trivial conversion going on here. I'd suggest to replace it with `Fusion::IndexVector qs(1, q);` or `Fusion::IndexVector qs{q};`.

src/Simulation/Native/src/external/fusion.hpp
```diff
@@ -58,21 +65,58 @@ class Fusion{
 		handle_controls(empty_matrix, empty_vec, {}); // remove all current control qubits (this is a GLOBAL factor)
 	}
 	
+	IndexSet get_target_set() const& {
```
irinayat-MS, line 68: same comment as in fused.hpp: this might be equivalent to `const IndexSet& get_target_set()` declaration but it's more traditional to do it with type modifiers unfront. Also make function const: `const IndexSet& get_target_set() const`

src/Simulation/Native/src/external/fusion.hpp
```diff
@@ -58,21 +65,58 @@ class Fusion{
...
+		std::set<Index> tempSet;
+		for (unsigned elem : qubits) {
+			if (mapFromOldLocToNewLoc.find(elem) != mapFromOldLocToNewLoc.end()) {
+				tempSet.insert(mapFromOldLocToNewLoc.at(elem));
+			}
+		}
+		qubits.swap(tempSet);
+	}
+
+	void update_target_set(std::unordered_map<unsigned, unsigned> mapFromOldLocToNewLoc) {
```
irinayat-MS, line 94: My bad, in the code snippet I suggested for the refactor, I forgot to change these signatures to `const ...&`. Could you pleaes do it to avoid the extra copy?

src/Simulation/Native/src/external/fusion.hpp
```diff
@@ -58,21 +65,58 @@ class Fusion{
...
+		remap_qubits(target_set_, mapFromOldLocToNewLoc);
+	}
+
+	void set_ctrl_set(std::unordered_map<unsigned, unsigned> mapFromOldLocToNewLoc) {
```
irinayat-MS, line 98: Maybe name it similar to the above `update_ctrl_set`? Also, why not `remap_target_set` and `remap_ctrl_set`?

src/Simulation/Native/src/external/fused.hpp
AniketDalvi: This needs to be set as const, as I am calling it in flush() which is const. Yes, the setter is supposed to change the object - but in this case by change here we are changing the contents of the container, not the container or its location.

src/Simulation/Native/src/simulator/wavefunction.hpp
irinayat-MS: Why not add a test to local_test.cpp? I agree that the test infrastructure in the native simulator is far from being convenient, but it does allow to create instances of Wavefunction class and execute public methods on them, which would give you direct access to the code you are adding. Given the complexity of the math here I am not comfortable proving its correctness by review and system-level tests only.

src/Simulation/Native/src/external/fused.hpp
irinayat-MS: As a part of _public class interface_ a const setter doesn't make much sense, whatever the implementation details of other methods might be. If clients outside of the class need to use the setter (for example, tests), make it non-const for them and inside `flush()` do the assignment directly, without going through the setter.

src/Simulation/Native/src/simulator/wavefunction.hpp
```diff
@@ -90,6 +91,52 @@ class Wavefunction
...
+        Fusion fg = fused_.get_fusedgates();
```
irinayat-MS, line 95: You are still creating copies in these three lines (changing the getters to const& allows the callers to choose whether to copy or not, and these callers have chosen to copy). I find the data-model rather confusing: are we copying because we need snapshot of the data? because there is multi-threaded access? for exception safety? for some other reason?

src/Simulation/Native/src/simulator/wavefunction.hpp
```diff
@@ -90,6 +91,52 @@ class Wavefunction
...
+        auto itemsToFuse = fg.get_items();
+        auto ctrlSet = fg.get_ctrl_set();
+        //getting all qubits to move to lower end of the wfn
+        if (itemsToFuse.size() > 0) {
```
irinayat-MS, line 99: nit: `if (!itemsToFuse.empty())` (`empty()` is more likely to be O(1) than `size()`)

src/Simulation/Native/src/simulator/wavefunction.hpp
```diff
@@ -90,6 +91,52 @@ class Wavefunction
...
+            std::vector<unsigned> unionOfAllQubitsInUse;
+            std::unordered_set<unsigned> indicesSet; //set is introduced to guard against duplicate insertion and maintianing original order
```
irinayat-MS, line 101: I don't understand the ordering requirement. Suppose, you have two items to fuze, one with set {0, 2} and the other with set {1, 3}. `unionOfAllQubitsInUse` will end up looking like {0, 2, 1, 3} -- what order does this represent?

AniketDalvi: This order represents that quibt {0,2} are acted on first, and then {1,3}. This order helps ensure that the qubits being acted on first occupy the lowest end of the wavefunction, and so on. Although this requirement will be nullified as I make my next optimization change for the reorder routine, for the purpose of this PR it was decided that we maintain original order while reordering

src/Simulation/Native/src/simulator/wavefunction.hpp
```diff
@@ -90,6 +91,52 @@ class Wavefunction
...
+            for (int i = 0; i < itemsToFuse.size(); i++) {
+                auto tempIndices = itemsToFuse[i].get_indices();
+                for (unsigned j = 0; j < tempIndices.size(); j++) {
+                    if (indicesSet.count(tempIndices[j]) == 0) {
+                        unionOfAllQubitsInUse.push_back(tempIndices[j]);
+                        indicesSet.insert(tempIndices[j]);
+                    }
+                }
+            }
+            for (auto it = ctrlSet.begin(); it != ctrlSet.end(); ++it) {
```
irinayat-MS, line 111:
```
for (unsigned index : ctrlSet) {
    if (indicesSet.count(index) == 0) {
        unionOfAllQubitsInUse.push_back(index);
        indicesSet.insert(index);
    }
}
```
(I didn't try to compile the above but range loops should work for most of the std containers)

src/Simulation/Native/src/simulator/wavefunction.hpp
irinayat-MS: What if the sets overlap? {2, 3} and {0, 1, 2} will produce sequence of {2, 3, 0, 1} -- is this expected?
Also, are the indices inside an itemsToFuse[i].get_indices() sorted in some defined way?

src/Simulation/Native/src/simulator/wavefunction.hpp
```diff
@@ -90,6 +91,52 @@ class Wavefunction
...
+                    indicesSet.insert(*it);
+                }
+            }
+            //performing reorder
+            std::vector<qubit_t> indexLocs = qubits(unionOfAllQubitsInUse);
+            for (unsigned i = 0; i < indexLocs.size(); i++)
+            {
+                auto currLoc = indexLocs[i];
+                reorder_wavefunction(currLoc, i);
+                indexLocs = qubits(unionOfAllQubitsInUse);
```
irinayat-MS, line 123: is this repeated assignment inside the loop necessary?

src/Simulation/Native/src/simulator/wavefunction.hpp
```diff
@@ -90,6 +91,52 @@ class Wavefunction
...
+            }
+            //keeping old and new location in order to set it appropriately
+            std::unordered_map<unsigned, unsigned> old2newDict;
+            for (unsigned i = 0; i < unionOfAllQubitsInUse.size(); i++) {
+                old2newDict[unionOfAllQubitsInUse[i]] = indexLocs[i];
+            }
+
+            for (int i = 0; i < itemsToFuse.size(); i++) {
+                itemsToFuse[i].set_idx(old2newDict);
```
irinayat-MS, line 132: `remap_idx` seems to me to be a better name for what the function does

src/Simulation/Native/src/simulator/wavefunction.hpp
```diff
@@ -90,6 +91,52 @@ class Wavefunction
...
+                reorder_wavefunction(currLoc, i);
+                indexLocs = qubits(unionOfAllQubitsInUse);
+            }
+            //keeping old and new location in order to set it appropriately
+            std::unordered_map<unsigned, unsigned> old2newDict;
+            for (unsigned i = 0; i < unionOfAllQubitsInUse.size(); i++) {
+                old2newDict[unionOfAllQubitsInUse[i]] = indexLocs[i];
+            }
+
+            for (int i = 0; i < itemsToFuse.size(); i++) {
```
irinayat-MS, line 131: As far as I understand this code, it could be thought as:
1. get current fusion state
2. use the state to reorder the wave function and to create mapping dictionary (the fusion state doesn't change during this step)
3. remap the current fusion state using the dictionary

If this understanding is correct, the code could be refactored to be more efficient by
a. use const references to the current Fusion state as we only need to read it and it doesn't update during step2 above
b. do the collection of qubit ids and reorder the wave function (basically, as it is now)
c. add a single `void remap_fusion(const std::unordered_map<unsigned, unsigned>& old2newDict) {...}` method on fused_ to fully update all corresponding structures.

src/Simulation/Native/src/simulator/wavefunction.hpp
```diff
@@ -238,19 +285,61 @@ class Wavefunction
         rng_.seed(s);
     }
 
+    void reorder_wavefunction(unsigned qubitLoc, unsigned newPos) const
+    {
+
```
irinayat-MS, line 290: nit: please remove the blank line (and at the end of the function as well)

src/Simulation/Native/src/simulator/wavefunction.hpp
```diff
@@ -238,19 +285,61 @@ class Wavefunction
...
+    void reorder_wavefunction(unsigned qubitLoc, unsigned newPos) const
+    {
+        
+        // swap qubits in wfn between qubitLoc and newPos
+        if (newPos != qubitLoc)
+        {
+            // get id of qubit located at newPos and qubitLoc - getting index from the element
+            auto newQubitLocItr = std::find(qubitmap_.begin(), qubitmap_.end(), newPos);
+            assert(newQubitLocItr != qubitmap_.end());
+            auto newPosId = std::distance(qubitmap_.begin(), newQubitLocItr);
```
irinayat-MS, line 297: newPosId seems unused?

src/Simulation/Native/src/simulator/wavefunction.hpp
```diff
@@ -238,19 +285,61 @@ class Wavefunction
...
+        
+    }
+
     /// generic application of a gate
     template <class Gate>
     void apply(Gate const& g)
     {
-        fused_.apply(wfn_, g.matrix(), qubit(g));
+        //check flush condition
+        std::vector<qubit_t> cs;
```
irinayat-MS, line 324: nit: I'd inline both into the corresponding function calls:
```
if (fused_.shouldFlush(wfn_, std::vector<qubit_t>{}, g.qubit())) {
    flush();
}
fused_apply(wfn_, g.matrix(), g.qubit());
```

src/Simulation/Native/src/external/fused.hpp
AniketDalvi: Apologies, I did not really understand what you were trying to convey here completely. Unfortunately because the data structure fusedgates is private, I cannot make the assignments outside this class, thereby not allowing me to make assignments in flush() directly. Given this, the use of a setter is necessary to alter the datastructure. And as this setter is being called in a function which is declared const, the language does not allow me to have the setter to be non-const.

src/Simulation/Native/src/simulator/wavefunction.hpp
AniketDalvi: The reason we cannot do a single update after the loop is because every iteration of the reorder routine changes the mapping, and if this updated mapping is not used for each iteration it results in buggy logic.

irinayat-MS: ah, yes, you are reading from the updated indexLocks inside the loop... 

The code is extremely hard to understand and britle. For example, going back to the orderding questions I had above: the loop progresses through indexLocs array even though its content keeps changing -- what does it mean for the final reordering? what state will we be in for variours possible inputs? which ones of them are a win for perf and which aren't? can we end up reordering the same qubit back and forth?

src/Simulation/Native/src/external/fused.hpp
irinayat-MS: "funny" public interfaces like this usually indicate that there is an issue with design of data structures: basically, you are trying to put a square peg into a round hole. You have multiple choices to address it:
1. keep entrenching the previous design decision(s) and add public interfaces that don't conform to good practices (ugly)
2. keep the previous design decision intact and workaround by adding const_cast(s) (ugly but obvious to spot in code and keeps badness from spreading)
3. remove `const` from member functions that violate constness (easy, as any chickening out is, but honest :))
4. rethink the logical\representational constness model of your data structures and get it right (hard and, given where C++ is re constness, the results might not be that different from option 3)

Personally, for a quick solution I'd go with either 2 or 3.
Your call.

src/Simulation/Native/src/simulator/wavefunction.hpp
AniketDalvi: Yes that is expected! Also I believe the get_indices() is sorted as target first, followed by controls

src/Simulation/Native/src/external/fused.hpp
AniketDalvi: Thought about this a little. Given a couple of factors - what this branch is trying to achieve and considering that I have just a couple of weeks left, I think it will be good idea for me to create an issue outlining this very relevant issue you bring up which can be tackled later. The only reason I say that is because I feel this deviates a little from the original purpose of this branch, and that making this const change is going to take some time as it very deeply entrenched in the simulator codebase, and I am hoping to get a couple of more branches checked-in before the end of my internship, and want to leave enough time for that. Thank you!

src/Simulation/Native/src/external/fusion.hpp
```diff
@@ -183,17 +227,17 @@ class Fusion{
...
 			for (auto &item : items_)
 				add_controls(item.get_matrix(), item.get_indices(), new_ctrls);
 		}
 	}
 	
-	IndexSet set_;
-	ItemVector items_;
-	IndexSet ctrl_set_;
-	Complex global_factor_;
+	mutable IndexSet target_set_; //set of qubits being acted on
```
irinayat-MS, line 237: I still believe that it would be better to remove `const` from flush - it just makes no sense that *all* members of a class are mutable...

irinayat-MS:
Without the understanding of the impact of the reordering in the case of multiple fused groups I don't feel comfortable pushing this code into production. If you must merge it, please consider adding a configuration switch that allows running it locally (e.g. with benchmarks) but doesn't enable the code for the end users yet. There are also still a few specific comments that haven't been addressed.

However, assuming you've discussed this with others, I don't want to block the PR so I'm changing my status to "Approve changes" (as there seems to be no way to remove my change request without approving). This doesn't indicate my sign off, though. Please seek a sign off from your lead and/or from the people you worked with on the design of this change.

PR #294, opened 2020-07-02
Implement friendly URI option for Azure quantum machines

cesarzc:
This change implements the friendly URI property in the `CloudJob` class, which is used by Azure quantum machines.

Additionally, it enables the friendly URI option in the entry point driver for standalone executables.

### DIFF
--- a/src/Azure/Azure.Quantum.Client/JobManagement/CloudJob.cs
+++ b/src/Azure/Azure.Quantum.Client/JobManagement/CloudJob.cs
@@ -60,7 +60,7 @@ namespace Microsoft.Azure.Quantum
         /// <summary>
         /// Gets an URI to access the job.
         /// </summary>
-        public Uri Uri => throw new NotImplementedException();
+        public Uri Uri => GenerateUri();
 
         /// <summary>
         /// Gets the workspace.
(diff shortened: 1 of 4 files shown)


src/Azure/Azure.Quantum.Client/JobManagement/CloudJob.cs
```diff
@@ -91,5 +91,16 @@ public async Task CancelAsync(CancellationToken cancellationToken = default)
...
         }
+
+        private Uri GenerateUri()
+        {
+            if (!(this.Workspace is Workspace cloudWorkspace))
+            {
+                throw new NotSupportedException($"{typeof(CloudJob)}'s Workspace is not of type {typeof(Workspace)} and does not have enough data to generate URI");
+            }
+
+            var uriStr = $"https://ms.portal.azure.com/#@microsoft.onmicrosoft.com/resource/subscriptions/{cloudWorkspace.SubscriptionId}/resourceGroups/{cloudWorkspace.ResourceGroupName}/providers/Microsoft.Quantum/Workspaces/{cloudWorkspace.WorkspaceName}/job_management?microsoft_azure_quantum_jobid={Id}";
```
bettinaheim, line 102: Is there a tests that fails if this uri changes in the future?

cesarzc: If the URI of the service changes it does not fail but there various tests would fail if the URI in this implementation changes.

cesarzc: Sorry, I was a bit confused, there is no test for that yet. I will add one.

cesarzc: This specific test will be added in another repo.

PR #297, opened 2020-07-06
Simulator Classical Control Execution Tests

ScottCarda-MS:
Adds execution tests to ensure that the simulators can run code containing classical comparisons of measurement results.


src/Simulation/Simulators.Tests/TestProjects/HoneywellExe/ClassicallyControlledSupportTests.qs
```diff
@@ -0,0 +1,732 @@
...
+    }
+
+    operation LiteralOnTheLeft() : Unit {
+        let r = Zero;
+        if (Zero == r) {
+            SubOp1();
+        }
+    }
+
+    //operation GenericsSupport<'A, 'B, 'C>() : Unit {
```
bettinaheim, line 302: What is the reason this is commented out? Not that that would be an issue - just curious.

src/Simulation/Simulators.Tests/TestProjects/HoneywellExe/ClassicallyControlledSupportTests.qs
```diff
@@ -0,0 +1,732 @@
...
+            SubOp3();
+            let temp = 4;
+            using (q = Qubit()) {
+                let temp2 = q;
+            }
+        }
+    }
+
+    operation LiftLoops() : Unit {
+        let r = Zero;
```
bettinaheim, line 29: Would you mind modifying some of these to just actually allocate and measure a qubit? Now that we actually tests execution, it seems like a good idea to check that all works as expected with allocations and qubit use as well.

src/Simulation/Simulators.Tests/TestProjects/IonQExe/MeasurementSupportTests.qs
```diff
@@ -0,0 +1,30 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+
+namespace Microsoft.Quantum.Simulation.Testing.IonQ.MeasurementSupportTests {
+
+    open Microsoft.Quantum.Intrinsic;
+    
+    operation MeasureInMiddle() : Unit {
```
bettinaheim, line 9: Perfect, thanks!

src/Simulation/Simulators.Tests/TestProjects/QCIExe/ClassicallyControlledSupportTests.qs
```diff
@@ -0,0 +1,732 @@
...
+            SubOp3();
+            let temp = 4;
+            using (q = Qubit()) {
+                let temp2 = q;
+            }
+        }
+    }
+
+    operation LiftLoops() : Unit {
+        let r = Zero;
```
bettinaheim, line 29: Same in this file; if you could just copy over the adaptions I mentioned for Honeywell, that would be great.

src/Simulation/Simulators.Tests/TestProjects/HoneywellExe/ClassicallyControlledSupportTests.qs
ScottCarda-MS: When this is not commented out, even without the corresponding test, I get the error:
CS0305	Using the generic type '_0937a8eea43e4297a95fbffc1a4e8b2f_GenericsSupport<__A__, __B__, __C__>' requires 3 type arguments.

src/Simulation/Simulators.Tests/TestProjects/HoneywellExe/ClassicallyControlledSupportTests.qs
ScottCarda-MS: I can add the tests that IonQ has to Honeywell and QCI. Would that address this?

ScottCarda-MS: Oh I see what you mean, instead of `let r = Zero;`, you want `let r = M(q);`, right?

ScottCarda-MS: Maybe that should be tested separately anyway though? If I leave the tests the way they are and add the tests from IonQ, then, if something goes wrong with the measurement on a qubit, just the new tests would fail and it would be more informative. Or if something goes wrong with the conditionals, then just the relevant tests would fail and the tests from IonQ would still pass. I think this would be more informative have more atomicity in the tests. I will add the tests from IonQ to Honeywell and QCI.

ScottCarda-MS: I've added the tests from IonQ that will check that measurement of qubits works.

bettinaheim: Adding the tests for IonQ sounds good, and adding a separate tests where you actually branch based on the measurement outcome for an allocated qubit would also be a good thing to do.

src/Simulation/Simulators.Tests/TestProjects/HoneywellExe/ClassicallyControlledSupportTests.qs
bettinaheim: That is something that should be fixed then, though - depending on what the source of the issue is as part of this PR (if it is related the the targeting) or in a separate PR (if it is unrelated and also occurs for a project that is not targeted). Could you please investigate which one it is?

src/Simulation/Simulators.Tests/TestProjects/HoneywellExe/ClassicallyControlledSupportTests.qs
ScottCarda-MS: Done

src/Simulation/Simulators.Tests/TestProjects/HoneywellExe/ClassicallyControlledSupportTests.qs
ScottCarda-MS: This issue is not related to targeting, but seems to be a bug in the C# generation for generic callables. I will file an issue.

ScottCarda-MS: Here is the issue for this: https://ms-quantum.visualstudio.com/Quantum%20Program/_workitems/edit/17245
I'm going to close out this conversation as this is not related to the purpose of this PR.

ScottCarda-MS: I've also added ToDos in the code for this.

samarsha: Since non-employees can't see ADO work items, could you create an issue on the qsharp-runtime GitHub repo?

ScottCarda-MS: I've added the following issue for this: #299

bettinaheim: @ScottCarda-MS The issue is related to targeting as the same code compiles fine if you don't target the project.

PR #300, opened 2020-07-08
Remove Negative Tests from Execution Testing

ScottCarda-MS:
Removing the negative execution tests from the classically controlled execution testing. These tests contain behavior that is not supported by their targets, by definition, and so cause the builds to fail.

### DIFF
--- a/src/Simulation/Simulators.Tests/TestProjects/HoneywellExe/ClassicallyControlledSupportTests.qs
+++ b/src/Simulation/Simulators.Tests/TestProjects/HoneywellExe/ClassicallyControlledSupportTests.qs
@@ -60,40 +60,11 @@ namespace Microsoft.Quantum.Simulation.Testing.Honeywell.ClassicallyControlledSu
             let temp = 2;
         }
     }
-    
-    operation DontLiftReturnStatements() : Unit {
-        let r = Zero;
-        if (r == Zero) {
-            SubOp1();
-            return ();
-        }
-    }
-
-    function SubFunc1() : Unit { }
-    function SubFunc2() : Unit { }
-    function SubFunc3() : Unit { }
-
-    function DontLiftFunctions() : Unit {
-        let r = Zero;
-        if (r == Zero) {
-            SubFunc1();
-            SubFunc2();
-            SubFunc3();
-        }
-    }
 
     operation LiftSelfContainedMutable() : Unit {
         let r = Zero;
         if (r == Zero) {
             mutable temp = 3;
-            set temp = 4;
-        }
-    }
-
-    operation DontLiftGeneralMutable() : Unit {
-        let r = Zero;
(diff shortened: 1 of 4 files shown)


src/Simulation/Simulators.Tests/TestProjects/HoneywellExe/ClassicallyControlledSupportTests.qs
```diff
@@ -23,724 +23,658 @@ namespace Microsoft.Quantum.Simulation.Testing.Honeywell.ClassicallyControlledSu
...
-            SubOp1();
-            return ();
-        }
-    }
-
-    function SubFunc1() : Unit { }
-    function SubFunc2() : Unit { }
-    function SubFunc3() : Unit { }
-
-    function DontLiftFunctions() : Unit {
```
bettinaheim, line 76: Why is this one removed? This should be valid code.

src/Simulation/Simulators.Tests/TestProjects/QCIExe/ClassicallyControlledSupportTests.qs
```diff
@@ -60,27 +60,6 @@ namespace Microsoft.Quantum.Simulation.Testing.QCI.ClassicallyControlledSupportT
...
-            SubOp1();
-            return ();
-        }
-    }
-
-    function SubFunc1() : Unit { }
-    function SubFunc2() : Unit { }
-    function SubFunc3() : Unit { }
-
-    function DontLiftFunctions() : Unit {
```
bettinaheim, line 76: Same here.

src/Simulation/Simulators.Tests/TestProjects/HoneywellExe/ClassicallyControlledSupportTests.qs
ScottCarda-MS: This code generates the following error:
`QS5024: Measurement results cannot be compared here. The execution target HoneywellProcessor only supports comparing measurement results as part of the condition of an if- or elif-statement in an operation.`
This error seems appropriate and tells me this code is not appropriate as an execution test.

src/Simulation/Simulators.Tests/TestProjects/QCIExe/ClassicallyControlledSupportTests.qs
ScottCarda-MS: See other comment.

src/Simulation/Simulators.Tests/TestProjects/HoneywellExe/ClassicallyControlledSupportTests.qs
bettinaheim: Ah, I see - DontLiftFunctions itself is a function.

PR #301, opened 2020-07-08
Add Runtime Metadata to Operations

theRoughCode:
This PR moves some code from [this PR](https://github.com/microsoft/iqsharp/pull/195) to this repo.

We add a `GetRuntimeMetadata` method to `ICallable`s that contains metadata to be used by the `ExecutionPathTracer` in that the iqsharp repo. The reason for moving the logic over here is to prevent circular dependencies between repos and to increase the modularity of defining custom `RuntimeMetadata` for each operation.

The default `RuntimeMetadata` is defined in `Operations.cs`, and any custom logic is overridden in the operation's specific namespace.

Tests are added that test the output of `GetRuntimeMetadata` of each of the custom logic we defined.


src/Simulation/Core/RuntimeMetadata.cs
```diff
@@ -0,0 +1,60 @@
...
+        /// </summary>
+        public bool IsControlled { get; set; }
+
+        /// <summary>
+        /// True if operation is a measurement operation.
+        /// </summary>
+        public bool IsMeasurement { get; set; }
+
+        /// <summary>
+        /// True if operation is composed of multiple operations.
```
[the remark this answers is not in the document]
theRoughCode: This would be like an `ApplyToEach` statement. Currently, it's not being used yet as I'm still working on that feature.

src/Simulation/Core/RuntimeMetadata.cs
```diff
@@ -0,0 +1,60 @@
...
+        /// List of control registers.
+        /// </summary>
+        public IEnumerable<Qubit> Controls { get; set; } = new List<Qubit>();
+
+        /// <summary>
+        /// List of target registers.
+        /// </summary>
+        public IEnumerable<Qubit> Targets { get; set; } = new List<Qubit>();
+    }
+}
```
cgranade, line 60: [nit] It's generally a good idea to have blank lines at the ends of files.
```suggestion
}

```

src/Simulation/Core/TypeExtensions.cs
```diff
@@ -144,6 +146,38 @@ public static Type[] GetTupleFieldTypes(this Type arg)
...
+                    {
+                        argString = f.GetValue(args)?.ToString();
+                    }
+
+                    return argString;
+                })
+                .WhereNotNull();
+
+            return argsStrings.Any()
+                ? $"({string.Join(",", argsStrings)})"
```
cgranade, line 177:
```suggestion
                ? $"({string.Join(", ", argsStrings)})"
```

src/Simulation/Core/TypeExtensions.cs
```diff
@@ -144,6 +146,38 @@ public static Type[] GetTupleFieldTypes(this Type arg)
...
+                        var nestedArgs = f.GetValue(args);
+                        if (nestedArgs != null) argString = f.FieldType.ArgsToString(nestedArgs);
+                    }
+                    // Add field as an argument if it is not a Qubit type
+                    else if (!f.FieldType.IsQubitsContainer())
+                    {
+                        argString = f.GetValue(args)?.ToString();
+                    }
+
+                    return argString;
```
cgranade, line 172:
```suggestion
                    // If field is a tuple, recursively extract its inner arguments and format as a tuple string.
                    if (f.FieldType.IsTuple())
                    {
                        var nestedArgs = f.GetValue(args);
                        return (nestedArgs != null) ? f.FieldType.ArgsToString(nestedArgs) : null;
                    }
                    // Add field as an argument if it is not a Qubit type
                    else if (!f.FieldType.IsQubitsContainer())
                    {
                        return f.GetValue(args)?.ToString();
                    }

                    return null;
```

src/Simulation/Core/Udts/UDTPartial.cs
```diff
@@ -62,6 +64,9 @@ public ICallable Partial(object partialTuple)
             return (ICallable)Activator.CreateInstance(partialType, new object[] { this, partialTuple });
         }
 
+        /// <inheritdoc/>
+        public RuntimeMetadata? GetRuntimeMetadata(IApplyData args) => throw new NotImplementedException();
```
cgranade, line 68: In all other cases, you've returned null when runtime metadata is unavailable; given that, is throwing an exception expected behavior? Will this cause any breakages, for instance, when considering snippets like `Mapped(Complex(_, 0.0), realParts))`?

src/Simulation/QsharpCore/Intrinsic.cs
```diff
@@ -0,0 +1,93 @@
...
+    {
+        /// <inheritdoc/>
+        public override RuntimeMetadata? GetRuntimeMetadata(IApplyData args) => null;
+    }
+
+    public partial class ResetAll
+    {
+        /// <inheritdoc/>
+        public override RuntimeMetadata? GetRuntimeMetadata(IApplyData args) => null;
+    }
```
cgranade, line 92: Is there a reason these two don't have associated metadata?

src/Simulation/Core/Operations/Adjoint.cs
```diff
@@ -72,6 +74,15 @@ public AdjointedOperation(Operation<I, O> op) : base(op.Factory)
 
         public override IApplyData __dataOut(QVoid data) => data;
 
+        /// <inheritdoc/>
+        public override RuntimeMetadata? GetRuntimeMetadata(IApplyData args)
+        {
+            var baseMetadata = this.BaseOp.GetRuntimeMetadata(args);
+            if (baseMetadata == null) return null;
+            baseMetadata.IsAdjoint = true;
```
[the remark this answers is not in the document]
anpaz-msft: I think Chris has a point. Maybe the right thing is to make it `baseMetadata.IsAdjoint = not baseMetadata.IsAdjoint`?

src/Simulation/Core/Operations/Controlled.cs
```diff
@@ -110,6 +113,25 @@ public ControlledOperation(Operation<I, O> op) : base(op.Factory)
...
+                var (controls, baseArgs) = ctrlArgs;
+                var baseMetadata = this.BaseOp.GetRuntimeMetadata(this.BaseOp.__dataIn(baseArgs));
+                if (baseMetadata == null) return null;
+                baseMetadata.IsControlled = true;
+                baseMetadata.Controls = controls.Concat(baseMetadata.Controls);
+                return baseMetadata;
+            }
+            else
+            {
+                Console.WriteLine($"Failed to retrieve runtime metadata for {this.ToString()}.");
```
anpaz-msft, line 130: This should really be a `Logger`.
Maybe instead, make a `Debug.Assert` for the condition?

src/Simulation/Core/Udts/UDTPartial.cs
anpaz-msft: I agree, just return `null` instead?

src/Simulation/Core/TypeExtensions.cs
```diff
@@ -144,6 +146,38 @@ public static Type[] GetTupleFieldTypes(this Type arg)
             }
         }
 
+        /// <summary>
+        /// Given a <see cref="Type"/> and its arguments, format its non-qubit arguments into a string.
+        /// Returns null if no arguments found.
+        /// </summary>
+        public static string? ArgsToString(this Type t, object args)
```
anpaz-msft, line 153: wondering if this can/should be incorporated to enhance the debugger proxy.

src/Simulation/Core/TypeExtensions.cs
```diff
@@ -144,6 +146,38 @@ public static Type[] GetTupleFieldTypes(this Type arg)
...
+        {
+            var argsStrings = t.GetFields()
+                .Select(f =>
+                {
+                    var argString = null as string;
+
+                    // If field is a tuple, recursively extract its inner arguments and format as a tuple string.
+                    if (f.FieldType.IsTuple())
```
anpaz-msft, line 161: what about arrays? does this work well with them?

src/Simulation/Simulators.Tests/RuntimeMetadataTests.cs
```diff
@@ -0,0 +1,180 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
anpaz-msft, line 1: Please add more tests for the Partial, Adjoint+Adjoint, Controlled, and user-defined operations.

src/Simulation/Core/Operations/Adjoint.cs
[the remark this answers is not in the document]
theRoughCode: Oh interesting, I guess the correct way to be to flip the value of `IsAdjoint`?

src/Simulation/Core/Udts/UDTPartial.cs
theRoughCode: I guess I wasn't sure what this was used for. I implemented partial operations in src/Simulation/Core/Operations/OperationPartial.cs. Is this similar (i.e. do I reuse that logic) or just return `null`?

src/Simulation/QsharpCore/Intrinsic.cs
theRoughCode: These operations aren't used in the visualization, so I return null to indicate that I don't intend on processing these operations. However, if there is another use for `RuntimeMetadata` in the future, it might make more sense to include a `Display=false` attribute or something of that flavour.

cgranade: I'd suggest that including resets in visualization would make a lot of sense; resetting a qubit or not can significantly change the meaning of a quantum program.

src/Simulation/Core/Operations/Controlled.cs
theRoughCode: Would that look like a `Debug.Assert(false, ...)` or more like `Debug.Assert(args.Value is ValueTuple<IQArray<Qubit>, I>)`? I tried the latter but it got pretty ugly with handling `null`s after the assert statement (i.e. have to handle the case when `args.Value` is not of the required type even though the `assert` statement should've caught it).

src/Simulation/Core/TypeExtensions.cs
theRoughCode: What ideas do you have regarding this?

src/Simulation/QsharpCore/Intrinsic.cs
theRoughCode: I see, would it make sense to implement this in a separate PR? cc-ing @anpaz-msft.

src/Simulation/Core/TypeExtensions.cs
theRoughCode: Good catch, I don't think so. I'll add tests for this function to make sure it works with arrays.

src/Simulation/Core/TypeExtensions.cs
```diff
@@ -144,6 +147,55 @@ public static Type[] GetTupleFieldTypes(this Type arg)
             }
         }
 
+        /// <summary>
+        /// Given an <see cref="object"/>, retrieve its non-qubit fields as a string.
+        /// Returns null if no non-qubit fields found.
+        /// </summary>
+        public static string? GetNonQubitArguments(this object o)
```
theRoughCode, line 154: Renamed this from `ArgsToString` to more correctly describe what this method does.

src/Simulation/QsharpCore/Intrinsic.cs
theRoughCode: We decided to remove this override and render it as a default gate visualization.

src/Simulation/Core/RuntimeMetadata.cs
```diff
@@ -56,5 +63,69 @@ public class RuntimeMetadata
...
+            
+            return this.Label == other.Label && this.FormattedNonQubitArgs == other.FormattedNonQubitArgs &&
+                this.IsAdjoint == other.IsAdjoint && this.IsControlled == other.IsControlled &&
+                this.IsMeasurement == other.IsMeasurement && this.IsComposite == other.IsComposite;
+        }
+
+        public override int GetHashCode()
+        {
+            // Stringify qubits, concatenate as string, and get resulting hashcode
+            var controlsHash = string.Join(",", this.Controls.Select(q => q.ToString())).GetHashCode();
```
anpaz-msft, line 106:
```suggestion
            var controlsHash = string.Join(",", this.Controls.Select(q => q?.ToString() ?? "")).GetHashCode();
```

src/Simulation/Core/RuntimeMetadata.cs
```diff
@@ -56,5 +63,69 @@ public class RuntimeMetadata
...
+            // If both children are not null, compare each child element-wise and return
+            // false if any of them are not equal
+            if (!IsBothNull(this.Children, other.Children))
+            {
+                if (this.Children.Count() != other.Children.Count() ||
+                    this.Children.Zip(other.Children, ListEquals<RuntimeMetadata>).Contains(false))
+                    return false;
+            }
+            
+            return this.Label == other.Label && this.FormattedNonQubitArgs == other.FormattedNonQubitArgs &&
```
anpaz-msft, line 98: I would check this first, so you can quickly return `False` if any of these trivial properties fail to be equal. 
Then check the most complex properties one by one.

src/Simulation/Core/RuntimeMetadata.cs
```diff
@@ -37,10 +39,15 @@ public class RuntimeMetadata
         /// <summary>
         /// True if operation is composed of multiple operations.
         /// </summary>
+        /// </summary>
+        /// <remarks>
+        /// Currently not used as this is intended for compositeoperations,
```
anpaz-msft, line 44: What does this even mean in the base class?

src/Simulation/Core/RuntimeMetadata.cs
```diff
@@ -0,0 +1,131 @@
...
+
+#nullable enable
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.Quantum.Simulation.Core
+{
+    public class RuntimeMetadata
```
anpaz-msft, line 12: You should add a description of what the RuntimeMetadata represents.

src/Simulation/Simulators.Tests/RuntimeMetadataTests.cs
```diff
@@ -0,0 +1,671 @@
...
+                IsAdjoint = false,
+                IsControlled = true,
+                IsMeasurement = false,
+                IsComposite = false,
+                Children = null,
+                Controls = new List<Qubit>() { control1, control2 },
+                Targets = new List<Qubit>() { target },
+            };
+
+            Assert.True(op.GetRuntimeMetadata(args) == expected);
```
anpaz-msft, line 215: Use `Assert.Equals` instead

src/Simulation/Simulators.Tests/RuntimeMetadataTests.cs
```diff
@@ -0,0 +1,671 @@
...
+                IsAdjoint = false,
+                IsControlled = false,
+                IsMeasurement = false,
+                IsComposite = false,
+                Children = null,
+                Controls = new List<Qubit>() { },
+                Targets = new List<Qubit>() { target },
+            };
+
+            Assert.True(op.GetRuntimeMetadata(args) == expected);
```
anpaz-msft, line 237: Use `Assert.Equals` instead

src/Simulation/Simulators.Tests/RuntimeMetadataTests.cs
```diff
@@ -0,0 +1,671 @@
...
+        }
+
+        [Fact]
+        public void ControlledUDT()
+        {
+            IQArray<Qubit> controls = new QArray<Qubit>(new[] { new FreeQubit(0) });
+            Qubit target = new FreeQubit(1);
+            var op = new QuantumSimulator().Get<Tests.Circuits.FooUDTOp>().Controlled;
```
anpaz-msft, line 460: although possible in C#, this (a controlled UDT) is not a valid Q# scenario. We should probably delete this test.

src/Simulation/Simulators.Tests/RuntimeMetadataTests.cs
```diff
@@ -0,0 +1,671 @@
...
+                Children = null,
+                Controls = new List<Qubit>() { },
+                Targets = new List<Qubit>() { target },
+            };
+
+            Assert.True(op.GetRuntimeMetadata(args) == expected);
+        }
+
+        [Fact]
+        public void AdjointUDT()
```
anpaz-msft, line 526: Similarly, an "adjoint" UDT makes no sense. We should remove this test.

src/Simulation/Core/RuntimeMetadata.cs
theRoughCode: Not sure I understand the question. Do you mean the default case? In that case this would be `false`.

theRoughCode:
There was a weird Github bug where my changes weren't showing after merging master in. I tried rebasing onto master instead but again, the changes weren't showing up on this PR specifically. Turns out this is a [known bug for years](https://github.com/isaacs/github/issues/750) and the solution involved having to switch the base branches.

PR #305, opened 2020-07-09
Include native files in Simulators project

anpaz-msft:
The native files for the simulator, were getting included in the nuget package via the nuspec and not the csproj, which breaks when we want to replace package with project references for the simulators.
This makes sure the native files are also part of the project.


src/Simulation/Simulators/Microsoft.Quantum.Simulators.nuspec.template
```diff
@@ -17,11 +17,9 @@
   </metadata>
 
   <files>
-    <file src="..\Native\osx\*" target="runtimes\osx-x64\native" exclude="**\*.pdb" />
-    <file src="..\Native\linux\*" target="runtimes\linux-x64\native" exclude="**\*.pdb" />
-    <file src="..\Native\win10\*" target="runtimes\win-x64\native" exclude="**\*.pdb" />
-    <file src="..\Native\build\$Configuration$\Microsoft.Quantum.*" target="runtimes\win-x64\native" exclude="**\*.pdb" />
-    <file src="..\Native\build\$Configuration$\Microsoft.Quantum.*.pdb" target="runtimes\win-x64\native" />
+    <file src="..\Native\osx\*" target="runtimes\osx-x64\native" />
```
swernli, line 20: This looks like we are now including PDBs where before we weren't. Since these are all dlls compiled from within this repo, is the idea that the PDB doesn't represent anything sensitive?

anpaz-msft: Exactly. There is nothing sensitive since it's all open source, and having the PDBs greatly helps with debugging.

Issue #299, opened 2020-07-07
C# compilation error for invoking generic callables


ScottCarda-MS:
When executing this program:
```
namespace test {
    @EntryPoint()
    operation Main() : Result {
        Generic<Double, Int, Int>();
        return Zero;
    }
    operation Generic<'A, 'B, 'C>() : Unit { }
}
```
The following error occurs:
`error CS0305: Using the generic type 'Generic<__A__, __B__, __C__>' requires 3 type arguments`

When inspecting the generated C# for this file, under the `Main` partial class, in the `Init` method, there is the following line:
`this.Generic = this.Factory.Get<ICallable<QVoid, QVoid>>(typeof(Generic));`

This line uses the `Generic` type without any of the type arguments, which is causing the error. If the line is changed to
`this.Generic = this.Factory.Get<ICallable<QVoid, QVoid>>(typeof(Generic<Double, Int32, Int32>));`
then the error goes away.

ScottCarda-MS:
Fixed with PR #306.

PR #306, opened 2020-07-09
Generic Parameter Bug Fix

ScottCarda-MS:
Changes the method in the C# generation for determining if a callable is a generic callable. The old behavior counts the number of type parameters referenced in the types of the regular parameter list and the return type. This causes a bug to occur when type parameters are listed for a callable that are not used in their parameter list or return type. The new behavior simply counts the length of the type parameters list for the callable. See issue #299.

### DIFF
--- a/src/Simulation/CsharpGeneration/SimulationCode.fs
+++ b/src/Simulation/CsharpGeneration/SimulationCode.fs
@@ -93,9 +93,7 @@ module SimulationCode =
     let isGeneric context (n: QsQualifiedName) =
         if context.allCallables.ContainsKey n then
             let signature = context.allCallables.[n].Signature
-            let tIn = signature.ArgumentType
-            let tOut = signature.ReturnType
-            hasTypeParameters [tIn;tOut]
+            not signature.TypeParameters.IsEmpty
         else
             false
 
@@ -861,9 +859,7 @@ module SimulationCode =
             let opName = if sameNamespace then n.Name.Value else n.Namespace.Value + "." + n.Name.Value
             if isGeneric context n then
                 let signature = context.allCallables.[n].Signature
-                let tIn = signature.ArgumentType
-                let tOut = signature.ReturnType
-                let count = (getTypeParameters [tIn;tOut]).Length
+                let count = signature.TypeParameters.Length
                 sprintf "%s<%s>" opName (String.replicate (count - 1) ",")
             else
                 opName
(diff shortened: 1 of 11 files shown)


bettinaheim:
Please wait with merging for @anpaz-msft to take a look at this one, since I believe there might be some perf considerations here.

src/Simulation/CsharpGeneration/SimulationCode.fs
```diff
@@ -93,9 +93,7 @@ module SimulationCode =
     let isGeneric context (n: QsQualifiedName) =
         if context.allCallables.ContainsKey n then
             let signature = context.allCallables.[n].Signature
-            let tIn = signature.ArgumentType
-            let tOut = signature.ReturnType
-            hasTypeParameters [tIn;tOut]
```
anpaz-msft, line 98: do we still need the `hasTypeParameters` method?

src/Simulation/CsharpGeneration/SimulationCode.fs
```diff
@@ -93,9 +93,7 @@ module SimulationCode =
...
+            not signature.TypeParameters.IsEmpty
```
anpaz-msft, line 96: I suppose `signature.TypeParameters` includes both, `tIn` and `tOut`?

src/Simulation/CsharpGeneration/SimulationCode.fs
ScottCarda-MS: The hasTypeParameters method is still used in some other parts of the code in this file, which doesn't seem to be related to the bug.

ScottCarda-MS: It can still be useful for determining if a given list of types contains references to type parameters.

src/Simulation/CsharpGeneration/SimulationCode.fs
ScottCarda-MS: The `signature.TypeParameters` is different from the ArgumentType or the ReturnType. Instead of counting the number of unique type parameter references in the ArgumentType and ReturnType of the callable to determine if it is generic, the new way I am presenting is to count the number of type parameters listed as part of the signature, which seems more straightforward to me as a means of checking if something is generic. This also avoids the underlying problem of the bug, which was when type parameters of the callable are not found in the argument type or return type, they would not be counted and the generated class would not be correct, causing a build-time error for the Q# code.

PR #310, opened 2020-07-14
Update to C++17

swernli:
This updates the CMake build configuration for the native simulator to use and require C++17. This will allow us to support C++17 specific features, such as `std::optional` and `std::any`.


src/Simulation/Native/src/SafeInt.hpp
```diff
@@ -72,8 +77,8 @@ Please read the leading comments before using the class.
 #define CPLUSPLUS_STD CPLUSPLUS_98
 #endif // Determine C++ support level
 
-#if (SAFEINT_COMPILER == CLANG_COMPILER || SAFEINT_COMPILER == GCC_COMPILER) && CPLUSPLUS_STD < CPLUSPLUS_11
-#error Must compile with --std=c++11, preferably --std=c++14 to use constexpr improvements
+#if (SAFEINT_COMPILER == CLANG_COMPILER || SAFEINT_COMPILER == GCC_COMPILER) && CPLUSPLUS_STD < CPLUSPLUS_17
+#error Must compile with --std=c++17
```
swernli, line 81: Note that this changes the build requirement for clang and gcc to C++17, where previously this file would have accepted and attempted to handle anything down to C++11. That said, the CMakeLists.txt for all the code under src\Native already had C++14 as the required standard, so I don't believe our support of C++11 was genuine here. Given that after this switch we will start incorporating C++17 only features, it felt appropriate to update this requirement too.

swernli:
@dbwz8 Yup, I tested these changes over here: https://github.com/microsoft/qsharp-runtime/pull/309. That included an actual usage of `std::optional` so that I could be sure that the Linux and Mac pipelines built as expected, and manually inspected the pipeline logs to double check.

PR #313, opened 2020-07-17
Use new position type in C# generation

samarsha:
This is the sister PR for microsoft/qsharp-compiler#523, updating the position type in the Q# syntax tree from `int * int` to `Position`.

### DIFF
--- a/src/Simulation/CsharpGeneration/SimulationCode.fs
+++ b/src/Simulation/CsharpGeneration/SimulationCode.fs
@@ -274,7 +274,7 @@ module SimulationCode =
         override this.OnStatement (node:QsStatement) =
             match node.Location with
             | Value loc ->
-                let (current, _) = loc.Offset
+                let current = loc.Offset.Line
                 parent.LineNumber <- parent.StartLine |> Option.map (fun start -> start + current + 1) // The Q# compiler reports 0-based line numbers.
             | Null ->
                 parent.LineNumber <- None // auto-generated statement; the line number will be set to the specialization declaration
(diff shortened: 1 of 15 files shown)


src/Simulation/CsharpGeneration/SimulationCode.fs
```diff
@@ -274,7 +274,7 @@ module SimulationCode =
         override this.OnStatement (node:QsStatement) =
             match node.Location with
             | Value loc ->
-                let (current, _) = loc.Offset
+                let current = loc.Offset.Line
                 parent.LineNumber <- parent.StartLine |> Option.map (fun start -> start + current + 1) // The Q# compiler reports 0-based line numbers.
```
ScottCarda-MS, line 278: Is the `+ 1` here still necessary?

src/Simulation/CsharpGeneration/SimulationCode.fs
```diff
@@ -1003,12 +1003,12 @@ module SimulationCode =
             | Null -> []
             | Value location -> [
                 // since the line numbers throughout the generated code are 1-based, let's also choose them 1-based here
```
ScottCarda-MS, line 1005: Why are line numbers in the generated code 1-based? And if they are, should we adopt the a similar strategy to what you did in the compiler, to use a zero-based system throughout the logic and convert at the very end?

src/Simulation/CsharpGeneration/SimulationCode.fs
samarsha: Yes, I think so, since `LineNumber` is one-based.

src/Simulation/CsharpGeneration/SimulationCode.fs
samarsha: I think it's referring to the [`#line`](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives/preprocessor-line) markers added to the generated code.

The `LineNumber` property that we define should probably be zero-based, and we can add 1 when the `#line` marker is actually generated, but it looks like the code is doing something weird where when `LineNumber` is `Some 0`, `#line hidden` is generated instead of an actual line number.

I'm a little afraid to change that and would prefer to keep the main refactoring on the compiler repo - the changes here were just the bare minimum needed to adapt to the compiler changes.

PR #314, opened 2020-07-20
Allow operations in arguments for `GetNonQubitArgumentsAsString`

theRoughCode:
Currently, an `ICallable` in would match the `IApplyData` case in `GetNonQubitArgumentsAsString`. This would throw a `NullException` because its `IApplyData.Value` is `null`. This PR adds a new check for `ICallable`s and returns its name as a string.

### DIFF
--- a/src/Simulation/Simulators.Tests/TypeExtensionsTest.cs
+++ b/src/Simulation/Simulators.Tests/TypeExtensionsTest.cs
@@ -34,6 +34,16 @@ namespace Microsoft.Quantum.Simulation.Simulators.Tests
         }
 
         [Fact]
+        public void OperationTypes()
+        {
+            var op = new QuantumSimulator().Get<Intrinsic.H>();
+            Assert.Equal("H", op.GetNonQubitArgumentsAsString());
+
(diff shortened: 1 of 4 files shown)


src/Simulation/Simulators.Tests/TypeExtensionsTest.cs
```diff
@@ -33,6 +33,16 @@ public void BasicTypes()
...
         }
 
+        [Fact]
+        public void OperationTypes()
+        {
+            var op = new QuantumSimulator().Get<Intrinsic.H>();
+            Assert.Equal("H", op.GetNonQubitArgumentsAsString());
+
+            var op2 = new QuantumSimulator().Get<Intrinsic.CNOT>();
+            Assert.Equal("CNOT", op2.GetNonQubitArgumentsAsString());
```
anpaz-msft, line 43: shouldn't this be "X"?

theRoughCode: It's a CNOT because that's the name of the operation being passed in, not the `RuntimeMetadata`.

PR #320, opened 2020-08-03
EncourageReuse option in QubitManager.

avasch01:
Currently Qubit Manager tries to reuse released qubits as soon as possible and prefers it to allocating fresh unused ones. This is good for some scenarios, however, in other scenarios it may be preferrable to first allocate fresh unused qubits, and only then try to reuse released ones (oldest released first). Both behaviors are now possible via a backwards-compatible optional parameter in QubitManager constructor.

### DIFF
--- a/src/Simulation/Common/QubitManager.cs
+++ b/src/Simulation/Common/QubitManager.cs
@@ -52,7 +52,11 @@ namespace Microsoft.Quantum.Simulation.Common
         /// <summary>
         /// Creates and initializes QubitManager that can handle up to numQubits qubits
         /// </summary>
-        public QubitManager(long qubitCapacity, bool mayExtendCapacity = false, bool disableBorrowing = false, bool encourageReuse = true)
+        public QubitManager(
+            long qubitCapacity, 
+            bool mayExtendCapacity = false, 
+            bool disableBorrowing = false, 
+            bool encourageReuse = true)
         {
             MayExtendCapacity = mayExtendCapacity;
             EncourageReuse = encourageReuse;
@@ -312,11 +316,9 @@ namespace Microsoft.Quantum.Simulation.Common
                 } 
                 else
                 {
-                    if (MayExtendCapacity)
-                    {
-
-                    }
-                    else
+                    // If we are allowed to extend capacity we will never reuse this qubit, 
+                    // otherwise we need to add it to the free qubits list.
+                    if (!MayExtendCapacity)
                     {
                         if (qubits[freeTail] != None)
                         {
--- a/src/Simulation/Common/QubitManager.cs
+++ b/src/Simulation/Common/QubitManager.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
+using Microsoft.Quantum.Intrinsic;
 using Microsoft.Quantum.Simulation.Core;
 using Microsoft.Quantum.Simulation.Simulators.Exceptions;
 
@@ -25,11 +26,13 @@ namespace Microsoft.Quantum.Simulation.Common
         long AllocatedForBorrowing; // All qubits allocated only for borrowing, will be marked with this number or higher.
         long[] qubits; // Tracks the allocation state of all qubits.
         long free; // Points to the first free (unallocated) qubit.
+        long freeTail; // Points to the last free (unallocated) qubit. Only valid iff (!EncourageReuse).
         long numAllocatedQubits; // Tracking this for optimization.
         long numDisabledQubits; // Number of disabled qubits.
 
         // Options
         bool MayExtendCapacity;
+        bool EncourageReuse;
         public bool DisableBorrowing { get; }
 
         const long MaxQubitCapacity = long.MaxValue - 3;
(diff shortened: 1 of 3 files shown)


src/Simulation/Simulators.Tests/QubitManagerTests.cs
```diff
@@ -152,6 +152,91 @@ public void TestQubitManager()
             }
         }
 
+        /// <summary>
+        /// Test for QubitManager.
+        /// </summary>
+        [Fact]
+        public void TestQubitManagerDiscouragingReuse()
```
irinayat-MS, line 159: Split into two tests so it will be easier to localize test failures?

src/Simulation/Common/QubitManagerTrackingScope.cs
```diff
@@ -44,8 +44,8 @@ public List<Qubit> Locals
         private Stack<StackFrame> operationStack; // Stack of operation calls.
         private StackFrame curFrame; // Current stack frame - all qubits in current scope are listed here.
 
-        public QubitManagerTrackingScope(long qubitCapacity, bool mayExtendCapacity = false, bool disableBorrowing = false) 
-            : base(qubitCapacity, mayExtendCapacity, disableBorrowing)
+        public QubitManagerTrackingScope(long qubitCapacity, bool mayExtendCapacity = false, bool disableBorrowing = false, bool encourageReuse = true)
```
irinayat-MS, line 47: nit: as far as I know we don't have rules about page width in C# projects, but it's really inconvenient to work with or review wide code on a laptop... Any objections to staying within 120 chars width?

src/Simulation/Common/QubitManager.cs
```diff
@@ -65,6 +69,7 @@ public QubitManager(long qubitCapacity, bool mayExtendCapacity = false, bool dis
             Debug.Assert(this.qubits[NumQubits - 1] == None);
 
             free = 0;
+            freeTail = NumQubits - 1;
```
irinayat-MS, line 72: Given it's only valid for EncourageReuse=false, would it make sense to set it to -1 (or some other 'invalid' sentinel value) when EncourageReuse=true?

src/Simulation/Common/QubitManager.cs
```diff
@@ -25,11 +26,13 @@ public class QubitManager
         long AllocatedForBorrowing; // All qubits allocated only for borrowing, will be marked with this number or higher.
         long[] qubits; // Tracks the allocation state of all qubits.
         long free; // Points to the first free (unallocated) qubit.
+        long freeTail; // Points to the last free (unallocated) qubit. Only valid iff (!EncourageReuse).
```
irinayat-MS, line 29: For my education: why do we need dynamic Allocated/Disabled markers rather than using fixed sentinel values (to avoid remarking the qubits when the manager is extended)?

src/Simulation/Common/QubitManager.cs
```diff
@@ -300,8 +306,31 @@ protected virtual void ReleaseOneQubit(Qubit qubit, bool usedOnlyForBorrowing)
...
-                free = qubit.Id;
+                if (EncourageReuse) { 
+                    qubits[qubit.Id] = free;
+                    free = qubit.Id;
+                } 
+                else
+                {
+                    if (MayExtendCapacity)
+                    {
+
```
irinayat-MS, line 317: Add a comment _why_ in this case there is nothing to do?

src/Simulation/Common/QubitManager.cs
irinayat-MS: I'd also suggest renaming `free` to `freeHead` to match `freeTail` (and maybe expanding the comment, explaining that this is a singly-linked free list).

src/Simulation/Simulators.Tests/QubitManagerTests.cs
avasch01: Yes, the same allocation pattern is being tested, just in another function. I am trying to avoid having too many test functions. We already have about 300 and navigating that list is a concern. So I am trying to keep small related tests together. This way it is also easier to understand they are doing the same. See for yourself, you did notice the same pattern in two blocks because they were in the same function, but you didn't notice the same pattern in a function right above it.

src/Simulation/Common/QubitManagerTrackingScope.cs
avasch01: Sure. Done.

src/Simulation/Common/QubitManager.cs
avasch01: I don't see why. It is only used in discourage reuse cases, but it is valid always. It is OK either way but it will take more churn to maintain -1 value. Every time you set it you have to do an if? Not worth it.

src/Simulation/Common/QubitManager.cs
avasch01: Ok, makes sense.

src/Simulation/Simulators.Tests/QubitManagerTests.cs
irinayat-MS: I didn't look at the function above, because that code didn't change -- that's an inherent danger of code reviews. If I was working with code (to change it, etc.) I'd look through test cases' names and this is where having well named separate tests helps to build the mental picture of what is covered and what matters. The other benefit of smaller tests is that when a test fails you get a localized failure, identified by the test's name. Might not be a big deal on a local machine when you can rerun tests under debugger, but important for CIs.

src/Simulation/Common/QubitManager.cs
irinayat-MS: The comment for freeTail was saying that it's only valid when EncourageReuse=false. When a state of a class is allowed to be inconsistent, I like making it more explicit with "invalid" markers so can assert on them and spot problems under debugger easier. In this case, probably should just change to comment to say "only _used_ when EncourageReuse=false".

PR #324, opened 2020-08-06
Remove duplicate qubits from RuntimeMetadata.Targets

theRoughCode:
This PR fixes a bug where there can be duplicate qubits in `RuntimeMetadata.Targets` if the same qubit is passed in to an operation multiple times.

### DIFF
--- a/src/Simulation/Core/Operations/Operation.cs
+++ b/src/Simulation/Core/Operations/Operation.cs
@@ -16,7 +16,7 @@ namespace Microsoft.Quantum.Simulation.Core
     {
         O Apply(I args);
 
-        ICallable<P,O> Partial<P>(Func<P, I> mapper);
+        ICallable<P, O> Partial<P>(Func<P, I> mapper);
     }
 
     /// <summary>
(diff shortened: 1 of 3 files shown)


src/Simulation/Core/Operations/Operation.cs
```diff
@@ -78,13 +78,18 @@ public Operation(IOperationFactory m) : base(m)
...
+        {
+            var targets = args.GetQubits() ?? new List<Qubit>();
+            // Remove duplicate qubits
+            targets = new HashSet<Qubit>(targets).ToList();
+            return new RuntimeMetadata()
             {
                 Label = ((ICallable)this).Name,
                 FormattedNonQubitArgs = args.GetNonQubitArgumentsAsString() ?? "",
-                Targets = args.GetQubits() ?? new List<Qubit>(),
+                Targets = targets,
```
cgranade, line 90: You may be able to use the `Enumerable.Distinct` extension method here instead of manually constructing a `HashSet`.

theRoughCode: Oh nice! I'll give it a shot, thanks!

PR #325, opened 2020-08-06
Implement GetRuntimeMetadata for ApplyIfElse variants

theRoughCode:
This PR is a first step at implementing classical operations as part of the goal of visualizing Q# programs (more details [here](https://github.com/microsoft/iqsharp/issues/158)).

In this PR, we:

1. Create a new `IsConditional` flag that is true if the operation is a conditional operation.
2. Create a new `ResultMeasured` class derived from the `Result` class that contains the measured qubit.
3. Implements the `GetRuntimeMetadata` methods for `ApplyIfElseR`, `ApplyIfElseRC`, `ApplyIfElseRA`, and `ApplyIfElseRCA`.

Once end-to-end testing of this flow (classically-controlled operations) works, we can duplicate this functionality for other operations such as `ApplyIfOne` and `ApplyIfZero`.

NOTE: This should be merged in after #324 as this is branched off of that PR.


src/Simulation/QsharpCore/ClassicalControl.cs
```diff
@@ -0,0 +1,136 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using Microsoft.Quantum.Simulation.Core;
+
+namespace Microsoft.Quantum.Simulation.QuantumProcessor.Extensions
+{
+    public partial class ApplyIfElseR<__T__, __U__>
+    {
```
theRoughCode, line 14: The code within `GetRuntimeMetadata` below is the same across all 4 `ApplyIfElseR` variants. Is there a way to avoid code duplication here? 

I tried a prototype with a new inherited `IApplyIf` class and having each variant inherit from it, but couldn't get it to work.

anpaz-msft: You can create a static method and have all of them call the same method.

src/Simulation/Core/RuntimeMetadata.cs
```diff
@@ -42,12 +42,16 @@ public class RuntimeMetadata
...
-        /// </summary>
         /// <remarks>
         /// This is used in composite operations, such as <c>ApplyToEach</c>.
         /// </remarks>
         public bool IsComposite { get; set; }
 
+        /// <summary>
+        /// True if operation is a classically-controlled conditional operation.
+        /// </summary>
+        public bool IsConditional { get; set; }
```
anpaz-msft, line 53: I'm not sure I like this name/description for this.

theRoughCode: Is the name not descriptive enough or is it the description that is not clear?
This flag is `true` when the operation is an If-Else statement, like `ApplyIfElseR` and will contain 2 `Children` (one for each measurement value).

bettinaheim:
@anpaz-msft Is this piece still needed? I thought we had merged everything related to visualization.

bettinaheim:
Closing this PR since I believe this is no longer needed.

Issue #132, opened 2020-01-22
Code generation failure when looping over array in an adjointable operation


cgranade:
## Describe the bug
The following operation fails to compile with C# errors:
```Q#
operation PrepareCatState(register : Qubit[]) : Unit is Adj + Ctl {
    H(register[0]);
    for (target in register[1...]) {
        CNOT(register[0], target);
    }
}
```
![image](https://user-images.githubusercontent.com/31516/72867931-5742a300-3c95-11ea-9711-c82865f82af0.png)

Modifying the operation to not be adjointable works, as does storing the array slice in a local variable before looping:
```Q#
operation PrepareCatState(register : Qubit[]) : Unit is Adj {
    H(register[0]);
    let targets = register[1...];
    for (target in targets) {
        CNOT(register[0], target);
    }
}
```

Using an array function also works around the issue:
```Q#
open Microsoft.Quantum.Arrays;
operation PrepareCatState(register : Qubit[]) : Unit is Adj {
    H(register[0]);
    for (target in Rest(register)) {
        CNOT(register[0], target);
    }
}
```

## To Reproduce
See reproduction at: https://mybinder.org/v2/gist/cgranade/471654e721ff80ca8de8c80a954f6d20/master

## System information
![image](https://user-images.githubusercontent.com/31516/72867909-46922d00-3c95-11ea-9e6f-a2e0ba58ea7a.png)

ScottCarda-MS:
Fixed with PR #330

PR #330, opened 2020-08-12
Slice Bug Fix

ScottCarda-MS:
Replaced the nullable member access with regular member access for Slice.


ScottCarda-MS:
> lgtm. what was this fixing?

Issue #132

PR #331, opened 2020-08-13
Use processor architecture to determine what code needs to be generated

bettinaheim:
This fixes an issue when targeting an executable that uses type parametrized callables from libraries. In that case, an error similar to that currently occurs: 

test.qs(35,181): error CS0234: The type or
namespace name '_5bdf0e77d6ac45daad30c7cc43f70e0b_ApplyToEach' does not exist
in the namespace 'Microsoft.Quantum.Canon' (are you missing an assembly reference?)
[/home/oneqbit/aq_tests/test_ionq/test.csproj]


swernli:
Hmm... looks like the linux build is getting stuck because of a duplicate file. Could this be a casing issue?

PR #332, opened 2020-08-13
Fix asserts in QubitManager

swernli:
Removing the asserts from `ExtendQubitArray` that can now be hit during testing due the new flag added in #320


swernli:
The build succeeded on rerun, but is just not correctly displaying here. You can find the results here:
https://dev.azure.com/ms-quantum-public/Microsoft%20Quantum%20(public)/_build/results?buildId=12900&view=results

PR #336, opened 2020-08-13
Protect simulator accessors via shared lock

### LINKED ISSUE #335: Simulator tests intermittently fail with Access Violation crashes
## Describe the bug
Our simulator tests in src\Simulation\Simulators.Tests sometimes crash with an Access Violation when calling into the native simulator code.

## To Reproduce
The crash is intermittent, but most often observed by `QSimMultiThreading` test, since that explicitly tests the multithreaded support.

## Expected behavior
Tests should execute successfully

## Actual behavior
Test host process crashes intermittently.

**Additional context**
Because `psis` is a global vector, it's shared across all the tests in the process, which xUnit will run in parallel. If a new test starts and goes to allocate a simulator and there are no empty slots in the vector, it triggers a call to `push_back`, and if another thread is accessing the vector at the same time it will get an access violation. `SetSeed` and the like follow the same pattern where it takes the simulator id as the first argument, then uses it like this: `psis[id]->seed(s)`
If the new size of the vector is greater than it's capacity, the whole internal storage is reallocated. From the push_back docs:

Data races
The container is modified.
If a reallocation happens, all contained elements are modified.
Otherwise, no existing element is accessed, and concurrently accessing or modifying them is safe.

swernli:
This fixes an intermittent crash by putting shared locks around the access of the `psis` vector, where create and destroy use an exclusive lock of the same mutex

Fixes #335

### DIFF
--- a/src/Simulation/Native/src/simulator/factory.cpp
+++ b/src/Simulation/Native/src/simulator/factory.cpp
@@ -86,7 +86,7 @@ namespace Microsoft
         _psis[id].reset();
       }
 
-      MICROSOFT_QUANTUM_DECL std::shared_ptr<SimulatorInterface> get(unsigned id)
+      MICROSOFT_QUANTUM_DECL std::shared_ptr<SimulatorInterface>& get(unsigned id)
       {
         std::shared_lock<std::shared_mutex> shared_lock(_mutex);
 
--- a/src/Simulation/Native/src/simulator/factory.cpp
+++ b/src/Simulation/Native/src/simulator/factory.cpp
@@ -5,6 +5,7 @@
 #include "config.hpp"
 #include "util/cpuid.hpp"
 #include <iostream>
+#include <shared_mutex>
 
 namespace Microsoft
 {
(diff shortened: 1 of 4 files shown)


swernli:
I've introduced a bug somewhere with how the new `get` method is working. Changing this PR to draft while I track it down.

src/Simulation/Native/src/simulator/factory.cpp
```diff
@@ -31,7 +32,8 @@ namespace Microsoft
   {
     namespace Simulator
     {
-      mutex_type _mutex;
```
swernli, line 34: @dbwz8 This code used to use the #define that resolves to `omp_mutex` when OpenMP support is enabled. However, this code doesn't interact with any other OpenMP code/classes, so I think it should be safe to use the std class instead, especially since we compile with C++17 across all platforms now. Please let me know if this assumption is safe, and if not we can look at other solutions!

dbwz8: Looks OK to me.

PR #339, opened 2020-08-14
Use culture-dependent string for decimals in RuntimeMetadataTests

samarsha:
`GetRuntimeMetadata` returns strings containing decimals in the format of the current culture, so `2.1` would become `"2.1"` in an English culture but `"2,1"` in a German culture, for example. However, the tests use a hard-coded string `"2.1"`, causing the tests to fail when the current culture uses a comma for decimals. While the tests pass on the build servers because they use an English culture, they can fail when run locally during development.

This PR might not be the best solution - maybe `GetRuntimeMetadata` should return a culture-invariant string, or all the unit tests should automatically run in the invariant culture regardless of local settings - but it does work around the immediate issue for me.


theRoughCode:
Oh that's interesting, I didn't know decimals are represented differently based on locale. This LGTM, thanks Sarah!

Issue #46, opened 2019-08-28
Code generation failure when mapping Q# namespaces to C# namespaces 


vadym-kl:
## Describe the bug
Consider two files `A.B.qs` and `B.qs` with namespaces `A.B` and `B` and operation `Foo` defined in `B.qs`. Fully qualified call `B.Foo()` in `A.B.qs` is valid Q# code that leads to invalid C# code, 
because in C# `B.Foo()` is equivalent to `A.B.Foo()` in this case.

## To Reproduce
First file:
```qsharp
namespace Buggy {

    operation Foo() : Unit is Adj {
    }
}
```
Second file:
```
namespace BuggyBuggy.Buggy {

    operation Foo2 () : Unit {
        Buggy.Foo();
    }   
}
```

## Expected behavior
Successful compilation of Q# program.

## Actual behavior
C# compilation error 
```
 error CS0234: The type or namespace name 'Foo' does not exist in the namespace 'BuggyBuggy.Buggy' (are you missing an assembly reference?)
```

## System information
- OS: Windows

**Additional context**
Not critical, the simple workaround is to avoid the situation by naming namespaces differently.

filipjelic:
@vadym-kl  May I take this one? It would be my first dev contribution to MS 😄

swernli:
Fixed by #346

PR #346, opened 2020-08-17
Fixing namespace references using "global::"

swernli:
Fixes Code generation failure when mapping Q# namespaces to C# namespaces  #46


swernli:
Me neither! Thanks @samarsha for pointing it out!

PR #349, opened 2020-08-19
Fix naming conflicts with generated C# members

### LINKED ISSUE #37: C# code generation generates invalid code when accessing UDT items named `Data`.
## Describe the bug
When a UDT has a named item whose name is `Data`, accessing that item causes C# code generation to display a warning:

```
Microsoft (R) Build Engine version 16.3.0-preview-19377-01+dd8019d9e for .NET Core
Copyright (C) Microsoft Corporation. All rights reserved.

  Restore completed in 23.26 ms for C:\Users\<redacted>\tmp\runtime-bug\runtime-bug.csproj.
  You are using a preview version of .NET Core. See: https://aka.ms/dotnet-core-preview
obj\qsharp\src\Operations.g.cs(27,31): warning CS0108: 'Foo.Data' hides inherited member 'QTuple<IQArray<long>>.Data'. Use the new keyword if hiding was intended. [C:\Users\<redacted>\tmp\runtime-bug\runtime-bug.csproj]
  runtime-bug -> C:\Users\<redacted>\tmp\runtime-bug\bin\Debug\netcoreapp2.1\runtime-bug.dll
```

If the UDT contains named items other than `Data` as well, then the code generation step fails entirely:

```
Microsoft (R) Build Engine version 16.3.0-preview-19377-01+dd8019d9e for .NET Core
Copyright (C) Microsoft Corporation. All rights reserved.

  Restore completed in 24.7 ms for C:\Users\<redacted>\tmp\runtime-bug\runtime-bug.csproj.
  You are using a preview version of .NET Core. See: https://aka.ms/dotnet-core-preview
obj\qsharp\src\Operations.g.cs(27,31): warning CS0108: 'Foo.Data' hides inherited member 'QTuple<(IQArray<long>, string)>.Data'. Use the new keyword if hiding was intended. [C:\Users\<redacted>\tmp\runtime-bug\runtime-bug.csproj]
obj\qsharp\src\Operations.g.cs(27,44): error CS1061: 'IQArray<long>' does not contain a 
definition for 'Item1' and no accessible extension method 'Item1' accepting a first argument of type 'IQArray<long>' could be found (are you missing a using directive or an assembly reference?) [C:\Users\<redacted>\tmp\runtime-bug\runtime-bug.csproj]        
obj\qsharp\src\Operations.g.cs(28,45): error CS1061: 'IQArray<long>' does not contain a 
definition for 'Item2' and no accessible extension method 'Item2' accepting a first argument of type 'IQArray<long>' could be found (are you missing a using directive or an assembly reference?) [C:\Users\<redacted>\tmp\runtime-bug\runtime-bug.csproj]        
obj\qsharp\src\Operations.g.cs(29,45): error CS1061: 'IQArray<long>' does not contain a 
definition for 'Item1' and no accessible extension method 'Item1' accepting a first argument of type 'IQArray<long>' could be found (are you missing a using directive or an assembly reference?) [C:\Users\<redacted>\tmp\runtime-bug\runtime-bug.csproj]        
obj\qsharp\src\Operations.g.cs(30,37): error CS1061: 'IQArray<long>' does not contain a 
definition for 'Item2' and no accessible extension method 'Item2' accepting a first argument of type 'IQArray<long>' could be found (are you missing a using directive or an assembly reference?) [C:\Users\<redacted>\tmp\runtime-bug\runtime-bug.csproj]        
obj\qsharp\src\Operations.g.cs(34,26): error CS1061: 'IQArray<long>' does not contain a 
definition for 'Item1' and no accessible extension method 'Item1' accepting a first argument of type 'IQArray<long>' could be found (are you missing a using directive or an assembly reference?) [C:\Users\<redacted>\tmp\runtime-bug\runtime-bug.csproj]        
obj\qsharp\src\Operations.g.cs(35,26): error CS1061: 'IQArray<long>' does not contain a 
definition for 'Item2' and no accessible extension method 'Item2' accepting a first argument of type 'IQArray<long>' could be found (are you missing a using directive or an assembly reference?) [C:\Users\<redacted>\tmp\runtime-bug\runtime-bug.csproj]  
```

In either case, when calling from IQ#, the kernel dies with no error message printed to the notebook or the console.

## To Reproduce
```Q#
open Microsoft.Quantum.Arrays;

newtype Foo = (Data: Int[]);

function Example() : Foo[] {
    return Mapped(Foo, [[1, 2], [3, 4, 5], new Int[0], [16]]);
}

function Boom() : Int {
    let foos = Example();
    return foos[0]::Data[0];
}
```
Renaming `Data` → `NotData`, the example works as expected.

See https://gist.github.com/cgranade/ee913899e0b40e8ee8cd43b82632a454 for IQ# version.

## System information
- OS: Windows 10

- .NET Core Version: 3.0.100-preview8-013656

- IQ# Version:  0.8.1907.1701

**Additional context**

### LINKED ISSUE #50: C# code generation fails for UDTs with named items called `Item*`.
## Describe the bug
I will readily admit that no one should ever write this code, but if one defines a UDT with named items such as `Item1` and `Item2`, compiling the generated C# code fails. In particular, compilation fails even if the affected UDT is never actually used.

## To Reproduce
```Q#
newtype Evil = (
    Item2: Int,
    Item1: String
);
```

image

### LINKED ISSUE #241: Some valid Q# callable names produce invalid C# code
The classes generated for Q# callables contains several built-in members with names that are not surrounded by `__`. This means they are valid Q# identifiers. If you declare a Q# callable with the same name as one of these members, the C# code does not compile because the member name conflicts with the constructor name.

Here is an example:

```
function Init() : Unit {
    Message("Hello quantum world!");
}
```

produces this C# compile error:

> Program.qs(22,30): error CS0542: '"Init": member names cannot be the same as their enclosing type.

Here is a list of names I tried that are valid Q# but produce invalid C#:

* `Body`
* `AdjointBody` if the operation is adjointable.
* `ControlledBody` if the operation is controllable.
* `ControlledAdjointBody` if the operation is controllable and adjointable.
* `Info`
* `Init`
* `Run`
* `__dataIn`
* `__dataOut`
* The fully qualified name of any callable used, with dots removed, such as `MicrosoftQuantumIntrinsicMessage` for the example above.

### LINKED ISSUE #243: Callables with the same fully-qualified name modulo dots cannot be called together
Consider the following valid Q# program:

```
namespace Program {
    @EntryPoint()
    function Main() : Unit {
        FooBar.Baz();
        Foo.Bar.Baz();
    }
}

namespace FooBar {
    function Baz() : Unit { }
}

namespace Foo.Bar {
    function Baz() : Unit { }
}
```

This produces a C# compile error:

> obj\qsharp\src\Program.g.cs(105,43): error CS0102: The type "Main" already contains a definition for "FooBarBaz".

The problem is that C# generation creates properties in the containing callable's generated class for both of them, but it strips out dots from their fully-qualified names, so they end up with the same name:

```
        protected ICallable<QVoid, QVoid> FooBarBaz
        {
            get;
            set;
        }

        protected ICallable<QVoid, QVoid> FooBarBaz
        {
            get;
            set;
        }
```

### LINKED ISSUE #247: In the generated code, local variable may shadow a field, resulting in incorrect code
Consider this program:

```
namespace Example {
    @EntryPoint()
    operation Test() : Unit {
        using (q = Qubit()) {
            let MicrosoftQuantumIntrinsicX = Z;
            X(q);
            // Zero or One, that is the question.
            Message($"Result: {M(q)}");
            Reset(q);
        }
    }
}
```

Because of the way the C# code is generated, it prints `Zero`, but it should print `One`.

samarsha:
This PR updates the naming scheme in the generated C# to avoid conflicts with Q# identifiers. It assumes the Q# compiler reserves `__` and `_.` anywhere in an identifier or namespace name, and `_` at the end of namespace names.

C# generation also has its own list of reserved words. If these words are used, there is no error, but the name has `__` appended to it to avoid a conflict:

* `Data`
* `Deconstruct`
* `Info`
* `Run`
* `ItemN` where `N` is a natural number
* The name of the enclosing operation, function, or UDT class

Any built-in member names that are defined in a generated class need to be:

1. In the reserved word list, or
2. Surrounded on both sides by `__`, or
3. Implemented via an explicit interface.

The protected properties in an operation class that are generated for referenced callables use this encoding:

* If the property name uses the fully-qualified name of the callable, `.` is replaced with `__`.
* Otherwise, `__` is appended to the end (but not the beginning, to avoid conflicts with built-in property names like `__Body__`).

TODO:

- [x] Update tests.
- [x] Add test cases for resolved issues.
- [x] Run an end-to-end build.
- [ ] Enable the `Data` test after it is removed from the compiler's reserved words.

Related issues:

* Resolves #37.
* Resolves #50.
* Resolves #241.
* Resolves #243.
* Resolves #247.


samarsha:
Also, tip for reviewers: this PR looks scarier than it is. :) The main changes are in:

* AbstractCallable.cs
* EntryPoint.fs
* Function.cs
* MemberNames.qs
* Operation.cs
* SimulationCode.fs

The rest I think should only be trivial renamings in response to these changes.

PR #355, opened 2020-08-22
Skip Csharp Generation of references with test names if not using test name

swernli:
This takes advantage of the functionality introduced in https://github.com/microsoft/qsharp-compiler/pull/581 that allows rewrite steps to know if `ExposeReferencesViaTestNames` was specified for the current compilation. Then CsharpGeneration rewrite step can only re-generate C# for references that both have a test name defined AND are being loaded by that test name. This fixes a bug where callables with a test name attribute would end up colliding with themselves if they were referenced without test name.

This was tested by temporarily adding a test name attribute to `Microsoft.Quantum.Intrinsic.Measure`, which would cause a compilation failure without these changes and compiles successfully with these changes.

### DIFF
--- a/src/Simulation/CsharpGeneration/SimulationCode.fs
+++ b/src/Simulation/CsharpGeneration/SimulationCode.fs
@@ -1624,8 +1624,8 @@ module SimulationCode =
             if globalContext.ExposeReferencesViaTestNames then
                 let asOption = function | Value _ -> Some nsElement | _ -> None
                 match nsElement with
-                | QsCallable c as e -> SymbolResolution.TryGetTestName c.Attributes
-                | QsCustomType t as e -> SymbolResolution.TryGetTestName t.Attributes
+                | QsCallable c -> SymbolResolution.TryGetTestName c.Attributes
+                | QsCustomType t -> SymbolResolution.TryGetTestName t.Attributes
                 |> asOption
             else None
         let context = {globalContext with fileName = Some dllName.Value}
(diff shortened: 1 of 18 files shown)


src/Simulation/CsharpGeneration/SimulationCode.fs
```diff
@@ -1619,13 +1619,15 @@ module SimulationCode =
...
-    let loadedViaTestNames (dllName : NonNullable<string>) globalContext =
+    let loadedViaTestNames (dllName : NonNullable<string>) (globalContext : CodegenContext) =
         let isLoadedViaTestName nsElement =
-            let asOption = function | Value _ -> Some nsElement | _ -> None
-            match nsElement with
-            | QsCallable c as e -> SymbolResolution.TryGetTestName c.Attributes
-            | QsCustomType t as e -> SymbolResolution.TryGetTestName t.Attributes
-            |> asOption
+            if globalContext.ExposeReferencesViaTestNames then
+                let asOption = function | Value _ -> Some nsElement | _ -> None
```
ScottCarda-MS, line 1625: What value does this function operate on? `function` should be short-hand for `match x with`, but I don't know what the "x" is in this case.

src/Simulation/CsharpGeneration/SimulationCode.fs
```diff
@@ -1619,13 +1619,15 @@ module SimulationCode =
...
+                match nsElement with
+                | QsCallable c as e -> SymbolResolution.TryGetTestName c.Attributes
+                | QsCustomType t as e -> SymbolResolution.TryGetTestName t.Attributes
```
ScottCarda-MS, line 1628:
```suggestion
                | QsCallable c -> SymbolResolution.TryGetTestName c.Attributes
                | QsCustomType t -> SymbolResolution.TryGetTestName t.Attributes
```
It doesn't look like the `e` is referenced anywhere anymore, so I don't think there is a need to keep it.

PR #358, opened 2020-08-25
Create QsharpCore targeting package

swernli:
This continues to refactor the existing core package to enable support for alternate targeting packages based on different sets of native capabilities.

Related to #249.


src/Simulation/Simulators.Tests/QuantumTestSuite/AssertProbMultiQubit.qs
```diff
@@ -8,6 +8,41 @@ namespace Microsoft.Quantum.Simulation.TestSuite {
     open Microsoft.Quantum.Simulation.TestSuite.Math;
     
     
+    internal operation flipToBasis (basis : Int[], qubits : Qubit[]) : Unit is Adj + Ctl {
```
samarsha, line 11: Is there a naming convention for internal operations, or should they use the same as public ones, e.g. `internal operation FlipToBasis`?

src/Simulation/Simulators.Tests/QuantumTestSuite/AssertProbMultiQubit.qs
```diff
@@ -8,6 +8,41 @@ namespace Microsoft.Quantum.Simulation.TestSuite {
...
+        if (Length(qubits) != Length(basis))
+        {
+            fail $"qubits and stateIds must have the same length";
```
samarsha, line 14:
```suggestion
            fail "qubits and stateIds must have the same length";
```

src/Simulation/TargetDefinitions/Intrinsic/Z.qs
```diff
@@ -0,0 +1,24 @@
...
+    ///         1 & 0 \\\\
+    ///         0 & -1
+    ///     \end{bmatrix}.
+    /// \end{align}
+    ///
+    /// # Input
+    /// ## qubit
+    /// Qubit to which the gate should be applied.
+    operation Z (qubit : Qubit) : Unit is Adj + Ctl {
+        body intrinsic;
```
samarsha, line 21: Does the TargetDefinitions folder contain the particular intrinsics and decompositions for the QuantumSimulator? Or is it more general than that?

src/Simulation/TargetDefinitions/Intrinsic/Z.qs
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
samarsha, line 1: I might have missed something earlier, but what is the reason to put each operation in its own file?

src/Simulation/Simulators.Tests/QuantumTestSuite/AssertProbMultiQubit.qs
swernli: I'm actually not sure. Previously this was `_flipToBasis` and that was acceptable, but it does make it seem like a local variable instead of a defined operation, so now that you've drawn my attention to it I'm inclined to change it!

src/Simulation/TargetDefinitions/Intrinsic/Z.qs
swernli: It's more general than that, and relates to the other question you had about separating callables into individual files. See my answer there.

src/Simulation/TargetDefinitions/Intrinsic/Z.qs
swernli: It's not obvious when there is only one target definition how this would be used, but you can check out a draft PR I have showing the ongoing work on the Type2 definition to see how this extends to multiple packages: #353. My intent with that draft is to close it and make a fresh one off of a cleaned up branch once this PR gets in. I'm still working on finding ways to reduce renames and duplications in order to make that PR more legible.

To describe the idea briefly, the set of intrinsics and decompositions for any and all target definitions should live in these folders. Having them in separate files then allows us to define a new target package via a list of `QsharpCompile` references to the specific subset of intrinsics and decompositions that make up that target definition (see [Type2.Package.props](https://github.com/microsoft/qsharp-runtime/pull/353/files#diff-233678157037e3e3be828fb427f29d49) as an example of this). This way we can easily reuse intrinsics or decompositions across targets, with the trade off being an explosion in the number of files. As you noted, this does make the first package look kinda silly here, but doing two at once would be way too many files and make the PR very hard to review.

samarsha: Thanks for the explanation, I think it answered my question of what happens if there are multiple decompositions for a given operation: it looks like there are multiple files that define the same operation, and targets only pick one of them.

PR #361, opened 2020-08-27
Adding scheduler enhancement for simulator performance boost

swernli:
This pulls in work done by @AniketDalvi, @dbwz8, and myself to enhance the native simulator with gate scheduling and clustering, leading to significant performance increases (originally part of the feature/simulator-enhancement branch). It also includes new tests that verify the performance, and a folder with benchmark projects to help with future performance comparisons. In our benchmarking, this led to between a 4x and 6x improvement, and could be significantly more depending on the host environment or circuit composition.


SupremacyBenchmark/privateBuild/Program.cs
```diff
@@ -0,0 +1,71 @@
+namespace quantum
```
dbwz8, line 1: We might want to put some explanitory comments in the test harness before it gets posted publicly,

SupremacyBenchmark/privateBuild/parseLog.py
```diff
@@ -0,0 +1,50 @@
+import re
```
dbwz8, line 1: Probably want a comment that this program will take the output of the tests and generate a summary for analysis as an excel spreadsheet

SupremacyBenchmark/privateBuild/runTest.csv
```diff
@@ -0,0 +1,252 @@
+CSV,test,loop,secs,gates,THREADS,FUSESPAN,FUSEDEPTH,Gates/sec
```
dbwz8, line 1: File can be removed

SupremacyBenchmark/privateBuild/runTest2.csv
```diff
@@ -0,0 +1,211 @@
+CSV,test,loop,secs,gates,THREADS,FUSESPAN,FUSEDEPTH,Gates/sec
```
dbwz8, line 1: Remove temp file

src/Simulation/Native/Makefile
```diff
@@ -0,0 +1,364 @@
+# CMAKE generated file: DO NOT EDIT!
```
dbwz8, line 1: Makefiles are generated by CMake... should be checked in

src/Simulation/Native/W15a_15.log
```diff
@@ -0,0 +1,95 @@
+@@@DBG: Generic
```
dbwz8, line 1: There are LOTS of temp files checked in. Let me know what branch to go to (when you're ready) and I'll prune them.

SupremacyBenchmark/privateBuild/parseLog.py
swernli: I added this as a ToDo on the PR to add a readme with guidance for using these benchmark projects and scripts.

src/Simulation/Native/Makefile
swernli: Is this still a concern? The file is present in the change, but perhaps you comment meant to say "shouldn't" be checked in? I could remove that file if that is preferable.

dbwz8: Yeah... shouldn't (sorry)

SupremacyBenchmark/privateBuild/parseLog.py
dbwz8: Good idea.

src/Simulation/Native/src/simulator/dbw_test.cpp
```diff
@@ -0,0 +1,231 @@
...
+
+    std::chrono::system_clock::time_point start = std::chrono::system_clock::now();
+    int itvl = loops / 10;
+    for (int i = 0; i < loops; i++) {
+        for (int j = 0; j < prb.size(); j++) {
+            auto qs = prb[j];
+            uint32_t cs[2];
+            switch (qs.size()) {
+            case 0: // Need to force a flush (end of cluster)
+                //Flush(sim_id);
```
swernli, line 202: @dbwz8 Why is this flush commented out? I noticed in the debug version of this code (now removed) the flush was not commented out, and the comment above this seems to indicate that the force flush is needed. Which is correct?

dbwz8: This is for testing without a scheduler (when clusters are pre-computed and loaded from a file). The real scheduler does the flush at the correct time (at the end of each cluster). This Flush is only necessary during DBWDBG (when we might be testing without the real scheduler). In fact, the Flush() call is only defined if DBWDBG is defined, so it had to be commented out. It can be removed in the "real" code.

swernli: Thanks for the explanation! I've removed the commented out lines.

PR #367, opened 2020-09-04
Add Type1 target package and simulator support

swernli:
This change adds the Type1 target package for Q# targets that support a limited number of control operations. The change includes support for simulating a Type1 package and tests that verify unitary gate behavior for that package. This part of the ongoing work described in #249.


swernli:
@bettinaheim @cgranade Something that is worth calling out in these changes is the introduction of `Microsoft.Quantum.Intrinsic.ApplyUncontrolledH` and similar operations for the native gates supported in Type1. Initially I'd hoped to have these marked as `internal` so that we wouldn't have to fully document them, but in order for a `body intrinsic` operation to be implemented outside of its own binary it cannot be marked as `internal`. Given that two of our goals are a) keeping the target packages as pure Q#, and b) allowing external authors to implement simulators in their own repos without always having to fork this one, there isn't a good way around the requirement that intrinsic callables be public. I did my best to document these appropriately so that we can have a chance to accept them as-is. But I'm certainly open to feedback and ideas regarding their identifiers, namespaces, etc.

By way of background, the need to introduce these was that the target gate set only supports singly controlled X and singly controlled Z, and no other controlled operations. So anywhere that our API has support for the Controlled specialization that accepts arbitrary number of controls, it had to be replaced with a decomposition on top of singly controlled variants.

src/Simulation/TargetDefinitions/Decompositions/Measure.qs
```diff
@@ -0,0 +1,71 @@
...
+            within {
+                MapPauli(qubits[0], PauliZ, bases[0]);
+            }
+            apply {
+                set res = M(qubits[0]);
+                PreparePostM(res, qubits[0]);
+            }
+        }
+        else {
+            using( q = Qubit() ) {
```
cgranade, line 57: This looks to use an additional spare qubit when measuring a multi-qubit Pauli operator; that's completely valid, but there's a tradeoff between that and finding the right Clifford to map back to a single-qubit measurement, then calling the measurement within that Clifford.

cgranade (commented):
Only taken a rough skim so far, but left what comments I have so far; I did have one quick question, though:

> Initially I'd hoped to have these marked as internal so that we wouldn't have to fully document them, but in order for a body intrinsic operation to be implemented outside of its own binary it cannot be marked as internal.

Is that a limitation on the Q# source, or the generated C# source? If the latter, is that the case even with the `InternalsVisibleTo` C# attribute?

src/Simulation/Simulators.Type1.Tests/OperationsTestHelperSimSupport.cs
```diff
@@ -0,0 +1,34 @@
...
+        {
+            var simulators = new SimulatorBase[] { new QuantumSimulator() };
+
+            foreach (var s in simulators)
+            {
+                InitSimulator(s);
+
+                test(s);
+
+                if (s is IDisposable sim)
```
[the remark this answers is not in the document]
swernli: I suppose it couldn't hurt... The dispose should eventually get called when the simulator is garbage collected anyway, but having it be in a finally is more direct and predictable behavior.

src/Simulation/Simulators.Type1/FindNuspecReferences.ps1
```diff
@@ -0,0 +1,103 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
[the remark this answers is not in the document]
swernli: I'll look through the scripts and make sure to clean these up.

src/Simulation/Simulators/QuantumSimulator/ApplyControlledZ.cs
```diff
@@ -0,0 +1,36 @@
...
+
+using System;
+using System.Runtime.InteropServices;
+using Microsoft.Quantum.Simulation.Core;
+
+namespace Microsoft.Quantum.Simulation.Simulators
+{
+    public partial class QuantumSimulator
+    {
+        public class QSimApplyControlledZ : Intrinsic.ApplyControlledZ
```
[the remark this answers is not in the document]
swernli: We are decomposing at the Q# level as much as possible right now, and that's part of what made these necessary. Because the `Controlled` specialization allows for any number of controls, I need to have a callable that is specifically the singly controlled gate and no `Controlled` specialization. Otherwise it won't match what the target architecture actually supports. It makes sense from a pure instruction translation standpoint, but it looks very odd and redundant when layered on top of our simulator that supports more than this.

src/Simulation/TargetDefinitions/Decompositions/ApplyControlledX.qs
```diff
@@ -0,0 +1,39 @@
...
+    /// ## target
+    /// Target qubit for the CNOT gate.
+    ///
+    /// # Remarks
+    /// Equivalent to:
+    /// ```qsharp
+    /// CNOT(control, target);
+    /// ```
+    @EnableTestingViaName("Test.TargetDefinitions.ApplyControlledX")
+    operation ApplyControlledX (control : Qubit, target : Qubit) : Unit is Adj {
```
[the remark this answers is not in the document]
swernli: Totally agreed. It's definitely something I was planning on calling out as part of the feature branch merge into master (at the latest). In total between both this Type1 work and the previous Type2 work, there are 16 new public callables in the Microsoft.Quantum.Intrinsic namespace: ApplyControlledX, ApplyControlledZ, ApplyUncontrolledH, ApplyUncontrolledRx, ApplyUncontrolledRy, ApplyUncontrolledRz, ApplyUncontrolledS, ApplyUncontrolledT, ApplyUncontrolledX, ApplyUncontrolledY, ApplyUncontrolledZ, IsingXX, IsingYY, IsingZZ, CheckQubitUniqueness, and RotationAngleValidation. That's a lot of redundant callables! Of those, only the three Ising gates are ones I think have a decent argument to be public, and even there I'd be fine if we instead decided to keep them internal. But as I'll go into in the other comment thread, we don't have a good solution for extensible internal intrinsics.

src/Simulation/TargetDefinitions/Decompositions/Measure.qs
swernli: Good point! The hope here is that as we find alternative decompositions for certain operations, we can introduce those here and allow users to select between them by choosing different targeting packages. How we can do that in a way that doesn't cause an explosion in the number of targeting packages is another question though...

swernli:
> > Initially I'd hoped to have these marked as internal so that we wouldn't have to fully document them, but in order for a body intrinsic operation to be implemented outside of its own binary it cannot be marked as internal.
> 
> Is that a limitation on the Q# source, or the generated C# source? If the latter, is that the case even with the `InternalsVisibleTo` C# attribute?

It is just a C# limitation, but even using `InternalsVisibleTo` only solves the problem for me and not future authors. Because `InternalsVisibleTo` requires putting the binary name and signing key on the class that defines the internal, it would mean putting on the target package the names of all binaries that implement it. For binaries in this repo that's easy enough, but then it becomes much harder for someone working in their own repo on a custom simulator or simulator extension to implement the target package intrinsics. Ideally, we wanted to land at a design that would allow for extensibility or custom implementation in a library without having to fork this repo or commit directly to it. But for now that seems to put us in the awkward position of needing to make public each of these arguably internal, helper callables, and then we get stuck supporting them.

src/Simulation/Simulators/QuantumSimulator/ApplyControlledZ.cs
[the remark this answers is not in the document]
swernli: To answer that a slightly different way, the Type1 package captures targeting a gate set is that is significantly more constrained than the one exposed by our simulator. As a result, implementing these limited intrinsics on top of our simulator requires using these C# classes as a shim that exposes less functionality. It's essentially the opposite of optimization, and it shows: simulating certain operations using this Type1 package are noticeably slower because it will use a complex decomposition, possibly including additional qubits and entanglers, instead of just calling directly into an optimized simulator API.

src/Simulation/Simulators.Type1/FindNuspecReferences.ps1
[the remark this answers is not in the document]
swernli: Actually, it looks like we have a habit of putting this almost everywhere. I see 620 files that include this in their comment just above the `Licensed under the MIT License` comment. Should we be pruning all of those? If so, I can file an issue for follow up rather that put all of that into this PR.

src/Simulation/Simulators.Type1/FindNuspecReferences.ps1
```diff
@@ -0,0 +1,103 @@
...
+    Write-Host "$target exists. Skipping generating new one."
+    exit
+ }
+
+
+# Start with the nuspec template
+$nuspec = [xml](Get-Content (Join-Path $PSScriptRoot "Microsoft.Quantum.Simulators.Type1.nuspec.template"))
+$dep = $nuspec.CreateElement('dependencies', $nuspec.package.metadata.NamespaceURI)
+
+function Add-PackageReferenceIfNew($ref)
```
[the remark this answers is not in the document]
swernli: Agreed (in fact, PowerShell is one of the few languages where it's not just a style difference but it can have functional implications as well, but that's a soap box for another time ;)). I updated this in the Type1 and Type2 FindNuspecReferences.ps1 files, but not in the existing ones that I haven't introduced or edited, for the sake of keeping the feature branch focused on decomposition related work.

swernli:
@cgranade @bettinaheim  I suppose I could use the `InternalsVisibleTo` as a temporary work around, where the changes in the feature branch do not introduce any new public APIs at the cost of having the target packages only be implementable in this repo for now, and then file a separate issue covering the need to allow implementation of a target package from another repo. That would punt the problem farther down the road, and we'd need to decide if we want to allow merging the feature branch into master with that work still pending or if we want to have a solution for this problem before accepting the target package feature into master.

cgranade:
> @cgranade @bettinaheim I suppose I could use the `InternalsVisibleTo` as a temporary work around, where the changes in the feature branch do not introduce any new public APIs at the cost of having the target packages only be implementable in this repo for now, and then file a separate issue covering the need to allow implementation of a target package from another repo. That would punt the problem farther down the road, and we'd need to decide if we want to allow merging the feature branch into master with that work still pending or if we want to have a solution for this problem before accepting the target package feature into master.

I agree that approach would kick the can down the road, but I honestly think that may be for the best in this case. Introducing new public Q# APIs comes at a large cost in terms of long-term maintenance, such that kicking the can gives us a bit of an opportunity to come up with an approach that doesn't commit us longer-term. That's just my 2¢, though, so please take with an appropriate grain of salt (apologies for mixed metaphors).

samarsha:
> But for now that seems to put us in the awkward position of needing to make public each of these arguably internal, helper callables, and then we get stuck supporting them.

Since they're intended to be implemented by simulators, it seems to me like they are actually part of the public API? Is the issue only that it shouldn't be exposed to Q# code, or that it shouldn't be exposed at all (in Q# or .NET)?

cgranade:
> > But for now that seems to put us in the awkward position of needing to make public each of these arguably internal, helper callables, and then we get stuck supporting them.
> 
> Since they're intended to be implemented by simulators, it seems to me like they are actually part of the public API? Is the issue only that it shouldn't be exposed to Q# code, or that it shouldn't be exposed at all (in Q# or .NET)?

I think part of the problem is that they are part of the public C# API, but shouldn't be part of the public Q# API; from the perspective of a quantum program, these APIs are an implementation detail internal to the simulation runtime, if I understand correctly.

samarsha:
I see, so we want the operation to be internal in Q#, but for C# generation to generate a public class. That is technically possible to do if we change C# generation, but it seems like there should be a cleaner way.

cgranade:
> I see, so we want the operation to be internal in Q#, but for C# generation to generate a public class. That is technically possible to do if we change C# generation, but it seems like there should be a cleaner way.

Agreed, there should be; hence why I tend to lean towards kicking the can whilst we brainstorm on and design that better way, rather than commiting to new public APIs in lieu of that.

swernli:
Agreed on all points, thanks @cgranade and @samarsha! I'll make the update in this PR to switch these new additions to internal and add this specific issue to the checklist in #249 so we can review it before the feature branch gets merged. For now, I'll treat any and all callabes that would be API additions as internal to avoid any new public API surface, and then we can debate the merits of exposing and fully documenting them on a case by case basis at a later date.

src/Simulation/TargetDefinitions/Decompositions/ApplyControlledX.qs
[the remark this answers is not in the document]
swernli: To summarize the temporary resolution we arrived at below, we are going to return to the question of the right way to balance extensibility vs making helper intrinsics public at a later date, and for now I've moved all 16 of these new callabes to be internal. That way they don't require an API review (yet). I've added this as a checklist item to #249 with the understanding that it may bet separated out into its own issue later.

swernli:
Ok, this seems like it should be ready for merge into the feature branch. Please let me know if there are any other things I should address in this PR and/or track in the overall issue for this feature!

src/Simulation/QsharpCore/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,9 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+// Allow the test assembly to use our internal methods
+[assembly: InternalsVisibleTo("Microsoft.Quantum.Simulators" + SigningConstants.PUBLIC_KEY)]
```
samarsha, line 9: Regarding the comment, this is not a test assembly, right?

src/Simulation/Common/Simulators.Type1.Dev.props
```diff
@@ -0,0 +1,42 @@
...
+    </None>
+  </ItemGroup>
+  
+  <Target Name="BeforeCsharpCompile">
+    <ItemGroup>
+      <Compile Include="$(GeneratedFilesOutputPath)**/*.g.cs" Exclude="@(Compile)" AutoGen="true" />
+    </ItemGroup>
+  </Target>
+
+</Project>
```
samarsha, line 42: This looks like it's identical to Simulators.Type2.Dev.props except for the project reference. Is it possible to factor these two files out into a common props file?

src/Simulation/Simulators.Type1.Tests/App.config
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <appSettings>
+    <add key="xunit.methodDisplay" value="method"/>
+  </appSettings>
+<startup><supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.6.1"/></startup></configuration>
```
samarsha, line 6: I am not sure what this file does, but the formatting in it looks off.

src/Simulation/Simulators.Type1.Tests/Tests.Microsoft.Quantum.Simulators.Type1.csproj
```diff
@@ -0,0 +1,57 @@
+﻿<Project Sdk="Microsoft.Quantum.Sdk/0.12.20082705-beta">
```
samarsha, line 1: Same question as the props file - this is almost identical to the Type2 csproj, is it possible to factor more of it out into a common file?

src/Simulation/QsharpCore/Properties/AssemblyInfo.cs
swernli: Good catch, this comment was copied but not updated correctly!

src/Simulation/Common/Simulators.Type1.Dev.props
swernli: I did try a variation like that, where all the simulators shared a single .props file that differentiated between which packages to load via conditionals. It does work, but the trade off for the reduced duplication is some extra complexity in seeing how files are loaded where. I ended up going back to this version in order to be more explicit. I could probably be convinced to address this in a follow up though ;)

src/Simulation/Simulators.Type1.Tests/App.config
swernli: Hmm... that is odd. It looks like all the simulator tests have this file in the same way, so it must have just been copied over and over again. I could fix them all here to make them more readable.

src/Simulation/Simulators.Type1.Tests/Tests.Microsoft.Quantum.Simulators.Type1.csproj
swernli: Honestly, while I think it could, I think at some point the reduced duplication is a diminishing return for the increased complexity. Eventually, I hope we can get to the point where all the simulators are unified into a single project, but for now I think I prefer to have them explicitly separate.

src/Simulation/Common/Simulators.Type1.Dev.props
samarsha: I wasn't thinking of conditionals, but a shared props file that has everything except the type1/2 project references. Then the type1/2 projects would be something like

```xml
<Project>
  <Import Project="Common.props" />
  <ItemGroup>
    <ProjectReference Include="$(EnlistmentRoot)src\Simulation\Type2Core\Microsoft.Quantum.Type2.Core.csproj" />
    <ProjectReference Include="$(EnlistmentRoot)src\Simulation\Simulators.Type2\Microsoft.Quantum.Simulators.Type2.csproj" Condition=" '$([MSBuild]::NormalizePath($(EnlistmentRoot)src\Simulation\Simulators.Type2))' != '$(MSBuildProjectDirectory)'" IncludeInSimulatorPackage="true" />
  </ItemGroup>
</Project>
```

Does that not work?
    <ProjectReference Include="$(EnlistmentRoot)src\Simulation\Type2Core\Microsoft.Quantum.Type2.Core.csproj" />

src/Simulation/Simulators.Type1.Tests/App.config
samarsha: It's a minor issue, so I'm OK with not fixing all of them right now. :)

src/Simulation/Common/Simulators.Type1.Dev.props
swernli: Oh, I see what you are suggesting... yeah, that does seem like it could work! I'll try pulling all the type specific stuff into the respective projects and then using the props file for only the things they have in common. Good thinking!

src/Simulation/Simulators.Type1.Tests/Tests.Microsoft.Quantum.Simulators.Type1.csproj
swernli: I'll see if the approach you suggested for the props file will also work here. Thanks!

src/Simulation/Common/Simulators.Type1.Dev.props
swernli: Finally had the chance to clean this up, and it looks much better this way. Thanks for the suggestion!

swernli:
This is once again building and passing the tests :). I'd like to try merging it by EOD.

swernli:
@samarsha @cgranade @alan-geller @bettinaheim Any chance I could get sign off on this so I can get it merged into the feature branch? Thanks!

src/Simulation/Simulators.Type1/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,9 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+// Allow the simulator assembly to use our internal methods
+[assembly: InternalsVisibleTo("Tests.Microsoft.Quantum.Simulators.Type1" + SigningConstants.PUBLIC_KEY)]
```
samarsha, line 9: This one is a test assembly. :) Although personally I think the comment is kind of redundant, so I don't mind if it's removed either.

src/Simulation/Simulators/QuantumSimulator/ApplyUncontrolledZ.cs
```diff
@@ -0,0 +1,35 @@
...
+
+
+            public QSimApplyUncontrolledZ(QuantumSimulator m) : base(m)
+            {
+                this.Simulator = m;
+            }
+
+            public override Func<Qubit, QVoid> __Body__ => (q1) =>
+            {
+                Simulator.CheckQubit(q1); ;
```
samarsha, line 27:
```suggestion
                Simulator.CheckQubit(q1);
```

src/Simulation/Simulators/QuantumSimulator/Extensions.cs
```diff
@@ -39,8 +39,12 @@ public static void InitBuiltinOperations<T>(this AbstractFactory<T> factory, Typ
 
             InitBuiltinOperations(factory, t.BaseType);
 
+            var overrideTypes = t.GetNestedTypes(
+                System.Reflection.BindingFlags.Public | 
+                System.Reflection.BindingFlags.NonPublic);
```
samarsha, line 44: `NonPublic` because of the internal intrinsics?

src/Simulation/TargetDefinitions/Decompositions/ExpUtil.qs
```diff
@@ -0,0 +1,31 @@
...
+namespace Microsoft.Quantum.Intrinsic {
+    open Microsoft.Quantum.Diagnostics;
+
+    @EnableTestingViaName("Test.TargetDefinitions.ExpUtil")
+    internal operation ExpUtil (paulis : Pauli[], theta : Double, qubits : Qubit[], rotation : ((Pauli, Qubit) => Unit is Adj + Ctl)) : Unit is Ctl {
+        if (Length(paulis) != Length(qubits)) { fail "Arrays 'paulis' and 'qubits' must have the same length"; }
+        if (Length(paulis) == 1) {
+            rotation(paulis[0], qubits[0]);
+        }
+        else { // Length(paulis) > 1
```
samarsha, line 13: Or `Length(paulis) == 0`? Does the `qubits[0]` reference below fail in that case?

src/Simulation/TargetDefinitions/Decompositions/CCNOTFromCCZ.qs
```diff
@@ -0,0 +1,39 @@
...
+            // [Page 15 of arXiv:1206.0758v3](https://arxiv.org/pdf/1206.0758v3.pdf#page=15)
+            within {
+                H(target);
+            }
+            apply {
+                Controlled Z([control1, control2], target);
+            }
+        }
+        controlled (ctls, ...) {
+            Controlled X (ctls + [control1, control2], target);
```
samarsha, line 35: Why does the controlled specialization not use the same decomposition?

src/Simulation/TargetDefinitions/Decompositions/ExpUtilFromIsing.qs
```diff
@@ -0,0 +1,46 @@
...
+                } elif (paulis[0] == PauliY) {
+                    IsingYY(theta / 2.0, qubits[0], qubits[1]);
+                } elif (paulis[0] == PauliZ) {
+                    IsingZZ(theta / 2.0, qubits[0], qubits[1]);
+                } else {
+                    fail "Type2 decompositions do not support PauliI as an input to Exp";
+                }
+            }
+        }
+        else { // Length(paulis) > 2
```
samarsha, line 29: Or `Length(paulis) == 0`?

samarsha (approved):

src/Simulation/Simulators.Type1/Properties/AssemblyInfo.cs
swernli: Oops! My find replace got a little too aggressive...

src/Simulation/TargetDefinitions/Decompositions/CCNOTFromCCZ.qs
swernli: That's certainly a possibility, and should end up being correct too, but I think the intent of the original author of these decompositions was to connect multi-controlled `CCNOT` back to `X` explicitly so that it was easier to follow the flow of the decomposition. That way much the same way that `Controlled X` with 2 control qubits is exactly `CCNOT` it is obvious that `CCNOT` with n control qubits is identical to `Controlled X` with n + 2 control qubits.

There's certainly opportunity in the future for experimentation and refinement of these decompositions, either in-place or as new target packages altogether, but for now I'd like to get this version in so we can continue to iterate on top of it.

src/Simulation/Simulators/QuantumSimulator/Extensions.cs
swernli: Yes, exactly.

src/Simulation/Simulators/QuantumSimulator/ApplyUncontrolledZ.cs
swernli: Nice catch, thanks!

src/Simulation/TargetDefinitions/Decompositions/ExpUtil.qs
swernli: Thankfully no, because this utility is only called by `Exp` in ExpFromExpUtil and `ExpFrac` in ExpFracFromExpUtil with a check:
```qsharp
            if (Length(paulis) != 0) {
```
or
```qsharp
            let (newPaulis, newQubits) = RemovePauliI(paulis, qubits);

            if (Length(newPaulis) != 0) {
                ExpUtil(newPaulis, theta , newQubits, R(_, -2.0 * theta, _));
            }
            else {
                ApplyGlobalPhase(theta);
            }
```
Where `RemovePauliI` correctly handles empty lists.

src/Simulation/TargetDefinitions/Decompositions/ExpUtilFromIsing.qs
swernli: Same as above; this utility is only called by `Exp` and `ExpFrac` that already check for an empty Pauli array.

src/Simulation/TargetDefinitions/Decompositions/ExpUtil.qs
samarsha: Maybe a quick comment mentioning this would be good, since it's potentially a gotcha if another place uses this operation without checking for length == 0?

PR #369, opened 2020-09-07
Adding handling of branching based on measurement results to SimulatorBase

bettinaheim:
This PR adds the handling of branching based on measurement results to SimulatorBase, exposing the virtual methods 
- StartBranchingBasedOnMeasurement,
- ExecuteBranchingBasedOnMeasurement, and
- EndBranchingBasedOnMeasurement. 

The PR removes the duplicated logic in the QuantumSimulator and the Tracer.
It also fixes the incorrect event spawned by Borrow, and contains other minor changes regarding which members are virtual, private, and protected. 

Since there are more PRs coming that impact the API, it may make sense to merge this to a feature branch. However, I am also trying to avoid having several feature branches with a large amount of changes that require coordination (the decomposition work comes to mind), and this change should be fairly lightweight.


swernli (approved):
LGTM!

PR #370, opened 2020-09-07
Unifying the two qubit managers

bettinaheim:
This is still a draft, since I have yet to confirm whether the API is sufficient or whether we need to keep more.


src/Simulation/Common/QubitManager.cs
```diff
@@ -74,185 +171,224 @@ public class QubitManager
...
-                else
+
+                long oldNumQubits = NumQubits;
+                long oldNone = None;
+                long oldAllocated = Allocated;
+                long oldDisabled = Disabled;
+                long oldAllocatedForBorrowing = AllocatedForBorrowing;
+                long[] oldQubitsArray = this.qubits;
+
+                UpdateQubitCapacity(oldNumQubits * 2); // This changes Allocated, Disabled, AllocatedForBorrowing markers.
```
swernli, line 333: When allocating a single new qubit, why do we update the capacity by 2x?

src/Simulation/Common/QubitManager.cs
```diff
@@ -49,6 +61,91 @@ public class QubitManager
...
 
+        private void UpdateQubitCapacity(long newQubitCapacity)
+        {
+            Debug.Assert(newQubitCapacity < MaxQubitCapacity);
+            NumQubits = newQubitCapacity;
+            None = NumQubits;
+            Allocated = NumQubits + 1;
+            Disabled = NumQubits + 2;
+            AllocatedForBorrowing = NumQubits + 3;
+            this.qubits = new long[NumQubits];
```
swernli, line 72: Based on the way `UpdateQubitCapacity` is called during allocate, this array doubles in size for each allocation, but it's not clear to me that this extra space is needed.

src/Simulation/Common/SimulatorBase.cs
```diff
@@ -429,8 +425,7 @@ public GetQubitsAvailableToBorrow(SimulatorBase m) : base(m)
                 sim = m;
             }
 
-            public override Func<QVoid, long> __Body__ => (arg) => sim.QubitManager.GetParentQubitsAvailableToBorrowCount() +
-                                                               sim.QubitManager.GetFreeQubitsCount();
+            public override Func<QVoid, long> __Body__ => (arg) => sim.QubitManager.GetQubitsAvailableToBorrowCount(1);
```
swernli, line 428: If passing `1` means get the value for the current stack frame, perhaps it's worth having that as the default value for the parameter?

src/Simulation/Simulators.Tests/Circuits/GetAvailableTest.qs
```diff
@@ -9,22 +9,22 @@ namespace Microsoft.Quantum.Simulation.Simulators.Tests.Circuits {
         mutable innerBorrow = 0;
         borrowing (qb = Qubit())
         {
-            set innerBorrow = GetQubitsAvailableToBorrow();
+            set innerBorrow = GetQubitsAvailableToBorrow() + GetQubitsAvailableToUse();
         }
-        return (GetQubitsAvailableToBorrow(), innerBorrow);
+        return (GetQubitsAvailableToBorrow() + GetQubitsAvailableToUse(), innerBorrow);
```
swernli, line 14: Are these changes that add together the available to use and available to borrow counts because the contract on the `GetQubitsAvailableToBorrow` method changed or because the way the test executes changed?

swernli (commented):
I have a few clarifying questions to help put the refactor in context (some of which are regarding functionality that was not changed in the refactor). Thanks!

src/Simulation/Common/SimulatorBase.cs
bettinaheim: Technically, passing 1 gets the value for the parent stack frame rather than the current one. The reason this needs to be 1 here however is that in Q# you would have a call to this operation and then *within* the called operation, the value is evaluated, meaning to get the accurate result you would need to know not the qubits in the current stack frame (which is inside GetQubitsAvailableToBorrow), but rather the qubits in the parent (which is where GetQubitsAvailableToBorrow is called).

src/Simulation/Simulators.Tests/Circuits/GetAvailableTest.qs
bettinaheim: I changed the interpretation of GetQubitsAvailableToBorrow (and the doc comment on it as well). Before by definition it returned not just all qubits that you could borrow, but also all qubits that weren't allocated/used yet. Since one interpretation can be converted to the other via  + or - GetQubitsAvailabeToUse() the change is certainly not necessary. I merely took the opportunity to do this switch as part of revising the Q# core library in general. @cgranade, @alan-geller for input on which one is better. All uses that I could come up with would require knowing how many qubits can be borrowed and not how many there are, but we should certainly be mindful to not change things unless there is a good reason to and maybe that is not a good enough reason.

src/Simulation/Common/QubitManager.cs
bettinaheim: I believe the capacity is only doubled if if free == None (meaning no more qubits can be allocated without exceeding the current capacity), and MayExtendCapacity is true.

src/Simulation/Common/QubitManager.cs
bettinaheim: See the reply above.

src/Simulation/Common/QubitManager.cs
```diff
@@ -17,22 +16,35 @@ namespace Microsoft.Quantum.Simulation.Common
...
+        /// </summary>
+        protected long NumAllocatedQubits;
+        /// <summary>
+        /// The number of disabled qubits.
+        /// </summary>
+        protected long NumDisabledQubits;
+        /// <summary>
+        /// Tracks the allocation state of all qubits.
+        /// </summary>
+        protected long[] qubits;
```
cgranade, line 36: Based on the discussion below about growing this, would `List<long>` be better for growing capacity dynamically?

src/Simulation/Simulators/QuantumProcessor/QuantumProcessorDispatcher.cs
```diff
@@ -27,19 +26,19 @@ public IQuantumProcessor QuantumProcessor
...
-        /// <param name="qubitManager">An instance of a class implementing <see cref="IQubitManager"/> interface. If the parameter is null <see cref="QubitManagerTrackingScope"/> is used.</param>
+        /// <param name="quantumProcessor">An instance of a class implementing <see cref="IQuantumProcessor"/> interface to be wrapped.</param>
+        /// <param name="qubitManager">An instance of a class implementing <see cref="IQubitManager"/> interface. If the parameter is null <see cref="QubitManager"/> is used.</param>
         /// <param name="randomSeed">A seed to be used by Q# <a href="https://docs.microsoft.com/qsharp/api/qsharp/microsoft.quantum.intrinsic.random">Microsoft.Quantum.Intrinsic.Random</a> operation.</param>
-        public QuantumProcessorDispatcher(IQuantumProcessor? quantumProcessor = null, IQubitManager? qubitManager = null, int? randomSeed = null)
+        public QuantumProcessorDispatcher(IQuantumProcessor quantumProcessor, IQubitManager? qubitManager = null, int? randomSeed = null)
```
cgranade, line 32: This may have release note impact and warrant a roll-forward of the minor version number, since it's a breaking change to the interface for this class, and could break ) third-party simulators not tracked in our cross-repo build.

src/Simulation/Simulators.Tests/Circuits/GetAvailableTest.qs
cgranade: I think either way makes sense; as you note, it's really easy to convert between. I could see, for instance, a pure-Q# library operation `GetQubitsAvailableToBorrowOrUse` to make the distinction clear.

That said, this is a breaking change that we can't make a deprecation stub for (the operation didn't change names, to wit). I strongly suspect these two operations are rarely used enough that we should be OK, but it would take a bump in minor version and a release note line item.

cgranade (commented):
Thanks for tagging me into the discussion; left a few thoughts by way of early feedback. Thanks!

src/Simulation/Common/IQubitManager.cs
```diff
@@ -3,41 +3,37 @@
 
 namespace Microsoft.Quantum.Simulation.Common
 {
-    using Microsoft.Quantum.Simulation.Core;
     using System.Collections.Generic;
+    using Microsoft.Quantum.Simulation.Core;
 
     public interface IQubitManager
     {
+        bool DisableBorrowing { get; }
```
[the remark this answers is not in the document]
bettinaheim: Yes, they did.

src/Simulation/Common/QubitManager.cs
```diff
@@ -74,185 +171,224 @@ public class QubitManager
...
+                curFrame = new StackFrame();
             }
+        }
 
-            for (long i = oldNumQubits; i < NumQubits; i++)
-            {
-                this.qubits[i] = i + 1;
-            }
-            Debug.Assert(this.qubits[NumQubits - 1] == None);
+        private class QubitNonAbstract : Qubit
```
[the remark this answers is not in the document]
bettinaheim: Ech, I was really hoping to get rid of the abstract for the Qubit class (it doesn't have any abstract members or similar) and then delete this code (that already existed before this PR). For reasons I haven't figured out yet, everything starts to fail if I remove the abstract from that base class however.

src/Simulation/Simulators/QuantumProcessor/QuantumProcessorDispatcher.cs
bettinaheim: There are plenty of breaking changes in this PR (like removing an entire class), so the idea is to accumulate these breaking changes across the runtime repo and then get them over with. The goal is to walk out with an API we are ready to commit to maintaining.

src/Simulation/Common/QubitManager.cs
swernli: Ah, I see that now! Thanks!

src/Simulation/Common/QubitManager.cs
```diff
@@ -17,22 +16,35 @@ namespace Microsoft.Quantum.Simulation.Common
...
+        protected long NumDisabledQubits;
+        /// <summary>
+        /// Tracks the allocation state of all qubits.
+        /// </summary>
+        protected long[] qubits; 
+
+        private long None; // Indicates the end of the list of free qubits. If we reach it - we are out of qubits.
+        private long Allocated; // All qubits allocated by user will be marked with this number.
+        private long Disabled; // All qubits disabled via DisableQubit interface will be marked with this number.
+        private long AllocatedForBorrowing; // All qubits allocated only for borrowing, will be marked with this number or higher.
```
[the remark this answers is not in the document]
alan-geller: Adding more: actually the state is a bit more complicated, since a qubit might be Allocated in one frame and Borrowed in another. I *think* that there are really 3 states: Free, Allocated, and Disabled. The StackFrame class should hold a list of the qubits that have are currently borrowed in that frame. To see if a qubit is currently borrowed, look to see if it's borrowed in the current frame, since that's the only borrowing that matters (if it's borrowed in a higher frame, so what? it can still be borrowed again, and it can't be released, so it's functionally just Allocated).

src/Simulation/Common/QubitManager.cs
```diff
@@ -355,91 +494,90 @@ public virtual void Release(IQArray<Qubit> qubitsToRelease)
...
 
-            numAllocatedQubits--;
-            Debug.Assert(numAllocatedQubits >= 0);
-        }
+        #region Borrow
 
         /// <summary>
-        /// Disables a given qubit.
-        /// Once a qubit is disabled it can never be reallocated.
+        /// Returns a qubit that is free, not disabled, and not listed as in use, whose id is in [minId, maxId).
```
alan-geller, line 506: It looks like it actually will never return a free qubit, only an allocated qubit that's available for borrowing (not in scope in the current frame).

src/Simulation/Common/QubitManager.cs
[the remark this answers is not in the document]
alan-geller: OK, it's even a bit more complicated: if you allocate a qubit just for borrowing, you need to mark it as taken (AllocatedForBorrowing) so you don't return it to an allocation deeper in the stack. That's OK, except that if you allow re-borrowing -- and why not? -- you'd need to keep a "borrowed count" to know when the `Return` should set the state back to Free. I think you can keep this in the borrowed frame by tracking (borrowed qubit, prior state) pairs instead of just borrowed qubits.

Admittedly at this point there's a list of qubit states, a stack or queue of free qubits, and a data structure (probably a dictionary) tracking borrowed qubits in each stack frame, all just to replace one array of longs. Arguably it's not really that much simpler, although at least each data structure has a clear, specific purpose.

src/Simulation/Common/QubitManager.cs
```diff
@@ -355,91 +494,90 @@ public virtual void Release(IQArray<Qubit> qubitsToRelease)
...
+                borrowed.Modify(curBorrowed, Borrow(inUse, minId: lastBorrowedId + 1));
+                lastBorrowedId = borrowed[curBorrowed].Id;
             }
+
+            if (numBorrowed < numToBorrow)
+            { // Not enough qubits to borrow. Allocate what was not borrowed.
+                for (long i = numBorrowed; i < numToBorrow; i++)
+                {
+                    borrowed.Modify(i, Allocate(usedOnlyForBorrowing: true));
+                    if (borrowed[(int)i] == null)
```
alan-geller, line 562: It might be worth having a method something like

```csharp
bool CanAllocate(int count) { ... }
```

that tells you whether or not it's possible to allocate `count` qubits, looking at the length of the free list and suchlike. That way here (and in another couple of places) you wouldn't have to check for allocation failures, but instead could fail before trying to allocate any new qubits.

src/Simulation/Simulators/QuantumProcessor/Return.cs
```diff
@@ -16,16 +16,18 @@ public class QuantumProcessorDispatcherReturn : Intrinsic.Return
 
             public override void Apply(Qubit q)
             {
-                long count = (sim.QubitManager.ToBeReleasedAfterReturn(q) ? 1 : 0);
-                sim.QuantumProcessor.OnReturnQubits(new QArray<Qubit>(q), count);
+                long allocatedBefore = sim.QubitManager.GetAllocatedQubitsCount();
                 sim.QubitManager.Return(q);
+                long allocatedAfter = sim.QubitManager.GetAllocatedQubitsCount();
+                sim.QuantumProcessor.OnReturnQubits(new QArray<Qubit>(q), allocatedBefore - allocatedAfter);
```
alan-geller, line 22: A broader question: If we think there is value to these events in the general case, then I would move the events into SimulatorBase and move firing them into the qubit manager. If they're special-purpose events that we don't want to make generally available, then we can leave them here.

alan-geller (approved):
A few small comments, a couple of bigger ones. 
This is a long-overdue simplification -- thanks!!!

src/Simulation/Common/QubitManager.cs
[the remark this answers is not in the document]
bettinaheim: The virtual properties don't seem to be overridden anywhere and making them non-virtual compiles fine. Removing the abstract from the class somehow results in something wanting a parameterless constructor that the class doesn't have ("No parameterless constructor defined for type 'Microsoft.Quantum.Simulation.Core.Qubit'.").

src/Simulation/Common/QubitManager.cs
```diff
@@ -49,6 +61,86 @@ public class QubitManager
...
+                        _locals = new List<Qubit>();
+                    }
+
+                    return _locals;
+                }
+            }
+        }
+
+        protected readonly Stack<StackFrame> operationStack; // Stack of operation calls.
+        protected StackFrame curFrame; // Current stack frame - all qubits in current scope are listed here.
```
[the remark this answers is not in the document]
bettinaheim: No reason aside from that I copied that code over and didn't change it. I make it a single stack as you suggested.

src/Simulation/Common/QubitManager.cs
```diff
@@ -74,185 +166,224 @@ public class QubitManager
...
+            public QubitNonAbstract(int id) : base(id) { }
+        }
 
-            if (free == oldNone)
+        /// <summary>
+        /// May be overriden to create a custom Qubit object of a derived type.
+        /// </summary>
+        /// <param name="id">unique qubit id</param>
+        /// <returns>a newly instantiated qubit</returns>
+        public virtual Qubit CreateQubitObject(long id)
```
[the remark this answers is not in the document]
bettinaheim: Then the conversion long to int just needs to happen in a bunch of other places (since Q# has Int64 by default). Hence having it throw here seems somewhat cleaner to me in this case. The alternative would be to make the entire qubit manager int based, and then have the checks at the intersection with Q# (meaning Allocate, Borrow, GetQubitsAvailableToUse, and GetQubitsAvailableToBorrow would need to do this check).

src/Simulation/Common/QubitManager.cs
```diff
@@ -74,185 +166,224 @@ public class QubitManager
...
+        }
+
         private bool IsAllocatedForBorrowing(long id)
         {
             return qubits[id] >= AllocatedForBorrowing;
         }
 
-        // Returns true if qubit needs to be released: It has been allocated for borrowing and is not borrowed any more.
-        private bool BorrowingRefCountIsOne(long id)
+        public virtual long GetQubitsAvailableToBorrowCount(int stackFrame)
```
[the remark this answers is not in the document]
bettinaheim: Since I don't think this method is called very often or with a large number here, I am not too concerned about the overhead of the temporary stack. Having the option to query the available qubits to borrow for the current frame makes a lot of sense to me, and I could imaging being able to query other frames would be useful - that is if we are not talking about the scenario where these are invoked from within Q# but rather directly by something interacting with the manager.

src/Simulation/Common/QubitManager.cs
```diff
@@ -355,91 +494,90 @@ public virtual void Release(IQArray<Qubit> qubitsToRelease)
...
-        public virtual long GetQubitsAvailableToBorrowCount()
-        {
-            return numAllocatedQubits;
-        }
-
-        public virtual long GetParentQubitsAvailableToBorrowCount()
-        {
-            return GetQubitsAvailableToBorrowCount();
-        }
+                while (IsFree(curQubit) || IsDisabled(curQubit))
```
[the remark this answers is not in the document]
bettinaheim: sure.

src/Simulation/Common/QubitManager.cs
bettinaheim: fixed

src/Simulation/Simulators/QuantumProcessor/Return.cs
bettinaheim: There are already such events in the SimulatorBase. The QuantumProcessor has a method OnReturnQubits that it adds as handler to the corresponding OnReturnQubits event spawned by the simulator base.

src/Simulation/Common/QubitManager.cs
bettinaheim: Quite silly, but since we are attempting to actually enable nullable, querying such a method and then assuming that none of the allocated qubits is null would not make the nullable check happy.

src/Simulation/Common/QubitManager.cs
```diff
@@ -74,185 +166,224 @@ public class QubitManager
...
+        /// Disables a given qubit.
+        /// Once a qubit is disabled it can never be reallocated.
+        /// </summary>
+        public virtual void Disable(Qubit qubit)
         {
-            if (IsAllocatedForBorrowing(id))
+            qubits[qubit.Id] = Disabled;
+            NumDisabledQubits++;
+
+            NumAllocatedQubits--;
```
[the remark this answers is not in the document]
bettinaheim: I really don't know what the intent is of Disable. It looks quite powerful as is, with the qubit being permanently and instantaneously unavailable independent on whether it is visible in the current scope, is in use by multiple things (e.g. is borrowed) or anything else.

src/Simulation/Common/QubitManager.cs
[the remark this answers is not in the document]
bettinaheim: While I broadly agree that it could be solved differently, I did focus on refactoring the public API rather then rewriting the code from scratch. :) 
I haven't looked close enough to have formed a full judgement on what I think the best internal representation could be that ensures both performance and readability. For this PR, since these are private handles (intentionally), I would like to proceed without doing a full refactoring since the code is in a working state and the priority needs to be to get to a stable API we can commit to.

src/Simulation/Common/QubitManager.cs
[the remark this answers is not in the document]
bettinaheim: No reason, just copied over that way. I changed it to be one stack.

src/Simulation/Common/QubitManager.cs
```diff
@@ -49,6 +61,91 @@ public class QubitManager
...
+            AllocatedForBorrowing = NumQubits + 3;
+            this.qubits = new long[NumQubits];
+        }
+
+        // stack frame management needed to support borrowing qubits
+        #region StackFrameManagement
+
+        protected class StackFrame
+        {
+            internal IApplyData Data;
```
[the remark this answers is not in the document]
bettinaheim: I enabled nullable. What is somewhat suboptimal is that any kind of indirection for a null check will cause a warning.

src/Simulation/Common/QubitManager.cs
```diff
@@ -74,185 +171,224 @@ public class QubitManager
...
+                    else if (oldQubitsArray[i] == oldDisabled)
+                    {
+                        // Disabled qubits are marked differently now.
+                        this.qubits[i] = Disabled;
+                    }
+                    else if (oldQubitsArray[i] >= oldAllocatedForBorrowing)
+                    {
+                        // This updates refCounts
+                        this.qubits[i] = (oldQubitsArray[i] - oldAllocatedForBorrowing) + AllocatedForBorrowing;
+                    }
```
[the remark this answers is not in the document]
bettinaheim: The compiler doesn't seem happy with that, telling me that a constant value is expected.

src/Simulation/Common/QubitManager.cs
bettinaheim: The intent is that growing should happen fairly infrequently (it is doubling size every time, and that is only if MayExtendCapacity is explicitly enabled).

src/Simulation/Common/QubitManager.cs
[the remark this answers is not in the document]
alan-geller: Makes sense to me. We can add this code to the "eventual refactor" list.

src/Simulation/Common/QubitManager.cs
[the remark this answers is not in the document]
alan-geller: OK, makes sense.

src/Simulation/Common/QubitManager.cs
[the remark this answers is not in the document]
alan-geller: OK. It's not so much a perf issue as a code complexity issue, but it's not particularly complicated as it is :-).

src/Simulation/Common/QubitManager.cs
alan-geller: Ah... If you have to check anyway, might as well just check, like you're doing, then.

src/Simulation/Simulators/QuantumProcessor/Return.cs
alan-geller: OK... I kind of hate chained event handlers, but that's *way* out of scope of this PR.

src/Simulation/Common/QubitManager.cs
```diff
@@ -74,185 +166,224 @@ public class QubitManager
...
+                    operationStack.Push(f);
                 }
             }
-            return false;
+            return DisableBorrowing ? 0 : NumAllocatedQubits - QubitsInUseCount(frame);
         }
 
-        // Returns true if qubit needs to be released: It has been allocated for borrowing and is not borrowed any more.
-        private bool DecreaseBorrowingRefCount(long id)
+        public long GetFreeQubitsCount()
```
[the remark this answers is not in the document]
bettinaheim: Agreed. The question is just how heavy we want to be on the API change side.

src/Simulation/Common/QubitManager.cs
```diff
@@ -74,185 +166,224 @@ public class QubitManager
...
-                else
-                {
-                    qubits[id]--;
-                    return false;
-                }
-            } else
+            return NumQubits - NumDisabledQubits - NumAllocatedQubits;
+        }
+
+        public long GetAllocatedQubitsCount()
```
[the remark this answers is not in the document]
bettinaheim: same

src/Simulation/Common/QubitManager.cs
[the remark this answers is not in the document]
bettinaheim: It got even simpler with having the current frame be part of the stack.

src/Simulation/Simulators/QuantumProcessor/Return.cs
bettinaheim: This additional indirection should become obsolete as part of the larger refactoring, such that only the events in the SimulatorBase remain.

src/Simulation/Common/QubitManager.cs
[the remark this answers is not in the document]
bettinaheim: I added that change.

src/Simulation/Simulators.Tests/Circuits/GetAvailableTest.qs
bettinaheim: I'll keep a list of breaking changes.

src/Simulation/Common/Factory.cs
```diff
@@ -13,7 +13,7 @@ namespace Microsoft.Quantum.Simulation.Common
     ///     It also provides a mechanism to register overrides when a given
     ///     type should be replaced by a subclass.
     /// </summary>
-    public abstract class AbstractFactory<T>
+    public class Factory<T>
     {
```
anpaz-msft, line 17: This is technically a breaking change.

src/Simulation/Common/IQubitManager.cs
```diff
@@ -3,41 +3,37 @@
...
+
+        bool IsValid(Qubit qubit);
+        bool IsFree(Qubit qubit);
+        bool IsDisabled(Qubit qubit);
+
+        long FreeQubitsCount { get; }
+        long AllocatedQubitsCount { get; }
+        IEnumerable<long> AllocatedIds();
```
anpaz-msft, line 19: if we switch the rest to be properties, why is this a method?

src/Simulation/Common/IQubitManager.cs
```diff
@@ -3,41 +3,37 @@
 
 namespace Microsoft.Quantum.Simulation.Common
 {
-    using Microsoft.Quantum.Simulation.Core;
     using System.Collections.Generic;
+    using Microsoft.Quantum.Simulation.Core;
 
     public interface IQubitManager
     {
```
anpaz-msft, line 10: This also contain breaking changes. We'll need to update the minor (I believe Sarah did some other breaking changes recently, so it's probably a good opportunity)

src/Simulation/Common/IQubitManager.cs
```diff
@@ -3,41 +3,37 @@
...
+        bool DisableBorrowing { get; }
+
+        bool IsValid(Qubit qubit);
+        bool IsFree(Qubit qubit);
+        bool IsDisabled(Qubit qubit);
+
+        long FreeQubitsCount { get; }
+        long AllocatedQubitsCount { get; }
+        IEnumerable<long> AllocatedIds();
+        long QubitsAvailableToBorrowCount(int stackFrame = 0);
```
anpaz-msft, line 20: Having this as a method really breaks the symmetry. Is it really necessary?
Do we really need a method to pass the stackFrame? doesn't seem to be something that should live in the interface, maybe it can be a public method in the implementation?

anpaz-msft (commented):
Have we checked that there are no perf regressions with these changes?
They also include breaking changes, so we should update the minor.

src/Simulation/Common/IQubitManager.cs
bettinaheim: Yes.

src/Simulation/Common/IQubitManager.cs
bettinaheim: What symmetry? 
Before there were two methods, one for stackFrame = 0 and one for stackFrame = 1 that were needed as part of the interface. Now there is one method with an argument that makes it more general (I suspect at some point we may want higher up stack frames, so making it more general rather than  having it a separate method seems like a good choice to me.

bettinaheim:
> Have we checked that there are no perf regressions with these changes?
> They also include breaking changes, so we should update the minor.

There is no functional or algorithmic change; this is merely an API change. We certainly monitor perf for everything.

src/Simulation/Common/IQubitManager.cs
bettinaheim: The two counts require no computation; they are already fields on the implementation and Count properties are common for other data types. This one on the other hand likely requires computation and seems closer to the other methods.

PR #372, opened 2020-09-11
Add RequiresCapability attribute

samarsha:
The `Capability` attribute is used by the capability inference feature (see microsoft/qsharp-compiler#600), which automatically adds the attribute to callables when a library is compiled. The attribute can also be used to override the inferred capability.

The value of the attribute is the name of the capability level, which means the names will be part of the public Q# API. However, the current names "Unknown," "QPRGen1," and "QPRGen0" are temporary, so we need to finalize those before this PR should be merged.

Alternatively, the `Capability` attribute could use an `Int` instead of a `String` to avoid picking names, but I think that would be more cryptic - and because there are only a finite number of levels, a full `Int` doesn't seem right. `String` is also not quite right, but we don't have discriminated unions or enums yet.

We might also want to consider what happens if there are capabilities along multiple dimensions, where a single value is not enough to represent the required capabilities accurately.

TODO:

- [x] Rename `Capability` to `RequiresCapability`
- [x] Move the attribute out of the `Microsoft.Quantum.Core` namespace
- [x] Add a `Reason : String` item to the attribute
- [x] Finalize names of capability levels "Unknown," "QPRGen1," and "QPRGen0"


alan-geller:
One nice thing about using a string is that you could use multiple appearances of this attribute to indicate multiple, additive capabilities in the multi-dimensional case. It's a bit harder with an integer. Strings are definitely more sensitive to typos -- a DU or enum would be nice to have for this -- but they're more flexible as well.

That said, I would try to avoid having many orthogonal capabilities if we can. It makes things much harder to understand than if there's a simple linear ordering. If we can't make the simple linear ordering work, then we'll deal with that, but I think it makes sense to start with linear capability levels and break that only if it's really necessary.

samarsha:
I had to add a capability override to `Reset` to fix an issue with IonQ thinking `Reset` used unsupported capabilities:

```qsharp
@Capability("QPRGen0")
operation Reset (target : Qubit) : Unit {
    if (M(target) == One) {
        X(target);
    }
}
```

Since `@Capability("QPRGen0")` looks kind of strange without context, I'm thinking of adding a required `Reason : String` property, something like:

```qsharp
@Capability("QPRGen0", "Reset is replaced by a supported operation on all execution targets.")
operation Reset (target : Qubit) : Unit {
    if (M(target) == One) {
        X(target);
    }
}
```

src/Simulation/QsharpCore/Core.qs
```diff
@@ -30,6 +30,21 @@ namespace Microsoft.Quantum.Core {
     @Attribute()
     newtype Deprecated = (NewName : String);
 
+    /// # Summary
+    /// Compiler-recognized attribute used to mark a callable with the runtime capabilities it requires.
+    ///
+    /// # Input
+    /// ## Level
+    /// The name of the runtime capability level required by the callable.
```
cgranade, line 38: This API documentation section will not currently be formatted in documentation (see https://github.com/microsoft/qsharp-compiler/issues/113). If https://github.com/microsoft/qsharp-compiler/pull/594 is adopted, the format for named items in UDTs will be finalized as the following:
```suggestion
    /// # Named Items
    /// ## Level
    /// The name of the runtime capability level required by the callable.
```

src/Simulation/QsharpCore/Core.qs
```diff
@@ -30,6 +30,21 @@ namespace Microsoft.Quantum.Core {
...
+    /// 
+    /// # Remarks
+    /// The valid capability level names, in order of increasing capabilities, are:
+    ///
+    /// 1. QPRGen0
+    /// 2. QPRGen1
+    /// 3. Unknown
+    @Attribute()
+    newtype Capability = (Level : String);
```
cgranade, line 47: Would it make sense to include in the name of this attribute that this isn't a capability itself, but a marker that a given capability is required? E.g.: `RequiredCapability`?

src/Simulation/QsharpCore/Core.qs
```diff
@@ -30,6 +30,21 @@ namespace Microsoft.Quantum.Core {
...
+    /// Compiler-recognized attribute used to mark a callable with the runtime capabilities it requires.
+    ///
+    /// # Input
+    /// ## Level
+    /// The name of the runtime capability level required by the callable.
+    /// 
+    /// # Remarks
+    /// The valid capability level names, in order of increasing capabilities, are:
+    ///
+    /// 1. QPRGen0
```
cgranade, line 43: Is there any documentation available for what features are included in each of these capability levels?

cgranade (commented):
I realize this is still in draft, but just wanted to provide early feedback in the hopes that it would be helpful. If I could ask, if this new attribute is primarily intended to be emitted by compiler rewrite steps, would it make sense to move it out of Microsoft.Quantum.Core so as to not add noise to autocomplete suggestions for all Q# source files? Something like `Microsoft.Quantum.Targeting`, perhaps?

src/Simulation/QsharpCore/Core.qs
cgranade: Very possibly, yeah. We don't yet have an official format for named items in UDTs due to https://github.com/microsoft/qsharp-compiler/issues/113, so it may be best to hold off updating until we have a format decided on.

samarsha:
> If I could ask, if this new attribute is primarily intended to be emitted by compiler rewrite steps, would it make sense to move it out of Microsoft.Quantum.Core so as to not add noise to autocomplete suggestions for all Q# source files? Something like `Microsoft.Quantum.Targeting`, perhaps?

That might work. It is intended to be used by users to override the compiler if needed, but that should happen rarely, since ideally the compiler will infer the right capability almost all of the time.

src/Simulation/QsharpCore/Core.qs
samarsha: I went ahead and renamed `Input` to `Named Items` for both this attribute and `Deprecated`.

src/Simulation/QsharpCore/Core.qs
samarsha: I added descriptions to the doc comment here.

src/Simulation/QsharpCore/Core.qs
samarsha: Yeah, that is probably a better name. I think I would slightly prefer `RequiresCapability`.

cgranade: That kind of points to attributes-as-adjectives more generally, yeah.

samarsha:
For reference, here's my proposed API of the attribute (which is different from the current code - wanted to get feedback before I change it here and in the compiler):

```qsharp
newtype RequiresCapability = (String : Level, String : Reason);
```

src/Simulation/QsharpCore/Intrinsic.qs
```diff
@@ -657,24 +658,26 @@ namespace Microsoft.Quantum.Intrinsic {
...
     /// Given a single qubit, measures it and ensures it is in the |0⟩ state
     /// such that it can be safely released.
     ///
     /// # Input
-    /// ## qubit
+    /// ## target
     /// The qubit whose state is to be reset to $\ket{0}$.
+    @RequiresCapability(
+        "BasicQuantumFunctionality",
+        "Reset is replaced by a supported implementation on all execution targets."
```
bettinaheim, line 672: Not for this PR, but we need a better way of handling that, since while this is true for all of the targets we provide, there is nothing that would indicate to someone defining their own target that this is required. I'll create a task to do that once we have the decompositions merged.

src/Simulation/QsharpCore/Targeting.qs
```diff
@@ -0,0 +1,35 @@
...
+    /// ## BasicMeasurementFeedback
+    /// Measurement results can be compared for equality only in if-statement conditional
+    /// expressions in operations. The block of an if-statement that depends on a result cannot
+    /// contain set statements for mutable variables declared outside the block, or return
+    /// statements.
+    ///
+    /// ## FullComputation
+    /// No runtime restrictions. Any Q# program can be executed.
+    @Attribute()
+    newtype RequiresCapability = (Level : String, Reason : String);
```
bettinaheim, line 34: I will put in a vote for RequiredCapability rather than RequiresCapability.

bettinaheim (approved):

src/Simulation/QsharpCore/Targeting.qs
```diff
@@ -0,0 +1,35 @@
...
+    ///
+    /// ## BasicMeasurementFeedback
+    /// Measurement results can be compared for equality only in if-statement conditional
+    /// expressions in operations. The block of an if-statement that depends on a result cannot
+    /// contain set statements for mutable variables declared outside the block, or return
+    /// statements.
+    ///
+    /// ## FullComputation
+    /// No runtime restrictions. Any Q# program can be executed.
+    @Attribute()
```
cgranade, line 33: An `/// # Example` section may help here.
```suggestion
    /// # Example
    /// The following snippet defines an operation `Reset` that requires the `BasicQuantumFunctionality` capability:
    /// ```Q#
    /// @RequiresCapability(
    ///     "BasicQuantumFunctionality",
    ///     "Reset is replaced by a supported implementation on all execution targets."
    /// )
    /// operation Reset(target : Qubit) : Unit {
    ///     // ...
    /// }
    /// ```
    @Attribute()
```

cgranade (commented):
Strictly speaking, this should go through Q# API review for approval at the start of next month's development cycle; that said, I think given that this is a very small change that's mainly focused on a fairly niche use case, I'm OK to approve directly in this PR (@msoeken and @guenp, do you concur?), modulo a few comments about API documentation.

src/Simulation/QsharpCore/Targeting.qs
cgranade: I personally am more in favor of "Requires," since it reads more like an adjective modifying the noun or verb, but that's just my 2¢.

samarsha:
> Strictly speaking, this should go through Q# API review for approval at the start of next month's development cycle; that said, I think given that this is a very small change that's mainly focused on a fairly niche use case, I'm OK to approve directly in this PR (@msoeken and @guenp, do you concur?), modulo a few comments about API documentation.

We could hold it until next month to go through API review. @bettinaheim Is there a reason to get it in this month?

guenp:
@samarsha @cgranade I suggest to merge this and have a post-discussion in the API review meeting and in the (perhaps unlikely event) where we want to change anything we can do a follow-up PR

src/Simulation/QsharpCore/Targeting.qs
samarsha: Other attributes in Q# describe what something is or does, rather than what something has. A callable with the `@EntryPoint` attribute *is* an entry point, a callable or type with the `@Deprecated` attribute *is* deprecated. A callable with the `@RequiredCapability` attribute is not itself a required capability - it only *has* a required capability. So for consistency I would prefer at least `@HasRequiredCapability` or `@RequiresCapability` for short.

bettinaheim: I am fine with RequiresCapability.

src/Simulation/QsharpCore/Targeting.qs
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+/// # Summary
+/// This namespace includes Q# core functions and operations.
```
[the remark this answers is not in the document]
samarsha: Oops, forgot to update the description.

bettinaheim:
> @bettinaheim Is there a reason to get it in this month?

@samarsha We are good to pull that in as per Guen's suggestion and having talked to Chris. Please go ahead with merging this.

src/Simulation/QsharpCore/Targeting.qs
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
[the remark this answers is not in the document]
samarsha: Out of curiosity, do you have a link to the guidelines for copyright headers?

src/Simulation/QsharpCore/Targeting.qs
samarsha: I'm not opposed to an example, but I'm not sure we should use an example that is only a short-term workaround until the decompositions for `Reset` are integrated into the runtime. But I can't think of an actual example that seems like good practice and a recommended usage - the compiler really should infer it correctly... @bettinaheim?

cgranade: Ah, good point. Maybe something in the remarks explaining that users shouldn't use directly, then?

samarsha: I added a note about that.

PR #377, opened 2020-09-16
E2E test.

anpaz-msft:


anpaz-msft:
Confirmed that E2E works correctly.

PR #379, opened 2020-09-18
Added new EmptyArray function.

cgranade:
This PR adds a new EmptyArray function to help deprecate the `new` keyword, and in keeping with the improvements at https://github.com/microsoft/QuantumLibraries/issues/313.


src/Simulation/QsharpCore/Arrays/Empty.cs
```diff
@@ -0,0 +1,19 @@
...
+using Microsoft.Quantum.Simulation.Core;
+
+namespace Microsoft.Quantum.Arrays
+{
+    public partial class EmptyArray<__TElement__>
+    {
+        public class Native : EmptyArray<__TElement__>
+        {
+            public Native(IOperationFactory m) : base(m) { }
+            public override Func<QVoid, IQArray<__TElement__>> __Body__ => (arg) =>
```
msoeken, line 14:
```suggestion
            public override Func<QVoid, IQArray<__TElement__>> __Body__ => _ =>
```

msoeken (approved):
Looks great to me. Thanks!

msoeken:
I was wondering what happens if we call `Default<'T[]>()` for some `'T`? Would that also return an empty array?

cgranade:
See https://github.com/microsoft/iqsharp/pull/325 for a discussion of build failures.

cgranade:
> I was wondering what happens if we call `Default<'T[]>()` for some `'T`? Would that also return an empty array?

At the moment, [there's a bug](https://github.com/microsoft/qsharp-compiler/issues/589) that causes `Default<'T[]>()` to return a null array (which isn't even a concept in Q#) when run against the simulation runtime. More generally, though, this new function could allow us to deprecate `Default` entirely in favor of using array functions like `EmptyArray`, `ConstantArray`, and so forth. See further discussion at https://github.com/microsoft/qsharp-compiler/issues/589.

PR #380, opened 2020-09-22
Improving the logging of WorkspaceClientException to include more actionable information

ricardo-espinoza:
Improving the logging of WorkspaceClientException to include more informative data in its message from its inner exception.

This is a follow up to #378 . 
 - Instead of providing the full stack of the exception in the output, we extract more relevant fields from the source RestErrorException and provide them to the user.

### DIFF
--- a/src/Azure/Azure.Quantum.Client/Exceptions/WorkspaceClientException.cs
+++ b/src/Azure/Azure.Quantum.Client/Exceptions/WorkspaceClientException.cs
@@ -89,8 +89,7 @@ namespace Microsoft.Azure.Quantum.Exceptions
                 formattedException += $"Server Error: {ex.Message}{Environment.NewLine}";
 
                 // Handle specific types of exceptions for additional data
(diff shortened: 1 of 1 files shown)


src/Azure/Azure.Quantum.Client/Exceptions/WorkspaceClientException.cs
```diff
@@ -67,9 +68,36 @@ public WorkspaceClientException()
...
+        private static string FormatInnerException(Exception ex)
+        {
+            string formattedException = string.Empty;
+            if (ex != null)
+            {
+                formattedException += $"Server Error: {ex.Message}{Environment.NewLine}";
+
+                // Handle specific types of exceptions for additional data
+                RestErrorException restErrorException = ex as RestErrorException;
+                if (restErrorException != null)
```
cesarzc, line 93:
```suggestion
                if (ex is RestErrorException restErrorException)
```

cesarzc (approved):

PR #383, opened 2020-09-22
Add Type3 targeting package.

swernli:
This package is almost identical to Type1, except it uses PreparePostMNoop.qs since Type3 targets don't need any special clean up instructions after measurement to allow qubit reuse, and has support for native uncontrolled SWAP gate.

This also adds a test case for native SWAP and fixes a bug in the Type2 native SWAP support.

This is part of the work to enable the feature described in #249.


swernli:
This work is out of date with the recent design updates, I'll pick it back up and update it once we get things merged to main.

PR #384, opened 2020-09-23
Dmitryv/consistent width depth

DmitryVasilevsky:
Creating compatible width-depth pair.
Use configuration parameter to optimize for width or for depth.
Optimizing for width may increase depth due to qubit reuse creating additional dependencies. Optimizing for width may increase width as qubits are not reused and fresh ones are always allocated. Added test.
Moved depth data to DepthCounterBase and stopped storing it in the qubits as data needs to be saved across release/allocate.

New simulator doesn't support the old way of getting data as multiple metric collectors are now supported each with its own data format. To get compatible pair among other data in the new framework the closest way is
            string csvSummary = sim.ToCSV()[MetricsCountersNames.depthCounter];
Alternatively one can use the following to get just width and depth
            sim.GetOperationMetric<TDepthWidth>("Width")
            sim.GetOperationMetric<TDepthWidth>("Depth")
Depth counter should be enabled and the mode of operation should be set (OptimizeDepth either true or false)


src/Simulation/QCTraceSimulator.Tests/Circuits/MetricsCalculationTests.qs
```diff
@@ -26,7 +26,35 @@ namespace Microsoft.Quantum.Simulation.QCTraceSimulatorRuntime.Tests {
...
+        //    O--O   
+        //    |  |   
+        //           
+        // Width 4, Depth 2.
+        // Alternatively q2 is reused for q4 (or q3)
+        // Width 3, Depth 3.
+
+        using (q1 = Qubit()) {
+            using (q2 = Qubit()) {
+                Controlled Z([q2], q1);
```
cgranade, line 47: Controlled Z is a Clifford operation; why is a non-zero 𝑇-depth reported for a sequence of Clifford operations?

src/Simulation/QCTraceSimulator.Tests/PrimitiveOperationsCounterTest.cs
```diff
@@ -120,6 +120,30 @@ void TDepthTest()
...
+            sim = RunDepthWidth(optimizeDepth: false);
+            Assert.Equal(3, sim.GetOperationMetric<TDepthWidth>("Width"));
+            Assert.Equal(3, sim.GetOperationMetric<TDepthWidth>("Depth"));
+        }
+
+        TracerSimulator RunDepthWidth(bool optimizeDepth) {
+            NewTracerConfiguration config = new NewTracerConfiguration();
+            config.UseDepthCounter = true;
+            config.OptimizeDepth = optimizeDepth;
+            config.TraceGateTimes = NewTraceGateTimes.ControlledZOnly;
```
cgranade, line 139:
```suggestion
            var config = new NewTracerConfiguration()
            {
                UseDepthCounter = true,
                OptimizeDepth = optimizeDepth,
                TraceGateTimes = NewTraceGateTimes.ControlledZOnly
            };
```

src/Simulation/QCTraceSimulator.Tests/PrimitiveOperationsCounterTest.cs
```diff
@@ -120,6 +120,30 @@ void TDepthTest()
...
+            TracerSimulator sim = new TracerSimulator(config);
+            QVoid res = TDepthWidth.Run(sim).Result;
```
cgranade, line 141: Since the result is insignificant, accessing `Result` is only being used for its side-effect (namely, waiting for completion of the Task). That can be made more explicit by calling `Wait`.
```suggestion
            TDepthWidth.Run(sim).Wait();
```

src/Simulation/Simulators/NewTracer/MetricCollectors/DepthCounterBase.cs
```diff
@@ -111,30 +145,33 @@ void IQubitTrackingTarget.OnReleaseQubits(IQArray<Qubit> qubits)
             this.CurrentState.ReleasedQubitsTime = System.Math.Max(this.CurrentState.ReleasedQubitsTime, releasedTime);
         }
 
-        //TODO: implement time storing as dictionary id->time an clear entries on release
-
         void IQubitTrackingTarget.OnAllocateQubits(IQArray<Qubit> qubits)
         {
+            foreach (Qubit qubit in qubits) {
+                this.MaxQubitId = System.Math.Max(this.MaxQubitId, qubit.Id);
+            }
```
cgranade, line 152:
```suggestion
            this.MaxQubitId = System.Math.Max(this.MaxQubitId, qubits.Max(qubit => qubit.Id));
```

src/Simulation/Simulators/NewTracer/MetricCollectors/DepthCounterBase.cs
```diff
@@ -111,30 +145,33 @@ void IQubitTrackingTarget.OnReleaseQubits(IQArray<Qubit> qubits)
...
         }
 
-        double IQubitTraceSubscriber<double>.NewTracingData(long id)
-        {
-            return 0.0d;
+        object? IQubitTraceSubscriber.NewTracingData(long id) {
+            // TODO: This function is not needed - this class doesn't need to store any data in qubits.
+            // However, infrastructure requires this function
+            return null;
         }
```
cgranade, line 163:
```suggestion
        object? IQubitTraceSubscriber.NewTracingData(long id)
        {
            // TODO: This method is not needed - this class doesn't need to store any data in qubits.
            // However, infrastructure requires this method
            return null;
        }
```

cgranade (commented):
I realize that this is still in draft, but I saw the notification go by and had a quick question about the test case added by this PR. Thanks for the help!

src/Simulation/QCTraceSimulator.Tests/Circuits/MetricsCalculationTests.qs
DmitryVasilevsky: Good question! The new tracer is configurable. For this test it is specifically configured to treat CZ as depth 1. I'll add a comment about this.

src/Simulation/QCTraceSimulator.Tests/PrimitiveOperationsCounterTest.cs
DmitryVasilevsky: Yes, this is copy-paste. I'll fix this.

src/Simulation/QCTraceSimulator.Tests/Circuits/MetricsCalculationTests.qs
DmitryVasilevsky: Updated comment.

src/Simulation/QCTraceSimulator.Tests/PrimitiveOperationsCounterTest.cs
DmitryVasilevsky: Updated code to use "simplified" initialization. Such initialization typically prevents line-by-line debugging and makes quoting portions of code harder, but this isn't a significant drawback here.
Explicit type is still used - easier to understand. When looking up variable, less time and mental effort required to figure out its type.

src/Simulation/QCTraceSimulator.Tests/PrimitiveOperationsCounterTest.cs
DmitryVasilevsky: Changed to Wait() and removed next line alltogether, it was not needed.

src/Simulation/Simulators/NewTracer/MetricCollectors/DepthCounterBase.cs
DmitryVasilevsky: Updated formatting.

src/Simulation/Simulators/NewTracer/MetricCollectors/DepthCounterBase.cs
DmitryVasilevsky: Updated formatting.
Loop makes it clear each qubit is touched. Lambda function adds unnecessary complexity to the code and hides computation complexity of the algorithm.

vadym-kl (approved):

avasch01 (approved):

bettinaheim:
Closing this PR for now, since we pulled the changes to main instead.

PR #385, opened 2020-09-23
Removed Bug2212 file, which has an invalid cycle in it.

ScottCarda-MS:
Removed a Q# file with an unsupported, invalid cycle in it.


bettinaheim (approved):
This scenario will no longer be supported and result in a compilation failure.

PR #387, opened 2020-09-28
Check to limit nr of threads for small problems

bettinaheim:


swernli (approved):
Changes seem good to me. Have we verified that these boundaries don't make the perf of small problems on Windows too much worse than it was before the original improvements get checked in?

src/Simulation/Native/src/external/fused.hpp
```diff
@@ -145,7 +145,10 @@ class Fused
             if (envNT == NULL) { // If the user didn't force the number of threads, make an intelligent guess
                 int nMaxThrds = std::thread::hardware_concurrency();        // Logical HW threads
                 if (nMaxThrds > 4) nMaxThrds/= 2;                           // Assume we have hyperthreading (no consistent/concise way to do this)
-                if (wfnCapacity < 1u << 20) {
+                if (wfnCapacity < 1ul << 14)      nMaxThrds = 1;
+                else if (wfnCapacity < 1ul << 16) nMaxThrds = 2;
```
rmshaffer, line 149: Minor note: looks like this line would actually _increase_ `nMaxThrds` from 1 to 2 if running on a single-core CPU?

dbwz8: No, but I haven't seen a 1 CPU machine in a long time. Easy to add the check if desired.

rmshaffer: Up to you. As you say, this is probably a rare case in practice.

PR #391, opened 2020-10-07
Revert "Performance improvement to QubitManager for Borrowing (#190)"

anpaz:
I'm reverting commit #190. This caused a huge drop performance for the runtime performance tests we run as part of the QDK end-to-end validation, specifically for Toffoli gates which is the only one that actually assign thousands of qubits. Speficially, the test that normally took about **800ms**, jumped to take normally **8.3s** and spikes of up to 50s, so at least a 10x perf drop. 
After reverting the change, the numbers look good again:
image


src/Simulation/Common/QubitManagerTrackingScope.cs
```diff
@@ -176,7 +176,7 @@ public virtual Qubit Borrow()
             return base.Borrow(1, ConstructExcludedQubitsArray())[0];
         }
 
-        public virtual IQArray<Qubit> Borrow(long numToBorrow)
```
DmitryVasilevsky, line 179: Keeping this virtual is important for exploration experiments and layout benchmarks.

msoeken (changes requested):
Looks good to me, but it would be good to modify the comment that changing loop direction can increase efficiency, and also add some more comments that we tried this change and why it did not work.

irinayat-MS:
@anpaz : Andres, I've effectively reverted #190 with #435. Could probably close this PR now?

bettinaheim:
Closing this PR since it is addressed with #435.

Issue #206, opened 2020-05-18
Simulator should allow release of measured Qubits


swernli:
** Please describe what you would like the feature to accomplish.**
The current simulator implementation requires that all qubits be reset to the `|0>` state before release, and treats it as a runtime error if this is not the case. The reset can be accomplished via an actual call to `Microsoft.Quantum.Intrinsic.Reset` or via other appropriate operations that put the qubit into the `|0>` state. The intent is to make sure no quantum state is "leaked" by releasing a qubit that may still be carrying information via superposition or entanglement. However, this goal can also be met by allowing release of a qubit that has been measured with regard to any single basis and then had no further operations performed on it.

** Describe the solution you'd like **
The simulator check for qubits on release should be updated to allow release of a qubit if it is `|0>` or if it has been measured with regard to any basis.

** Describe alternatives you've considered **
An alternative could be removing the simulator check for qubit state on release entirely. However, that would be undesirable given the benefit for algorithm validation that this check affords. Without any check, it would be too easy to write algorithms in Q# that did not correctly account for cleaning up qubits that might contain relevant program state.

** Additional context **
The following sample code demonstrates the behavior:
```
namespace Test {
    open Microsoft.Quantum.Intrinsic;

    @EntryPoint()
    operation Execute() : Result {
        using (q = Qubit()) {
            H(q);
            return M(q);
        }
    }
}
```
Because the `M` will result in collapsing the state to Zero about half the time, this code will execute successfully half of the time but fail with the following error the rest of the time:
```
Unhandled exception. Microsoft.Quantum.Simulation.Simulators.Exceptions.ReleasedQubitsAreNotInZeroState: Released qubits are not in zero state.
 ---> Test.Execute on C:\Users\swern\Programming\experimental\Wgate\Program.qs:line 8

Unhandled exception: Microsoft.Quantum.Simulation.Simulators.Exceptions.ReleasedQubitsAreNotInZeroState: Released qubits are not in zero state.
   at Microsoft.Quantum.Simulation.Simulators.QuantumSimulator.QSimQubitManager.ReleaseOneQubit(Qubit qubit, Boolean usedOnlyForBorrowing)
   at Microsoft.Quantum.Simulation.Common.QubitManager.Release(Qubit qubit)
   at Microsoft.Quantum.Simulation.Common.SimulatorBase.Release.Apply(Qubit q)
[...]

PR #393, opened 2020-10-08
Fix bug in logic for releasing qubits after joint measurement

swernli:
To allow release of qubits that have been collapsed by measurement, we mark them with `IsMeasured` equal to true (see #206). However, joint measurement does not collapse the states of the input qubits and might actually create additional entanglement, so those qubits should not be marked as valid for release.

This change fixes the logic by only marking the input to `Measure` when it is a single qubit, and adds a test that validates the exception that should be thrown when trying to release qubits after joint measurement.

### DIFF
--- a/src/Simulation/Simulators/QuantumSimulator/Measure.cs
+++ b/src/Simulation/Simulators/QuantumSimulator/Measure.cs
@@ -31,10 +31,11 @@ namespace Microsoft.Quantum.Simulation.Simulators
                 {
                     throw new InvalidOperationException($"Both input arrays for {this.GetType().Name} (paulis,qubits), must be of same size");
                 }
-                foreach (Qubit q in qubits)
+                if (qubits.Length == 1)
                 {
-                    //setting qubit as measured to allow for release
-                    q.IsMeasured = true;
+                    // When we are operating on a single qubit we will collapse the state, so mark
+                    // that qubit as measured.
+                    qubits[0].IsMeasured = true;
                 }
                 return Measure(Simulator.Id, (uint)paulis.Length, paulis.ToArray(), qubits.GetIds()).ToResult();
             };
--- a/src/Simulation/Simulators.Tests/Circuits/CoreOperations.qs
+++ b/src/Simulation/Simulators.Tests/Circuits/CoreOperations.qs
@@ -342,6 +342,7 @@ namespace Microsoft.Quantum.Tests.CoreOperations {
 
 // Using a different namespace, so tests can be auto-discovered.
 namespace Microsoft.Quantum.Simulation.Simulators.Tests.Circuits {
+    open Microsoft.Quantum.Simulation.Simulators.Tests.Circuits.Generics;
     open Microsoft.Quantum.Intrinsic;
     open Microsoft.Quantum.Convert;
     open Microsoft.Quantum.Diagnostics;
@@ -556,6 +557,14 @@ namespace Microsoft.Quantum.Simulation.Simulators.Tests.Circuits {
             X(q);
             let r = M(q);
             // Should not raise an exception
+        }
+    }
+
+    operation ReleaseMeasureMultipleQubitCheck() : Unit {
+        using (qs = Qubit[2]) {
+            ApplyToEach(H, qs);
+            let r = Measure([PauliZ, PauliZ], qs);
+            // Should raise an exception
         }
     }
 
--- a/src/Simulation/Simulators.Tests/QuantumSimulatorTests/QubitReleaseTest.cs
+++ b/src/Simulation/Simulators.Tests/QuantumSimulatorTests/QubitReleaseTest.cs
@@ -33,6 +33,15 @@ namespace Microsoft.Quantum.Simulation.Simulators.Tests
             await ReleaseMeasuredQubitCheck.Run(sim);
         }
 
+        //test to check that qubits cannot be released after multiple qubit measure
+        [Fact]
+        public async Task MeasuredMultipleQubitsReleaseTest()
+        {
+            var sim = new QuantumSimulator();
+
+            await Assert.ThrowsAsync<ReleasedQubitsAreNotInZeroState>(() => ReleaseMeasureMultipleQubitCheck.Run(sim));
+        }
+
         //test to check that qubit that is released and reallocated is in state |0>
         [Fact]
         public async Task ReallocateQubitInGroundStateTest()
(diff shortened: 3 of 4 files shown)


irinayat-MS:
What is the exact promise of the release-after-measurement? Is it just "best effort" on the simulator's side or a guarantee that if post-measurement the state isn't a superposition then it's OK to release? Because a joint measurement _might_ collapse a state to one of the basis vectors. For example, if you have a superposition like |00>+|10>, the joint ZZ measurement would collapse it to either |00> or |01>.

src/Simulation/Simulators/QuantumSimulator/Measure.cs
```diff
@@ -31,10 +31,11 @@ public QSimMeasure(QuantumSimulator m) : base(m)
                 {
                     throw new InvalidOperationException($"Both input arrays for {this.GetType().Name} (paulis,qubits), must be of same size");
                 }
-                foreach (Qubit q in qubits)
+                if (qubits.Length == 1)
```
ScottCarda-MS, line 34: Does it matter what basis the qubit is measured in? Do you want to release after _any_ single qubit measurement?

ScottCarda-MS (approved):

swernli:
The original motivation behind the exception at runtime was to allow the simulator to perform some validation of algorithms. The simulator will always collapse the state of a qubit that is being released (it actually internally performs the measurement to do so). But that may have a side effect on other qubits if the qubit that is reset is still entangled. With the original check, the runtime would throw if a qubit was released that is not exactly in the |0⟩ state, but we did the work to  relax that to allow release of any qubit that is either |0⟩ or measured, since measurement means it cannot be entangled. But even there, we didn't consider joint measurement, which would not guarantee the qubit is no longer entangled. With this change, I think we landed at what we originally wanted to do to address #206: qubits that we can reasonably conclude are no longer entangled and are in a well-known basis state are safe to release.

There is actually some good context for the decision and the approach in Aniket's PR, #231, so I'd recommend checking that out for some details from Chris and Andres.

src/Simulation/Simulators/QuantumSimulator/Measure.cs
swernli: Yeah, I this @cgranade put this best in the discussion for #231:

>> Releasing a Qubit measured in a different base, the qubit is still in super-position on the Z bases but it doesn't throw:
>>    ```
>>      using (q = Qubit()) {
>>          H(q);
>>          let _ = Measure([PauliX], [q]);
>>      }
>> ```
> Quick comment, but this should be entirely OK, as the qubit is still in a deterministically known state at the point that it is released, such that appropriate instructions can be automatically added to reset it to the |0⟩ state.

swernli:
I think I need to close and reopen this to get it the pipelines to rerun...

src/Simulation/Simulators/QuantumSimulator/Measure.cs
ScottCarda-MS: This has been answered is @swernli 's post below.

swernli:
Oh, never mind. The pipelines not running appears to be an ongoing issue:
https://developercommunity.visualstudio.com/content/problem/1211721/azp-run-does-not-find-associated-pipelines-for-pr.html

irinayat-MS:
> The original motivation behind the exception at runtime was to allow the simulator to perform some validation of algorithms. 

Do I understand it correctly, that the purpose of the feature is to allow the user omit explicit reset in the cases when the runtime can verifiably prove the qubit being released isn't entangled? I.e. it's _OK_ if the user still has to call Reset explicitly even if a joint measurement might have unentangled the qubit (|00>+|10>+|01> => |00> if ZZ measured Zero) but it's _a bug_ to allow the user omit reset when the qubit is actually entangled, right?

swernli:
@irinayat-MS Yes, that's correct, and good way to summarize it. The one addition I'd make is that we also allow for what I think of as a "logical reset" ie: calling `X` when the qubit is known to be in the |1〉 or `H` when it is known to be in the |+〉 state. I think this is implied by your use of just the word "reset" instead of the operation `Reset` but I figured it's worth calling out.

irinayat-MS (approved):
lgtm

cgranade:
> > The original motivation behind the exception at runtime was to allow the simulator to perform some validation of algorithms.
> 
> Do I understand it correctly, that the purpose of the feature is to allow the user omit explicit reset in the cases when the runtime can verifiably prove the qubit being released isn't entangled? I.e. it's _OK_ if the user still has to call Reset explicitly even if a joint measurement might have unentangled the qubit (|00>+|10>+|01> => |00> if ZZ measured Zero) but it's _a bug_ to allow the user omit reset when the qubit is actually entangled, right?

Apologies for re-opening the thread, but I wanted to provide a little bit of clarification if I could. It's entirely possible for two qubits to be entangled with each other, but yet still in a fully deterministic state that allows the compiler to efficiently and automatically reset them. In particular, consider the following snippet:

```Q#
using ((left, right) = (Qubit(), Qubit()) {
    let xx = Measure([PauliX, PauliX], [left, right]);
    let zz = Measure([PauliZ, PauliZ], [left, right]);
}
```

At the point of deallocation, the state of `[left, right]` is guaranteed to be one of the four Bell pairs, conditioned on the values of `xx` and `zz`. Using the stabilizer formalism, it's efficient to compute the sequence of operations needed to unprepare each of the four states.

While I don't suggest that we necessarily support that scenario, I find it helpful in separating out the requirement that qubits be unentangled from the requirement that the compiler can efficiently and automatically unprepare a register into the |00…0⟩ state so that they're ready for the next allocation.

bettinaheim:
@ScottCarda-MS @swernli Looks like monomorphization is throwing on this one; is this fixed or are you on that?

ScottCarda-MS:
> @ScottCarda-MS @swernli Looks like monomorphization is throwing on this one; is this fixed or are you on that? I just updated the branch, respawning the build.

Looks like it is an instance of the issue on compiler repo: https://github.com/microsoft/qsharp-compiler/issues/689
There is a PR up for that on the same repo: https://github.com/microsoft/qsharp-compiler/pull/694
Once that is in, it should fix the issue experienced here.

ScottCarda-MS:
> > @ScottCarda-MS @swernli Looks like monomorphization is throwing on this one; is this fixed or are you on that? I just updated the branch, respawning the build.
> 
> Looks like it is an instance of the issue on compiler repo: [microsoft/qsharp-compiler#689](https://github.com/microsoft/qsharp-compiler/issues/689)
> There is a PR up for that on the same repo: [microsoft/qsharp-compiler#694](https://github.com/microsoft/qsharp-compiler/pull/694)
> Once that is in, it should fix the issue experienced here.

The PR has been merged in, so once you merged in from main, the issue should be resolved.

swernli:
Thanks @bettinaheim and @ScottCarda-MS for keeping an eye on this and getting it merged!

PR #394, opened 2020-10-08
QIR runtime initial commit

irinayat-MS:
I'm looking into how to enable CIs but outside of that the code is ready for review. Please refer to readme.md in QirRuntime folder for more details on the structure of the project, how to build/test, etc.

@bettinaheim could you please add other reviewers as appropriate?


src/QirRuntime/README.md
```diff
@@ -0,0 +1,115 @@
...
+
+*Building from Visual Studio and VS Code is **not** supported.
+Running cmake from the editors will likely default to MSVC or clang-cl and fail.*
+
+### Linux via WSL pre-reqs
+
+1. On the host Windows machine install WLS.
+2. In WSL install clang/llvm and the other tools:
+    1. NB: the default version of clang/llvm on WSL's Ubuntu are likely to be too old, it's recommended you get the latest packages.
+    2. $ sudo apt install clang
```
swernli, line 48: Thanks for the detailed instructions! I'm starting off my review by trying out the local compile and test. Just as a heads up, I found that using the Ubuntu 20.04 image meant I also had to install cmake and clang-tidy, so it might be worth calling those out here.

src/QirRuntime/experimental/IonqTranslator.cpp
```diff
@@ -0,0 +1,665 @@
...
+        will call back for each step, even if it might come back to it (R1Frac -> R1).
+    ==============================================================================*/
+    class CCanonicalDecomposer final : public CQuantumApiBase
+    {
+        // The decomposer assumes that owner's lifetime exceeds its own.
+        IQuantumApi* owner = nullptr;
+
+        static double CalculateAngle(long numerator, long power)
+        {
+            return -numerator * M_PI / std::pow(2.0, power - 1);
```
swernli, line 181: Looks like this produces a warning-as-error on the newer clang-tidy version:
`error: narrowing conversion from 'long' to 'double' [bugprone-narrowing-conversions,-warnings-as-errors]`. The same error crops up on line 226 as well.

src/QirRuntime/README.md
```diff
@@ -0,0 +1,115 @@
...
+To execute all tests, including QIR and managed interop (the interop is Windows only) run `test.py` from the project's root folder:
+
+- (Windows) `python test.py [nobuild] [debug/release]`
+- (Linux) `python3 test.py [nobuild] [debug/release]`
+
+The script will trigger incremental build unless `nobuild` options is specified.
+
+*NB*: the test script currently doesn't detect build failures, so in case of build failure it will run tests against the previously built binaries.
+
+All test binaries and their dependencies are copied by the build into *install* folder: `build/[Windows|Linux]/[Debug|Release]/bin` and should be run from there (otherwise the tests might fail to load the shared libraries they depend on). On **Linux** `test.py` adds the install folder to LD_LIBRARY_PATH for the duration of the script. If you'd like to run the tests directly, add the path for the session manually. On WLS it might look like this: `$export LD_LIBRARY_PATH=/mnt/d/repos/llvm-project/qtracer/build/Linux/Debug/bin:${LD_LIBRARY_PATH}`.
```
swernli, line 68:
```suggestion
All test binaries and their dependencies are copied by the build into *install* folder: `build/[Windows|Linux]/[Debug|Release]/bin` and should be run from there (otherwise the tests might fail to load the shared libraries they depend on). On **Linux** `test.py` adds the install folder to LD_LIBRARY_PATH for the duration of the script. If you'd like to run the tests directly, add the path for the session manually. On WSL it might look like this: `$export LD_LIBRARY_PATH=/mnt/d/repos/llvm-project/qtracer/build/Linux/Debug/bin:${LD_LIBRARY_PATH}`.
```

src/QirRuntime/public/ChunkedStore.hpp
```diff
@@ -0,0 +1,129 @@
...
+#include <vector>
+
+#include "BitStates.hpp"
+
+namespace quantum
+{
+    /*==============================================================================
+        Provides dynamically extendable storage for TItem_WithId (usually, qubits)
+        that never moves previously allocated items, meaning the address of the item
+        can be safely used to access it untill it's released.
```
swernli, line 13:
```suggestion
        can be safely used to access it until it's released.
```

src/QirRuntime/public/IQuantumApi.hpp
```diff
@@ -0,0 +1,143 @@
...
+
+namespace quantum
+{
+    struct IQuantumApi
+    {
+        virtual ~IQuantumApi() {}
+
+        // TODO: Only some simulators 'know' the state -- is it OK to have this method on IQuantumApi?
+        // The callback will be invoked on each basis vector (in the standard computational basis) in little-endian
+        // order untill it returns `false` or the state is fully dumped.
```
swernli, line 15:
```suggestion
        // order until it returns `false` or the state is fully dumped.
```

irinayat-MS:
Too much overhead to collect feedback on the initial implementation of the QIR Runtime. Will switch from feature branch to a personal branch.

src/QirRuntime/experimental/IonqTranslator.cpp
irinayat-MS: Changed `long power` to `int power` throughout the interfaces because we don't expect the clients to need super-large power values and providing support for them isn't trivial.

PR #396, opened 2020-10-09
Initial implementation of QIR Runtime

irinayat-MS:
Please see `readme.md` in `src/QirRuntime` folder for details on the structure of the project and how to build/test it. 

Using a non-feature branch for this review to allow for quick turnaround on PR feedback. When done with the PR we can switch the branch to become a feature branch, subject to the corresponding policies.

@bettinaheim, please add other reviewers as appropriate.


src/QirRuntime/public/QuantumApiBase.hpp
```diff
@@ -0,0 +1,260 @@
...
+        {
+            throw std::logic_error("not_implemented");
+        }
+        virtual std::string DumpQubit(Qubit qubit) override
+        {
+            throw std::logic_error("not_implemented");
+        }
+
+        // Shortcuts
+        void CX(Qubit control, Qubit target) override
```
swernli, line 31: I'm still a little confused why these shortcut methods exist, and why they are decomposed here instead of in some other place. Long term, I think we'd want to see all decompositions removed in favor of pure Q# decomposition in the targeting package (see my other comment). For now, I think I would be more comfortable with these if they were moved to the `CCanonicalDecomposer` class so that all the decompositions are in the same place and don't start spreading out. That way, this base class has purely "not implemented" logic_error versions of every function and be a truly common base that can be used by any later implementation.

swernli (approved):
There are a few elements that I have questions on and/or would like to have follow up conversation regarding, but I really like the structure and design overall. Thanks for all this work; I'm exciting to see it get in and continue to develop/iterate!

swernli:
One more idea I forgot to include in the review: might it be worth renaming the code in the experimental folder to something other than Ionq*? I know that was the inspiration, but especially if we decide to align that with the new target packages, it can just be a sample of a generic target implementation and does not have to be tied to a specific provider. Food for thought, and not something that I imagine has to be decided on/address as part of this PR.

src/QirRuntime/lib/QIR/utils.cpp
```diff
@@ -0,0 +1,54 @@
...
+    void quantum__rt__fail(QirString* msg) // NOLINT
+    {
+        std::string str = msg->str;
+        quantum__rt__string_unreference(msg);
+        throw std::runtime_error(str);
+    }
+
+    // Include the given message in the computation's execution log or equivalent.
+    // TODO: should we allow the user to register their own output?
+    void quantum__rt__message(QirString* msg) // NOLINT
```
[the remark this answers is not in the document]
irinayat-MS: Yes, a very valid point and I was on the fence about it. The RT defines all APIs in a target-agnostic way. It happens, that when targeting simulators, we delegate some of the APIs to the simulators (e.g. qubit allocation) because it makes sense for these targets. We could pass message call to the simulator as well. However, we don't have to (as this implementation shows) and I'd like to keep the IQuantumApi as thin as possible. On the other hand, I agree that dumping to std::cout might not be suitable for all situations... We should probably have a notion of execution context that incorporates the target simulator and configurations for things like message/fail (as I'm also not convinced that throwing from `fail` is always the correct choice). I'd suggest we leave this implementation in for now and review it later when we have better understanding of the ways QIR is produced/consumed and the requirements that places on the runtime.

src/QirRuntime/experimental/IonqTranslator.cpp
```diff
@@ -0,0 +1,664 @@
...
+                }
+                *err << "]}";
+            }
+        }
+    };
+
+    /*==============================================================================
+        CIonqSimulator
+    ==============================================================================*/
+    class CIonqSimulator final : public CQuantumApiBase
```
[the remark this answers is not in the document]
irinayat-MS: Notice, that we are hiding the concrete type of the simulator from the clients: the factory methods return pointers to single interface and to allow the client access to the other interface we'd have to either implement some kind of COM-like QI or require the clients to assemble the simulator from the parts they create themselves. I've chosen the latter per "prefer composition over inheritance" principle. The part I don't like about this approach is that one cannot actually mix-and-match translators and simulators. At least, not the ones we provide right now -- the simulators _assume_ the ITranslator given to them has particular runtime type and statically casts to it. Oh, I guess, there is an approach #3 when the client _knows_ which `IQuantumApi`s are also `ITranslator`s and does the cast on their side, but it would be even less safe than what we have right now. Also, please keep in mind that the design has to support "consuming the runtime via shared library" scenario which limits us to C-style signatures for QAPI_* functions.

I'm not 100% happy with the current design and open to reviewing/changing it, provided we can meet the requirements outlined above.

src/QirRuntime/experimental/IonqTranslator.cpp
```diff
@@ -0,0 +1,664 @@
...
+        unitaries without any hardware specific optimizations.
+
+        A quantum processor, that wants to avail of the decompositions, should
+        implement the primitive operations, pass itself to the decomposer and then
+        delegate to it calls that should be decomposed.
+
+        NB: because the owner opts-in for each decomposition separately the decomposer
+        will call back for each step, even if it might come back to it (R1Frac -> R1).
+    ==============================================================================*/
+    class CCanonicalDecomposer final : public CQuantumApiBase
```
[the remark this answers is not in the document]
irinayat-MS: I'd like to keep this example for now as it demonstrates how IQuantumApi can be used to layer/reuse partial implementations. Given it's in experimental folder and we are not planning to provide a "real" IonQ target here any time soon I'm not worried it will be in the way of the new patterns for decomposition.

Per your other comment, I'll rename the simulator to avoid IonQ reference and make it clear it's just an exercise in using the layering + ITranslator.

src/QirRuntime/experimental/TracingSimulator.cpp
```diff
@@ -0,0 +1,364 @@
...
+                this->resourceTranslator->stats.cCircuitDepth = -1;
+            }
+        }
+
+        void X(Qubit qubit) override
+        {
+            UpdateQubitsAndCircuitDepth(qubit);
+            this->resourceTranslator->stats.cX++;
+        }
+        void ControlledX(long cControls, Qubit* controls, Qubit target) override
```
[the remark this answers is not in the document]
irinayat-MS: Yes, this is not a very useful resource tracer :) 
For a real one I think we should provide a way to configure which gates to count...

src/QirRuntime/public/QuantumApiBase.hpp
irinayat-MS: I've been asked to follow the incubation IQuantumApi as close as possible and it happened to differentiate between singly versus multi-controlled operations. @avasch01 would have more context.

irinayat-MS:
> There are a few elements that I have questions on and/or would like to have follow up conversation regarding, but I really like the structure and design overall. Thanks for all this work; I'm exciting to see it get in and continue to develop/iterate!

Thank you for the review! Please feel free to reactivate any of the comments if you'd like to follow up on any of them. For design decisions without immediate solutions, I can either add a corresponding section to the readme or, maybe, log issues?

src/QirRuntime/public/QuantumApiBase.hpp
swernli: I think the API itself is good (especially considering I can already see think of a few places where specific targets would want to use these over the multi-controlled version), I just disagree with having the decomposition live in this particular base class. I'm just in favor of having a base class where every method is "not implemented" and leaving the implementation of any supported methods to the child classes. But this is not a big deal for this PR and the pattern of where decompositions live and how they imply what implementations are present and native to the target is something that we will need to do more review and design on later anyway.

For context, over in the feature/decomp branch I ended up needing to define an internal, singly-controlled variant for `X` and `Z` as well as versions that did not support any controls (see [ApplyControlledX](https://github.com/microsoft/qsharp-runtime/blob/feature/decomp/src/Simulation/TargetDefinitions/Intrinsic/ApplyControlledX.qs) as an example). I would've called this `CX` except that this is already taken by a library operation, [Microsoft.Quantum.Canon.CX](https://github.com/microsoft/QuantumLibraries/blob/b802b254b7cb98049573a367ed399e23ad784e59/Standard/src/Canon/CommonGates.qs#L38). As we figure out how to integrate the `IQuantumAPI` with the target packages and the runtime, it may mean moving some of these callables out of the libraries and into the runtime to allow them to be treated as target-overridable intrinsics and redefining the operations I put into these first target packages to make them line up with the simulator API (possibly making them public instead of internal in the process). But all that is a much bigger cross-repo change that we would need to coordinate later. Suffice it to say, I think the API design is good, I just think it implies the need for this larger design update and refactor at some point in the future :).

swernli:
> For design decisions without immediate solutions, I can either add a corresponding section to the readme or, maybe, log issues?
That's a good question... I think logging issues for design discussion/integration work is probably the way to go, though they will be a little tricky to capture as an issue since they are not a bug with current functionality nor a discrete feature request for new behavior in the product. I leave it to your judgement how you want to summarize for the issue, but the main thing I know I'm interested in tracking is the opportunities for continued revision of this API along with the target packages and the code generation @alan-geller is working on. Each of those pieces is trying to define a view of the same concept: what are the native capabilities of this target (hardware or simulator) and how does Q# call into them? Getting those to line up well will make all of this work nicely together, I'm just not sure yet which adjustments need to be made to which pieces yet!

dbwz8 (commented):
I did a (very) quick pass through all the C++ files (50). I didn't see any obvious red flags. Everything looks consistent and clean. There is some disparity in the amount of commenting (really heavy in some places and non-existent in others). It might be good to "bulk up" some of the files that are light on comments. I also noted (at least) one place where memcpy is being used. Is that still the best/safest call to make to manipulate memory (yes, I know it's the fastest ;))?

All in all, very nice code.

dbwz8 (approved):
Given my comments, I see no reason not to Approve

src/QirRuntime/experimental/CMakeLists.txt
```diff
@@ -4,8 +4,8 @@
 add_executable(experimental-unittests
   driver.cpp
   IntervalsColoringTests.cpp
-  IonqTranslator.cpp
-  IonqTranslatorTests.cpp
+  CircuitToJasonPrinter.cpp
```
swernli, line 7: nitpick: Jason is usually just for a person's name, while JSON or Json would be used for the format.

src/QirRuntime/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+The project is using catch2 for all native tests, including QIR. `<test_binary> -help` provides details on how to run a subset of the tests and other options.
+
+All native tests are fully integrated with CTest. The coverage is the same as when using `test.py` or running the test binaries individually, but the results are logged into the corresponding `build/[Windows|Linux]/[Debug|Release]/bin/<test_binary_name>_results.xml` file. To trigger tests this way, navigate into `build/[Windows|Linux]/[Debug|Release]` folder and run `ctest`. No configuration options required.
+
+## QIR Bridge and Runtime
+
+This project contains implementation of the QIR runtime per the specifications at [https://github.com/microsoft/qsharp-language/tree/main/Specifications/QIR] and the translation layer between the QIR and the IR, generated by Clang from the native code. We call the translation layer "QIR Bridge". This project also provides an optional implementation of intrinsics from `__quantum__qis*` namespace that are used in QIR, generated from Q#.
```
bettinaheim, line 85:
```suggestion
This project contains an implementation of the QIR runtime per the specifications at [https://github.com/microsoft/qsharp-language/tree/main/Specifications/QIR] and the translation layer between the QIR and the IR, generated by Clang from the native code. We call the translation layer "QIR Bridge". This project also provides an optional implementation of intrinsics from `__quantum__qis*` namespace that are used in QIR, generated from Q#.
```

src/QirRuntime/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+There are two ways to compile and run the QIR files against the runtime.
+
+1. Link against the runtime libraries *statically*. For the example of this approach see `test/QIR-static` tests. It allows the client to access the target simulator directly, if so desired.
+1. Link against the *shared qdk* library. The example of this approach can be found in `test/QIR-dynamic` folder. In the future we'll provide fully self-contained packages of the runtime to enable this workflow completely outside of the current project.
+
+QIR's architecture assumes a single target, whether that be hardware or a particular simulator. As a result, there is no provision in the QIR specifications to choose a target dinamically. To connect QIR to the simulators from this runtime, we provide `SetCurrentQuantumApiForQIR(IQuantumApi*)` method. When linking against shared qdk, the method will be invoked automatically when creating an execution context. If linking statically, the client can call the method directly with the simulator of their choice. Switching simulators while executing QIR isn't supported, however, and would yield undefined behaviour.
```
bettinaheim, line 94:
```suggestion
QIR's architecture assumes a single target, whether that be hardware or a particular simulator. As a result, there is no provision in the QIR specifications to choose a target dinamically. To connect QIR to the simulators from this runtime, we provide `SetCurrentQuantumApiForQIR(IQuantumApi*)` method. When linking against the shared qdk library, the method will be invoked automatically when creating an execution context. If linking statically, the client can call the method directly with the simulator of their choice. Switching simulators while executing QIR isn't supported, however, and would yield undefined behavior.
```

src/QirRuntime/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+### Building from IR files
+CMake doesn't support using LLVM's IR files as input so instead we invoke Clang directly from custom commands to create utility libs that can be linked into other targets using their absolute paths.
+
+*NB*: Compiling from IR has fewer checks than compiling from C++. For example, IR doesn't support overloading so declarations
+and definitions of functions are matched by name, without taking in account the arguments. This means that a build might
+succeed with mismatched signatures between caller/callee which will likely lead to crashes and other bugs at runtime.
+
+**QIR runtime is work in progress. Current known limitations are as follows:**
```
bettinaheim, line 103:
```suggestion
**The QIR runtime is work in progress. Current known limitations are as follows:**
```

src/QirRuntime/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+QIR's architecture assumes a single target, whether that be hardware or a particular simulator. As a result, there is no provision in the QIR specifications to choose a target dinamically. To connect QIR to the simulators from this runtime, we provide `SetCurrentQuantumApiForQIR(IQuantumApi*)` method. When linking against shared qdk, the method will be invoked automatically when creating an execution context. If linking statically, the client can call the method directly with the simulator of their choice. Switching simulators while executing QIR isn't supported, however, and would yield undefined behaviour.
+
+### Building from IR files
+CMake doesn't support using LLVM's IR files as input so instead we invoke Clang directly from custom commands to create utility libs that can be linked into other targets using their absolute paths.
+
+*NB*: Compiling from IR has fewer checks than compiling from C++. For example, IR doesn't support overloading so declarations
+and definitions of functions are matched by name, without taking in account the arguments. This means that a build might
```
bettinaheim, line 100:
```suggestion
and definitions of functions are matched by name, without taking into account the arguments. This means that a build might
```

src/QirRuntime/README.md
```diff
@@ -0,0 +1,127 @@
...
+succeed with mismatched signatures between caller/callee which will likely lead to crashes and other bugs at runtime.
+
+**QIR runtime is work in progress. Current known limitations are as follows:**
+
+1. All functionality related to BigInt type (including `__quantum__rt__bigint_to_string`) NYI.
+2. QIR is assumed to be single threaded. No effort was made to make the bridge and runtime thread safe.
+3. Strings are implemented as a thin wrapper over std::string with virtually no optimizations.
+4. `__quantum__rt__string_create` currently doesn't conform to the spec (it expects a null terminated string rather than a string of specified length).
+5. Variadic functions (e.g. `__quantum__rt__array_create`) require platform specific bridges. Currently implemented bridge is for Windows.
```
bettinaheim, line 109:
```suggestion
5. Variadic functions (e.g. `__quantum__rt__array_create`) require platform specific bridges. The currently implemented bridge is for Windows.
```

src/QirRuntime/README.md
```diff
@@ -0,0 +1,127 @@
...
+3. cd build
+4. cmake -G Ninja -DCMAKE_BUILD_TYPE=Release ..
+5. cmake --build .
+
+Or you can run `build.py` script from the root of the project. The default options for the script are `make debug`.
+
+- (Windows) `python build.py [make/nomake] [debug|release] [ir]`
+- (Linux) `python3 build.py [make/nomake] [debug|release] [ir]`
+
+The scrip will place the build artefacts into `build/[Windows|Linux]/[Debug|Release]` folder.
```
cesarzc, line 28:
```suggestion
The script will place the build artifacts into `build/[Windows|Linux]/[Debug|Release]` folder.
```

src/QirRuntime/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+### Windows pre-reqs
+
+1. Install Clang, Ninja and CMake from the public distros.
```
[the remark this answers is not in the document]
irinayat-MS: Those are public distros we don't maintain so their location might change. User can easily discover them online (the user might also have their preferred packaging apps, etc.)

src/QirRuntime/README.md
```diff
@@ -0,0 +1,127 @@
...
+2. Add all three to your/system `%PATH%`.
+3. Install VS 2019 (Clang uses MSVC's standard library on Windows).
```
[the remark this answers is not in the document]
irinayat-MS: The native simulator requires VS's toolchain as well (it's built with msvc compiler) so this is not a new requirement for working with the repo as a whole. I don't know whether it's possible to install the toolchain without installing VS (from a quick search online, I believe the answer is "no").

src/QirRuntime/README.md
```diff
@@ -0,0 +1,127 @@
...
+4. Install .NET
```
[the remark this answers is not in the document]
irinayat-MS: Clarified: ".NET Core 3.1". As with other prerequisites I don't want to maintain the links to public distros as those are external to us and might change.

src/QirRuntime/README.md
```diff
@@ -0,0 +1,127 @@
...
+4. <_optional_>
+    1. Either clone LLVM's repo and build LLVM locally to produce your own `opt.exe` and other IR tools or download LLVM's dev package.
+5. Install clang-tidy and clang-format if your Clang/LLVM packages didn't include the tools.
```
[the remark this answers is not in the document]
irinayat-MS: The releases of llvm from https://releases.llvm.org/ include both of these so the user, who does the mainstream setup, shouldn't need to do anything here. And if the user is sophisticated enough to install LLVM/Clang from somewhere else, I think it's OK to expect them to figure out the additional tools on their own.

src/QirRuntime/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+See [https://code.visualstudio.com/docs/remote/wsl] on how to use VS Code with WSL.
+
+### Generating IR
+
+To generate IR of the native components as part of the build:
+
+1. Install LLVM (see above) so you have opt.exe and other tools that work with IR.
+2. Modify ProduceIR.cmake to point to these tools. (Or copy them into a separate folder and add that folder to `%PATH%.` If you are building LLVM locally, don't add the whole target build folder to `%PATH%` because it would likely confuse Clang.)
+3. To produce IR once pass `ir` option to `build.py`. To do that on every build, uncomment setting `${GENERATE_IR}` variable in the root `CMakeLists.txt`.
```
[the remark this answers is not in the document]
irinayat-MS: Clang. This is the IR from our native code. I thought we would need it for optimizations and occasionally I used it to see what Clang does with native code when developing the bridge. We might remove this later but it was such a pain to enable it that I'd rather keep it in until dead sure it's not being used

src/QirRuntime/lib/QIR/callables.cpp
```diff
@@ -0,0 +1,189 @@
...
+    %3 = call %Callable* @__quantum__rt__callable_create(
+        [4 x void (%TupleHeader*, %TupleHeader*, %TupleHeader*)*]* @UpdateAnsatz,
+        %TupleHeader* null)
+
+==============================================================================*/
+typedef void (*t_CallableEntry)(QirTupleHeader*, QirTupleHeader*, QirTupleHeader*);
+struct QirCallable
+{
+    static int constexpr Adjoint = 1;
+    static int constexpr Controlled = 1 << 1;
```
cesarzc, line 57: These ints seem to be used to apply them to the `appliedFunctor` mask.

Maybe these pattern could be implemented in a slightly safer way by creating an enum that represents functors and inside the `ApplyFunctor` method do the logic to transform from enum value to bitmask.

The caveat is that the max enum value cannot be larger that the number of bits in an `int` but that can be ensured through an static assert.

cesarzc: Another advantage of this approach would be that the following runtime assert in `ApplyFunctor` wouldn't be necessary:
`assert(functor == QirCallable::Adjoint || functor == QirCallable::Controlled)`

src/QirRuntime/lib/QIR/callables.cpp
```diff
@@ -0,0 +1,189 @@
...
+    void Invoke(QirTupleHeader* args, QirTupleHeader* result)
+    {
+        assert(this->appliedFunctor < QirCallable::TableSize);
+        this->function_table[this->appliedFunctor](capture, args, result);
+    }
+
+    void ApplyFunctor(int functor)
+    {
+        assert(functor == QirCallable::Adjoint || functor == QirCallable::Controlled);
+        this->appliedFunctor ^= functor;
```
cesarzc, line 117: Is it intended that the functor is cleared if when trying to apply it is already present?

If it is, a comment would be great since my expectation would when applying would be to make sure that the functor bit is present regardless if it was already set before invoking this method.

If not, the more appropriate operator would be `|=`.

alan-geller: The correct behavior, I think, is to clear the functor for Adjoint, but not for Controlled -- Adjoint Adjoint op is the same as op, but Controlled Controlled op is not the same as op, it's the same as Controlled op except that it requires some additional argument munging.

src/QirRuntime/lib/TracingContext/TracingContext.cpp
```diff
@@ -0,0 +1,103 @@
...
+    assert(context == nullptr);
+
+    std::ostringstream os;
+    g_stats.Print(os);
+    std::cout << os.str();
+    g_stats = {0};
+}
+
+// Qubit management
+EXPORTQAPI Qubit QAPI_AllocateQubit(QuantumExecutionContext* context) // NOLINT
```
cesarzc, line 22: For parameters that are not referenced inside functions, if available it would be great to use a macro like `UNREFERENCED_PARAMETER` if available to express intent.

src/QirRuntime/experimental/CircuitToJsonPrinter.cpp
```diff
@@ -0,0 +1,659 @@
...
+
+#include "ExperimentalSimFactory.hpp"
+#include "ITranslator.hpp"
+
+#include "BitStates.hpp"
+#include "QuantumApiBase.hpp"
+
+using namespace std;
+
+namespace quantum
```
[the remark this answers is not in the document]
irinayat-MS: I've changed the coding style rule for namespaces from lower_case to CamelCase and will nest `Quantum` namespace in `Microsoft` namespace, but this is an example of feedback that better to be provided as early as possible as the change affects almost every file in the project, including formatting... :(

src/QirRuntime/experimental/CircuitToJsonPrinter.cpp
```diff
@@ -0,0 +1,659 @@
...
+                {
+                    error.Print(*err);
+                }
+                *err << "]}";
+            }
+        }
+    };
+
+    /*==============================================================================
+        CCircuitPrintingSimulator
```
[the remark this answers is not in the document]
irinayat-MS: It's a common readability practice in native code bases.

src/QirRuntime/experimental/CircuitToJsonPrinter.cpp
```diff
@@ -0,0 +1,659 @@
...
+        unitaries without any hardware specific optimizations.
+
+        A quantum processor, that wants to avail of the decompositions, should
+        implement the primitive operations, pass itself to the decomposer and then
+        delegate to it calls that should be decomposed.
+
+        NB: because the owner opts-in for each decomposition separately the decomposer
+        will call back for each step, even if it might come back to it (R1Frac -> R1).
+    ==============================================================================*/
+    class CCanonicalDecomposer final : public CQuantumApiBase
```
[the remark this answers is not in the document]
irinayat-MS: The class demonstrates a particular technique. I've clarified the comments.

src/QirRuntime/lib/QIR/callables.cpp
irinayat-MS: Given that the values are local to the class and not exposed outside of it, I don't believe the increased safety of the two calls to ApplyFunctor outweighs the hassle of converting enums into bitfields.

src/QirRuntime/lib/QIR/callables.cpp
irinayat-MS: Yes: A+A = I, A + C = C+ A = CA,  CA + A = C, but I haven't thought through the C + C case. Thank you for catching it!

irinayat-MS: @alan-geller, I've fixed the `make_controlled` functor to behave as C + C = C but I don't think the runtime should be messing with the arguments as the tuple types are opaque to the runtime. Could you please take a look and either update the spec with what kind of argument munging is expected from the runtime or change the QIR generation code so I could enable the corresponding tests? Thanks!

alan-geller: I'll specify the behavior in the documentation. Unfortunately the argument munging can't be avoided, but I think all that is necessary is to know the length of the tuple (and possibly not event that, I don't remember).

irinayat-MS: But why the combining of the controls cannot happen during QIR generation? At the moment QIR creates the nested tuples, why cannot it flatten the controls rather than nesting them?

anpaz:
> CodeFlow for GitHub does not download executable files. If you trust this file and would like to run it, please download it yourself through GitHub.

these files should not be committed and should be generated as part of the build.
if you do want to keep these files please add a README.md explaining why they are here and how to update them.

---
Refers to: src/QirRuntime/externals/QuantumSimulator/libMicrosoft.Quantum.Simulator.Runtime.dylib:1 in 7eb68d5. [](commit_id = 7eb68d53fefc50b49b68fa7b8544eb1ec6a3d08d, deletion_comment = False)

build/build.ps1
```diff
@@ -8,14 +8,44 @@ $all_ok = $True
 
 if ($Env:ENABLE_NATIVE -ne "false") {
     Write-Host "##[info]Build Native simulator"
```
anpaz, line 10: All this block should go into its own `build-native.ps1` so we can also use it in the e2e build.

anpaz: More specifically, what I would do is break this into two `build.ps1` script that go into the `src\Simulation\Native` and `src\QirRuntime` accordingly, so you can call them separately when working on either project independently.
Then I would create a `build\build-native.ps1` script that gets called by this `build\build.ps1` which the only thing it does is called the other two. That way it will be easier to integrate with the e2e builds.

src/QirRuntime/build.py
```diff
@@ -0,0 +1,95 @@
+import sys, os, platform, subprocess, datetime
```
anpaz, line 1: Is this still needed? Looks like the logic has been incorporated to `build.ps1`

anpaz (changes requested):

bootstrap.ps1
```diff
@@ -11,6 +11,20 @@ Push-Location (Join-Path $PSScriptRoot "src/Simulation/Simulators")
     .\FindNuspecReferences.ps1
 Pop-Location
 
+# install prereqs for native builds
+if ($Env:ENABLE_NATIVE -ne "false") {
+    if (-not (Test-Path Env:AGENT_OS) -or ($Env:AGENT_OS.StartsWith("Win"))) {
+        choco install llvm
```
anpaz, line 17: these pre-reqs should be listed in the global README.md, or a point in the README about how to skip building QIR

build/build.ps1
```diff
@@ -8,14 +8,44 @@ $all_ok = $True
...
         $script:all_ok = $False
     }
+
+    Write-Host "##[info]Build QIR Runtime"
+    $oldCC = $env:CC
+    $oldCXX = $env:CC
+    if (-not (Test-Path Env:AGENT_OS) -or ($Env:AGENT_OS.StartsWith("Win"))) {
+        $env:CC = "C:\Program Files\LLVM\bin\clang.exe"
+        $env:CXX = "C:\Program Files\LLVM\bin\clang++.exe"
+        $llvmExtras = (Join-Path $PSScriptRoot "../externals/LLVM")
```
anpaz, line 24: this path doesn't seem to be right.

build/build.ps1
```diff
@@ -8,14 +8,44 @@ $all_ok = $True
...
     if ($LastExitCode -ne 0) {
         Write-Host "##vso[task.logissue type=error;]Failed to build Native simulator."
         $script:all_ok = $False
     }
+
+    Write-Host "##[info]Build QIR Runtime"
+    $oldCC = $env:CC
+    $oldCXX = $env:CC
+    if (-not (Test-Path Env:AGENT_OS) -or ($Env:AGENT_OS.StartsWith("Win"))) {
+        $env:CC = "C:\Program Files\LLVM\bin\clang.exe"
```
anpaz, line 22: how does this normally get installed? Should be listed in README.md

build/build.ps1
irinayat-MS: Yeah... I wonder how come it worked...

build/build.ps1
irinayat-MS: The prerequisites are currently listed in the readme in QirRuntime folder on the assumption that at the moment it's completely additive and not required for anything else in qsharp-runtime, so someone will opt-in to contribute to QirRuntime. I'll look into providing the controls to exclude it from the local builds by default while allowing us to run CI. Ideally, the CI should trigger only when there are changes to QirRuntime itself and to the native simulator. Is that possible?

bootstrap.ps1
```diff
@@ -11,6 +11,20 @@ Push-Location (Join-Path $PSScriptRoot "src/Simulation/Simulators")
...
+        choco install ninja
+        refreshenv
+    } else {
+        #brew install llvm # this seems to mess up native simulator build, probably because of STD libs
+        brew install ninja
```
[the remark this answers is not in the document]
irinayat-MS: I'm not an expect on Linux. Wouldn't most Linux users prefer installing with sudo? Is the goal of these build scripts to setup the local dev environment fully as well as the environment on the cloud build machines?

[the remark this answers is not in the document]
cgranade: As written, this line gets called when `bootstrap.ps1` is run from non-Windows hosts; if we expect Linux-based developers to use `sudo` to install pre-requisites, then I'd suggest modifying the condition for this line to make sure that it doesn't run on Linux hosts. Something like [`if (-not $IsLinux)`](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_automatic_variables?view=powershell-7#islinux) could work, perhaps?

src/QirRuntime/build.py
irinayat-MS: I'd like to keep build.py/test.py for the convenience of local development. If maintaining both these and powershell scripts scripts proves to be a burden we can deprecate.

bootstrap.ps1
irinayat-MS: For now I suggest we have ENABLE_QIRRUNTIME environment variable which is set in the CI builds but not locally so by default people won't need to deal with the prereqs and the build at all or even be aware of it. And if someone decides to contribute to the QirRuntime, its own readme outlines the required build/test steps. As the project matures and becomes required part of qsharp-runtime we will promote the corresponding info to the top.

build/build.ps1
irinayat-MS: We will look into facilitating setup later, when QirRuntime becomes required part of the qsharp-runtime. Right now nobody is expected to depend on it.

bootstrap.ps1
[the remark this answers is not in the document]
irinayat-MS: For now I'll hide all QirRuntime related steps from the mainstream development under ENABLE_QIRRUNTIME environment variable. As the project matures we'll need to look into making setup more robust on all supported platforms.

src/QirRuntime/build.py
irinayat-MS: I'd like to keep build/test python scripts for the convenience of the local development. If maintaining both python and powershell scripts proves to be a burden in the future we can deprecate the former.

src/QirRuntime/public/QuantumApiBase.hpp
```diff
@@ -0,0 +1,263 @@
...
+{
+namespace Quantum
+{
+    class CQuantumApiBase : public IQuantumApi
+    {
+      public:
+        // Qubit management
+        Qubit AllocateQubit() override
+        {
+            throw std::logic_error("not_implemented");
```
alan-geller, line 17: Is the expectation that they will change to not throw sometime soon? If these are essentially pure virtual methods in an abstract base class, then why not have them be pure virtual methods in an abstract base class?

Throwing a "not implemented" exception on a permanent basis means that you're claiming to implement some interface that you don't really implement -- you're lying to the compiler, either about your class (it doesn't really implement the interface) or the interface (it's much bigger than it should be) or both. See [here](https://www.dotnetperls.com/notimplementedexception) for a mild explanation, or [here](https://lostechies.com/jimmybogard/2008/10/12/notimplementedexception-and-the-interface-segregation-principle/) for more of a rant on the topic...

In this case, it seems that the only reason for this class to exist is to give the public virtual methods in IQuantumApi "not implemented" stub implementations. Assuming that any actual simulator will need to implement all of these, I think it would be better to just have the simulators inherit from IQuantumApi and leave methods "not implemented" in the actual simulator code during development, because then it's easy to see what work is remaining. If, instead, actual simulators don't need to actually implement all of these methods, then I would limit IQuantumAPI to only declare the methods that are actually necessary. Either way, I don't think this class is needed.

irinayat-MS: I am well aware of the conversation around pros and cons of providing this kind of base classes. As you can see in this project, they are super useful for testing, where most test IQuantumAPIs only need to override/mock a few methods. If we collectively agree that none of the production code should use this base class, I will eliminate it from the corresponding implementations.

alan-geller: I'm OK with it not being used in production code. What would be ideal would be a general IDE facility like VS has for C#, where you can start with an empty class that implements an interface and have VS generate the "not implemented" methods in the final class for you; I think that's rather out of scope for this PR, though :-)..

build/build.ps1
irinayat-MS: Have split the build script. Should I do the same with the test script as well?

anpaz (changes requested):

build/build.ps1
irinayat-MS: moved into build-native-simulator.ps1

---
In reply to: [508211204](https://github.com/microsoft/qsharp-runtime/pull/396#discussion_r508211204) [](ancestors = 508211204)

irinayat-MS: added Simulation\Native\build-native-simulator.ps1 and QirRuntime\build-qir-runtime.ps1

---
In reply to: [508214902](https://github.com/microsoft/qsharp-runtime/pull/396#discussion_r508214902) [](ancestors = 508214902)

build/build.ps1
```diff
@@ -6,16 +6,16 @@ $ErrorActionPreference = 'Stop'
...
-    Write-Host "##[info]Build Native simulator"
-    $nativeBuild = (Join-Path $PSScriptRoot "../src/Simulation/Native/build") 
-    cmake --build $nativeBuild --config $Env:BUILD_CONFIGURATION
-    if ($LastExitCode -ne 0) {
-        Write-Host "##vso[task.logissue type=error;]Failed to build Native simulator."
-        $script:all_ok = $False
-    }
-} else {
-    Write-Host "Skipping native. ENABLE_NATIVE variable set to: $Env:ENABLE_NATIVE."
+$nativeSimulator = (Join-Path $PSScriptRoot "../src/Simulation/Native")
```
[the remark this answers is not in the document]
irinayat-MS: Build of QIR Runtime doesn't depend on the native simulator, only tests do. I also don't think that it makes sense to bundle two basically unrelated components together based on the language they are implemented in. If qdk builds needs to exclude either of them, it should do so on a per component basis.

---
In reply to: [509744724](https://github.com/microsoft/qsharp-runtime/pull/396#discussion_r509744724) [](ancestors = 509744724)

bootstrap.ps1
```diff
@@ -11,6 +11,11 @@ Push-Location (Join-Path $PSScriptRoot "src/Simulation/Simulators")
     .\FindNuspecReferences.ps1
 Pop-Location
 
+Push-Location (Join-Path $PSScriptRoot "src/QirRuntime")
```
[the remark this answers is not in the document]
irinayat-MS: I think ENABLE_NATIVE is a misnomer. It controls a specific component: the fullstate simulator, which happens to be implemented in C++. QirRuntime is an independent component and users might want to build it whether or not they build the fullstate simulator. ENABLE_QIRRUNTIME variable is checked inside install-prereqs.ps1. I can move the check here if that's what you mean.

---
In reply to: [509743231](https://github.com/microsoft/qsharp-runtime/pull/396#discussion_r509743231) [](ancestors = 509743231)

src/QirRuntime/install-prereqs.ps1
```diff
@@ -0,0 +1,15 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+if ($Env:ENABLE_QIRRUNTIME -eq "true") {
```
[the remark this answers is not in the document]
irinayat-MS: I was on the fence on this... will move the checks from here (and the build scripts) into the driver

---
In reply to: [509753046](https://github.com/microsoft/qsharp-runtime/pull/396#discussion_r509753046) [](ancestors = 509753046)

bootstrap.ps1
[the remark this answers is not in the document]
irinayat-MS: Moved the check.

---
In reply to: [510413021](https://github.com/microsoft/qsharp-runtime/pull/396#discussion_r510413021) [](ancestors = 510413021,509743231)

anpaz (approved):

src/QirRuntime/experimental/CircuitToJsonPrinter.cpp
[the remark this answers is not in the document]
bettinaheim: @irinayat-MS It is entirely unclear what technique this is supposed to illustrate, and that kind of split seems like a bad design choice in any chase for the reasons I mentioned above. Could you please clarify your thinking to me?

src/QirRuntime/public/QuantumApiBase.hpp
```diff
@@ -0,0 +1,263 @@
...
+
+#include <exception>
+
+#include "IQuantumApi.hpp"
+
+namespace Microsoft
+{
+namespace Quantum
+{
+    class CQuantumApiBase : public IQuantumApi
```
bettinaheim, line 11: This class promotes building an backend for which certain methods permanently throw a runtime exception with the expectation that these methods are simply not used. We should not promote this, since this is rather fragile design. I would like to remove this.

irinayat-MS: This class allows quick experimentation with building a new simulator and also makes test code _much more_ readable. I can rename it to CQuantumApiNoneImplemented to make it more clear that the class isn't recommended to be used as a permanent base class in production code and replace it in the toffoli/fullstate simulators with direct inheritance from the interface and explicit "no supported" rather than "not implemented" throws on the relevant methods.

src/QirRuntime/experimental/CircuitToJsonPrinter.cpp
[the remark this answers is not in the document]
irinayat-MS: This design pattern is known as 'Composite' (with elements of 'Template Method' pattern). The tests show how the layering could be shared by multiple simulators and how the decomposition could be tested on its own without relying on a real simulator, which might be tricky.

src/QirRuntime/public/IQuantumApi.hpp
```diff
@@ -0,0 +1,146 @@
...
+{
+    struct IQuantumApi
+    {
+        virtual ~IQuantumApi() {}
+
+        // TODO: Only some simulators 'know' the state -- is it OK to have this method on IQuantumApi?
+        // The callback will be invoked on each basis vector (in the standard computational basis) in little-endian
+        // order until it returns `false` or the state is fully dumped.
+        typedef bool (*TGetStateCallback)(size_t /*basis vector*/, double /* amplitude Re*/, double /* amplitude Im*/);
+        virtual void GetState(TGetStateCallback callback) = 0;
```
alan-geller, line 19: To answer the TODO: no, it's not OK to have this method in the base interface. It's not just "only some": it's basically one (full-state simulator) that has a state that is accessible in this way, compared to many (CHP, density matrix, resource counter, ...) that don't.

src/QirRuntime/public/IQuantumApi.hpp
```diff
@@ -0,0 +1,146 @@
...
+
+        virtual std::string DumpQubit(Qubit qubit) = 0;
+
+        // Qubit management
+        virtual Qubit AllocateQubit() = 0;
+        virtual void ReleaseQubit(Qubit qubit) = 0;
+
+        // Shortcuts
+        virtual void CNOT(Qubit control, Qubit target)
```
alan-geller, line 28: Why is this alias needed?

cesarzc (approved):

src/QirRuntime/public/IQuantumApi.hpp
irinayat-MS: I don't feel strongly about those shortcuts and it's more here to collect feedback on various ways to shape the interface... I'd suggest to keep it for now so we can move on with the PR and file an issue that tracks finalizing the interface.

src/QirRuntime/public/IQuantumApi.hpp
```diff
@@ -0,0 +1,146 @@
...
+        {
+            CX(control, target);
+        }
+        virtual void CX(Qubit control, Qubit target) = 0;
```
alan-geller, line 32: The base class for all simulators shouldn't specify the instruction set. QIR by design doesn't specify an instruction set, so the standard simulation runtime shouldn't either.

src/QirRuntime/public/IQuantumApi.hpp
```diff
@@ -0,0 +1,146 @@
...
+
+        // Elementary operatons
+        virtual void X(Qubit target) = 0;
+        virtual void Y(Qubit target) = 0;
+        virtual void Z(Qubit target) = 0;
+        virtual void H(Qubit target) = 0;
+        virtual void S(Qubit target) = 0;
+        virtual void T(Qubit target) = 0;
+        virtual void SWAP(Qubit target1, Qubit target2) = 0;
+        virtual void Clifford(CliffordId cliffordId, PauliId pauli, Qubit target) = 0;
```
alan-geller, line 44: X, Y, Z, H, and S are all Cliffords. Why don't they get implemented by this method as well?

src/QirRuntime/public/IQuantumApi.hpp
```diff
@@ -0,0 +1,146 @@
...
+        // unitary is an array of columns: 2^n x 2^n. Columns are nullable, so for state prep use just the first column
+        virtual void Unitary(long numTargets, double** unitary, Qubit targets[]) = 0;
```
alan-geller, line 46: This only makes sense for a small number of specific simulators. It doesn't belong in a base class.

src/QirRuntime/public/IQuantumApi.hpp
irinayat-MS: Supporting QIR is only one of the goals of the simulators framework. QIR has the flexibility of using the interface for the qis methods or not.

src/QirRuntime/public/IQuantumApi.hpp
```diff
@@ -0,0 +1,146 @@
...
+
+        virtual void R(PauliId axis, Qubit target, double theta) = 0;
+        virtual void RFraction(PauliId axis, Qubit target, long numerator, int power) = 0;
+        virtual void R1(Qubit target, double theta) = 0;
+        virtual void R1Fraction(Qubit target, long numerator, int power) = 0;
+        virtual void Exp(long numTargets, PauliId paulis[], Qubit targets[], double theta) = 0;
+        virtual void ExpFraction(long numTargets, PauliId paulis[], Qubit targets[], long numerator, int power) = 0;
+
+        // Multicontrolled operations
+        virtual void ControlledX(long numControls, Qubit controls[], Qubit target) = 0;
```
alan-geller, line 58: Not all simulators can implement this, so it doesn't belong in a base class.
Also, why is CX different from ControlledX with numControls==1? Why isn't there just one method?

src/QirRuntime/public/IQuantumApi.hpp
irinayat-MS: As I replied above, I don't have any strong preference wrt to the convenience APIs in this interface and their signatures. If the consensus is that they are more confusing than helpful, let's remove them.

src/QirRuntime/public/IQuantumApi.hpp
```diff
@@ -0,0 +1,146 @@
...
+            Qubit controls[],
+            long numTargets,
+            PauliId paulis[],
+            Qubit targets[],
+            long numerator,
+            int power) = 0;
+
+        // Adjoint operations
+        virtual void SAdjoint(Qubit target) = 0;
+        virtual void TAdjoint(Qubit target) = 0;
```
alan-geller, line 111: Why only these? Why not the adjoint for other Cliffords or for rotations?

src/QirRuntime/public/CoreTypes.hpp
```diff
@@ -0,0 +1,59 @@
+#pragma once
+
+// The core types will be exposed in the C-interfaces for interop, thus no
+// namespaces or scoped enums can be used to define them.
+
+/*==============================================================================
+  Helper types
+==============================================================================*/
+enum TernaryBool
```
alan-geller, line 9: Just curious: when is this used? The Boolean type in QIR, at least, is always either true or false, never pending or undefined.

src/QirRuntime/public/BitStates.hpp
```diff
@@ -0,0 +1,78 @@
...
+#include <vector>
+
+namespace Microsoft
+{
+namespace Quantum
+{
+    /*==============================================================================
+        Provides dynamically extendable storage for packed bits
+    ==============================================================================*/
+    struct BitStates
```
alan-geller, line 13: Why can't you use `vector<bool>` for this? I think it's pretty equivalent, other than for `IsAny`, and I don't think that method is ever used in a performance-sensitive situation where simply iterating through all of the bits wouldn't be OK.

src/QirRuntime/public/ChunkedStore.hpp
```diff
@@ -0,0 +1,132 @@
...
+        uint64_t filled = 0;
+        TItem data[COUNT_ITEMS];
+
+        bool IsValidItem(long index)
+        {
+            return (index < COUNT_ITEMS) && (filled & (static_cast<uint64_t>(1) << index));
+        }
+    };
+
+    template <typename TItem_WithId> struct CChunkedStore
```
alan-geller, line 43: Why not just use `vector<TItem_WithId>`? Yes, it will be slower when it needs to reallocate, but as far as I can tell this class is only used for testing, so the perf hit isn't particularly meaningful, and this class may be slightly slower when accessing elements, so it may all balance out anyway?

src/QirRuntime/public/ChunkedStore.hpp
```diff
@@ -0,0 +1,132 @@
...
+#include <assert.h>
+#include <vector>
+
+#include "BitStates.hpp"
+
+namespace Microsoft
+{
+namespace Quantum
+{
+    /*==============================================================================
```
alan-geller, line 12: Why is this part of the public API? It seems like an internal implementation class.

src/QirRuntime/public/CoreTypes.hpp
irinayat-MS: Yes, QIR doesn't support deferred results but from the conversations with the researches the ability to defer measurements came up as a desirable feature. To be able to experiment with it but to avoid throwing from the C-API the enum was introduced. I'm not too happy about it, but thinking through the design here was one of the tasks that itself got, no pun intended, deferred.

src/QirRuntime/public/QubitAllocationPolicies.hpp
```diff
@@ -0,0 +1,48 @@
...
+using namespace std;
+namespace Microsoft
+{
+namespace Quantum
+{
+    /*==============================================================================
+        The default qubit manager reuses released qubit ids in "Last released -
+        first reused" order.
+    ==============================================================================*/
+    class CReuseLastReleasedQubitAllocationPolicy : public IIdAllocationPolicy
```
alan-geller, line 17: I assume there could be other implementations of this interface, so it's not clear to me why these two are considered part of the public API.

In general, I would expect a folder/project that held the header files for types that any code that uses a simulator or execution engine would need to know about, and a separate folder/project that defines the types that an implementation of a simulator or execution engine needs to have access to. I would expect actual implementations to be in yet other folders, either one per folder or perhaps a few lumped together if they're small. The current "public" folder seems to be a mix of external types, internal types, and specific implementation types.

src/QirRuntime/public/ResourceStatistics.hpp
```diff
@@ -0,0 +1,55 @@
+#pragma once
+#include <sstream>
+
+namespace Microsoft
+{
+namespace Quantum
+{
```
alan-geller, line 7: Shouldn't everything in this file be part of the resource estimator implementation, rather than part of the public API?

src/QirRuntime/public/ResourceStatistics.hpp
```diff
@@ -0,0 +1,55 @@
...
+namespace Quantum
+{
+    static const char* names[] = {"cX", "cCX", "cY",  "cCY",  "cZ", "cCZ", "cH",  "cCH",  "cR", "cCR",
+                                  "cS", "cCS", "cAS", "cCAS", "cT", "cCT", "cAT", "cCAT", "cM"};
+    static constexpr int counters = sizeof(names) / sizeof(char*);
+
+    struct ResourceStatistics
+    {
+        // these members must match names[] above
```
alan-geller, line 14: Why not just use a `map<string, long>` so the code isn't so fragile and limited?

src/QirRuntime/public/ResourceStatistics.hpp
```diff
@@ -0,0 +1,55 @@
...
+        long cX, cCX = 0;
```
alan-geller, line 15: Given that there's no methods to increment or read these counters, and that the counters are all public, I assume that means that users of this type just access and update the counters directly. This is very fragile and very difficult to extend and maintain.

src/QirRuntime/public/ResourceStatistics.hpp
```diff
@@ -0,0 +1,55 @@
...
+            os << "{";
+            os << "\"qubit_width\":" << cQubitWidth;
+
+            if (cCircuitDepth > 0)
+            {
+                os << ",\"circuit_depth\":" << cCircuitDepth;
+            }
+
+            os << ",\"statistics\":[";
+            long* fields = reinterpret_cast<long*>(this);
```
alan-geller, line 43: This is massively unsafe and error-prone, especially if I'm brave enough to derive from this class. There's no guarantee that there's nothing between what `this` points to and the first instance variable, and there's no guarantee that the instance variables wind up in memory in the same order they appear in the type. While one compiler may work that way, another might not, or the same compiler might work differently on different platforms or with different optimization options.
Either you should step through the fields explicitly, without a loop, in which case you don't need the string array and can avoid that fragility, or replace the whole thing with a `map` that you can iterate over cleanly.

alan-geller (changes requested):
I've only finished `public` so far, but I didn't want my comments to pile up forever.

src/QirRuntime/experimental/CircuitToJsonPrinter.cpp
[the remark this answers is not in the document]
bettinaheim: I see a lot of downsides with this pattern in general, and in this case specifically see hardly any upsides. There is no clear answer to "what piece is responsible for implementing the quantum instruction set" here, and it therefore directly violates the best practices for the compiler and runtime codebase. I understand that this is also the case for the current translator setup, which is a) causing a lot of grief, and b) why we are in the process of revising it.

src/QirRuntime/public/QuantumApiBase.hpp
irinayat-MS: The class is used by production tests and experimental features and their tests so it would create a wrong dependency direction to move it into either. While some tools do support refactoring of C++ code, as far as I know there is no widely accepted one yet. The class allows to start on a new simulator quickly, being able to compile and test it with minimal effort. Turning the code into production is also simple by replacing the inheritance from this class to interface and implementing everything that compiler would complain about.

src/QirRuntime/experimental/CircuitToJsonPrinter.cpp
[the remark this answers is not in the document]
bettinaheim: Following our wiki regarding C++ style rules, it doesn't seem to be encouraged: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#S-class
The practice seems related to the general use of Hungarian notation and based on a quick search it sounds like it may be common mostly in certain older code bases. Given that our native simulator doesn't follow that practice, we should make a call in general, and I would suggest to follow the page our wiki leads to and drop the C. I am happy to set up a broader decision for that, to document this call if you feel this is likely something that other people might wonder about.

src/QirRuntime/public/QuantumApiBase.hpp
alan-geller: Before we can merge this into main, I think we need to replace the interface as well as this class. Throwing "not supported" permanently instead of "not implemented" doesn't solve the problem, which is that the base interface currently pre-selects a specific quantum instruction set where the quantum instruction set should be completely open and unspecified,

A code generation solution for  like we're enabling for managed simulators is the right long-term solution, allowing every target to specify the quantum instruction set (and common gate decompositions) it supports in a standard way.

src/QirRuntime/lib/QAPI/QAPI.cpp
```diff
@@ -0,0 +1,243 @@
...
+EXPORTQAPI Result QAPI_UseZero(QuantumExecutionContext* context) // NOLINT
+{
+    return GetQuantumExecutor(context)->UseZero();
+}
+EXPORTQAPI Result QAPI_UseOne(QuantumExecutionContext* context) // NOLINT
+{
+    return GetQuantumExecutor(context)->UseOne();
+}
+
+// Shortcuts
```
bettinaheim, line 75: All the gate operations are specific for a particular instruction set (whereas allocations and deallocations are rather ubiquitous). It may hence make sense to split them out to reflect that. What is the idea here; What is QAPI to be used for in general? What are the general purpose pieces of the runtime and do you have the target specific pieces split out somewhere?

src/QirRuntime/lib/QIR/__quantum__rt.hpp
```diff
@@ -0,0 +1,291 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
bettinaheim, line 1: For my own understanding: Which files do have a "__" naming convention?

src/QirRuntime/lib/TracingContext/TracingContext.cpp
```diff
@@ -0,0 +1,106 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
bettinaheim, line 1: Given that the resources estimation tool is still experimental, how come this context isn't? How is it supposed to be used?

src/QirRuntime/public/CoreTypes.hpp
bettinaheim: That sounds like a solution rather than a need that was communicated to you by researchers. Could you please clarify what exactly the scenarios are that need to be supported where this would be useful, and then we can make an informed call on how to best support them.

src/QirRuntime/public/CoreTypes.hpp
```diff
@@ -0,0 +1,59 @@
...
+    PauliId_Z = 2,
+    PauliId_Y = 3,
+};
+
+static const char PauliChars[4] = {'I', 'X', 'Z', 'Y'};
+
+/*==============================================================================
+  Clifford operators
+==============================================================================*/
+enum CliffordId
```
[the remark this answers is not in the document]
cgranade: This looks to be a presentation of \overline{𝐶}₁ = 𝐶₁ / 𝑃₁ (that is, the group of Clifford operators modded out by the Pauli group, representing Clifford operations up to Pauli corrections)?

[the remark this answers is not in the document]
cgranade: If I may recommend an alternative presentation, the decomposition of 𝐶₁ into 𝑆, 𝑋, and 𝐸 used in the newsynth project is particularly nice for implementation of the full 24-member single-qubit Clifford group in terms of enums. https://hackage.haskell.org/package/newsynth

src/QirRuntime/public/QuantumApiBase.hpp
swernli: While I agree with the idea that this interface and related base classes will be heavily refactored to take advantage of the redesign of target packages and the introduction of the code generation components, neither of those parts are done yet. I don't think we should force this PR, which introduces code that is not yet part of the production workflow, to wait on all the other related work to be finished before it gets merged. I would be more in favor of merging this now in its working form, where it mimics the current patterns in the production infrastructure, and then iterating on it in main as the other work comes in. That way we don't continue developing in the mode where large chunks of code live in feature branches waiting for integration with other large chunks also in feature branches.

src/QirRuntime/experimental/ExperimentalSimFactory.hpp
```diff
@@ -0,0 +1,44 @@
...
+    // result values from each simulator after calling `GetResultValue` on a result, that
+    // represents a cumulative result of a measurement. If the buffer isn't provided (is null),
+    // the individual result values aren't communicated back to the caller.
+    std::unique_ptr<IQuantumApi> CreateLockstepSimulator(
+        std::vector<IQuantumApi*>&& simulatorsToRunInLockstep,
+        ResultValue* resultValuesBuffer /*size of the buffer must be equal the number of simulators*/);
+
+    // The circuit to json translator traces the quantum program and generates json that describes the program.
+    std::shared_ptr<ITranslator> CreateCircuitToJsonTranslator();
+    std::unique_ptr<IQuantumApi> CreateCircuitPrintingSimulator(std::shared_ptr<ITranslator> translator);
```
bettinaheim, line 28: Since this is under experimental, it's certainly fine if it's not perfectly documented. In terms of commenting when something is unclear: it is not clear to me what the intended duty/responsibility of the CircuitPrintingSimulator here is.

src/QirRuntime/public/IQuantumApi.hpp
```diff
@@ -0,0 +1,146 @@
...
+
+#include <complex>
+
+#include "CoreTypes.hpp"
+
+namespace Microsoft
+{
+namespace Quantum
+{
+    struct IQuantumApi
```
bettinaheim, line 11: I think it would be quite helpful go over the IQuantumAPI and QAPI in our Friday meeting. I'll catch up outside this PR to get more alignment on the precise API before that and I'll start a spec document to capture this (of course I am happy if you want to do that, but I believe you may be eager to focus on the work with Alex :) ).

bettinaheim (commented):
I finally made it through the PR. It's exciting to see how this is coming together!

src/QirRuntime/experimental/ITranslator.hpp
```diff
@@ -0,0 +1,23 @@
...
+
+namespace Microsoft
+{
+namespace Quantum
+{
+    // Currently, there is implicit expectation that the simulator knows the runtime type of the translator and can cast
+    // to it in order to access callbacks that allow the translator to accumulate necessary information. The client must
+    // create matching (simulator, translator) pair using the corresponding factory methods, but they don't get access
+    // to the inner-workings of the translator.
+    struct ITranslator
```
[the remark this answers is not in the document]
irinayat-MS: My choice of `translator` was driven by the fact that the class that implements this interface accumulates information from simulator's execution and then it can generate a new represantation of this information (that is, 'translate').

src/QirRuntime/experimental/QubitInterferences.cpp
```diff
@@ -0,0 +1,184 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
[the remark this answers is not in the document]
irinayat-MS: I hope we won't need to create subfolders under experimental. It would indicate we are doing too many experiments at once.

src/QirRuntime/lib/QAPI/QAPI.cpp
```diff
@@ -0,0 +1,243 @@
...
+using namespace Microsoft::Quantum;
+
+static IQuantumApi* GetQuantumExecutor(QuantumExecutionContext* context)
+{
+    return reinterpret_cast<IQuantumApi*>(context->hQuantumApi);
+}
+
+QuantumExecutionContext* QAPI_CreateContext_Toffoli() // NOLINT
+{
+    unique_ptr<IQuantumApi> qapi = CreateToffoliSimulator();
```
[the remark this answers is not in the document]
irinayat-MS: The factory method returns a unique_ptr, which is a better (imho) memory management choice for internal clients who will get all the benefits of automatic memory management it provides. The QuantumExecutionContext hasn't been thought through (copied from some preexisting experimental code) and is mostly just a stub to indicate that we might need to manage more environment data than a single simulator while exposing a primitive C-API for testing the scenario of consuming the runtime via dll rather than statically.

src/QirRuntime/lib/QIR/__quantum__rt.hpp
irinayat-MS: No convention. I wanted "__dummy.cpp" out of the way, and the headers got named after the corresponding prefixes in the QIR spec. I think the dummy is weird enough to keep as is, the headers should probably be renamed to quantum__qis.hpp and quantum__rt.hpp.

src/QirRuntime/lib/TracingContext/TracingContext.cpp
irinayat-MS: This context uses a linker seem to avoid virtual methods so LLVM could fold the loops with gate counts and it's not related to TracingSimulator. It should be moved into experimental.

src/QirRuntime/public/BitStates.hpp
irinayat-MS: The standard doesn't guarantee that `vector<bool>` is optimized for space.

src/QirRuntime/public/ChunkedStore.hpp
irinayat-MS: Initially I set up the project with two folders for headers: public and internal (in addition to the headers that are consumed only inside their own folder) but got feedback from Dimitry and Sasha that setup like this tends to cause maintainence issues and that there is no harm in over-sharing utilities headers.

src/QirRuntime/public/CoreTypes.hpp
irinayat-MS: https://ms-quantum.visualstudio.com/Quantum%20Program/_workitems/edit/14449 used IonQ's result placeholders as an example when it might not be possible to get value from a measurement and input from Vadym re avoiding throwing from the C-API. I don't think we've ever had a conversation about the scenarios the native runtime is supposed to support. There is no spec or design document for it, so in a way, all code in this project is a solution. To something.

src/QirRuntime/public/ResourceStatistics.hpp
irinayat-MS: Yes, the direct access was intended here to allow LLVM to optimize the code away and collapse the loops to a simple increment instructions based on the loop count. I played with it at the beginning of the project when the focus was on incubation of highly optimized native resource tracer. I'll move it into experimental together with the TracingContext.

src/QirRuntime/public/ResourceStatistics.hpp
irinayat-MS: I cannot use a map because it will surely prevent LLVM from optimizing increments of the counters. I could replace the field names with an enum and an array of longs which would make it somewhat safer, but I don't know whether that would play well with LLVM's optimization passes.

src/QirRuntime/public/SimFactory.hpp
```diff
@@ -0,0 +1,18 @@
...
+#include <vector>
+
+#include "IQuantumApi.hpp"
+
+namespace Microsoft
+{
+namespace Quantum
+{
+    // Toffoli Simulator
+    std::unique_ptr<IQuantumApi> CreateToffoliSimulator();
```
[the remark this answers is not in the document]
irinayat-MS: To allow creating a specific simulator to run the quantum program on.

src/QirRuntime/README.md
```diff
@@ -0,0 +1,127 @@
...
+4. cmake -G Ninja -DCMAKE_BUILD_TYPE=Release ..
+5. cmake --build .
+
+Or you can run `build.py` script from the root of the project. The default options for the script are `make debug`.
+
+- (Windows) `python build.py [make/nomake] [debug|release] [ir]`
+- (Linux) `python3 build.py [make/nomake] [debug|release] [ir]`
+
+The scrip will place the build artefacts into `build/[Windows|Linux]/[Debug|Release]` folder.
+
```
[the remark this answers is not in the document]
irinayat-MS: The project builds on Mac and all tests pass, including those that run QIR on the full state simulator.

src/QirRuntime/public/QuantumApiBase.hpp
bettinaheim: I believe this thread should be resolved with the plan we defined yesterday. To summarize: We will merge this PR before any of the automation infrastructure exists, and we will roughly align the structure of the code to how the code is expected to be structured in the the future when parts of it are autogenerated instead of manually implemented. This PR contains the manual implementation and execution support for one specific target package, and all further target packages will be supported once we have more automation built out. Let me know if there are still open questions.

src/QirRuntime/experimental/QubitInterferences.cpp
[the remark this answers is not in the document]
bettinaheim: :) That seems fair.

src/QirRuntime/lib/QAPI/QAPI.cpp
bettinaheim: My questions here were also answered yesterday. :) 
To recap (correct me if anything seems wrong): QAPI is to facilitate directly interacting with the runtime infrastructure without going via QIR. Since this is out of scope for now, we don't want to take on the burden of maintaining an additional API at this time. We will hence drop the QAPI from this PR. 
As far as the target specific quantum instruction set goes, we said we will split the IQuantumAPI into two pieces; one for the qubit and result handles that are independent on the quantum instruction set (though may vary depending on the used runtime/backend), and one with all the method that are specific to the instruction set supported by the quantum processor itself (including Message/logging, and asserts).

src/QirRuntime/experimental/TracingSimulator.cpp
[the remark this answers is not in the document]
irinayat-MS: removed experimental features

src/QirRuntime/public/QuantumApiBase.hpp
irinayat-MS: Split IQuantumApi into ISimulator (qubit and result management) and IQuantumGateSet both of which are currently implemented by the full state simulator. When the automatic generation of target gate sets is available, the latter of the two interfaces will be deprecated.

CQuantumApiBase is now named SimulatorStub and not used in production code.

src/QirRuntime/public/IQuantumApi.hpp
irinayat-MS: Full state simulator provides these as part of its public interface. I presume, this is because other operators are either self-adjoint and the adjoint can be created by modifying the arguments. Given the interface is temporary, I'll keep these for now to make sure all QIR can run against the full state simulator. In the future, the automatically generated base class will define the gates as it sees fit.

src/QirRuntime/public/CoreTypes.hpp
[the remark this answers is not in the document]
irinayat-MS: No provisions for decomposition in QIR Runtime anymore. Stefan's work on targeting will take care of it.

src/QirRuntime/public/IQuantumApi.hpp
irinayat-MS: Moved it to IDiagnostics

src/QirRuntime/public/QAPI.hpp
```diff
@@ -0,0 +1,223 @@
...
+// QuantumExecutionContext collects handles to internal interfaces, used to drive execution.
+// The meaning of handles in the array is defined by the concrete type of the main quantum api target.
+struct QuantumExecutionContext
+{
+    QAPI_HANDLE hQuantumApi;
+    // anything else?
+};
+
+// Factory methods for various simulators we provide.
+EXPORTQAPI QuantumExecutionContext* QAPI_CreateContext_Toffoli();   // NOLINT
```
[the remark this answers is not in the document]
irinayat-MS: No public C-API anymore

src/QirRuntime/lib/QAPI/QAPI.cpp
irinayat-MS: No public C-API anymore

src/QirRuntime/experimental/ExperimentalSimFactory.hpp
irinayat-MS: All experimental code has been removed

bettinaheim (approved):
It's great how this has taken shape!

alan-geller (approved):

PR #397, opened 2020-10-13
Return projects to original paths to fix e2e build

swernli:
The e2e build failures were due to the project path changes introduced by the feature work. This reverts the Simulation\Simulators.Core folder to Simulation\Simulators to fix e2e build.

This change is mostly file renames and path fixes, and the only change to functionality was a bug fix in `MResetY` to add the adjoint functor when calling `S`.


ScottCarda-MS (approved):
lgtm

swernli:
There are some failures in iqsharp python tests that I think will require some work to fix. I'd like to get this PR merged up to the feature branch and tackle those in a separate one to make that work easier to review in isolation.

Issue #141, opened 2020-02-06
Improve error messaging in Q# test & application projects 


vadym-kl:
## What you would like the feature to accomplish. ##
Currently when a single test fails the error message is verbose, hard to read and exposes C# code generation internals. 
See below for an example. Similar improvement would be nice to have in the code that is generated when creating Q# console application.

##  Additional context ##
Consider a test project with a single qsharp file with a single test that fails:
```qsharp
namespace qsharp_unit_test_example {
    
    open Microsoft.Quantum.Canon;
    open Microsoft.Quantum.Diagnostics;
    open Microsoft.Quantum.Intrinsic;

    @Test("QuantumSimulator")
    operation AllocateQubit () : Unit {
        
        using (q = Qubit()) {
            X(q); // the new line we added to make test fail
            Assert([PauliZ], [q], Zero, "Newly allocated qubit must be in |0> state."); //test fails here
        }
        
        Message("Test passed.");
    }
}
```
Then run `dotnet test` for the project and observe the following output:
```
[xUnit.net 00:00:01.12]     qsharp_unit_test_example.AllocateQubit+QuantumSimulator.AllocateQubit [FAIL]
  X qsharp_unit_test_example.AllocateQubit+QuantumSimulator.AllocateQubit [52ms]
  Error Message:
   System.AggregateException : One or more errors occurred. (Newly allocated qubit must be in |0> state.
        Expected:       0
        Actual: 1)
---- Microsoft.Quantum.Simulation.Core.ExecutionFailException : Newly allocated qubit must be in |0> state.
        Expected:       0
        Actual: 1
  Stack Trace:
     at System.Threading.Tasks.Task.ThrowIfExceptional(Boolean includeTaskCanceledExceptions)
   at System.Threading.Tasks.Task.Wait(Int32 millisecondsTimeout, CancellationToken cancellationToken)
   at System.Threading.Tasks.Task.Wait()
   at qsharp_unit_test_example.AllocateQubit.QuantumSimulator.AllocateQubit() in D:\repos\qsharp-unit-test-example\Tests.qs:line 16
----- Inner Stack Trace -----
   at Microsoft.Quantum.Simulation.Simulators.QuantumSimulator.QSimAssert.<get_Body>b__6_0(ValueTuple`4 _args)
   at Microsoft.Quantum.Simulation.Core.Operation`2.Apply(I a)
[...]

PR #399, opened 2020-10-15
Improve Error Messaging in Q# Unit Tests

ScottCarda-MS:
This creates clearer error messages when an Assert fails in Q# unit tests. This partially addresses #141 .


ScottCarda-MS:
The code change did not have the desired effect. It seems the error messages are much less complicated without this change.

PR #400, opened 2020-10-16
Add debug symbols to Microsoft.Quantum.Simulators package

cesarzc:
This change adds symbols for  `Microsoft.Quantum.Simulation.Common.dll` and `Microsoft.Quantum.Simulation.QCTraceSimulatorRuntime.dll` to the `Microsoft.Quantum.Simulators` package.


src/Simulation/Simulators/Microsoft.Quantum.Simulators.nuspec.template
```diff
@@ -20,6 +20,10 @@
     <file src="..\Native\osx\*" target="runtimes\osx-x64\native" />
     <file src="..\Native\linux\*" target="runtimes\linux-x64\native" />
     <file src="..\Native\win10\*" target="runtimes\win-x64\native" />
+    <file src="..\Common\bin\$Configuration$\**\Microsoft.Quantum.Simulation.Common.xml" target="lib" />
+    <file src="..\Common\bin\$Configuration$\**\Microsoft.Quantum.Simulation.Common.pdb" target="lib" />
+    <file src="..\QCTraceSimulator\bin\$Configuration$\**\Microsoft.Quantum.Simulation.QCTraceSimulatorRuntime.xml" target="lib" />
```
DmitryVasilevsky, line 25: Do you know why .DLL files are referenced below (for Microsoft.Quantum.Simulation for example) and not referenced here?

cesarzc: For `Microsoft.Quantum.Simulation.Common` and for `Microsoft.Quantum.Simulation.QCTraceSimulatorRuntime` DLLs are added by a script that takes this template to generate the final nuspec used for packaging.

Adding a reference to the nuspec from the template is straight forward because it only needs to append it, but adding a file is a bit more complicated. That is why opted for adding them directly to the nuspec template. However, for consistency it would be better to add them through the script so I will give it a try. The only concern that I have is the order of operations, in this case `pdb` and `xml` files not being available at the time the script is invoked.

anpaz: No, the script only adds package references. There is a comment on the script explaining what it does and why.

cesarzc: Thanks for the clarification Andres.

After a conversation with @bettinaheim we thought having the `pdb` and `xml` files explicitly added to the package is a good option since its contents is pretty stable and we want users to have an easy way to debug.

bettinaheim (approved):

PR #401, opened 2020-10-16
Create the correct default values of Q# types

### LINKED ISSUE #359: Default values in uninitialized arrays can contain null
## Describe the bug
The default values for a type `T` given by `(new T[1])[0]` or the equivalent library function `Default<T>()` do not match the documented values here:

https://docs.microsoft.com/en-us/quantum/user-guide/language/expressions#array-creation

## Expected behavior
The default values should match the documented values.

## Actual behavior
* `Default<String>()` is `null` but should be `""`
* `Default<Result>()` is `null` but should be `Zero`
* `Default<Range>()` is `0..1..0` but should be `1..1..0`
* `Default<'T[]>()` is `null` but should be `new 'T[0]`
* `Default<Unit>()` is `null` but should be `()`

**Additional context**
See microsoft/qsharp-compiler#589 for more examples.

samarsha:
Closes #359.


swernli (approved):
The reflection you are using here to get info on the underlying generic types seems reasonable to me. I don't know of any other ways to handle these kinds of types, and I've seen this approach recommended elsewhere (and used it myself for some .NET that wraps WMI). I'd definitely let others weigh in too, though I'm of the opinion this is the way C# is meant to handle generics.

src/Simulation/Core/Default.cs
```diff
@@ -0,0 +1,92 @@
...
+        private static object? OfArrayType(Type type) =>
+            type.IsGenericType && type.GetGenericTypeDefinition() == typeof(IQArray<>)
+                ? Activator.CreateInstance(typeof(QArray<>).MakeGenericType(type.GenericTypeArguments))
+                : null;
+
+        /// <summary>
+        /// If the given type is a Q# tuple type, returns the default tuple of that type, or null otherwise.
+        /// </summary>
+        private static object? OfTupleType(Type type) =>
+            type.IsGenericType && Tuples.Contains(type.GetGenericTypeDefinition())
```
bettinaheim, line 78: How are tuples with more than 8 items handled? And aren't they wrapped by QTuple?

src/Simulation/Core/QArray.cs
```diff
@@ -150,8 +149,7 @@ public void Extend(long newLength)
                 long oldLength = storage.Count;
                 for (int i = 0; i < (newLength - oldLength); i++)
                 {
-                    T obj = CreateDefault();
-                    storage.Add(obj);
+                    storage.Add(Default.OfType<T>());
```
bettinaheim, line 152: While I am sure the compiler should be smart enough to pull this out of the loop, it doesn't hurt to do that manually to make sure it is not evaluated each time.

bettinaheim (commented):
Could you please double check the perf tests? Thanks!

src/Simulation/Core/Default.cs
samarsha: `QTuple` is only used by UDTs, not plain Q# tuples. Tuples with more than 8 items are defined recursively in terms of tuples of arity 8 and smaller.

src/Simulation/Core/QArray.cs
samarsha: I can do it manually. Since C# is impure and `OfType` could potentially have side effects, I am not so sure that the compiler can perform common subexpression elimination here.

samarsha:
> Could you please double check the perf tests? Thanks!

This turns out to not be easy. Currently, the runtime performance tests aren't published in a format that's convenient to compare against results from main.

PR #402, opened 2020-10-19
Updated generated code to use try-catch in unit tests.

ScottCarda-MS:
This creates clearer error messages when an Assert fails in Q# unit tests. This partially addresses #141.

### DIFF
--- a/src/Simulation/CsharpGeneration.Tests/SimulationCodeTests.fs
+++ b/src/Simulation/CsharpGeneration.Tests/SimulationCodeTests.fs
@@ -3685,10 +3685,21 @@ namespace Microsoft.Quantum.Tests.UnitTests
                     baseSim.OnLog += this.Output.WriteLine;
                 }
 
-                sim.Run<UnitTest1, QVoid, QVoid>(QVoid.Instance).Wait();
-                if (sim is IDisposable disposeSim)
+                try
                 {
-                    disposeSim.Dispose();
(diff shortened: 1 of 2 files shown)


src/Simulation/CsharpGeneration.Tests/SimulationCodeTests.fs
```diff
@@ -3685,10 +3685,21 @@ namespace Microsoft.Quantum.Tests.UnitTests
...
 
-                sim.Run<UnitTest1, QVoid, QVoid>(QVoid.Instance).Wait();
-                if (sim is IDisposable disposeSim)
+                try
+                {
+                    sim.Execute<UnitTest1, QVoid, QVoid>(QVoid.Instance);
+                }
+                catch (Exception e)
+                {
+#line 22 "%%"
```
swernli, line 3694: What are these `#line 22 %%` markers for? I'm not familiar with that syntax.

ScottCarda-MS: This is so that the error message that gets throw is assigned a line number and source file position that is relevant to the Q# source code, rather than this C# generated code. The error will appear to come from the Q# because of this, rather than this C#, which is likely more useful to Q# developers who are debugging the error.

swernli: Oh, that's really cool. Thanks for explaining!

vadym-kl (approved):

PR #403, opened 2020-10-19
Move EmptyArray to QsharpFoundation

swernli:
Merging from master brought over `Microsoft.Quantum.Arrays.EmptyArray` which should be part of the QsharpFoundation in the decomposition refactor.


ScottCarda-MS (approved):
Seems reasonable

PR #404, opened 2020-10-20
Made Width and Depth compatible in current version of the tracer

DmitryVasilevsky:
Tracer component now produces compatible **"Width"** and **"Depth"** metrics, both computed in the depth counter. **"Width"** previously computed by the width counter is still reported as **"QubitCount"**. User can request width or depth optimized circuit to get compatible width and height in desired case. For example,
```
using(q = Qubit()) { T(q); }
using(q = Qubit()) { T(q); }
```
will result in Width=1, Depth=2 when OptimizeDepth is false, and Width=2, Depth=1 when OptimizeDepth is true. For compatibility **"QubitCount"** is kept and will still be reported as 1.


src/Simulation/EntryPointDriver.Tests/Tests.fs
```diff
@@ -486,6 +486,7 @@ let private resourceSummary =
      T               0   0
      Depth           0   0
      Width           1   1
+     ExtraWidth      1   1
```
bettinaheim, line 489: It's not clear to me what ExtraWidth is, and what the value is of tracking it?

src/Simulation/Simulators.Tests/ResourcesEstimatorTests.cs
```diff
@@ -166,14 +182,14 @@ public void VerifyTracingMultipleOperationsTest()
             Assert.Equal(1.0 + 0.0, data2.Rows.Find("T")["Sum"]);
             Assert.Equal(0.0 + 1.0, data2.Rows.Find("R")["Sum"]);
             Assert.Equal(0.0 + 1.0, data2.Rows.Find("Measure")["Sum"]);
-            Assert.Equal(2.0 + 3.0, data2.Rows.Find("Width")["Sum"]);
-            Assert.Equal(System.Math.Max(2.0, 3.0), data2.Rows.Find("Width")["Max"]);
+            Assert.Equal(2.0 + 3.0, data2.Rows.Find("ExtraWidth")["Sum"]);
```
bettinaheim, line 185: If Width is the metric that is compatible with Depth then it would make more sense to me to run the tracer once with depth optimization and once with width optimization and check the Depth and Width metrics for both cases instead of keeping the old numbers.

src/Simulation/Simulators/ResourcesEstimator/ResourcesEstimator.cs
```diff
@@ -75,10 +75,6 @@ public ResourcesEstimator(QCTraceSimulatorConfiguration config) : base(config)
             {
                 return null;
             }
-            else if (name == MetricsNames.WidthCounter.ExtraWidth)
```
bettinaheim, line 78: What was ExtraWidth in the old tracer? I don't think that was ever documented, but I may be wrong.

src/Simulation/QCTraceSimulator/QubitMetrics.cs
```diff
@@ -10,24 +10,14 @@ namespace Microsoft.Quantum.Simulation.QCTraceSimulatorRuntime
     /// </summary>
     public class QubitTimeMetrics
```
bettinaheim, line 11: Is this class still needed? It seems only the depth counter is using it, so it may make sense to absorb it into the depth counter rather than exposing it as part of the public API, especially since it now only contains the qubit Id.

src/Simulation/QCTraceSimulator/QubitAvailabilityTimeTracker.cs
```diff
@@ -0,0 +1,63 @@
...
+                QubitAvailableAt[(int)index] = value;
+            }
+        }
+
+        internal long GetMaxQubitId()
+        {
+            return MaxQubitId;
+        }
+
+        internal void MarkQubitIdUsed(long qubitId)
```
bettinaheim, line 56: Is it possible that this relies on exactly knowledge about the inner workings of the qubit manager, i.e. it relies on how the qubit manager currently allocates qubits?

bettinaheim:
The failing iqsharp build seems to be due to a test that simply needs to be updated.

src/Simulation/QCTraceSimulator/QubitAvailabilityTimeTracker.cs
```diff
@@ -0,0 +1,63 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
cgranade, line 1:
```suggestion
﻿// Copyright (c) Microsoft Corporation.
```
If I understand correctly (please confirm!), current guidance is to not include "All rights reserved" in the header for new source files.

src/Simulation/Simulators.Tests/ResourcesEstimatorTests.cs
```diff
@@ -120,24 +120,40 @@ public void DepthDifferentQubitsTest()
...
+
+            // Now run with depth optimization
+            data = RunDepthVersusWidthTest(true);
+            Assert.Equal(2.0, data.Rows.Find("T")["Sum"]);
+            Assert.Equal(1.0, data.Rows.Find("ExtraWidth")["Sum"]);
+            Assert.Equal(2.0, data.Rows.Find("Width")["Sum"]);
             Assert.Equal(1.0, data.Rows.Find("Depth")["Sum"]);
         }
 
+        private DataTable RunDepthVersusWidthTest(bool optimizeDepth) {
```
cgranade, line 148: [nit] Minor formatting comment, but generally the convention for C# sources is to have braces on the next line (though Q# and TypeScript still use braces on same line).
```suggestion
        private DataTable RunDepthVersusWidthTest(bool optimizeDepth)
        {

```

src/Simulation/Simulators/QCTraceSimulator/QCTraceSimulator.cs
```diff
@@ -98,6 +98,14 @@ public class QCTraceSimulatorConfiguration
         /// </summary>
         public bool UseWidthCounter = false;
 
+        /// <summary>
+        /// Constrols if depth or width optimization is favored.
+        /// If set to <c>true</c>, resulting circuit is optimized for depth by discouraging qubit reuse.
+        /// If set to <c>false</c>, resulting circuit is optimized for width be encouraging qubit reuse.
+        /// Optimization is only limited to reuse of qubits after they are released by user code.
+        /// </summary>
+        public bool OptimizeDepth;
```
cgranade, line 107: Should this have a default value, similar to other fields in the class?

src/Simulation/QCTraceSimulator/DepthCounter.cs
```diff
@@ -48,21 +52,27 @@ public class Metrics
...
+            return new QubitTimeMetrics(qubitId);
         }
 
         public void OnAllocate(object[] qubitsTraceData)
         {
+            QubitTimeMetrics[] qubitsMetrics = Utils.UnboxAs<QubitTimeMetrics>(qubitsTraceData);
+            foreach (QubitTimeMetrics metric in qubitsMetrics)
+            {
+                qubitAvailabilityTime.MarkQubitIdUsed(metric.QubitId);
+            }
```
cgranade, line 75: Is there a reason that `Utils.UnboxAs` is preferred to using `Enumerable.Cast`? E.g.:
```
            QubitTimeMetrics[] qubitsMetrics = qubitsTraceData.Cast<QubitTimeMetrics>().ToArray();
```
Relatedly, is an array needed at all here? The only usage is to immediately loop over its contents, such that you may be able to avoid allocating a new array:
```suggestion
            foreach (QubitTimeMetrics metric in qubitsTraceData.Cast<QubitTimeMetrics>())
            {
                qubitAvailabilityTime.MarkQubitIdUsed(metric.QubitId);
            }
```

cgranade (commented):
Thanks for opening this! I think the main thing that jumps out at me as a user of the trace simulator is that, as per @bettinaheim's comments, it's not really clear what ExtraWidth is and what that metric tracks about the execution of a quantum program.

src/Simulation/Simulators.Tests/ResourcesEstimatorTests.cs
```diff
@@ -120,24 +120,40 @@ public void DepthDifferentQubitsTest()
...
 
         /// <summary>
-        /// Documents that the width and depth statistics reflect independent lower
+        /// Documents that the ExtraWidth and Depth statistics reflect independent lower
         /// bounds for each (two T gates cannot be combined into a circuit of depth
-        /// one and width one).
+        /// one and width one). Width on the other hand is compatible with Depth.
         /// </summary>
         [Fact]
         public void DepthVersusWidthTest()
```
irinayat-MS, line 128: The current tests for width/depth (in `Simulators.Tests\Circuits\ResourcesEstimator.qs`) are rather simplistic. Would it make sense to add more coverage? For example, to cover operations that invoke other operations, etc.

src/Simulation/EntryPointDriver.Tests/Tests.fs
DmitryVasilevsky: I don't like this name either. This is what former "width" was called internally (see how it was renamed in ResourceEstimator.cs). We can come with a better name if necessary.
There's a value of tracking it (even if compatibility isn't an issue). Current width isn't a very useful metric for a single operation within execution, the old one is.

src/Simulation/QCTraceSimulator/DepthCounter.cs
```diff
@@ -23,11 +23,15 @@ class OperationCallRecord
             public double MaxOperationStartTime;
             public double ReleasedQubitsAvailableTime;
             public double ReturnedQubitsAvailableTime;
+            public long MaxQubitIdAtStart = -1;
```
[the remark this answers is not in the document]
DmitryVasilevsky: This is actually -1. One less than zero.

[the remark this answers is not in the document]
DmitryVasilevsky: As for API, I believe this is strictly private.

src/Simulation/QCTraceSimulator/DepthCounter.cs
```diff
@@ -23,11 +23,15 @@ class OperationCallRecord
...
             public QubitTimeMetrics[] InputQubitMetrics;
         }
 
         StatisticsCollector<CallGraphEdge> stats;
         Stack<OperationCallRecord> operationCallStack;
+        QubitAvailabilityTimeTracker qubitAvailabilityTime = new QubitAvailabilityTimeTracker(
+            initialCapacity: 128,
```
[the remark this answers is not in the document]
DmitryVasilevsky: It's initial capacity. Seems to be a reasonable number to pre-allocate for a simulation. Not to expensive and some simulations won't exceed this. If I'm not mistaken qubit manager reallocates 1024 qubits. Don't see any reason to make this something special.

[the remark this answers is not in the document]
irinayat-MS: any literal value is special. my point was that if there is already a notion of default qubit capacity elsewhere, we should probably use that, and if not, add a comment that the number was chosen effectively at random.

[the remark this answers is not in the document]
DmitryVasilevsky: At random, but from an interesting distribution :) Besides 128 is a nice number. Being close to the peak and a nice number does it.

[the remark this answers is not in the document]
DmitryVasilevsky: (I will add a comment)

src/Simulation/EntryPointDriver.Tests/Tests.fs
bettinaheim: @DmitryVasilevsky It would be good to understand why the width that is consistent with the depth is not a useful metric for a single operation? In particular, I think if you configure to optimize for width, I think you should recover the old width value (ExtraWidth), aren't you? I.e. if optimize for width is configured, aren't ExtraWidth and Width always the same? If not, where not and why?

src/Simulation/QCTraceSimulator/DepthCounter.cs
```diff
@@ -23,11 +23,15 @@ class OperationCallRecord
...
+            defaultAvailabilityTime: 0);
```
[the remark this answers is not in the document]
DmitryVasilevsky: Somehow I don't see benefits in case of 0, but I'll change it.

src/Simulation/QCTraceSimulator/QCTraceSimulatorCore.cs
```diff
@@ -31,6 +31,13 @@ public class QCTraceSimulatorCoreConfiguration
         /// of the call graph.
         /// </summary>
         public uint CallStackDepthLimit = uint.MaxValue;
+
+        /// <summary>
+        /// Constrols if depth or width optimization is favored.
+        /// If set to true, resulting circuit is optimized for depth by discouraging qubit reuse.
+        /// If set to false, resulting circuit is optimized for width be encouraging qubit reuse.
```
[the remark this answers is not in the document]
DmitryVasilevsky: Fixed.

src/Simulation/QCTraceSimulator/QubitAvailabilityTimeTracker.cs
```diff
@@ -0,0 +1,63 @@
...
+    internal class QubitAvailabilityTimeTracker
+    {
+        private double DefaultAvailabilityTime = 0;
+        private List<double> QubitAvailableAt;
+        private long MaxQubitId = -1;
+
+        internal QubitAvailabilityTimeTracker(int initialCapacity, double defaultAvailabilityTime)
+        {
+            DefaultAvailabilityTime = defaultAvailabilityTime;
+            QubitAvailableAt = new List<double>(initialCapacity);
```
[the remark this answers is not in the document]
DmitryVasilevsky: This is only the initial capacity, a memory reservation. It still contains no elements and the size is 0.

src/Simulation/QCTraceSimulator/QubitAvailabilityTimeTracker.cs
```diff
@@ -0,0 +1,63 @@
...
+        }
+
+         internal double this[long index]
```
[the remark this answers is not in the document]
DmitryVasilevsky: Fixed.

src/Simulation/QCTraceSimulator/QubitAvailabilityTimeTracker.cs
```diff
@@ -0,0 +1,63 @@
...
+                return DefaultAvailabilityTime;
+            }
+            set
+            {
+                if (index == QubitAvailableAt.Count)
+                {
+                    QubitAvailableAt.Add(value);
+                    return;
+                }
+                else if (index >= int.MaxValue)
```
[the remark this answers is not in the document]
DmitryVasilevsky: Qubit manager uses long for qubit ids. I used List<> to store information related to qubit ids. List is indexed by int, and this where the limitation comes from. Somewhere there has to be a conversion. This class is the single place where conversion and checks are made.

src/Simulation/QCTraceSimulator/QubitAvailabilityTimeTracker.cs
DmitryVasilevsky: It does not rely on qubit manager inner workings in general. It only relies on the fact that the same id will be assigned to the same underlying qubit that qubit manager manages. I.e. If a program does allocate-release-allocate-release, and qubit manager assigns the same underlying qubit for the second allocate-release cycle, it will have same id as in the first. This information (detecting reuse of qubits) is crucial for computing compatible depth and width. However, qubit _object_ generated for the second round may be different.

src/Simulation/QCTraceSimulator/QubitMetrics.cs
DmitryVasilevsky: I would definitely like to get rid of this class entirely, but I cannot because of IQCTraceSimulatorListener. This is a public interface used by many listeners including depth counter. It has "OnPrimitiveOperation" method, and qubits or qubit ids are not passed to this method, only attached metrics. I need ids in that method that's why I put them into metrics.
The new approach to tracer has this changed so it is easier to add qubit tracking there.

src/Simulation/Simulators.Tests/ResourcesEstimatorTests.cs
```diff
@@ -120,24 +120,40 @@ public void DepthDifferentQubitsTest()
...
         {
-            var sim = new ResourcesEstimator();
-
+            // Operation to execute:
             // using(q = Qubit()) { T(q); } using(q = Qubit()) { T(q); } (yes, twice)
-            DepthVersusWidth.Run(sim).Wait();
-            var data = sim.Data;
 
+            // First run with width optimization
+            DataTable data = RunDepthVersusWidthTest(false);
```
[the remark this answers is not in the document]
DmitryVasilevsky: Added named argument "optimizeDepth:".

src/Simulation/Simulators.Tests/ResourcesEstimatorTests.cs
DmitryVasilevsky: Updated.

src/Simulation/Simulators.Tests/ResourcesEstimatorTests.cs
DmitryVasilevsky: There's a DepthVersusWidthTest for exactly that purpose. This test is for tracking width in subroutines.

src/Simulation/Simulators/QCTraceSimulator/QCTraceSimulator.cs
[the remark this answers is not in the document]
DmitryVasilevsky: Good question. This translates directly to QubitManager option EncourageReuse so I decided to use just bool. At first I wanted to name it just EncourageQubitReuse, but it would have been a different layer of abstraction. So I named it OptimizeDepth as this was the major driver for this feature. Now that you pointed this out, this level of abstraction may have its own value and makes sense.

DmitryVasilevsky: A result of playing with different settings. Thanks for catching. Fixed.

src/Simulation/QCTraceSimulator/DepthCounter.cs
```diff
@@ -48,21 +52,27 @@ public class Metrics
         {
             public const string Depth = "Depth";
             public const string StartTimeDifference = "StartTimeDifference";
+            public const string Width = "Width";
```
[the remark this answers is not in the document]
DmitryVasilevsky: It wasn't static, but it can be made static and similar class in WidthCounter can also be made static. Updated.

src/Simulation/QCTraceSimulator/DepthCounter.cs
```diff
@@ -93,7 +103,8 @@ public void OnOperationEnd(object[] returnedQubitsTraceData)
             double[] metrics =
                 StatisticsRecord( 
                     Depth : operationEndTime - opRec.MaxOperationStartTime,
-                    StartTimeDifference: opRec.MaxOperationStartTime - opRec.MinOperationStartTime );
+                    StartTimeDifference: opRec.MaxOperationStartTime - opRec.MinOperationStartTime,
+                    Width: qubitAvailabilityTime.GetMaxQubitId() - opRec.MaxQubitIdAtStart );
```
irinayat-MS, line 107: I am confused here. If the QubitManager was configured to reuse qubits it's possible that the operation reuses all of the qubits it allocates, meaning the Width will be calculated as zero. But this doesn't mean the operation's width is zero, right?

src/Simulation/QCTraceSimulator/DepthCounter.cs
DmitryVasilevsky: Just moved this code from elsewhere. Enumerable is a good improvement!

src/Simulation/QCTraceSimulator/QubitAvailabilityTimeTracker.cs
```diff
@@ -0,0 +1,63 @@
...
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+
+namespace Microsoft.Quantum.Simulation.QCTraceSimulatorRuntime
+{
+    internal class QubitAvailabilityTimeTracker
+    {
+        private double DefaultAvailabilityTime = 0;
+        private List<double> QubitAvailableAt;
```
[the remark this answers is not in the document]
irinayat-MS: Actually, I believe it tracks the last time the qubit was used...

src/Simulation/QCTraceSimulator/QubitAvailabilityTimeTracker.cs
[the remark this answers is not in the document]
irinayat-MS: Ah. OK.

src/Simulation/QCTraceSimulator/DepthCounter.cs
DmitryVasilevsky: Correct about calculated with = 0. This is the ongoing discussion about the definition of Width, AddedWidth and how they should be documents.

src/Simulation/QCTraceSimulator/QubitAvailabilityTimeTracker.cs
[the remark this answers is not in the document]
DmitryVasilevsky: Yes.

src/Simulation/Simulators/ResourcesEstimator/ResourcesEstimator.cs
DmitryVasilevsky: I will resolve this in favor of the conversation in the beginning of the output, which discusses the name of this metrics. I'm inclining to make it LocalWidth.

src/Simulation/QCTraceSimulator/QubitAvailabilityTimeTracker.cs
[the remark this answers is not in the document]
irinayat-MS: wouldn't this indicate that using List to store the qubits' times might not be the best choice? implicitly, it relies on the QubitManager allocating the qubits more or less densily starting at zero, but I can imagine a QubitManager that might want to use the upper word in the `long` of the id to pack some additional information about a qubit, or randomizing the ids, or whatever other crazy idea someone inevitably will come up with, unless they are restrained by a spec :)

src/Simulation/QCTraceSimulator/QubitAvailabilityTimeTracker.cs
DmitryVasilevsky: This is a rename of the file, not a new one. Confirmed that "All rights reserved." should not be included. I will remove it.

src/Simulation/QCTraceSimulator/QubitAvailabilityTimeTracker.cs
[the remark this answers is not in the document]
DmitryVasilevsky: If someone decides to write a very special qubit manager one would want to write a very special tracer.

src/Simulation/EntryPointDriver.Tests/Tests.fs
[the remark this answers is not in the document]
DmitryVasilevsky: I will resolve this in favor of the general discussion on how this should be named and documented. ExtraWidth is a not the right name.

src/Simulation/Simulators/QCTraceSimulator/QCTraceSimulator.cs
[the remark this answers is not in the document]
DmitryVasilevsky: On the closer consideration - it will complicate code. This value is passed through layers (QCTraceSimulatorConfiguration, QCTraceSimulatorCoreConfiguration) and user of the top layer may not need/want to know about low layer. So either two emums need to be introduced and converted or the user of the top layer will have to know about the low layer. Existing bool is simple and reasonable enough.

src/Simulation/QCTraceSimulator/DepthCounter.cs
```diff
@@ -105,33 +116,57 @@ public void OnOperationStart(HashedString name, OperationFunctor functorSpeciali
...
+        private double MaxAvailableTime(QubitTimeMetrics[] qubitTimeMetrics)
+        {
+            Debug.Assert(qubitTimeMetrics != null);
+            double max = 0;
+            foreach (QubitTimeMetrics metric in qubitTimeMetrics)
+            {
+                max = Max(max, qubitAvailabilityTime[metric.QubitId]);
+            }
+            return max;
+        }
```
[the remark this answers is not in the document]
DmitryVasilevsky: foreach makes it clear what's going on here instead of obscure methods and lambda function to compute per element. I will consider using enumerable though to avoid some array materialization in callers.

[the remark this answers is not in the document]
cgranade: I'm not sure I'd refer to core C# language features as obscure, but fair enough. Mostly my comment here was with respect to this was the duplicated code as a result.

src/Simulation/QCTraceSimulator/QubitAvailabilityTimeTracker.cs
```diff
@@ -0,0 +1,63 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+
+namespace Microsoft.Quantum.Simulation.QCTraceSimulatorRuntime
+{
+    internal class QubitAvailabilityTimeTracker
```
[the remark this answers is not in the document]
DmitryVasilevsky: Qubit Availability Time is the term that was used in the code. Time is the noun, it seems to be necessary.

src/Simulation/QCTraceSimulator/QubitAvailabilityTimeTracker.cs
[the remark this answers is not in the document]
DmitryVasilevsky: Added comments.

src/Simulation/Simulators.Tests/ResourcesEstimatorTests.cs
```diff
@@ -155,7 +171,7 @@ public void VerifyTracingMultipleOperationsTest()
             Assert.Equal(1.0, data1.Rows.Find("T")["Sum"]);
             Assert.Equal(0.0, data1.Rows.Find("R")["Sum"]);
             Assert.Equal(0.0, data1.Rows.Find("Measure")["Sum"]);
-            Assert.Equal(2.0, data1.Rows.Find("Width")["Sum"]);
+            Assert.Equal(2.0, data1.Rows.Find("ExtraWidth")["Sum"]);
```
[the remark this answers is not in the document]
DmitryVasilevsky: This tests were made for local width of operations. I'm keeping them for that purpose. ExtraWidth will be renamed.

src/Simulation/EntryPointDriver.Tests/Tests.fs
DmitryVasilevsky: I suggest QubitCount instead of ExtraWidth.

src/Simulation/Simulators.Tests/ResourcesEstimatorTests.cs
DmitryVasilevsky: I already modified DepthVersusWidthTest. I can add another test.

src/Simulation/EntryPointDriver.Tests/Tests.fs
bettinaheim: Sounds good. As per our discussion, I also assume that QubitCount will only show up in the CSV, but for the summary e.g. given by the ResourcesEstimator, only Depth and Width would be listed (applicable to the entire program).

DmitryVasilevsky: I've prepared a short summary describing Width, Depth and QubitCount.
I need to check the code for where QubitCount will show up.

src/Simulation/EntryPointDriver.Tests/Tests.fs
```diff
@@ -486,6 +486,7 @@ let private resourceSummary =
      T               0   0
      Depth           0   0
      Width           1   1
+     QubitCount      1   1
```
bettinaheim, line 489: This would necessarily be the same as Width in this case. Do you see any case where this would be different? 
If they are the same, could you please remove it to avoid unnecessary confusion? I.e. just looking at this, people might expect this to be the total number of qubits use (without reuse), which it isn't.

src/Simulation/QCTraceSimulator/DepthCounter.cs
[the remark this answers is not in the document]
bettinaheim: @DmitryVasilevsky What do you mean by this being private? It is part of a public field in a public class, so it is indeed part of the API. Would it make sense to make this field private or internal?

src/Simulation/QCTraceSimulator/QubitAvailabilityTimeTracker.cs
bettinaheim: My concern was that it might rely e.g. on consecutiveness of the numbering. But looking closer I believe this is not the case, correct?

DmitryVasilevsky: It does not rely on consecutiveness. But it's performance relies on IDs being a dense set.

src/Simulation/EntryPointDriver.Tests/Tests.fs
bettinaheim: Thanks! Just one last comment in that regard, see here: https://github.com/microsoft/qsharp-runtime/pull/404#discussion_r509773163

src/Simulation/QCTraceSimulator/DepthCounter.cs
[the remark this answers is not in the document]
DmitryVasilevsky: This is a nested class inside DepthCounter. It doesn't have any references from outside. It is only used as a type for a private field in DepthCounter. I don't see why this should be public. I'll make it private.
Having said that, this field is just as good (or bad) as others.

src/Simulation/EntryPointDriver.Tests/Tests.fs
DmitryVasilevsky: Yes it may be different. When we optimize for Depth, QubitCount will be smaller than Width. Also QubitCount does count borrowed qubits as extra qubits, but Width does not, as they are the same underlying qubits.

bettinaheim: @DmitryVasilevsky That borrowed qubits are included in QubitCount is important to document. Thanks!

bettinaheim (approved):
Thanks for adding this!

src/Simulation/QCTraceSimulator/QCTraceSimulatorCore.cs
```diff
@@ -30,6 +30,13 @@ public class QCTraceSimulatorCoreConfiguration
         /// of the call graph.
         /// </summary>
         public uint CallStackDepthLimit = uint.MaxValue;
+
+        /// <summary>
+        /// Constrols if depth or width optimization is favored.
```
avasch01, line 35: Controls

DmitryVasilevsky: Updated. Here and in another configuration class.

src/Simulation/Simulators/QCTraceSimulator/QCTraceSimulator.cs
```diff
@@ -98,6 +98,14 @@ public class QCTraceSimulatorConfiguration
         /// </summary>
         public bool UseWidthCounter = false;
 
+        /// <summary>
+        /// Constrols if depth or width optimization is favored.
```
avasch01: Controls

DmitryVasilevsky: Already fixed.

src/Simulation/QCTraceSimulator/QCTraceSimulatorCore.cs
```diff
@@ -30,6 +30,13 @@ public class QCTraceSimulatorCoreConfiguration
...
+        /// If set to true, resulting circuit is optimized for depth by discouraging qubit reuse.
+        /// If set to false, resulting circuit is optimized for width by encouraging qubit reuse.
+        /// </summary>
+        public bool OptimizeDepth = false;
     }
```
avasch01, line 40: I would call it OptimizeQubitReuseForDepth. Otherwise it is confusing. Reader immediately starts thinking about global algorithm optimization for depth.

avasch01 (approved):

src/Simulation/QCTraceSimulator/QCTraceSimulatorCore.cs
DmitryVasilevsky: I'd rather keep it as OptimizeDepth. It is described here in the comments and in the documentation so confusion should be minimal.

src/Simulation/EntryPointDriver.Tests/Tests.fs
DmitryVasilevsky: Addressed.

PR #406, opened 2020-10-22
Updating manifest for signed builds.

anpaz:


build/manifest.ps1
```diff
@@ -13,7 +27,8 @@
         "Microsoft.Quantum.Runtime.Core",
         "Microsoft.Quantum.Simulators",
         "Microsoft.Quantum.Xunit"
-    );
+    ) | ForEach-Object { Join-Path $Env:NUGET_OUTDIR "$_.$Env:NUGET_VERSION.nupkg" };
```
rmshaffer, line 30: Do the `.snupkg` files also need to be signed, if they exist? Should they be added here somehow? (This comment also applies to similar changes across all the repos.)

bettinaheim (approved):
Seems fine, aside from the build failure that still needs to be resolved.

build/manifest.ps1
anpaz: I think they do. Practically it doesn't matter as for nuget packages, as opposed to dlls, we submit for signing everything that we found in the NUGET_OUTDIR folder (which includes the `.snupkg`s). So, this is not really needed for signing.
We do use it to validate that the package is created on local builds. So for now, I'm just going to leave it like this.

PR #409, opened 2020-10-23
Native Simulator refactors (mostly in wavefunction.hpp)

irinayat-MS:
Format, unify naming, comment, document logical vs positional ids, remove some of the unnecessary copying -- but it has grown to be unreviewable, so creating the PR for reference only and will submit individual smaller refactors as separate PRs.

### DIFF
--- a/src/Simulation/Native/src/simulator/wavefunction.hpp
+++ b/src/Simulation/Native/src/simulator/wavefunction.hpp
@@ -10,138 +10,305 @@
 #include <iostream>
 #include <iterator>
 #include <limits>
+#include <list>
 #include <random>
-#include <vector>
-#include <unordered_set>
-#include <unordered_map>
 #include <string.h>
+#include <unordered_map>
+#include <unordered_set>
+#include <vector>
 
-#include "types.hpp"
 #include "gates.hpp"
+#include "types.hpp"
 
 #include "external/fused.hpp"
 
 namespace Microsoft
 {
-    namespace Quantum
+namespace Quantum
+{
+namespace SIMULATOR
+{
+namespace detail
+{
+inline std::size_t get_register(const std::vector<positional_qubit_id>& qs, std::size_t basis_state)
+{
+    std::size_t result = 0;
+    for (unsigned i = 0; i < qs.size(); ++i)
+        result |= ((basis_state >> qs[i]) & 1) << i;
+    return result;
+}
+
+inline std::size_t set_register(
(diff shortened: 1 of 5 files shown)


src/Simulation/Native/src/simulator/wavefunction.hpp
```diff
@@ -10,138 +10,305 @@
...
+                std::set_difference(
+                    nextQs.begin(), nextQs.end(), // Figure out if any of the future qubits aren't already seen by us
+                    myTouched.begin(), myTouched.end(), std::back_inserter(myDiff));
+                allInter.clear();
+                std::set_intersection(
+                    myDiff.begin(), myDiff.end(), // These are any new qubits that might have already been touched
+                    allTouched.begin(), allTouched.end(), std::back_inserter(allInter));
+                if (allInter.size() == 0)
+                { // If the new qubits are untouched... then this is allowed
+                    auto ret = std::make_pair(std::move(*it), myUnion);
```
swernli, line 183: This comment got lost in the refactor. I'm not sure if the word "Remove" is still appropriate on this specific line, since the erase happens on the line below, but I think preserving the idea will be helpful.
```suggestion
                    auto ret = std::make_pair(std::move(*it), myUnion); // Remove the future cluster and add it our cluster.
```

swernli (commented):
This clean up looks good! One thing I'd like to see before signoff is some detail on the perf impact, if any, and I have some tips on how to verify that that I'll add in stand-alone comments below.

swernli:
To verify the perf impact of these changes, there are two tests that you can use. 

One is in [dbw_test.cpp](https://github.com/microsoft/qsharp-runtime/blob/main/src/Simulation/Native/src/simulator/dbw_test.cpp#L1) and runs as part of the native tests executed by build\test.ps1 in this repo. The output from that test is the gates per second (GPS) of the loaded advantage circuit log when run through the simulator. You could try running the test once without your changes and once with your changes and compare the the difference, if any, in GPS on your machine.

The other test mechanism you can use is not a literal test, but rather the code in the AdvantageBenchmark folder at the root of the repo. There is a readme there that goes through what the benchmark is and how the code is expected to work. Right now, it's set up to compare an August beta build (0.12.20082705-beta) to the current tree. You could either just run the benchmark for the current tree in privateBuild without your changes and then again with your changes, or alternatively update the release build to point to the October release and then compare the that side by side with your changes. Either way, this should give a good end-to-end validation of the perf.

PR #413, opened 2020-10-26
Document logical versus positional qubit ids

irinayat-MS:
The bulk of the PR is adding new `using` declarations that document intent of the code wrt logical qubit ids versus the indexes of qubits in the wave function's basis (which I dubbed 'positional qubit id'). The positional id is an implementation detail and shouldn't be used outside of the wave function class so removed it from simulator and replaced the simulator test with a direct test of wave function. The PR also contains a couple renames to improve readability/consistency but there should be no functionality changes.


swernli (approved):
LGTM!

PR #414, opened 2020-10-26
Updates/fixes to decomposition feature branch

swernli:
This change merges in main, and also fixes two issues with decompositions:
 - The Reset capability fix was lost in the previous merge from master.
 - RFrac should allow negative powers of 2.

Both of these bugs were found during e2e build failure investigation.


swernli:
To view just the fixes without the merge from main, see the last two commits:
[Fix Reset decomposition capability requirement](https://github.com/microsoft/qsharp-runtime/pull/414/commits/a365c5f51fd73ec5e252a2577aeefda5e4c0e371)
and
[Fix RFrac decomposition to allow negative exponentiation](https://github.com/microsoft/qsharp-runtime/pull/414/commits/af1ae7391dcd1fcc4e6bb578f6b18d8132ffb3c2)

src/Simulation/TargetDefinitions/Decompositions/RFrac.qs
```diff
@@ -41,7 +41,7 @@ namespace Microsoft.Quantum.Intrinsic {
     /// ```
     @EnableTestingViaName("Test.TargetDefinitions.RFrac")
     operation RFrac (pauli : Pauli, numerator : Int, power : Int, qubit : Qubit) : Unit is Adj + Ctl {
-        let angle = ((-2.0 * PI()) * IntAsDouble(numerator)) / IntAsDouble(2 ^ power);
+        let angle = ((-2.0 * PI()) * IntAsDouble(numerator)) / (2.0 ^ IntAsDouble(power));
```
cesarzc, line 44: Was it not compiling before this fix or was it doing something undesired at runtime?

swernli: There is a test in one of the other repos that executes a full spread of inputs to `RFrac` in order to verify the behavior, and it was crashing on negative exponents. It turns out that there are two implementations of the `Pow` extension, [one that operates on integers](https://github.com/microsoft/qsharp-runtime/blob/main/src/Simulation/Core/Extensions.cs#L41) and will throw if it is given a negative power (since it must return an integer), and [another that operates on doubles](https://github.com/microsoft/qsharp-runtime/blob/main/src/Simulation/Core/Extensions.cs#L36) and can support negative exponents since it can then return fractional results. So 
```qsharp
2 ^ -1
```
will fail with an exception, while
```qsharp
2.0 ^ -1.0
```
succeeds. Looking back, this is the same as the implementation of `RFrac` that is used by `Microsoft.Quantum.Intrinsic` today, and I think the only reason this was working before was because this operation was being overridden for certain execution targets, which affected the tests by side effect. With the new target packages and updates to the override behavior in the runtime this implementation was actually being called for the first time in the test and it found the bug.

cesarzc: Thanks for the explanation Stefan!

cesarzc (approved):

PR #415, opened 2020-10-27
Fixing negative exponentiation in ExpFrac

swernli:
Turns out I forgot to fix the same bug in ExpFrac as I already fixed in RFrac (see #414).

### DIFF
--- a/src/Simulation/TargetDefinitions/Decompositions/ExpFrac.qs
+++ b/src/Simulation/TargetDefinitions/Decompositions/ExpFrac.qs
@@ -31,6 +31,8 @@ namespace Microsoft.Quantum.Intrinsic {
     /// Register to apply the given rotation to.
     @EnableTestingViaName("Test.TargetDefinitions.ExpFrac")
     operation ExpFrac (paulis : Pauli[], numerator : Int, power : Int, qubits : Qubit[]) : Unit is Adj + Ctl {
+        // Note that power must be converted to a double and used with 2.0 instead of 2 to allow for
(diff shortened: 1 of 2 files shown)


src/Simulation/TargetDefinitions/Decompositions/ExpFrac.qs
```diff
@@ -31,7 +31,7 @@ namespace Microsoft.Quantum.Intrinsic {
     /// Register to apply the given rotation to.
     @EnableTestingViaName("Test.TargetDefinitions.ExpFrac")
     operation ExpFrac (paulis : Pauli[], numerator : Int, power : Int, qubits : Qubit[]) : Unit is Adj + Ctl {
-        let angle = (PI() * IntAsDouble(numerator)) / IntAsDouble(2 ^ power);
+        let angle = (PI() * IntAsDouble(numerator)) / (2.0 ^ IntAsDouble(power));
```
irinayat-MS, line 34: It wasn't clear to me how this change was related to negative exponentiation. Luckily, Cesar had the same question when you fixed RFrac and you gave him an excellent explanation :) I wonder however, whether it's worth a comment here _in the code_?

irinayat-MS (approved):

src/Simulation/TargetDefinitions/Decompositions/ExpFrac.qs
swernli: Seems like a good idea :)

PR #416, opened 2020-10-27
Update the embedded resource logical name

cesarzc:
This change updates the embedded resource logical name which reflects the change from Newtonsoft to Bond.

### DIFF
--- a/src/Quantum.Development.Kit/Props/QSharp.targets
+++ b/src/Quantum.Development.Kit/Props/QSharp.targets
@@ -24,7 +24,7 @@
     <Delete Files="@(QsharpCodegenFiles)" />
     <Exec Command="$(QscCommand)" />
     <ItemGroup>
-      <EmbeddedResource Include="$(QsharpOutDir)/$(QsharpTree).bson" LogicalName="__qsharp_data__.bson" Visible="false" />
+      <EmbeddedResource Include="$(QsharpOutDir)/$(QsharpTree).bson" LogicalName="__qsharp_data_bond_fast_binary__.bson" Visible="false" />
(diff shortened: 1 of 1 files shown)


src/Quantum.Development.Kit/Props/QSharp.targets
```diff
@@ -24,7 +24,7 @@
     <Delete Files="@(QsharpCodegenFiles)" />
     <Exec Command="$(QscCommand)" />
     <ItemGroup>
-      <EmbeddedResource Include="$(QsharpOutDir)/$(QsharpTree).bson" LogicalName="__qsharp_data__.bson" Visible="false" />
+      <EmbeddedResource Include="$(QsharpOutDir)/$(QsharpTree).bson" LogicalName="__qsharp_data_bond_fast_binary__.bson" Visible="false" />
```
bettinaheim, line 27: Do we ever intend to have a __qsharp_data_bond_slow_binary__? ;) Maybe __qsharp_data_bond__ might be enough specific. If you think we might need to change the bond serialization significantly enough to so something similar to this in the future, you could also go with something like a version number for the resource (e.g. __qsharp_data_bond_v1__).

bettinaheim (approved):

src/Quantum.Development.Kit/Props/QSharp.targets
cesarzc: I don't expect the serialization protocol to change but maybe the schema could change in such a way that it is incompatible with previous versions. Your suggestion of using a version number in the resource name gives us the flexibility we need in case the schema changes so I'll rename the resource to **qsharp_data_bond_v1**.

PR #417, opened 2020-10-27
Native Simulator: introduce catch2 test framework

irinayat-MS:
Native Simulator is lacking convenient testing facilities. I'm using catch2 for QirRuntime and so far has been happy with it. Catch2 also integrates well with CMake and CI. 

Note: I've set it up to allow coexistence of old-style tests and the new ones so we can migrate the tests over as needed and time permits.


swernli (approved):
Looks good! I was also impressed with catch2 in the QIR changes you worked on, so happy to see it incorporated here.

PR #418, opened 2020-10-28
Native Simulator: fix obviously unnecessary copying of large structures in wave function

irinayat-MS:
Replace passing by value of `GateWraper`s and `Cluster`s with passing by const& when read-only access is required. We still copy instances of these two classes _a lot_ and there will be one more refactor to address that.

### DIFF
--- a/src/Simulation/Native/src/simulator/wavefunction.hpp
+++ b/src/Simulation/Native/src/simulator/wavefunction.hpp
@@ -6,14 +6,11 @@
 #include <cassert>
 #include <complex>
 #include <ctime>
-#include <fstream>
 #include <iostream>
 #include <iterator>
 #include <limits>
 #include <random>
 #include <string.h>
-#include <unordered_map>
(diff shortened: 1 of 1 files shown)


swernli (approved):
These changes look good, and I expect that the existing native tests would fail on clustering bugs (as they did when this infrastructure was introduced over the summer), so if those tests are passing that gives us good confidence on correctness. I would still like @dbwz8 to take a look to make sure, but this seems like it's ready to me!

irinayat-MS:
> These changes look good, and I expect that the existing native tests would fail on clustering bugs (as they did when this infrastructure was introduced over the summer), so if those tests are passing that gives us good confidence on correctness. I would still like @dbwz8 to take a look to make sure, but this seems like it's ready to me!

Thank you for reviewing! I'm not concerned about this particular batch of changes because they are mostly mechanical. The next refactor of clustering is more meaningful and I will certainly involve Dave.

PR #420, opened 2020-10-28
Native Simulator: refactor clustering algorithm to copy less memory around (plus a bug fix)

irinayat-MS:
The change wasn't meant to improve the clustering algorithm, only to eliminate inefficiencies related to how it accessed the pertinent data-structures. However, have discovered a bug that might cause some pending gates to be dropped if the fuse limit was hit. The bug is documented by two of the new tests and is fixed in the last commit of this PR.

Local runs of dbw_test don't show much difference in GPS. The numbers from five runs before/after the change are (for the sake of space, the last loop only):

| Loops[   9] before | Loops[   9] after | 
| :------------- | :----------: | 
| GPS = 1.09e+04 | GPS = 1.29e+04 |
| GPS = 1.28e+04 | GPS = 1.34e+04 |
| GPS = 9.92e+03 | GPS = 1.13e+04 |
| GPS = 1.36e+04 | GPS = 1.16e+04 |
| GPS = 1.22e+04 | GPS = 1.26e+04 |


dbwz8 (approved):
Nice clean code. Good comments.

swernli (approved):
Great stuff! Thanks for including the dbw_test output table in the PR as well.

PR #421, opened 2020-10-28
Swernli/merge/feature/decomp

swernli:
This merges in the payload from feature/decomp, which includes the accumulated work for the targeting feature described in #249. These changes have been reviewed and tested e2e from the feature branch.

Several major changes included here:
- Microsoft.Quantum.Qsharp.Core.dll has some functionality split out into Microsoft.Quantum.Qsharp.Foundation.dll, which has it's own package that forms the basis for any target package.
- Microsoft.Quantum.Type1.Core.dll and Microsoft.Quantum.Type2.Core.dll are introduced as alternative target packages that make use of the Qsharp.Foundation but utilize a different set of decompositions and intrinsic gates.
- Each target package has it's own simulator, where currently the Type1 and Type2 simulators are not directly used anywhere other than in tests.
- The implementation of gate decompositions are broken out on a file-by-file basis inside of src/Simulation/TargetDefinitions to allow the rapid addition or recombination of these pieces into new target packages.

The goal is to introduce this new code hierarchy and design now so it can be extended and incorporated into the QIR work, while not changing any requirements or dependencies in a way that would cause breaking changes for developers building on top of the QDK and the existing full state simulator.


src/Simulation/Simulators.Impl/QuantumSimulator/SWAP.cs
```diff
@@ -0,0 +1,60 @@
...
+                this.Simulator = m;
+            }
+
+            public override Func<(Qubit, Qubit), QVoid> __Body__ => (args) =>
+            {
+                var (qubit1, qubit2) = args;
+                var ctrls = new QArray<Qubit>(qubit1);
+                Simulator.CheckQubits(ctrls, qubit2);
+
+                MCX(Simulator.Id, (uint)ctrls.Length, ctrls.GetIds(), (uint)qubit2.Id);
```
swernli, line 31: This is a bug in the native SWAP support that is fixed as part of #383, which will need to be retargeted as a PR against main once this PR is merged.

src/Simulation/TargetDefinitions/TargetPackages/Common.Package.props
```diff
@@ -0,0 +1,43 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <Import Project="..\..\Common\AssemblyCommon.props" />
+  <Import Project="..\..\Common\DebugSymbols.props" />
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.1</TargetFramework>
+    <QsharpDocsGeneration>true</QsharpDocsGeneration>
```
swernli, line 9: @cgranade Because this common props file is used by all of the targeting packages, I assume this means that we end up re-generating the docs for commonly defined intrinsics (like `X`) more than once. How does this affect the doc generation pipeline? Should I move this so that it is only true for Qsharp.Core and let that be the single source of generated docs?

build/manifest.ps1
```diff
@@ -37,9 +41,14 @@ $artifacts = @{
...
         ".\src\Simulation\QsharpCore\bin\$Env:BUILD_CONFIGURATION\netstandard2.1\Microsoft.Quantum.QSharp.Core.dll",
+        ".\src\Simulation\Type1Core\bin\$Env:BUILD_CONFIGURATION\netstandard2.1\Microsoft.Quantum.Type1.Core.dll",
+        ".\src\Simulation\Type2Core\bin\$Env:BUILD_CONFIGURATION\netstandard2.1\Microsoft.Quantum.Type2.Core.dll",
+        ".\src\Simulation\QsharpFoundation\bin\$Env:BUILD_CONFIGURATION\netstandard2.1\Microsoft.Quantum.QSharp.Foundation.dll",
         ".\src\Simulation\Simulators\bin\$Env:BUILD_CONFIGURATION\netstandard2.1\Microsoft.Quantum.Simulation.Common.dll",
         ".\src\Simulation\Simulators\bin\$Env:BUILD_CONFIGURATION\netstandard2.1\Microsoft.Quantum.Simulation.QCTraceSimulatorRuntime.dll",
         ".\src\Simulation\Simulators\bin\$Env:BUILD_CONFIGURATION\netstandard2.1\Microsoft.Quantum.Simulators.dll",
+        ".\src\Simulation\Simulators.Type1\bin\$Env:BUILD_CONFIGURATION\netstandard2.1\Microsoft.Quantum.Simulators.Type1.dll",
+        ".\src\Simulation\Simulators.Type2\bin\$Env:BUILD_CONFIGURATION\netstandard2.1\Microsoft.Quantum.Simulators.Type2.dll",
```
swernli, line 51: @anpaz These are new binaries that will start being produced and packed from this change. Do I need to do anything to add these to the signing handling you've built?

src/Simulation/TargetDefinitions/Intrinsic/Checks.qs
```diff
@@ -0,0 +1,38 @@
...
+    ///
+    /// # Description
+    /// Will fail if any qubits are repeated or null.
+    ///
+    /// # Input
+    /// ## qubits
+    /// The array of qubits to verify for uniqueness. In the controlled variant
+    /// the full list of qubits among targets and controls are verified to be unique.
+    @EnableTestingViaName("Test.TargetDefinitions.CheckQubitUniqueness")
+    internal operation CheckQubitUniqueness (qubits : Qubit[]) : Unit is Adj + Ctl{
```
bettinaheim, line 18: How come this is an operation? I believe it should always return the same result given the same input, right?

src/Simulation/TargetDefinitions/TargetPackages/QsharpCore.Package.props
```diff
@@ -0,0 +1,57 @@
+<?xml version="1.0" encoding="utf-8"?>
```
bettinaheim, line 1: I quite like how the used quantum instruction set becomes quite apparent this way.

src/Simulation/Type1Core/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,9 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+// Allow the simulator assembly to use our internal methods
+[assembly: InternalsVisibleTo("Microsoft.Quantum.Simulators.Type1" + SigningConstants.PUBLIC_KEY)]
```
bettinaheim, line 9: Why do the internals need to be visible to the simulator again? Because it needs to implement things that are intrinsic in Q#?

src/Simulation/TargetDefinitions/TargetPackages/Type1.Package.props
```diff
@@ -0,0 +1,61 @@
+<?xml version="1.0" encoding="utf-8"?>
```
bettinaheim, line 1: A very short readme in the TargetDefinitions folder would be useful such that people are not confused by the expectation that all files in the Decomposition and/or Intrinsic folder should belong to the same package.

src/Simulation/EntryPointDriver/Microsoft.Quantum.EntryPointDriver.csproj
```diff
@@ -26,7 +26,6 @@
   </ItemGroup>
 
   <ItemGroup>
-    <ProjectReference Include="..\Core\Microsoft.Quantum.Runtime.Core.csproj" />
```
bettinaheim, line 29: Why is this no longer needed when it was needed before (not sure why it was needed before)?

src/Simulation/QCTraceSimulator.Tests/Tests.Microsoft.Quantum.Simulation.QCTraceSimulatorRuntime.csproj
```diff
@@ -13,6 +13,7 @@
   <ItemGroup>
     <ProjectReference Include="..\..\Xunit\Microsoft.Quantum.Xunit.csproj" />
     <ProjectReference Include="..\CsharpGeneration\Microsoft.Quantum.CsharpGeneration.fsproj" IsQscReference="true" />
+    <ProjectReference Include="..\Simulators\Microsoft.Quantum.Simulators.csproj" />
```
bettinaheim, line 16: I don't have a good overview just looking at this PR how the project/package structure now works: Wasn't the reference before the other way round (i.e. the simulators project referencing the QCTraceSimulatorRuntime)? Why has that changed? How does that work if we were to ever ship each simulator separately?

Simulation.sln
```diff
@@ -1,534 +1,678 @@
...
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "HoneywellExe", "src\Simulation\Simulators.Tests\TestProjects\HoneywellExe\HoneywellExe.csproj", "{1448512E-132F-4DA8-BCBA-D98F16B31600}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "IonQExe", "src\Simulation\Simulators.Tests\TestProjects\IonQExe\IonQExe.csproj", "{55833C6C-6E91-4413-9F77-96B3A09666B8}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "QCIExe", "src\Simulation\Simulators.Tests\TestProjects\QCIExe\QCIExe.csproj", "{C015FF41-9A51-4AF0-AEFC-2547D596B10A}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "TargetedExe", "src\Simulation\Simulators.Tests\TestProjects\TargetedExe\TargetedExe.csproj", "{D292BF18-3956-4827-820E-254C3F81EF09}"
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{BC562DAE-FE2B-4A8C-880C-C546F83F99E4}"
```
bettinaheim, line 68: A couple of things that clearly aren't meant to be split across different solution folders got split. I can't really quickly capture the necessary edits in writing here, but I could make a push to this branch if you want to to bring the solution into shape again.

bettinaheim (commented):
One minor thing: There are a couple of */random.* files with Q#/C# implementation and in contrast to all other file names of similar file, the name is lower case. Could you rename those files to be upper case as well? (I unfortunately can't comment on files that have merely been renamed.)

bettinaheim:
To capture minimally what we had discussed today morning: 
1.) With this PR the default behavior changes for giving any target implementation precedence to what is defined in Q# to a strictly hierarchical structure where only what is declared as intrinsic in Q# will be provided by the target. We need to have the option to override this default behavior in order to support the current functionalities until we have fully built the alternative. The override needs to be possible both when using a driver to construct a target and also for standalone executables (a project configuration or a command line argument would both be fine, I think). This needs to go into this PR. 
2.) Properly splitting out the connecting piece for each instruction set from the target implementation of these instruction should allow to a) autogenerate the connecting piece, b) package the connecting piece with the instruction set and thus eliminate the InternalsVisibleTo used to expose internals to the simulator implementation (might make sense to do with this PR, tbd), and c) ultimately have a single simulator implementation that supports all three target packages (definitively not part of this PR).

build/manifest.ps1
```diff
@@ -24,8 +24,12 @@ $artifacts = @{
         "Microsoft.Quantum.Development.Kit",
         "Microsoft.Quantum.EntryPointDriver",
         "Microsoft.Quantum.QSharp.Core",
+        "Microsoft.Quantum.Type1.Core",
+        "Microsoft.Quantum.Type2.Core",
```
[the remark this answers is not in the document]
swernli: Yup, those exist and are over in #383 that hasn't been merged yet. Given we wanted to fix the e2e build and get the feature branch in, I think it makes sense to leave those out for now and then retarget that PR at master once this merge is done.

src/Simulation/QsharpCore/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,9 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+// Allow the simulator assembly to use our internal methods
+[assembly: InternalsVisibleTo("Microsoft.Quantum.Simulators" + SigningConstants.PUBLIC_KEY)]
```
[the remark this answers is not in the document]
swernli: We had some good sync about how we might remove this dependency, and I'm going to try that design now!

src/Simulation/QsharpFoundation/Diagnostics/AssertQubit.qs
```diff
@@ -47,7 +47,7 @@ namespace Microsoft.Quantum.Diagnostics {
     /// <xref:microsoft.quantum.diagnostics.assertqubitisinstatewithintolerance> allows for asserting
     /// arbitrary qubit states rather than only $Z$ eigenstates.
     operation AssertQubitWithinTolerance(expected : Result, q : Qubit, tolerance : Double) : Unit {
-        AssertProb([PauliZ], [q], expected, 1.0, $"Qubit in invalid state. Expecting: {expected} with tolerance {tolerance}", tolerance);
```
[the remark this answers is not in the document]
swernli: I did! They were annoying me, so I figured I was touching enough files that it was worth fixing :).

src/Simulation/Simulators.Impl/QuantumSimulator/ApplyUncontrolledH.cs
```diff
@@ -0,0 +1,35 @@
...
+
+using System;
+using System.Runtime.InteropServices;
+using Microsoft.Quantum.Simulation.Core;
+
+namespace Microsoft.Quantum.Simulation.Simulators
+{
+    public partial class QuantumSimulator
+    {
+        internal class QSimApplyUncontrolledH : Intrinsic.ApplyUncontrolledH
```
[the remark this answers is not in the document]
swernli: No, this one exists in parallel. H.cs has full support for arbitrary number of controls.

src/Simulation/Simulators.Impl/QCTraceSimulator/QCTraceSimulator.Checks.cs
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Runtime.InteropServices;
+using Microsoft.Quantum.Simulation.Core;
+
+namespace Microsoft.Quantum.Simulation.Simulators.QCTraceSimulators.Implementation
+{
+    public partial class QCTraceSimulatorImpl
```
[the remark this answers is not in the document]
swernli: This was done to make the APIs for the target packages behave like the existing ones for the negative tests (things like passing the same qubit in the controls array as the target). We discussed possibly changing the tests instead of exposing this validation logic in Q#, and I'm going to investigate that as part of the updates we discussed.

src/Simulation/Simulators.Impl/QuantumSimulator/ApplyControlledX.cs
```diff
@@ -0,0 +1,36 @@
...
+
+using System;
+using System.Runtime.InteropServices;
+using Microsoft.Quantum.Simulation.Core;
+
+namespace Microsoft.Quantum.Simulation.Simulators
+{
+    public partial class QuantumSimulator
+    {
+        internal class QSimApplyControlledX : Intrinsic.ApplyControlledX
```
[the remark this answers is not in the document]
swernli: It's in the Type1 target package, and it's internal, so not intended to be an addition to our Q# API surface.

src/Simulation/Simulators.Tests/App.config
```diff
@@ -3,4 +3,7 @@
   <appSettings>
     <add key="xunit.methodDisplay" value="method"/>
   </appSettings>
-<startup><supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.6.1"/></startup></configuration>
+  <startup>
+    <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.6.1"/>
```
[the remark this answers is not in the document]
swernli: I'm actually not sure if it's required... Let me double check that!

src/Simulation/Simulators.Tests/TestProjects/IntrinsicTests/IntrinsicTests.csproj
```diff
@@ -3,15 +3,13 @@
...
     <IsPackable>false</IsPackable>
-    <ExposeReferencesViaTestNames>true</ExposeReferencesViaTestNames>
     <!-- we will provide our own -->
     <CsharpGeneration>false</CsharpGeneration>
     <IncludeQsharpCorePackages>false</IncludeQsharpCorePackages>
   </PropertyGroup>
 
   <ItemGroup>
     <ProjectReference Include="..\..\..\CsharpGeneration\Microsoft.Quantum.CsharpGeneration.fsproj" PrivateAssets="All" IsQscReference="true" />
-    <ProjectReference Include="..\..\..\QsharpCore\Microsoft.Quantum.QSharp.Core.csproj" />
```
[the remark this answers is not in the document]
swernli: We had a few places that included project references that were not called directly but rather used via other packages that already include them as a dependency. Removing them makes the dependency tree a little easier to see.

src/Simulation/Simulators.Tests/TestProjects/IntrinsicTests/IntrinsicTests.csproj
```diff
@@ -3,15 +3,13 @@
   <PropertyGroup>
     <TargetFramework>netcoreapp3.1</TargetFramework>
     <IsPackable>false</IsPackable>
-    <ExposeReferencesViaTestNames>true</ExposeReferencesViaTestNames>
```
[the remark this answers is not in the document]
swernli: Let me double check... I thought I had kept those tests around but it's possible I took them out by mistake.

src/Simulation/Simulators.Impl/QuantumSimulator/ApplyControlledX.cs
```diff
@@ -0,0 +1,36 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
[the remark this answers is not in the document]
swernli: Right now the Tracer and Toffoli simulator only work with the Qsharp.Core package and not with any of the target packages, so they don't have to implement those.

src/Simulation/TargetDefinitions/Decompositions/Reset.qs
```diff
@@ -0,0 +1,25 @@
...
+    /// # Summary
+    /// Given a single qubit, measures it and ensures it is in the |0⟩ state
+    /// such that it can be safely released.
+    ///
+    /// # Input
+    /// ## qubit
+    /// The qubit whose state is to be reset to $\ket{0}$.
+    @RequiresCapability(
+        "BasicQuantumFunctionality",
+        "Reset is replaced by a supported implementation on all execution targets."
```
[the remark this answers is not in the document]
swernli: Yes, exactly. Because the translations don't actually use the target decompositions yet, this work around is still required, but it should hopefully be removable soon!

src/Simulation/Type1Core/Properties/AssemblyInfo.cs
swernli: Yeah, hopefully this can be removed before the PR merges, per our design idea of having one extra layer of indirection.

src/Simulation/TargetDefinitions/Intrinsic/Checks.qs
swernli: Yeah, this can definitely be a function, or better yet, removed (fingers crossed).

src/Simulation/TargetDefinitions/TargetPackages/Type1.Package.props
swernli: Agreed. There is an outstanding checkbox for that in #249, but I'm torn on whether I want to include that as part of this PR or in a follow up. It's probably worth the effort to do it now so that folks can find their way around the rearranged code right away!

src/Simulation/EntryPointDriver/Microsoft.Quantum.EntryPointDriver.csproj
swernli: It wasn't needed before, so cleaning it up makes the dependency tree easier to track.

src/Simulation/QCTraceSimulator.Tests/Tests.Microsoft.Quantum.Simulation.QCTraceSimulatorRuntime.csproj
swernli: I'll need to double check this one; I may have added this speculatively when it isn't strictly necessary.

Simulation.sln
swernli: I'd definitely appreciate some help getting the solution back into proper form again. Because I don't use Visual Studio, this is just a file that helps trigger a build on my machine, but I don't want to mess up the solution view for Visual Studio users who depend on a specific structure here.

swernli:
> 1.) With this PR the default behavior changes for giving any target implementation precedence to what is defined in Q# to a strictly hierarchical structure where only what is declared as intrinsic in Q# will be provided by the target. We need to have the option to override this default behavior in order to support the current functionalities until we have fully built the alternative. The override needs to be possible both when using a driver to construct a target and also for standalone executables (a project configuration or a command line argument would both be fine, I think). This needs to go into this PR.

Regarding this, I've made the changes in this commit: https://github.com/microsoft/qsharp-runtime/pull/421/commits/53d204f5495c14ed923ad4601255689eb2f57e1c. As part of that, the old behavior of C# always overriding the Q# (even if there was a Q# implementation) is now the default, and C# only overriding Q# that is `body intrinsic` is opt-in. `QuantumSimulator` opts into the new behavior, since I've rearchitected it to depend on this behavior (old behavior will fail tests), but any existing experiments that inherit from either `SimulatorBase` or `QuantumProcessorDispatcher` will have no change in behavior. Hopefully that ensures no one is broken by these updates!

swernli:
> 3.) We need to decide when we want to require distinct qubits. The current behavior is closer to only require them for operations that are implemented by the target machine. For this PR we should limit it to that, and we can consider how much value there is to being more strict. With that restrictions, it means one of the tests needs to be updated, since it indeed doesn't/shouldn't produce an error. Once the test is updated, it should be possible to remove CheckQubitUniqueness from the Q# API (we haven't actually gotten around to chatting about RotationAngleValidation - I think I need the reminder what it is used for first).

For this one, I went ahead and removed both `CheckQubitUniqueness` and `RotationAngleValidation` and updated the tests accordingly. I'm using a preprocessor constant to skip certain tests for the Type1 and Type2 simulators where we know the validation won't work given the decomposition structure. I had to restructure the tests a little big to get to them to run in a timely fashion, as the previous set up would continue using qubits after an exception in a way that could cause the wave function to get rather large at the cost of slow processing and huge amounts of memory. See https://github.com/microsoft/qsharp-runtime/pull/421/commits/53d204f5495c14ed923ad4601255689eb2f57e1c for the details of this update.

src/Simulation/Simulators.Impl/QCTraceSimulator/QCTraceSimulator.Checks.cs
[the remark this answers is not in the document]
swernli: I was able to remove the CheckQubitUniqueness and RotationAngleValidation helpers.

src/Simulation/Simulators.Impl/QuantumSimulator/SWAP.cs
swernli: To ensure that tests passed as expected, I had to pull in this portion of #383, so SWAP is now be fixed.

src/Simulation/Simulators.Tests/TestProjects/IntrinsicTests/IntrinsicTests.csproj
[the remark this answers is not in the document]
swernli: I did double check, and the tests that actually relied on this are still present and compile out of src\Simulation\Simulators.Tests\TestProjects\UnitTests\UnitTests.csproj, which still has `ExposeViaTestReferences` configured. I think this may have been copied in here accidentally and there wasn't any dependency on it, so I took it out.

src/Simulation/Type1Core/Properties/AssemblyInfo.cs
swernli: See the comment below, but the short version is that the work to refactor simulator to target package relationship is now captured by #432, at which point we should be able to remove this AssemblyInfo entirely.

swernli:
> 2.) Properly splitting out the connecting piece for each instruction set from the target implementation of these instruction should allow to a) autogenerate the connecting piece, b) package the connecting piece with the instruction set and thus eliminate the InternalsVisibleTo used to expose internals to the simulator implementation (might make sense to do with this PR, tbd), and c) ultimately have a single simulator implementation that supports all three target packages (definitively not part of this PR).

We've talked more about this approach, and have an idea to define this connecting piece as a common interface package that both the target and the simulator packages refer to, providing the indirection that allows the simulator to stop compiling directly against the target package. This is now captured by #432, and looks like it will be more work than will fit into this PR. We can merge this PR without this work, and then continue that refactor directly in main!

src/Simulation/QCTraceSimulator.Tests/Tests.Microsoft.Quantum.Simulation.QCTraceSimulatorRuntime.csproj
swernli: Ah, I remember now why this was needed: when I split out the three simulators I had to remove the simulator reference from src\Simulators\Common\Simulators.Dev.props and add it explicitly to any projects that needed to refer to the QsharpCore simulator. So it's not a new dependency, just now it's direct instead of being via a props file.

src/Simulation/Simulators.Tests/App.config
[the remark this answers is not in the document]
swernli: I think we just need it for the `xunit.methodDisplay` configuration, which would default to `classAndMethod` if not otherwise specified. It makes the unit test log output a little cleaner. See https://xunit.github.io/docs/configuration-files#methodDisplay.

src/Simulation/TargetDefinitions/TargetPackages/Type1.Package.props
swernli: Added a README in https://github.com/microsoft/qsharp-runtime/pull/421/commits/ef53ae73f4d684b0c32de074f992f370ef2e8fc4

Simulation.sln
swernli: @bettinaheim I think this is the last of the feedback I need to address. Would you prefer we fix this as a change to the working branch or can it be done after merging this PR?

swernli:
Closing this in favor of #442 that puts the contents back into the feature branch. I'm going to collect some more work there first and then reopen a PR to main.

PR #423, opened 2020-10-29
Make it easier to add new new entry point driver tests

samarsha:
This changes the entry point driver test cases from being indexed by their position in the file to being indexed by a name. This makes it easier to add new tests without needing to update the indices of all the tests that come after the newly added test.

It also adds a new test to make sure that an entry point callable with the same name as the parent namespace is supported (see microsoft/qsharp-compiler#711).

### DIFF
--- a/src/Simulation/EntryPointDriver.Tests/Tests.fs
+++ b/src/Simulation/EntryPointDriver.Tests/Tests.fs
@@ -34,7 +34,7 @@ let private intrinsicFile = Path.GetFullPath "Intrinsic.qs"
 let private testFile = Path.GetFullPath "Tests.qs"
 
 /// The namespace used for the test cases.
-let private testNamespace = "Microsoft.Quantum.EntryPointDriver.Tests"
+let private testNamespace = "EntryPointTest"
 
 /// The test case for the given test number.
 let private testCase =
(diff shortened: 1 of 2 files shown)


src/Simulation/EntryPointDriver.Tests/Tests.qs
```diff
@@ -5,308 +5,317 @@
 // No Options
 //
 
-namespace Microsoft.Quantum.EntryPointDriver.Tests {
+// --- Returns Unit
```
cesarzc, line 8: It would be useful to add a comment at the start of this file describing that `// ---` is used as a separator for test cases and that the string that follows the separator is interpreted as the name of the test case.

src/Simulation/EntryPointDriver.Tests/Tests.fs
```diff
@@ -34,16 +33,21 @@ let private intrinsicFile = Path.GetFullPath "Intrinsic.qs"
 let private testFile = Path.GetFullPath "Tests.qs"
 
 /// The namespace used for the test cases.
-let private testNamespace = "Microsoft.Quantum.EntryPointDriver.Tests"
+let private testNamespace = "EntryPointTest"
 
-/// The test case for the given test number.
+/// The test case for the given test name.
```
cesarzc, line 38: It also would be useful to have a more detailed description here on how the test file is organized such that it is more clear to the reader how a test case is obtained. This also makes it easier to add more test cases for people that are new to the codebase.

bettinaheim (approved):

PR #425, opened 2020-10-31
Native Simulator: state prep emulation

irinayat-MS:
This change introduces state prep emulation API and implements the trivial case of total state injection.

To avoid the annoyance of compiling local_test.cpp twice, migrated all tests in this file to catch2.


src/Simulation/Native/src/simulator/wavefunction.hpp
```diff
@@ -549,6 +550,40 @@ class Wavefunction
         return kernels::jointprobability(wfn_, bs, get_qubit_positions(qs));
     }
 
+    /// \pre: Qubits, listed in `q`, must be unentangled and in state |0>.
```
cgranade, line 553:
```suggestion
    /// \pre: Qubits, listed in `q`, must be unentangled and in the |00…0⟩ state.
```

[nit]: If `q` is in the |00…0⟩ state, that implies that they are unentangled. I agree it's still helpful to emphasize, though.

src/Simulation/Native/src/simulator/wavefunction.hpp
```diff
@@ -549,6 +550,40 @@ class Wavefunction
...
+        assert((static_cast<size_t>(1) << qubits.size()) == amplitudes.size());
+
+        if (num_qubits_ != qubits.size()) throw std::runtime_error("Subsystem state injection not supported (yet)");
+
+        flush();
+
+        // Check prerequisites.
+        for (logical_qubit_id q : qubits)
+        {
+            if (!kernels::isclassical(wfn_, get_qubit_position(q))
```
cgranade, line 567: Apologies for the naive question, but what does `kernels::isclassical` check? It looks from context like it's checking that the qubit is in a computational basis state?

src/Simulation/Native/src/simulator/wavefunction.hpp
```diff
@@ -549,6 +550,40 @@ class Wavefunction
         return kernels::jointprobability(wfn_, bs, get_qubit_positions(qs));
     }
 
+    /// \pre: Qubits, listed in `q`, must be unentangled and in state |0>.
+    /// place these `n` qubits into superposition of 2^n basis vectors with (re, im) amplitudes, where the order of
+    /// qubits in array `q` defines (big or little endian?) order of the basis vectors.
```
cgranade, line 555: Would it be good to note in this API comment the temporary limitation that subsystem state injection is not yet supported? I could see either way, since that limitation is going away pretty soon.

src/Simulation/Native/src/simulator/local_test.cpp
```diff
@@ -320,7 +318,163 @@ TEST_CASE("Clustering", "[local_test]")
...
+        qs.push_back(sim.allocate());
+    }
+
+    const double amp = 1.0 / std::sqrt(n);
+    std::vector<ComplexType> amplitudes = {{0.0, 0.0}, {amp, 0.0}, {amp, 0.0}, {0.0, 0.0},
+                                           {amp, 0.0}, {0.0, 0.0}, {0.0, 0.0}, {0.0, 0.0}};
+    REQUIRE(N == amplitudes.size());
+
+    sim.InjectState(qs, amplitudes);
+    CHECK(sim.Measure({Gates::PauliZ, Gates::PauliZ, Gates::PauliZ}, qs));
```
cgranade, line 373: If I understand right, this checks that the parity of a 𝑍𝑍𝑍 measurement is One, correct? If so, that's definitely a necessary condition, but it doesn't fully specify the three-qubit W-state. It could be a bit challenging to use `sim.Measure` to check for the W-state since W-states are not stabilizer states in general...

cgranade (approved):
I think this looks really good, thanks for tagging me in! I just had some fairly minor questions, mostly about comments and test cases.

src/Simulation/Native/src/simulator/wavefunction.hpp
irinayat-MS: My rational for phrasing the comment like this was that for entangled qubits it doesn't make sense to talk about their own state. The qubits also might not be adjacent in the computational basis. Would it be clearer to say "Each qubit, listed in `q` must be unentangled and in state |0>"?

cgranade: I think that makes sense, yeah. As written, the comment may be misunderstood as requiring that the register consist of a single qubit, I guess? All fairly minor in any case.

src/Simulation/Native/src/simulator/wavefunction.hpp
irinayat-MS: Yes, that's my understanding. I've added a new test to document that.

src/Simulation/Native/src/simulator/local_test.cpp
irinayat-MS: I wanted to have a test with non-Cat-state injection to catch bugs related to little versus big endian order... but thinking more about it now it doesn't seem a good idea doing it like this because the failure won't be deterministic. And I later added the test that creates a Bell pair on two qubits out of three which does cover the original concern. I'll remove the W-state test.

PR #426, opened 2020-11-02
Enable test for using "Data" as a named item

samarsha:
This enables a commented out test for using "Data" as a named item, which is now supported since microsoft/qsharp-compiler#643 was merged recently.


cgranade (approved):

cgranade (approved):
💕

PR #428, opened 2020-11-04
Native Simulator: permute_basis tests and asserts

irinayat-MS:
I thought I could use `permute_basis` to rearrange the wave function to simplify injecting a state on a subsystem of qubits. After trying to figure out the intent of `permute_basis` I don't think it would be useful for my purposes but now have a question about it. The result of the method _depends_ on the order in which logical qubits are listed. I don't understand how to interpret this and how the client should define the `permutation_table`. I was expecting the list of logical qubits to be treated as an unordered set that defines the subsystem to be permuted.

The tests `permute_basis depends on the order of logical qubits` documents the order dependency. I also added a temporary comment inside `permute_basis` which should be replaced with a method-level comment that describes the contract before checking in. After resolving the ordering question, I'd like to merge this PR because it would be a pity to throw away the tests, comments and asserts I added while investigating the API.

The API is used in https://github.com/microsoft/Quantum/blob/main/samples/runtime/oracle-emulation/PermutationOracle.cs and I don't know who owns that code. @cgranade ?


cgranade:
Thanks for tagging me in, @irinayat-MS, looking now. As far as the permutation oracle code, that sample was one written by @msoeken to demonstrate how to use the simulator to emulate Q# operations.

Looking at that sample, I think the correct action of `permute_basis` does depend on the ordering of the register passed into that method, as that allows a Q# caller to refer to computational basis states of a given register without needing to know how that register is represented internal to the simulator.

Tracing through to confirm that impression, the sample uses `permute_basis` to implement the Q# operation at https://github.com/microsoft/Quantum/blob/main/samples/runtime/oracle-emulation/PermutationOracle.qs#L7:

```
    /// # Summary
    /// Apply a classical permutation oracle to two registers.
    ///
    /// # Description
    /// The effect of the oracle is a permutation of basis states according to
    /// the provided classical function:
    /// $$
    /// \begin{align}
    ///     \ket{x}\ket{y}\ket{w} \rightarrow \ket{x}\ket{f(x, y)}\ket{w},
    /// \end{align}
    /// $$
    /// with registers x, y, w and the oracle function f. 
    ///
    /// # Input
    /// ## oracle
    /// A function that defines the action of the oracle on the computational
    /// basis states of the two registers x, y. The mapping
    /// $$
    /// \begin{align}
    ///		$(x, y) \rightarrow (x, z=f(x, y))$
    /// \end{align}
    /// $$
    /// must be a bijective mapping on the basis states.
    /// ## xbits
    /// Input register x.
    /// ## ybits
    /// Output register y.
    operation PermutationOracle(oracle : ((Int, Int) -> Int), xbits : Qubit[], ybits : Qubit[]) : Unit
```

From there, the overload at https://github.com/microsoft/Quantum/blob/41c4fc38bdd71c24f917e15d324a4171bd0139ff/samples/runtime/oracle-emulation/PermutationOracle.cs#L26 calls `oracle` repeatedly to make a permutation table, so that the new array `permutation` is effectively a representation of `oracle` as a re-ordering of computational basis state indices. If I understand correctly, those computational basis state indices are with respect to interpreting the `qs` register as a little-endian encoded integer, such that to describe a CNOT (swap |11⟩ ↔ |10⟩, interpreted in little-endian as |3⟩ ↔ |1⟩) a `permutation` of `{0, 3, 2, 1}` would be needed.

Do you agree, @msoeken?

irinayat-MS:
I'm still confused. Let's work through a very simple example of two qubits, q0 and q1. Suppose that the current wave function (not-normalized) is 0*|00> + 1*|01> + 2*|10> + 3*|11> (the wave function does depend on how the qubits are mapped to the positions in the standard computational basis but this detail isn't exposed to the users). Suppose, I'd like to transpose |00> and |01> basis vectors, thus I define the permutation_table as {1,0,2,3}

What are the expected states of the wave function after each of the calls and why?
```
permute_basis({q0,q1}, 4, permutation_table);
permute_basis({q1,q0}, 4, permutation_table);
```
The effective results I'm seeing now are:
1*|00> + 0*|01> + 2*|10> + 3*|11>
2*|00> + 1*|01> + 0*|10> + 3*|11>
And I don't understand why the second one is correct.

cgranade:
I could also be confused, so I agree it's good to walk through!

In the first example, the state of the Q# register `[q0, q1]` is interpreted as 0 |0⟩ + 1 |2⟩ + 2 |1⟩ + 3 |3⟩ (interpreting 01₂ = 2₁₀ in little-endian notation, such that the state of `q0` labels the ones-digit). Thus, applying the permutation 0 → 1, 1 → 0, 2 → 2, 3 → 3 gives the final state 0 |1⟩ + 1 |2⟩ + 2 |0⟩ + 3 |3⟩. Back into bitstring labels, that gives 0 |10⟩ + 1 |01⟩ + 2 |00⟩ + 3 |11⟩, which you have as the final state for the second example.

On the other hand, taking the Q# register `[q1, q0]`, I get that the initial state is interpreted as 0 |0⟩ + 1 |1⟩ + 2 |2⟩ + 3 |3⟩, such that the same permutation gives 0 |1⟩ + 1 |0⟩ + 2 |2⟩ + 3 |3⟩, for a final state of 0 |01⟩ + 1 |00⟩ + 2 |10⟩ + 3 |11⟩ (your first example).

I'm not quite sure why the two examples are flipped, but I think it at least makes sense that the application of the permutation depends on the interpretation of the register specified by the first argument to `permute_basis`.

irinayat-MS:
I believe, there might be a disconnected between how you view the wave function and the internal representation of the simulator. This is what I'm observing in the native simulator:

```
    SimulatorType sim;
    logical_qubit_id q0 = sim.allocate();
    logical_qubit_id q1 = sim.allocate();
    sim.X(q0);
```
```
yields internal representation as:
q0 (0) maps to position 0
q1 (1) maps to position 1
0:(0,0) 1:(1,0) 2:(0,0) 3:(0,0)
```
The last row above lists the amplitudes in the wave function by index: <index>:<(re, im)>. In this case the state is 1*|01>

When I'm saying I've injected state 0*|00> + 1*|01> + 2*|10> + 3*|11>, I mean that the internal state of the simulator is set to
```
q0 (0) maps to position 0
q1 (1) maps to position 1
0:(0,0) 1:(1,0) 2:(2,0) 3:(3,0)
```

if in this state I do
```
    size_t permutation[4] = {1, 0, 2, 3};
    sim.psi.permute_basis({q0, q1}, 4, permutation);
``` 
The internal state is changed to
```
q0 (0) maps to position 0
q1 (1) maps to position 1
0:(1,0)1:(0,0)2:(2,0)3:(3,0)  = 1*|00> + 0*|01> + 2*|10> + 3*|11>
```
That is, |00> and |01> are transposed. So it seems that in this call {q0, q1}'s interpretation matches their current positional order in the internal representation.

but if in state 0:(0,0) 1:(1,0) 2:(2,0) 3:(3,0) I do
```
    size_t permutation[4] = {1, 0, 2, 3};
    sim.psi.permute_basis({q1, q0}, 4, permutation);
```
The internal state is changed to
```
q0 (0) maps to position 0
q1 (1) maps to position 1
0:(2,0)1:(1,0)2:(0,0)3:(3,0) = 2*|00> + 1*|01> + 0*|10> + 3*|11>
```
per your reasoning above, the order {q1, q0} should have been interpreted as the permutation on the positions of q1 and q0 in the internal representation so the state is assumed to be 0*|00> + 2*|01> + 1*|10> + 3*|11>, then apply the requested transposition of |00> and |01> to arrive at: 2*|00> + 0*|01> + 1*|10> + 3*|11>. If "undo" the positioning reordering, then will get to the final 2*|00> + 1*|01> + 0*|10> + 3*|11>.

That is, the order of logical qubits defines permutation of the basis U that derives from reodering of the qubits.
The permutation table defines an arbitrary permutation of the standard computational basis V where the qubits are used in the order they were listed in.
The final operator applied is Adjoint(U)*V*U.

I believe, I can document the API now :D

cgranade:
> When I'm saying I've injected state 0*|00> + 1*|01> + 2*|10> + 3*|11>, I mean that the internal state of the simulator is > set to
>
> ```
> q0 (0) maps to position 0
> q1 (1) maps to position 1
> 0:(0,0) 1:(1,0) 2:(2,0) 3:(3,0)
> ```

Ah, I think that's the disconnect, in that it looks like you've used big-endian ordering to interpret the injection of 0 |00⟩ + 1 |01⟩ + 2 |10⟩ + 3 |11⟩. I'd have expected the internal state from that injection to be `0:(0,0) 1:(2,0) 2:(1,0) 3:(3,0)` following a little-endian ordering of computational basis labels.

msoeken:
> Do you agree, @msoeken?

That code was written by @jgukelberger, as fas as I know.

cgranade:
> > Do you agree, @msoeken?
> 
> That code was written by @jgukelberger, as fas as I know.

Ah, thanks for the correction, I had gotten confused with the other permutation-of-basis-states sample, sorry.

jgukelberger:
Thanks for tagging me. Looks like you figured it out - thanks @cgranade for providing a detailed explanation! Indeed, the order of the qubits provided by the user indicates how they interpret them as a register, and the permutation table specifies how the register's computational basis states are to be permuted.

irinayat-MS:
@jgukelberger would you mind reviewing the comments/asserts/tests I've added for the API and signing off on the PR? There should be no functional changes, purely documentation and testing. Thank you!

jgukelberger:
@irinayat-MS yes, looking through it right now :)

src/Simulation/Native/src/simulator/wavefunction.hpp
```diff
@@ -719,37 +734,74 @@ class Wavefunction
         return kernels::subsytemwavefunction(wfn_, get_qubit_positions(qs), qubitswfn, tolerance);
     }
 
-    // apply permutation of basis states to the wave function
+    /// Apply the unitary operator that permutes the standard computational basis of the subsystem, defined by the
+    /// provided qubits. `qs` lists the qubits of the subsystem in little-endian order (the front qubit in the list
+    /// corresponds to the last position of the standard computational basis). `table_size` must be equal 2^n, where n
+    /// is the number of qubits in `qs` `permutation_table` must point to an array of `table_size` integers that
+    /// describes a permutation of set {0, 1, ... , 2^n -1}. `adjoint` specifies whether should apply the adjoint of the
+    /// unitary.
```
jgukelberger, line 742: (minor language nit)
```suggestion
    /// is the number of qubits in `qs`. `permutation_table` must point to an array of `table_size` integers that
    /// describes a permutation of set {0, 1, ... , 2^n -1}. `adjoint` specifies whether to apply the adjoint of the
    /// unitary.
```

jgukelberger (approved):
Looks good to me. Thanks for adding all these tests and docs!

PR #429, opened 2020-11-04
Native Simulator: state prep emulation - part 2 (partial state injection)

irinayat-MS:
This PR completes the native side of state emulation. The remaining work is calling it from the libraries.

Added two micro benchmarks to compare perf of state injection between partial/total injection and quantum state preparation. 
The benchmarks show that it is worth handling total state injection separately and that compared to preparing the same state using quantum operations (assuming they are finalized) there is a considerable win (x10) when injecting state that touches many basis vectors and a slight win (~10%) when injected state is sparse.

The results of the benchmarks raise the following questions:
- can the current implementation of state injection be improved and is it worth pursuing? (e.g. it might improve injection perf if the API is changed/extended to support sparse state vectors rather than passing in all amplitudes, pass in an array of non-zero amplitude indexes and array of the corresponding values)
- what states are good candidates for benchmarking this feature? are the two cases representative enough?

For the overall perf numbers we should also measure how much time state injection would save on Q# side (decomposing the state into quantum operations, etc.) 
 
== below section is outdated (the benchmarks didn't flush after quantum ops) ==
The conclusion is that total injection is worth optimizing comparing to partial injection but for the simple cases of these benchmarks quantum state preparation _way_ outperforms state injection.

For 20 qubits typical ranges are:

Quantum state preparation:  <500 microseconds
    Total state injection:      15000 - 20000 microseconds
  Partial state injection:      30000 - 60000 microseconds

The partial state injection also shows much higher variance (probably due to more often missing caches?).


src/Simulation/Native/src/simulator/wavefunction.hpp
```diff
@@ -553,35 +553,46 @@ class Wavefunction
...
-            wfn_[i] = amplitudes[i];
+            const size_t num_states = wfn_.size();
+            std::vector<positional_qubit_id> positions = get_qubit_positions(qubits);
+            const size_t mask = kernels::make_mask(positions);
+            WavefunctionStorage wfn_new(num_states);
+
+            for (size_t basis_index = 0; basis_index < num_states; basis_index++)
+            {
+                const size_t injected_index = detail::get_register(positions, basis_index);
+                const size_t original_term = detail::set_register(positions, mask, 0, basis_index);
```
cgranade, line 592: I know they're internal and not part of the PR, but would it be possible to add some documentation comments for `get_register` and `set_register` to help explain what they do?

src/Simulation/Native/src/simulator/wavefunction.hpp
```diff
@@ -553,35 +553,46 @@ class Wavefunction
...
-        for (size_t i = 0; i < wfn_.size(); i++)
+        else
         {
-            wfn_[i] = amplitudes[i];
+            const size_t num_states = wfn_.size();
+            std::vector<positional_qubit_id> positions = get_qubit_positions(qubits);
+            const size_t mask = kernels::make_mask(positions);
+            WavefunctionStorage wfn_new(num_states);
+
+            for (size_t basis_index = 0; basis_index < num_states; basis_index++)
```
cgranade, line 589: If I understand correctly, the approach here is to:

- write out the state as Σᵢ (𝑎ᵢ |𝑖⟩ ⊗ |0⟩), where the first register is the qubits not being injected and where the second register should have the state being injected. Each 𝑎ᵢ can be found using `wfn_[detail::set_register(...)]`.
- inject the state Σᵢ 𝑎ᵢ |𝑖⟩ ⊗ |φ⟩ = Σᵢ,ⱼ (𝑎ᵢ 𝑏ⱼ |𝑖⟩ ⊗ |𝑗⟩), where |φ⟩ = Σⱼ 𝑏ⱼ |𝑗⟩ is the state to be injected and where {𝑏ⱼ}ⱼ is given by `amplitudes`. Since each term |𝑖⟩ ⊗ |𝑗⟩ is a computational basis state, that can be injected by setting the amplitudes of `wfn_new` directly for each index (𝑖, 𝑗) to be 𝑎ᵢ 𝑏ⱼ.

src/Simulation/Native/src/simulator/wavefunction.hpp
```diff
@@ -553,35 +553,46 @@ class Wavefunction
...
                 throw std::runtime_error("Cannot prepare state of entangled qubits or if they are not in state |0>");
             }
         }
 
-        // Reorder the qubits to the positions that match the wave function provided by the user.
-        for (unsigned i = 0; i < qubits.size(); i++)
+        if (qubits.size() == num_qubits_)
         {
-            qubitmap_[qubits[i]] = i;
+            // For full state injection we can simply copy the user's wave function into our store and reorder the
```
cgranade, line 574: This optimization should also hold if the subregister not being injected is in a single computational basis state. That could even possibly be extended further to use sparsity in the computational basis, perhaps?

cgranade (commented):
I didn't look at the unit tests yet, just wanted to make sure I understood the basics of the approach first and so filed a couple questions to that effect. Thanks!

src/Simulation/Native/src/simulator/wavefunction.hpp
irinayat-MS: Please see https://github.com/microsoft/qsharp-runtime/pull/428

src/Simulation/Native/src/simulator/wavefunction.hpp
alexeib2: Hmm, the bechmarking results are fairly surprising if I understand them correctly. They are not consistent with my previous state preparation experience. Can we do an experiment based on an equivalent of the following pseudo-code:
double[] v = new double[2^20]; double s2 = 0.0; for (size_t ix = 0; ix < 2^20; ix++) { 
double x= (double) rand(); s2+=x*x; v[ix]=x;} double nu = 1.0/sqrt(s2); for (size_t jx = 0; jx < 2^20; jx++) v[ jx]*=nu;}
After which we (a) feed v into the StatePreparation API and (b) feed v into the state injection API 
?
I would like to to understand where and how much time is spent in each scenario.

irinayat-MS: I believe I figured out why the micro benchmarks on the native side were skewed: the simulator was caching the quantum operations so the wave function wasn't actually being updated. After forcing the flush the numbers are much closer to the expectations:

For creating the equal superposition of all basis vectors, the injection performs similar to the quantum preparation on systems with small number of qubits (< 10) and about x5 faster for systems with 20 qubits. Partial state injection yields smaller perf boost and might be worse than quantum state preparation when injecting a state on small subsystem into a large system. Basically, we might need heuristics at the Q# library level when deciding whether to inject or to prepare.

I haven't done the plumbing yet from Q# to native to measure it but the native micro benchmarks look good to me now and I think we should proceed with this change.

src/Simulation/Native/src/simulator/wavefunction.hpp
irinayat-MS: When injecting a partial state we don't know what the complementary qubits are doing and to find out that the current wave function is sparse would require us to traverse it. And even if the current wave function is just a single term we'd still have to compute the registers before assigning the new amplitudes. Basically, I don't believe there is an optimization opportunity here but please do reactivate if I've missed your point.

src/Simulation/Native/src/simulator/wavefunction.hpp
```diff
@@ -553,35 +553,52 @@ class Wavefunction
...
+            wfn_.assign(amplitudes.begin(), amplitudes.end());
         }
-
-        // For full state injection we can simply copy the user's wave function into our store.
-        for (size_t i = 0; i < wfn_.size(); i++)
+        else
         {
-            wfn_[i] = amplitudes[i];
+            const int64_t num_states = static_cast<int64_t>(wfn_.size());
+            std::vector<positional_qubit_id> positions = get_qubit_positions(qubits);
```
swernli, line 585: This is a very minor optimization, but now this branch ends up iterating over `qubits` to convert them into positional ids, and the precheck on each entry in `qubits` above iterates over the whole list too. Perhaps this positional vector should be created at the top of the function and used in both the precheck and here?

swernli (approved):
The changes look good to me! One minor optimization is possible below, though even that is a trade-off that seems fine if it stays as-is.

src/Simulation/Native/src/simulator/wavefunction.hpp
swernli: This would be a trade-off, I suppose, and I don't know which is better: the cost of always allocating the vector to save one additional loop through the `qubits` input, or only allocating the positional vector when we need it for subsystem injection at the cost of iterating through `qubits` twice. I don't expect `qubits` to be very big, so I'm not sure which side to err on.

irinayat-MS: I honestly don't think it matters here because, as you said, qubits array isn't going to be large, but getting the positions vector before doing the prereq loop will make the loop a bit cleaner looking, so why not? :)

irinayat-MS:
@cgranade -- Chris, are you done with the review? Would you like more time to look at it before I merge?

src/Simulation/Native/src/simulator/wavefunction.hpp
```diff
@@ -552,36 +552,57 @@ class Wavefunction
...
+            }
+            wfn_.assign(amplitudes.begin(), amplitudes.end());
         }
-
-        // For full state injection we can simply copy the user's wave function into our store.
-        for (size_t i = 0; i < wfn_.size(); i++)
+        else
         {
-            wfn_[i] = amplitudes[i];
+            // The current state can be thought of as Sum(a_i*|i>|0...0>), after the state injection it will become
```
cgranade, line 584: Thanks for adding this comment block, I think that makes the logic here a lot easier to follow!

cgranade (approved):

PR #431, opened 2020-11-09
Update for compiler API change that removes NonNullable

samarsha:
See microsoft/qsharp-compiler#729.

### DIFF
--- a/build/ci.yml
+++ b/build/ci.yml
@@ -10,7 +10,7 @@ pr:
 
 variables:
   Build.Major: 0
-  Build.Minor: 11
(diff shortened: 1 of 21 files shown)


build/ci.yml
```diff
@@ -10,7 +10,7 @@ pr:
 
 variables:
   Build.Major: 0
-  Build.Minor: 11
+  Build.Minor: 13
```
bettinaheim, line 13: I suggest to set this to 14, since we have some breaking changes coming in with this month.

bettinaheim (approved):

PR #433, opened 2020-11-10
Native Simulator: change prereq check in state injection to return false instead of throw

irinayat-MS:
I misunderstood specifications of PrepareArbitraryState. The API guarantees that if the qubits, the state is being injected for, are unentangled and each is in state |0>, then they will be transformed into the requested state, otherwise, the API will perform _a valid albeit undefined_ unitary operator but it shouldn't throw.

Considered options:
1. Ignore the failed prereq and apply the same logic as we have now for state update -> Cannot do this, because it might end up not being a unitary operation in the partial state injection case (*)
2. Do nothing (apply identity transform to the state) -> We currently rely on Q# to do adjoint operator, which means the base and adjoint must match but they wouldn't in this case (**)
3. Bail out and let Q# know that it has to do the quantum state preparation -> implemented in this change

(*) Given how state setting is implemented now. Conceivably, there might be other algorithms that are both efficient and guarantee unitariness even if prereqs aren't satisfied? Also, could we still do this for total state injection?
(**) I don't believe it's easy/efficient to implement emulation of adjoint state prep, but even if it were, the unitary we apply cannot depend on the state of qubits because the base/adjoint might be applied in different states.

Note: The bug was found in course of reviewing https://github.com/microsoft/QuantumLibraries/pull/370


irinayat-MS:
Oops. Wrong branch. Will republish the PR in a moment.

PR #435, opened 2020-11-10
Release qubits in array in the order they were allocated in

irinayat-MS:
For borrowing QubitManager tracks qubits allocated in each frame(=scope) by adding them in order of allocation into `List<Qubit>` on `StackFrame` class. `Release(Qubit)` method then removes the qubit from the list using [List.Remove](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.remove?view=net-5.0) method, which has to traverse the list to find the element to be removed. Thus, when removing in reverse to allocation order (as it happens in case of arrays), the list is fully traversed on each call, yielding O(n^2) complexity. One might have expected the remove at the tail rather than at the head of the list to be faster because it wouldn't require shifting of the remaining elements, and it is, if using `List.RemoveAt` method. However, apparently C# runtime optimizes the shift, because search time dominates even when working with `List<int>` and even more so when the type defines custom equality operation.

For the attached standalone ListRemoveExperiment (`WithCustomEquality` class mimics Qubit in qsharp-runtime) I'm seeing the following times (release builds for lists with 10000 items):

1. Remove at head on ints: 3.98 
2. Remove at tail on ints: 0.0241 (x100 faster than 1 because doesn't have to shift the elements)
3. Remove value forward on ints: 3.2857 (same as 1, because minimal search overhead, being slightly faster is a fluke)
4. Remove value backward on ints: 18.3455 (x5 slower than 3 -- search dominates the shift)

5. Remove at head on custom type: 13.0862 (slower than 1, probably because the custom type is bigger than int)
6. Remove at tail on custom type: 0.0274 (same as 2, as expected)
7. Remove value forward on custom type: 14.3319 (same as 5, because minimal search overhead)
8. Remove value backward on custom type: 546.5933 (x40 slower than 7 -- search trumps the shift)

The above dynamics from the standalone app translate into QubitManage almost exactly. It can be observed in qdk perf tests, or by running a test like the one below in qsharp-runtime directly. All simulators that enable borrowing are affected.

```
        [Fact]
        public void test_perf()
        {
            var count = 10000;
            var sim = new ToffoliSimulator();
            var allocate = sim.Get<Intrinsic.Allocate>();
            var release = sim.Get<Intrinsic.Release>();

            var qs = allocate.Apply(count);
            release.Apply(qs);
        }
```

[ListRemoveExperiment.txt](https://github.com/microsoft/qsharp-runtime/files/5533538/ListRemoveExperiment.txt)

### DIFF
--- a/src/Simulation/Common/QubitManager.cs
+++ b/src/Simulation/Common/QubitManager.cs
@@ -488,9 +488,11 @@ namespace Microsoft.Quantum.Simulation.Common
                 return;
             }
 
-            for (long i = qubitsToRelease.Length-1; i>=0; i--)
-            { // Going from the end is more efficient in case we are tracking scope.
-                this.Release(qubitsToRelease[i], wasUsedOnlyForBorrowing: false);
+            // When tracking scopes for borrowing, it's more efficient to release qubits in the same order
+            // they were allocated in (which in case of allocated arrays is from left to right).
+            foreach (var qubit in qubitsToRelease)
+            {
+                this.Release(qubit, wasUsedOnlyForBorrowing: false);
             }
         }
 
@@ -663,9 +665,11 @@ namespace Microsoft.Quantum.Simulation.Common
                 return;
             }
 
-            for (long i = qubitsToReturn.Length - 1; i >= 0; i--)
-            { // Going from the end is more efficient in case we are tracking scope.
-                this.Return(qubitsToReturn[i]);
+            // When tracking scopes for borrowing, it's more efficient to release qubits in the same order
+            // they were allocated in (which in case of borrowed arrays is from left to right if we had to allocate).
+            foreach (var qubit in qubitsToReturn)
+            {
+                this.Return(qubit);
             }
         }
     }
--- a/src/Simulation/Simulators.Tests/CoreTests.cs
+++ b/src/Simulation/Simulators.Tests/CoreTests.cs
@@ -95,8 +95,8 @@ namespace Microsoft.Quantum.Simulation.Simulators.Tests
                 testOneBody(1, (1, q0));
                 testOneBody(1, (1, q1));
                 testOneBody(1, (1, q4));
-                testOneBody(3, (1, q5));
-                testOneBody(0, (1, q6));
+                testOneBody(1, (1, q5));
+                testOneBody(2, (1, q6));
                 testOneBody(2, (2, q2));
                 testOneBody(2, (2, q3));
                 testOneBody(1, (3, q0));
@@ -114,8 +114,8 @@ namespace Microsoft.Quantum.Simulation.Simulators.Tests
                 testOneAdjoint(1, (1, q0));
                 testOneAdjoint(1, (1, q1));
                 testOneAdjoint(1, (1, q4));
-                testOneAdjoint(3, (1, q5));
-                testOneAdjoint(0, (1, q6));
+                testOneAdjoint(2, (1, q5));
+                testOneAdjoint(1, (1, q6));
                 testOneAdjoint(2, (2, q2));
                 testOneAdjoint(2, (2, q3));
                 testOneAdjoint(1, (3, q2));
@@ -131,8 +131,8 @@ namespace Microsoft.Quantum.Simulation.Simulators.Tests
                 testOneCtrl(1, (1, q0));
                 testOneCtrl(1, (1, q1));
                 testOneCtrl(1, (1, q4));
-                testOneCtrl(2, (1, q5));
-                testOneCtrl(0, (1, q6));
+                testOneCtrl(1, (1, q5));
+                testOneCtrl(1, (1, q6));
                 testOneCtrl(0, (2, q0));
                 testOneCtrl(0, (2, q1));
                 testOneCtrl(2, (2, q2));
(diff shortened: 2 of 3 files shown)


irinayat-MS:
@DmitryVasilevsky & @avasch01 & @bettinaheim : the PR that introduced this perf regression (https://github.com/microsoft/qsharp-runtime/pull/190) mentions benchmarks that were improved by the change. What were the scenarios?

bettinaheim:
Thanks, @irinayat-MS, for analyzing that! 
@DmitryVasilevsky should be able to provide the scenario that benefitted - I believe the case that benefitted is executed on a custom simulator and I suspect adapting that simulator should allow to preserve performance for that case while doing the change in this PR. If I am wrong and the benefit also applies to one of our released simulators we should add a perf test to capture that.

bettinaheim (approved):

alan-geller:
While this particular change makes sense given the underlying implementation, to me it seems that having the performance be sensitive to the order in which qubits get released is probably a sign that the data structure we're using is the wrong data structure. Reading through the description, I wonder why we're using a list at all. Mightn't a data structure like a dictionary or a sorted list work better? If there's a need to keep some externally-defined order, other than by qubit ID, then a doubly-linked list with a dictionary of pointers on the side might be appropriate.

irinayat-MS:
Before experimenting with other data structures I'd like to understand better the scenarios we are optimizing for. The current choice of data structure is close to ideal for allocating and then releasing a single large array of qubits in a given scope.

cgranade:
> Before experimenting with other data structures I'd like to understand better the scenarios we are optimizing for. The current choice of data structure is close to ideal for allocating and then releasing a single large array of qubits in a given scope.

I agree that's an important scenario to optimize for, but I would also expect other allocation and deallocation patterns to be relatively common as well. For instance, in considering reversible implementations of classical logic (i.e., oracle synthesis with pebbling), temporary registers of qubits may be allocated and deallocated fairly frequently to hold intermediate results needed for uncomputation steps. On the other hand, variational algorithms such as the QCCC classifier packaged with the QML library will more commonly allocate and then deallocate a small number of qubits many times.

irinayat-MS:
> For instance, in considering reversible implementations of classical logic (i.e., oracle synthesis with pebbling), temporary registers of qubits may be allocated and deallocated fairly frequently to hold intermediate results needed for uncomputation steps. On the other hand, variational algorithms such as the QCCC classifier packaged with the QML library will more commonly allocate and then deallocate a small number of qubits many times.

It would be great to have those scenarios added to qperf suite because then we can measure rather than speculate. For example, if the temporary registers are small or allocated/deallocated in different scopes, I doubt we'd see any perf impact of how the frame stores locals (but again, I'm just speculating here).

cgranade:
> It would be great to have those scenarios added to qperf suite because then we can measure rather than speculate.

Agreed entirely, and I'm happy to help in getting those scenarios added.

irinayat-MS:
> I'm happy to help in getting those scenarios added.

I believe, it should be possible to add more tests into https://github.com/microsoft/qdk/tree/main/Tests/Perf/Library. If you can give me a standalone file like this for any of the important scenarios that aren't covered yet, I'll attempt extending the qdk's perf runs (with Andres' blessing, of course).

bettinaheim:
@cesarzc For your awareness regarding additional perf tests for the runtime.

alan-geller:
> > I'm happy to help in getting those scenarios added.
> 
> I believe, it should be possible to add more tests into https://github.com/microsoft/qdk/tree/main/Tests/Perf/Library. If you can give me a standalone file like this for any of the important scenarios that aren't covered yet, I'll attempt extending the qdk's perf runs (with Andres' blessing, of course).

That sounds like an excellent approach. Real data is a wonderful thing. Thanks, Irina and Chris!

DmitryVasilevsky:
What prompted the original change is that a very simple test was very slow with borrowing. It was also relatively slow without borrowing.

```
      using (q = Qubit[1000000]) {
          Message("+1000000");
      }
      Message("-1000000");

```

irinayat-MS:
> What prompted the original change is that a very simple test was very slow with borrowing. It was also relatively slow without borrowing.

Hm. This is exactly the case, covered by the current qperf tests, albeit the tests go up to only 50000 qubits. Conceivably, there might be a perft cliff on larger data (if C# CRT is not optimizing shifts as well in big lists) but I'm not seeing it in my local micro benchmarks (numbers are in micro sec):
```
Remove at head on ints: 88403.6272
Remove at tail on ints: 6.4346
Remove value forward on ints: 88496.8341
Remove value backward on ints: 158196.9465
Remove at head on custom type: 207272.69
Remove at tail on custom type: 6.8854
Remove value forward on custom type: 207194.2313
Remove value forward on custom type: has been running for more than 10 min and not finished yet.  
```
Therefore, I propose the following: 
1) commit this change to get back to baseline performance;
2) add Chris' scenarios and let them collect perf data for a few builds;
3) look into how to speedup massive allocations/releases of qubit arrays, either by using different data structures to store locals or by batching the operation.

DmitryVasilevsky:
@irinayat-MS agreed. The original change made some important scenarios slower so let's get back to baseline and re-evaluate the fix.

DmitryVasilevsky (approved):

Issue #436, opened 2020-11-12
Change IWorkspace.GetSasUriAsync return type to Task<Uri>


mirounga:
** Please describe what you would like the feature to accomplish.**
Currently the method returns string that is immediately converted to Uri object by the callers.
I suggest keeping it strongly typed and use the Uri class throughout.

** Describe the solution you'd like **
Change the return type to Task<Uri>

** Describe alternatives you've considered **

** Additional context **

bettinaheim:
That seems reasonable, thanks for filing.

PR #438, opened 2020-11-12
Using Uri class to represent URI

mirounga:
Fix for issue #436


bettinaheim:
@mirounga Thanks for implementing this change!
@cesarzc Could you please take a quick look and advice on integration with the service?

src/Azure/Azure.Quantum.Client/generated/Models/SasUriResponse.cs
```diff
@@ -10,8 +10,8 @@
 
 namespace Microsoft.Azure.Quantum.Client.Models
 {
+    using System;
```
george-moussa, line 13: this file is auto-generated from swagger, we should see if swagger can generate Uri directly but editing manually will be overwritten by next version

src/Azure/Azure.Quantum.Client/generated/Models/JobDetails.cs
```diff
@@ -12,6 +12,7 @@ namespace Microsoft.Azure.Quantum.Client.Models
 {
     using Microsoft.Rest;
```
george-moussa, line 13: Same for this file, it's auto-generated

PR #441, opened 2020-11-14
Add --location option to standalone executable submit command

cesarzc:
This change adds a `--location` option to the standalone executable `submit` command. This new option provides the ability to specify the location of the default endpoint.


src/Simulation/EntryPointDriver/Driver.cs
```diff
@@ -253,6 +255,12 @@ internal static class Driver
         internal static readonly OptionInfo<Uri?> BaseUriOption = new OptionInfo<Uri?>(
             ImmutableList.Create("--base-uri"), default, "The base URI of the Azure Quantum endpoint.");
 
+        /// <summary>
+        /// The base URI option.
+        /// </summary>
+        internal static readonly OptionInfo<string?> LocationOption = new OptionInfo<string?>(
+            ImmutableList.Create("--location"), default, "The location to use with the default endpoint.");
```
anpaz, line 262: You should indicate what the default is (westus) in the help message, or does that happen automatically?

anpaz (approved):

anpaz:
I don't see where the default location gets applied. Just wanted to make sure it is indeed optional and if not provided we're using westus.

src/Simulation/EntryPointDriver/Azure.cs
```diff
@@ -267,10 +272,27 @@ internal sealed class AzureSettings
...
+                return AadToken is null
+                    ? new Workspace(Subscription, ResourceGroup, Workspace, baseUri: BaseUri)
+                    : new Workspace(Subscription, ResourceGroup, Workspace, AadToken, baseUri: BaseUri);
+            }
+            else if (Location != null)
+            {
+                return AadToken is null
+                    ? new Workspace(Subscription, ResourceGroup, Workspace, location: Location)
+                    : new Workspace(Subscription, ResourceGroup, Workspace, AadToken, location: Location);
+            }
```
samarsha, line 288: Can we make the `--base-uri` and `--location` options mutually exclusive (i.e. providing both of them is an error) rather than ignoring `--location` if `--base-uri` is provided?

src/Simulation/EntryPointDriver/Driver.cs
cesarzc: Since in principle the driver itself is agnostic about the default endpoint, I think it should also be agnostic about the default location. I added the default location and endpoint in the `Constants` class in `src/Azure/Azure.Quantum.Client/Utility/Constants.cs`.

src/Simulation/EntryPointDriver/Azure.cs
cesarzc: I added the validation to the `submit` command that makes `--base-uri` and `--location` mutually exclusive. Since the validation now throws an error when message when both options are included, I think it is ok to leave this logic as is since the `AzureSettings` object is kind of independent and this logic expresses the settings' precedence.

cesarzc:
> 
> 
> I don't see where the default location gets applied. Just wanted to make sure it is indeed optional and if not provided we're using westus.

The default is applied when creating the Workspace. It can be seen in the constructors at `src/Azure/Azure.Quantum.Client/JobManagement/Workspace.cs`.

src/Azure/Azure.Quantum.Client/JobManagement/Workspace.cs
```diff
@@ -267,5 +313,8 @@ public async Task<string> GetSasUriAsync(string containerName, string blobName =
                 jobId,
                 inner);
         }
+
+        private static string NormalizeLocation(string location) =>
+            string.Concat(location.Where(c => !char.IsWhiteSpace(c))).ToLower();
```
rmshaffer, line 318: In IQ#, we also check that the resulting string is a valid hostname. Otherwise the `Uri` that is created may be invalid or point to some totally different domain. (e.g., imagine the user provides `mydomain.com/quantum` as a `--location` parameter.) See here:
https://github.com/microsoft/iqsharp/blob/a31142b969f4c49b82338c3ea61b8edbde339d52/src/AzureClient/AzureEnvironment.cs#L73-L78
Obviously this is purely a usability thing, not a security thing, since the user already has the ability to override the URI via `--base-uri`. Up to you whether you want to do that.

src/Simulation/EntryPointDriver/Options.cs
```diff
@@ -80,5 +80,36 @@ public static Option<T> CreateOption<T>(string name, string description)
                     Arity = ArgumentArity.OneOrMore
                 }
             };
+
+        /// <summary>
+        /// Creates a prefix and alias tuple from a string representing an option alias provided through the command line
+        /// following the convention that muti-character aliases start with '--' and single-character aliases start with '-'.
+        /// </summary>
+        /// <param name="commandLineAlias">Option alias provided through the command line.</param>
+        /// <returns>TODO.</returns>
```
rmshaffer, line 89: Reminder that there is a `TODO` here.

rmshaffer (approved):

src/Azure/Azure.Quantum.Client/JobManagement/Workspace.cs
samarsha, line 318: Why is whitespace not allowed but other invalid URI characters are?

src/Simulation/EntryPointDriver/Driver.cs
```diff
@@ -99,11 +100,15 @@ public async Task<int> Run(string[] args)
...
             AddOptionIfAvailable(submit, BaseUriOption);
+            AddOptionIfAvailable(submit, LocationOption);
             AddOptionIfAvailable(submit, JobNameOption);
             AddOptionIfAvailable(submit, ShotsOption);
             AddOptionIfAvailable(submit, OutputOption);
             AddOptionIfAvailable(submit, DryRunOption);
             AddOptionIfAvailable(submit, VerboseOption);
+            MarkOptionsAsMutuallyExclusive(
+                submit, 
+                new[] { BaseUriOption.Aliases.First(), LocationOption.Aliases.First()});
```
samarsha, line 111:
```suggestion
                new[] { BaseUriOption.Aliases.First(), LocationOption.Aliases.First() });
```

src/Simulation/EntryPointDriver/Driver.cs
```diff
@@ -208,6 +214,41 @@ private void AddOptionIfAvailable<T>(Command command, OptionInfo<T> option)
...
+        /// Adds a validator to the command such that the specified options cannot be simultaneously included.
+        /// </summary>
+        /// <param name="command">The command to add the validator to.</param>
+        /// <param name="primaryAliases">The primary aliases of the options to be marked as mutually exclusive.</param>
+        /// <remarks>Options must already be added to the command before invoking this method.</remarks>
+        private void MarkOptionsAsMutuallyExclusive(Command command, string[] primaryAliases) =>
+            command.AddValidator(result =>
+            {
+                var presentAliases = new List<string>();
+                foreach(var rawAlias in primaryAliases)
```
samarsha, line 228:
```suggestion
                foreach (var rawAlias in primaryAliases)
```

src/Simulation/EntryPointDriver/Options.cs
```diff
@@ -80,5 +80,36 @@ public static Option<T> CreateOption<T>(string name, string description)
...
+        internal static Tuple<string, string> PrefixAliasTuple(string commandLineAlias)
```
samarsha, line 90:
```suggestion
        internal static (string, string) PrefixAliasTuple(string commandLineAlias)
```

src/Simulation/EntryPointDriver/Options.cs
```diff
@@ -80,5 +80,36 @@ public static Option<T> CreateOption<T>(string name, string description)
...
+                prefix = singleCharacterAliasPrefix;
+                alias = commandLineAlias.Substring(prefix.Length);
+            }
+            else
+            {
+                prefix = commandLineAlias.Length > 1 ? multiCharacterAliasPrefix : singleCharacterAliasPrefix;
+                alias = commandLineAlias;
+            }
+
+            return new Tuple<string, string>(prefix, alias);
```
samarsha, line 112:
```suggestion
            return (prefix, alias);
```

src/Simulation/EntryPointDriver/Driver.cs
```diff
@@ -208,6 +214,41 @@ private void AddOptionIfAvailable<T>(Command command, OptionInfo<T> option)
                     $"The required option {option.Aliases.First()} conflicts with an entry point parameter name.");
             }
         }
+
+        /// <summary>
+        /// Adds a validator to the command such that the specified options cannot be simultaneously included.
+        /// </summary>
+        /// <param name="command">The command to add the validator to.</param>
+        /// <param name="primaryAliases">The primary aliases of the options to be marked as mutually exclusive.</param>
+        /// <remarks>Options must already be added to the command before invoking this method.</remarks>
```
samarsha, line 223: I'm not sure that this remarks section is true - as long as the validator is added before the command line is parsed, I don't think the order relative to the options being added matters?

src/Simulation/EntryPointDriver/Driver.cs
```diff
@@ -208,6 +214,41 @@ private void AddOptionIfAvailable<T>(Command command, OptionInfo<T> option)
...
+        private void MarkOptionsAsMutuallyExclusive(Command command, string[] primaryAliases) =>
+            command.AddValidator(result =>
+            {
+                var presentAliases = new List<string>();
+                foreach(var rawAlias in primaryAliases)
+                {
+                    var (prefix, alias) = Options.PrefixAliasTuple(rawAlias);
+                    var option = command.Options.Where(o => o.Aliases.Contains(alias)).FirstOrDefault();
```
samarsha, line 231: I think if instead of `o.Aliases`, you used `o.RawAliases` which includes the `--` or `-`, you could do this without needing `PrefixAliasTuple`?

src/Simulation/EntryPointDriver/Driver.cs
```diff
@@ -208,6 +214,41 @@ private void AddOptionIfAvailable<T>(Command command, OptionInfo<T> option)
...
+                    if (option == null)
+                    {
+                        continue;
+                    }
+
+                    var presentAlias = option.Aliases.Where(a => result.Children.Contains(a)).FirstOrDefault();
```
samarsha, line 237:
```suggestion
                    var presentAlias = option.Aliases.Where(result.Children.Contains).FirstOrDefault();
```

src/Simulation/EntryPointDriver/Driver.cs
```diff
@@ -208,6 +214,41 @@ private void AddOptionIfAvailable<T>(Command command, OptionInfo<T> option)
...
+                    if ((!string.IsNullOrEmpty(presentAlias)) &&
+                        (result.Children.GetByAlias(presentAlias).Tokens.Count > 0))
```
samarsha, line 239:
```suggestion
                    if (!string.IsNullOrEmpty(presentAlias) &&
                        result.Children.GetByAlias(presentAlias).Tokens.Count > 0)
```

src/Simulation/EntryPointDriver/Azure.cs
samarsha: Thanks. Can you state the precedence in the doc comments for the `BaseUri` and `Location` properties?

src/Azure/Azure.Quantum.Client/JobManagement/Workspace.cs
cesarzc: Removing whitespace characters was part of the feature specification and I will incorporate Ryan's suggestion of additionally checking it is a valid location through Uri.CheckHostName.

cesarzc: That's a good point, thanks for the suggestion, I'll incorporate it :)

src/Simulation/EntryPointDriver/Driver.cs
cesarzc: You are right! Thanks for catching this, I will remove the remarks section.

src/Simulation/EntryPointDriver/Driver.cs
cesarzc: Possibly, I was not aware of the RawAliases property. I'll give it a try and as you mentioned maybe the PrefixAliasTuple is not needed anymore.

src/Simulation/EntryPointDriver/Options.cs
cesarzc: Thanks for the catch!

src/Simulation/EntryPointDriver/Azure.cs
cesarzc: Added precedence to the doc comments of both properties.

src/Simulation/EntryPointDriver/Options.cs
cesarzc: Your suggestion of using the `RawAliases` property made this no longer needed :)

src/Simulation/EntryPointDriver/Options.cs
cesarzc: No longer needed \o/

src/Simulation/EntryPointDriver/Driver.cs
cesarzc: Updated.

src/Simulation/EntryPointDriver/Driver.cs
cesarzc: Using `o.RawAliases` now.

src/Simulation/EntryPointDriver/Options.cs
cesarzc: This method was removed in the latest iteration since it was no longer needed.

src/Azure/Azure.Quantum.Client/JobManagement/Workspace.cs
samarsha: Is there a link to the feature specification?

src/Simulation/EntryPointDriver/Azure.cs
```diff
@@ -235,9 +235,16 @@ internal sealed class AzureSettings
 
         /// <summary>
         /// The base URI of the Azure Quantum endpoint.
+        /// If both BaseUri and Location properties are not null, BaseUri takes precedence.
```
samarsha, line 238: The tags can help make sure the doc comment stays up to date if the property names change in the future:

```suggestion
        /// If both <see cref="BaseUri"/> and <see cref="Location"/> properties are not null, <see cref="BaseUri"/> takes precedence.
```

src/Simulation/EntryPointDriver/Azure.cs
```diff
@@ -235,9 +235,16 @@ internal sealed class AzureSettings
...
         /// </summary>
         public Uri? BaseUri { get; set; }
 
+        /// <summary>
+        /// The location to use with the default Azure Quantum endpoint.
+        /// If both BaseUri and Location properties are not null, BaseUri takes precedence.
```
samarsha, line 244:
```suggestion
        /// If both <see cref="BaseUri"/> and <see cref="Location"/> properties are not null, <see cref="BaseUri"/> takes precedence.
```

src/Simulation/EntryPointDriver/Driver.cs
```diff
@@ -208,6 +214,39 @@ private void AddOptionIfAvailable<T>(Command command, OptionInfo<T> option)
...
+                var presentAliases = new List<string>();
+                foreach (var rawAlias in primaryAliases)
+                {
+                    var option = command.Options.Where(o => o.RawAliases.Contains(rawAlias)).FirstOrDefault();
+                    if (option == null)
+                    {
+                        continue;
+                    }
+
+                    var presentAlias = option.RawAliases.Where(result.Children.Contains).FirstOrDefault();
```
samarsha, line 235:
```suggestion
                    var presentAlias = option?.RawAliases.Where(result.Children.Contains).FirstOrDefault();
```

samarsha (approved):

src/Azure/Azure.Quantum.Client/JobManagement/Workspace.cs
cesarzc: I added a validator to the location which passes the normalized value provided by the user to the `Uri.CheckHostName` method to verify it.

I sent a link to you regarding the details on the whitespace :)

cesarzc: This verification is now done through a validator in the command line option itself.

PR #442, opened 2020-11-16
Swernli/merge/feature/decomp

swernli:
This PR merges the working branch from #421 back into the feature/decomp branch so we can collect some more changes there. See that PR for the change history, which includes updates from CR feedback and several merges from main.


bettinaheim (approved):
I quickly skimmed it. Rather than doing a detailed review here, I'll take a closer look at the feature to main that follows this PR.

PR #444, opened 2020-11-17
QubitManager: use Dictionary for storing frame locals

irinayat-MS:
When scope is tracked for qubit allocations (i.e. when borrowing is enabled), time to release qubits in a frame grows quadratically on the number of qubits allocated in that frame. This is because the locals are stored in a list and we remove qubits one by one at the head of the list, which leads to shifting the whole list on each release.

I tried fixing this by releasing the qubits in reverse order and manually searching for the qubit to be released in the list in reverse order as well (otherwise Remove(qubit) method becomes quadratic as it has to search for the qubit first and the qubit is always at the end of the list). This solution removed the quadratic complexity and yielded ~300ms running time to allocate and release a single 10^6 qubit array. This is probably the fastest we can get with built-in library containers. However, this approach relies on the fact that Q# would allocate/release qubits in particular order, which should be a relatively safe assumption, given Q#'s `using` construct, but it also requires multiple methods of QubitManager to be aware of this ordering dependency, making maintenance of this code more difficult.

I also tried replacing the list with ordering-agnostic Dictionary and while it yielded not as good results in the 10^6 array allocation/release benchmark (~580ms), I believe the difference isn't large enough to warrant the complexity -- and fragility! -- of the tuned list approach above.

The perf results show small improvements on ToffoliChainTest50000 test, as expected, but somehow considerably better results in some other tests for DefaultSimulator and NoBorrowingSimulator, which doesn't make much sense especially for the latter, as the fix should have only affected scenarios when borrowing is enabled (improvements in default simulator are also surprising because it cannot run with the number of qubits necessary for the quadratic nature of StackFrame locals to manifest itself, that is, unless it was borrowing the same qubits _a lot_).

[QubitManager_Perf.xlsx](https://github.com/microsoft/qsharp-runtime/files/5554880/QubitManager_Perf.xlsx)


src/Simulation/Common/QubitManager.cs
```diff
@@ -544,7 +547,7 @@ internal IQArray<Qubit> Borrow(long numToBorrow, IEnumerable<Qubit> qubitsInUse)
         {
             var inUse = new Stack<Qubit>(qubitsInUse.Distinct().OrderByDescending(q => q.Id));
             var borrowed = QArray<Qubit>.Create(numToBorrow);
-            long numBorrowed = System.Math.Min(NumAllocatedQubits - inUse.Count, numToBorrow);
+            long numBorrowed = System.Math.Min(NumAllocatedQubits - inUse.Count, numToBorrow);
```
DmitryVasilevsky, line 550: Do you know why this line is marked as changed?

src/Simulation/Common/QubitManager.cs
```diff
@@ -81,20 +81,20 @@ protected class StackFrame
...
 
-            public List<Qubit> Locals
+            public HashSet<Qubit> Locals
             {
                 get
                 {
                     if (_locals == null)
                     {
-                        _locals = new List<Qubit>();
+                        _locals = new HashSet<Qubit>();
```
DmitryVasilevsky, line 97: Is Qubit class or record (i.e. by reference or by value)? Does default hash provider work well?

irinayat-MS: Qubit is a class and overrides GetHashCode (https://github.com/microsoft/qsharp-runtime/blob/main/src/Simulation/Core/Qubit.cs)

src/Simulation/Common/QubitManager.cs
irinayat-MS: Line ending. Normalized the file to use "LF" consistently. Thank you for pointing this out!

DmitryVasilevsky (dismissed):

DmitryVasilevsky:
I don't get it. How did it work before?

irinayat-MS:
> I don't get it. How did it work before?

It didn't. The benchmark I ran doesn't exercise the codepath that adds duplicated items to the locals list. Thankfully, the CI tests found the problem.

DmitryVasilevsky:
I mean before the hash set. But I can see now that there were multiple entries in this array with the same value.

irinayat-MS:
@alan-geller , @DmitryVasilevsky - ping? I'd like to merge this change in.

alan-geller (approved):
Looks fine to me.

DmitryVasilevsky (approved):

src/Simulation/Common/QubitManager.cs
```diff
@@ -127,7 +127,7 @@ protected IEnumerable<Qubit> QubitsInUse(StackFrame frame)
                 return Qubit.NO_QUBITS;
             }
 
-            return frame.Locals.Concat(frame.QubitsInArgument ?? Array.Empty<Qubit>())
+            return frame.Locals.Keys.Concat(frame.QubitsInArgument ?? Array.Empty<Qubit>())
```
DmitryVasilevsky, line 130: Are you sure the change works here. Before it was returning duplicates and now it does not.

irinayat-MS: All uses of this methods call Distinct on the returned IEnumerable so the change is correct, but I should probably clean it up to remove the necessity to call Distinct altogether. Thanks for pointing this out!

src/Simulation/Common/QubitManager.cs
```diff
@@ -368,8 +383,8 @@ public void Disable(IQArray<Qubit> qubitsToDisable)
             if (!DisableBorrowing)
             {
                 if (!operationStack.Peek().Locals.TryAdd(ret, 1))
-                {
```
DmitryVasilevsky: Here and in 3 other places. Code seems to be the same, yet it's marked as changed...

src/Simulation/Common/QubitManager.cs
```diff
@@ -43,6 +43,7 @@ public class QubitManager : IQubitManager
         private long AllocatedForBorrowing; // All qubits allocated only for borrowing, will be marked with this number or higher.
         private long free; // Points to the first free (unallocated) qubit.
         private long freeTail; // Points to the last free (unallocated) qubit. Only valid iff (!EncourageReuse).
+        private static HashSet<Qubit> EMPTY_SET = new HashSet<Qubit>();
```
DmitryVasilevsky, line 46: Suggest making this readonly and, if possible, create immutable hashset.

src/Simulation/Common/QubitManager.cs
irinayat-MS: Yes, line endings in this file weren't consistent. VS Code normalized them to "LF"

src/Simulation/Common/QubitManager.cs
```diff
@@ -120,23 +121,37 @@ public virtual void OnOperationEnd(ICallable _, IApplyData values)
...
         {
             if (DisableBorrowing || frame == null)
             {
-                return Qubit.NO_QUBITS;
+                return EMPTY_SET;
             }
 
-            return frame.Locals.Keys.Concat(frame.QubitsInArgument ?? Array.Empty<Qubit>())
-                .Where(q => q != null && !this.IsDisabled(q));
+            HashSet<Qubit> inUse = new HashSet<Qubit>();
```
DmitryVasilevsky, line 131: Would it make sense to pre-allocate a little? Maybe size of locals?

src/Simulation/Common/QubitManager.cs
irinayat-MS: Removed it. The optimization would be useful when borrowing is disabled to avoid allocating an empty set each time, however, we already check not to call QubitsInUse if borrowing is disabled, and I assume the null frame case isn't that common.

I'd rather not make the returned hashset from QubitsInUse immutable, because we create it afresh on each invocation and at this point, if the caller would like to modify it, they should be able to.

src/Simulation/Common/QubitManager.cs
irinayat-MS: yep, good idea

DmitryVasilevsky (approved):

PR #445, opened 2020-11-17
Fix the check for build agent to enable QIR Runtime CI

irinayat-MS:
The QIR Runtime CI hasn't been running because of an incorrect check in set-env.ps1. If they were, they would catch a build break on Linux and find find inconsistencies between tests setup across platforms.

The bootstrap logic is doing more than just preparing for the build. The bootstrap scripts are also inconsistent across platforms. To begin with, I've switched the CI build not to use bootstrap scripts. Those are still useful for setting the repo for local development and I don't understand all requirements here, so keeping them intact.


alan-geller (approved):

irinayat-MS:
@anpaz , could you please take a look?

bootstrap.ps1
```diff
@@ -2,6 +2,7 @@
 # Licensed under the MIT License.
 
 $ErrorActionPreference = 'Stop'
+& "$PSScriptRoot/build/set-env.ps1"
```
irinayat-MS, line 5: Need this to set ENABLE_QIRRUNTIME before we use it later in the script

build/pack.ps1
```diff
@@ -10,7 +10,7 @@ Write-Host "##[info]Copy Native simulator xplat binaries"
 pushd ../src/Simulation/Native
 If (-not (Test-Path 'osx')) { mkdir 'osx' }
 If (-not (Test-Path 'linux')) { mkdir 'linux' }
-$DROP = "$Env:DROP_NATIVE/src/Simulation/Native/build"
+$DROP = "$Env:DROP_NATIVE/src/Simulation/Native/build/drop"
```
irinayat-MS, line 13: To unify consumption of the native simulator library across platforms, I added 'install' command to the cmake files so the dll and tests can be "installed" into a known folder.

build/steps-init.yml
```diff
@@ -16,8 +16,8 @@ steps:
 
 
 ##
-# Bootstrap
+# Custom pre-reqs
 ##
-- pwsh: ./bootstrap.ps1
```
irinayat-MS, line 21: The purpose of bootstrap scripts is to prep the repository for local development, including building some of the components, which was a mess to re-use in CIs. So I introduced new prereq/build/test scrips for native simulator but I didn't touch any of the bootstrap logic.

src/QirRuntime/lib/QIR/callables.cpp
```diff
@@ -138,7 +139,7 @@ struct QirCallable
         {
             if (this->controlledDepth > 0)
             {
-                throw new std::exception("multi-level controlled operations not implemented yet");
+                throw new std::runtime_error("multi-level controlled operations not implemented yet");
```
irinayat-MS, line 142: This is the build-break on Linux that caused the discovery of the fact that CI build isn't running :)

build/prerequisites.ps1
```diff
@@ -0,0 +1,32 @@
...
+if ($Env:ENABLE_NATIVE -ne "false") {
+    Push-Location (Join-Path $PSScriptRoot "../src/Simulation/Native")
+        .\install-prereqs.ps1
+    Pop-Location
+} else {
+    Write-Host "Skipping installing prerequisites for native simulator because ENABLE_NATIVE variable set to: $Env:ENABLE_NATIVE"
+}
+
+# At the moment the QIR Runtime build isn't enabled locally by default.
+if ($Env:ENABLE_QIRRUNTIME -eq "true") {
```
anpaz, line 25: In the e2e build this is also enabled (and we said we didn't want it enabled on e2e yet). The problem is that the `set-env.ps1` sets it automatically when running on a build agent (regardless of the build).
My recommendation is to not set it automatically on the set-env.ps1 script. Instead, create a `Enable.QIR` variable in the ci.yml and set it to true there.

bootstrap.ps1
```diff
@@ -3,30 +3,32 @@
...
+    .\prerequisites.ps1
 Pop-Location
 
-if ($Env:ENABLE_QIRRUNTIME -eq "true") {
-    Push-Location (Join-Path $PSScriptRoot "src/QirRuntime")
-        .\install-prereqs.ps1
-    Pop-Location
-} else {
-    Write-Host "Skipping installing prerequisites for qir runtime because ENABLE_QIRRUNTIME variable set to: $Env:ENABLE_QIRRUNTIME."
+# Temporary hack until switch qdk build pipeline to use the new build scripts (as a result it will build the native
```
anpaz, line 12: I don't understand why this is necessary...

build/prerequisites.ps1
```diff
@@ -0,0 +1,32 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
anpaz, line 1: maybe this file should be named `install-prereqs.ps1` to keep it consistent.

build/pack.ps1
anpaz: actually, I think the change here is ok, as it seems we don't really copy the dlls here anymore, they are copied during bootstrap (so this might not be necessary at all). I'm a little confused why they are getting copied, though, since bootstrap.ps1 the last thing it does is call intsll-prereqs on QIR, and QIR doesn't have this code; any ideas of where this is coming from?
https://dev.azure.com/ms-quantum-public/Microsoft%20Quantum%20(public)/_build/results?buildId=19385&view=logs&j=663ee715-dcc5-5b8a-8432-471131753a80&t=ebac3767-3683-5916-990d-44e0e78421f8&l=1492

anpaz (changes requested):

src/QirRuntime/build-qir-runtime.ps1
```diff
@@ -1,34 +1,51 @@
...
-    New-Item -Path $qirRuntimeBuildFolder -ItemType "directory"
-}
+    if (($IsMacOS) -or ((Test-Path Env:AGENT_OS) -and ($Env:AGENT_OS.StartsWith("Darwin"))))
+    {
+        Write-Host "On MacOS build QIR Runtim using the default C/C++ compiler (should be AppleClang)"
+    }
+    elseif (($IsLinux) -or ((Test-Path Env:AGENT_OS) -and ($Env:AGENT_OS.StartsWith("Lin"))))
+    {
+        Write-Host "On Linux build QIR Runtime using Clang"
+        $env:CC = "/usr/bin/clang"
```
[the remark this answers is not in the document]
irinayat-MS: The variable has to be set differently for QIR Runtime and Native (they use different compilers). I do restore it at the end of the script. The pre-req script installs clang iff QIR Runtime build is enabled, so it doesn't seem particularly useful to validate here, or am I missing your point?

src/QirRuntime/build-qir-runtime.ps1
```diff
@@ -1,34 +1,51 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
```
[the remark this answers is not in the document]
irinayat-MS: build.py uses slightly different folder structure to avoid conflicts between win and wsl builds on the same machine. I can also run it from cmd rather than from PS which I find convenient (and linux users might prefer it to PS as well). it's for local development only and is more flexible if you'd like to do custom builds locally (e.g. easy to do just make)

README.md
```diff
@@ -40,13 +40,14 @@ To build on Windows:
         * **Desktop development with C++**
```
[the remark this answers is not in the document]
irinayat-MS: done

bootstrap.ps1
irinayat-MS: Before this change _bootstrap_ script would create build folders and run cmake to generate build definitions, and _build_ script would run "cmake --build". I placed both commands into the _build_ script to avoid spreading across multiple scripts the knowledge about target folders, etc. Because I don't know whether anything in e2e pipeline relies on the build folders being created I added the call to new build script here. I suppose, I can remove it and see if the pipeline gets unhappy.

build/pack.ps1
irinayat-MS: I am trying to solve two problems here:
1. depending on the context (ci/e2e/local) we might build either DEBUG or RELEASE build of native simulator but the consumers wouldn't know which one to pick up
2. the target folder of cmake build is created differently on Windows vs Linux and the install into "drop" folder helps unifying the handling across platforms.

My grasp of who copies what and when during the pipeline is rather shaky. The build kept failing and I just searched through the repo for mentions of "libMicrosoft.Quantum.Simulator.Runtime.so" and re-pointed them to pick up from the unified drop location.

build/prerequisites.ps1
irinayat-MS: Because it at the root I thought it might do more varied stuff than the individual components that only need to install prereqs. Actually, I'll rename them all to prerequisites.ps1...

build/prerequisites.ps1
irinayat-MS: Done. Setting the variable in ci.yml is exactly what I need. Thanks!

bootstrap.ps1
irinayat-MS: e2e pipeline does become unhappy. I don't think the pipeline calls the build.ps1 script at all, instead it invokes cmake directly from https://github.com/microsoft/qdk/blob/main/build/steps-native.yml. The yml should be modified to either do both cmake steps or to call build-native-simulator.ps1. To coordinate between the two pipelines, I've added the hack back. After the change is in, we can update qdk to use the new script (explicit cmake), and after that we can remove the hack.

build/ci.yml
```diff
@@ -13,6 +13,9 @@ variables:
   Build.Minor: 14
   Drops.Dir: $(Build.ArtifactStagingDirectory)/drops
   Drop.Native: $(System.DefaultWorkingDirectory)/xplat
+  CI: "true"
```
anpaz, line 16: not sure if a `CI` variable is needed, or just checking the `AGENT_NAME` is good enough.

src/QirRuntime/README.md
```diff
@@ -32,11 +32,9 @@ CI builds and tests are enabled for this project. The build has no external depe
...
 1. Install Clang, Ninja and CMake from the public distros.
 2. Add all three to your/system `%PATH%`.
-3. Install VS 2019 (Clang uses MSVC's standard library on Windows).
-4. Install .NET Core 3.1
-5. <_optional_> Either clone LLVM's repo and build LLVM locally to produce your own `opt.exe` and other IR tools or download LLVM's dev package.
-6. Install clang-tidy and clang-format if your Clang/LLVM packages didn't include the tools.
-7. <_optional_> To use build/test scripts install Python 3.8.
+3. Install VS 2019 and enable "Desktop development with C++" component (Clang uses MSVC's standard library on Windows).
+4. Install clang-tidy and clang-format if your Clang/LLVM packages didn't include the tools.
+5. <_optional_> To use build/test scripts install Python 3.8.
```
anpaz, line 37: Is Python still required? looks like you moved everything to PS1...

anpaz (approved):
lgtm.

src/QirRuntime/README.md
irinayat-MS: python script gives me more flexibility with local builds (e.g. just running make part of cmake) and it also creates windows/linux subfolders so WSL builds don't override the host (which is unnecessary in the cloud build). I'd like to keep it, at least for now.

build/ci.yml
irinayat-MS: it will be gone after e2e pipeline starts using the build script, and at this particular moment of pure greenness I'd rather not tinker with the PR any more :D

PR #446, opened 2020-11-18
Reuse some qubits when minimal depth is requested

DmitryVasilevsky:
This change enables reuse of some qubits when minimal depth is requested. Before all qubits were allocated fresh. Such reuse does not increase depth, but may not be optimal given minimal depth. This is to resolve customer issue as no reuse of qubits results in excessively high width.
* Qubit times are fixed at multi-qubit gates.
* Initial single-qubit gate chains are adjusted in time when multi-qubit gates are encountered.
* Qubit busy intervals are stored in two red-black trees (SortedSet) for left and right ends.
* Reuse is attempted before or after existing qubits when qubit is released.
* Complex time is used to account for primitive gates with depth 0, otherwise dependencies may be broken.
* Some care is used to handle depth - primitive qubit gate depth are of type double in the interface.


src/Simulation/Simulators.Tests/Circuits/ResourcesEstimator.qs
```diff
@@ -48,6 +48,21 @@ namespace Microsoft.Quantum.Simulation.Simulators.Tests
             T(q[0]);
         }
     }
+
+    operation QubitReuseWithOptimizedDepth() : Unit
```
bettinaheim, line 52: It would be good to add the case filed in the issue that spawned this work.

src/Simulation/QCTraceSimulator/SortedQubitPool.cs
```diff
@@ -0,0 +1,321 @@
...
+    /// <summary>
+    /// Time to track qubit availability (DepthTime). Time is also used to track dependencies:
+    /// Gates that happen later on a qubit cannot be scheduled before gates on the same qubit.
+    /// This is tracked transitively via time, but zero gate depth prevents this from happening.
+    /// This is solved by having additional time component - DependencyTime.
+    /// </summary>
+    internal struct ComplexTime
+    {
+        /// <summary>
+        /// Time used to track qubit availability time. Double to match configuration interface.
```
irinayat-MS, line 16: It's not clear what "availability time" is. Time when qubit was allocated? Time when the qubit was last used? Why qubit at all? The class doesn't seem to link to any particular qubit.

src/Simulation/QCTraceSimulator/SortedQubitPool.cs
```diff
@@ -0,0 +1,321 @@
...
+        /// </summary>
+        internal double DepthTime { get; private set; }
+
+        /// <summary>
+        /// This is only used for comparisons when DepthTime is equal. This time tracks dependencies.
```
irinayat-MS, line 21: It's not clear what it means to "track dependencies". Dependencies between what? What kind of dependencies?

src/Simulation/QCTraceSimulator/SortedQubitPool.cs
```diff
@@ -0,0 +1,321 @@
...
+        /// Depth times are assumed to be precise  for comparison.
+        /// </summary>
+        /// <param name="time">Time to subtract from this object.</param>
+        /// <returns>Result of subtraction.</returns>
+        internal ComplexTime Subtract(ComplexTime time) {
+            if (DepthTime < time.DepthTime) {
+                throw new ArgumentException("Result of ComplexTime subtraction is negative.");
+            }
+            if (DepthTime == time.DepthTime) {
+                return new ComplexTime(0, DependencyTime - time.DependencyTime);
```
irinayat-MS, line 61: Is it OK for the DependencyTime to be negative?

src/Simulation/QCTraceSimulator/SortedQubitPool.cs
```diff
@@ -0,0 +1,321 @@
...
+        internal static ComplexTime MaxValue { get; } = new ComplexTime(double.MaxValue, long.MaxValue);
+
+        /// <summary>
+        /// Zero ComplexTime.
+        /// </summary>
+        internal static ComplexTime Zero { get; } = new ComplexTime(0.0, 0);
+    }
+
+    /// <summary>
+    /// Stores qubits sorted by corresponding times. Instantiated to store interval start points and end points.
```
irinayat-MS, line 111: what interval?

src/Simulation/QCTraceSimulator/SortedQubitPool.cs
```diff
@@ -0,0 +1,321 @@
...
+            internal readonly long QubitId;
+            
+            /// <summary>
+            /// Time on which these nodes are sorted.
+            /// </summary>
+            internal ComplexTime Time; // This can be readonly if Sample handling is changed.
+
+            /// <summary>
+            /// Multiple nodes can correspond to the same time. They are added to the linked list via this field.
+            /// If qubits with the same time is common array may be more efficient than the linked list.
```
irinayat-MS, line 136: nit: If qubits with the same time _are_ common

src/Simulation/QCTraceSimulator/SortedQubitPool.cs
```diff
@@ -0,0 +1,321 @@
...
+                        MaxLowerBound = b.Time;
+                        MinUpperBound = b.Time;
+                    }
+                }
+                return result;
+            }
+        }
+
+        /// <summary>
+        /// QubitTimeNodes sorted by Time field. Nodes with the same Time field are stored in a linked list.
```
irinayat-MS, line 222: The comment isn't clear. Do you mean, that nodes with same time should _not_ be added to the set but should be tracked using the linked list facility of QubitTimeNode class?

src/Simulation/QCTraceSimulator/SortedQubitPool.cs
```diff
@@ -0,0 +1,321 @@
...
+        /// </summary>
+        private readonly SortedSet<QubitTimeNode> QubitTime;
```
irinayat-MS, line 224: QubitsTimeline? (otherwise it reads as a single time point of a single qubit)

src/Simulation/QCTraceSimulator/SortedQubitPool.cs
```diff
@@ -0,0 +1,321 @@
...
+        public void Add(long qubitId, ComplexTime time) {
+            QubitTimeNode newNode = new QubitTimeNode(qubitId, time);
+            if (QubitTime.Add(newNode)) {
+                // New item was added to the set, we are done.
+                return;
+            }
+            if (!QubitTime.TryGetValue(newNode, out QubitTimeNode existingNode)) {
+                // We cannot add, but we cannot find a node with the same value.
+                // Is it a floating point glitch? We'll just ignore this one.
+                // If this happens reuse will be hindered, but program will still work.
```
irinayat-MS, line 257: Assert? How would you notice that this is happening (if it does) and reuse isn't as efficient as it could be?

src/Simulation/QCTraceSimulator/SortedQubitPool.cs
```diff
@@ -0,0 +1,321 @@
...
+        }
+
+        /// <summary>
+        /// Find qubit with the specified time in the set. If such qubit is not present,
+        /// find either maximum value less than it or minimum value greater than it.
+        /// This function uses one call to SortedSet.TryGetValue() so it takes log(N) time, where N - number of nodes.
+        /// </summary>
+        /// <param name="requestedTime">Sample time to find</param>
+        /// <param name="getLowerBound">"true" to find maximum value less than sample</param>
+        /// <param name="actualTime">"true" if sample was found in the set.</param>
```
irinayat-MS, line 272: actualTime is ComplexTime, not bool. the comment was probably meant for return

src/Simulation/QCTraceSimulator/SortedQubitPool.cs
```diff
@@ -0,0 +1,321 @@
...
+        /// <returns></returns>
+        public bool FindBound(ComplexTime requestedTime, bool getLowerBound, out ComplexTime actualTime) {
+            // We use the following approach:
+            // We call function TryGetValue on a sorted set. If value is found, we go with it.
+            // Otherwise TryGetValue must make conclusion that the value is absent.
+            // Any comparison-based algorithm without caching must inspect both
+            // maximum value less than the sample and minimum value greater than the sample.
+            // So they msut be among the values seen since reset.
```
irinayat-MS, line 280: typo: msut -> must

src/Simulation/QCTraceSimulator/SortedQubitPool.cs
```diff
@@ -0,0 +1,321 @@
...
+                MinUpperBound = ComplexTime.MaxValue;
+                Sample = sample;
+            }
+
+            /// <summary>
+            /// Compares Time field of two nodes. Also updates MaxLowerBound and MinUpperBound.
+            /// </summary>
+            /// <returns>Result of comparison of Time field of QubitTimeNode</returns>
+            public int Compare(QubitTimeNode a, QubitTimeNode b) {
+                // Note that comparison of a and b nodes to Sample is "by reference" in this function.
```
irinayat-MS, line 196: I don't understand this part. Below, the Sample is allocated once with the default "0" qubit id and the code keeps overwriting its time when needed. How can this "by reference" comparison ever return 'true'?

src/Simulation/QCTraceSimulator/QubitMetrics.cs
```diff
@@ -8,14 +8,27 @@ namespace Microsoft.Quantum.Simulation.QCTraceSimulatorRuntime
     /// </summary>
     public class QubitTimeMetrics
     {
-        // TODO: Qubit Ids are already available in qubits, but DepthCounter doesn't have access to it
+        // Qubit Ids are already available in qubits, but DepthCounter doesn't have access to it
         // in OnPrimitiveOperation because it's not part of IQCTraceSimulatorListener interface.
         // Consider changing architecture to pass qubits rather than metrics in IQCTraceSimulatorListener.
         public long QubitId { get; }
 
+        // If qubit is not fixed in time, this is depth of single-qubit-gate chain on this qubit
```
irinayat-MS, line 16: It's not clear what it means for a qubit to be (or not to be) fixed in time.

src/Simulation/Simulators.Tests/Circuits/ResourcesEstimator.qs
DmitryVasilevsky: QubitReuseSequentialTest and QubitReuseSimultaneousTest are the tests for the specific customer issue.

src/Simulation/QCTraceSimulator/SortedQubitPool.cs
DmitryVasilevsky: Yes.

src/Simulation/Simulators.Tests/ResourcesEstimatorTests.cs
```diff
@@ -1,196 +1,236 @@
...
+        public void QubitReuseWithOptimizedDepthTest() {
+            QCTraceSimulators.QCTraceSimulatorConfiguration config = ResourcesEstimator.RecommendedConfig();
+            config.OptimizeDepth = true;
+            var sim = new ResourcesEstimator(config);
+
+            QubitReuseWithOptimizedDepth.Run(sim).Wait();
+
+            Assert.Equal(2.0, sim.Data.Rows.Find("QubitCount")["Sum"]);
+            Assert.Equal(2.0, sim.Data.Rows.Find("Width")["Sum"]);
+            Assert.Equal(1.0, sim.Data.Rows.Find("Depth")["Sum"]);
```
irinayat-MS, line 132: operation QubitReuseWithOptimizedDepth() has sequential block of:
    T(q2);
    CNOT(q1, q2);
Are CNOT's treated as taking zero time?

src/Simulation/Simulators.Tests/ResourcesEstimatorTests.cs
```diff
@@ -1,196 +1,236 @@
...
+
+        [Fact]
+        public void QubitReuseWithOptimizedDepthTest() {
+            QCTraceSimulators.QCTraceSimulatorConfiguration config = ResourcesEstimator.RecommendedConfig();
+            config.OptimizeDepth = true;
+            var sim = new ResourcesEstimator(config);
+
+            QubitReuseWithOptimizedDepth.Run(sim).Wait();
+
+            Assert.Equal(2.0, sim.Data.Rows.Find("QubitCount")["Sum"]);
```
irinayat-MS, line 130: What is the "QubitCount"? From the whitepaper my understanding was that for OptimizeDepth = true config we should produce two numbers: optimal minimal depth and compatible, as low as we can get, width.

src/Simulation/QCTraceSimulator/QubitMetrics.cs
DmitryVasilevsky: It's not about primary/secondary component. I updated the comment, thanks for some wording.

irinayat-MS: My point was that in the context of this file it's not clear what it means for the qubit to be "fixed or not fixed in time". Maybe refer to the whitepaper doc that explains it?

src/Simulation/QCTraceSimulator/DepthCounter.cs
```diff
@@ -27,15 +27,24 @@ private class OperationCallRecord
...
-            initialCapacity: 128, // Reasonable number to preallocate.
-            defaultAvailabilityTime: 0.0);
+        private readonly bool optimizeDepth = false;
+        private readonly StatisticsCollector<CallGraphEdge> stats;
+        private readonly Stack<OperationCallRecord> operationCallStack;
+
+        // Used when OptimizeDepth = false. We track availability time for each underlying qubits identified by id.
+        private readonly QubitAvailabilityTimeTracker qubitAvailabilityTime;
+
+        // Used when OptimizeDepth = true. We reuse qubits based on busy intervals.
```
irinayat-MS, line 37: It's mildly concerning that some of the fields of the class are only used depending on a value of another field. Are they at least valid with some meaningful defaults or must every use of them be conditioned on `optimizeDepth` setting? If the latter, maintenance of this code will be likely prone to introducing bugs :(

irinayat-MS (commented):
@DmitryVasilevsky, could you please resolve the comments you've addressed and also merge with the latest from main? (currently the PR shows changes to callgraph logic that don't seem related to depth/width computation)

src/Simulation/QCTraceSimulator/QubitMetrics.cs
irinayat-MS: Also, would it be possible to change the whitepaper from .doc to .md? It would be easier to provide feedback on it in this PR and later consume it directly via the github interface.

src/Simulation/QCTraceSimulator/DepthCounter.cs
```diff
@@ -46,6 +55,16 @@ public DepthCounter(IDoubleStatistic[] statisticsToCollect = null)
             OperationCallRecord opRec = new OperationCallRecord();
             opRec.OperationName = CallGraphEdge.CallGraphRootHashed;
             operationCallStack.Push(opRec);
+
+            this.optimizeDepth = optimizeDepth;
+            if (optimizeDepth) {
```
irinayat-MS, line 60: I believe, our C# style guide requires to place braces on separate lines. All of the new code in this file is affected.

src/Simulation/QCTraceSimulator/QubitMetrics.cs
```diff
@@ -8,14 +8,27 @@ namespace Microsoft.Quantum.Simulation.QCTraceSimulatorRuntime
...
+        // If qubit is fixed in time, this is the busy interval end time for this qubit
+        internal ComplexTime EndTime { get; set; }
+
+
+        // If qubit is not fixed in time, this is ComplexTime.MinValue
+        // If qubit is fixed in time, this is the busy interval start time
+        internal ComplexTime StartTime { get; set; }
```
irinayat-MS, line 23: Would it help to add a method like `bool IsFixedInTime() { /*validate that the settings on start/end are consistent and return whether the qubit is fixed in time based on the current values of the fields*/ }` to both document here and use in DepthCounter class?

src/Simulation/QCTraceSimulator/QubitMetrics.cs
DmitryVasilevsky: Wait until I push the updated comment. It should be better.

src/Simulation/QCTraceSimulator/DepthCounter.cs
DmitryVasilevsky: Yes, that's true. I use either qubitAvailabilityTime or the other two fields. Unused fields are left at default value of null. I'm not sure there's a better option short of creating two different depth counters for two different alternatives.

src/Simulation/QCTraceSimulator/DepthCounter.cs
DmitryVasilevsky: Yes, I will address this later.

src/Simulation/QCTraceSimulator/SortedQubitPool.cs
DmitryVasilevsky: Updated wording will be in the next push.
I thought about making this a general class, but it has too much ties with gate timing. Besides very few operations are defined to make it more general.

src/Simulation/QCTraceSimulator/SortedQubitPool.cs
DmitryVasilevsky: Updated wording will be in the next push.

src/Simulation/QCTraceSimulator/SortedQubitPool.cs
DmitryVasilevsky: No. We should always subtract smaller value from larger value. I do checks for DepthTime because it is double and floating point glitches may happen. I'll do it a little different to emphasize this.

DmitryVasilevsky: Updated wording will be in the next push.

src/Simulation/QCTraceSimulator/SortedQubitPool.cs
DmitryVasilevsky: Updated wording will be in the next push.

src/Simulation/QCTraceSimulator/SortedQubitPool.cs
DmitryVasilevsky: Comparison will return 'true' when one of the arguments of this compare is that same Sample instance allocated below. If it's another instance (even with the same values of fields) comparison by value will return false.

src/Simulation/QCTraceSimulator/SortedQubitPool.cs
DmitryVasilevsky: Updated to QubitsSortedByTime.

src/Simulation/QCTraceSimulator/SortedQubitPool.cs
DmitryVasilevsky: It's a standard .net SortedSet, it won't accept nodes with the same time. Updated wording will be in the next push.

src/Simulation/QCTraceSimulator/SortedQubitPool.cs
DmitryVasilevsky: Minimum width is not guaranteed anywhere, it's all based on heuristics. So the program will continue correctly with less reuse.
I believe we discourage debug asserts in runtime code. Added Debug.Assert nevertheless.

src/Simulation/QCTraceSimulator/QubitMetrics.cs
DmitryVasilevsky: I'm reluctant to converting .doc to .md. It has formulas and embedded images. I had troubles with both before. I'll check if they work here.

DmitryVasilevsky:
Consider converting .doc to .md

src/Simulation/Simulators.Tests/ResourcesEstimatorTests.cs
DmitryVasilevsky: Yes, with default timings only T gates are depth 1. The rest are depth 0.

src/Simulation/QCTraceSimulator/QubitMetrics.cs
irinayat-MS: Images can be checked-in alongside the md file and linked to using relative links. Don't know how to do formulas, but yours aren't complex, are they?

irinayat-MS: btw, the updated comment is perfect :)

src/Simulation/Simulators.Tests/ResourcesEstimatorTests.cs
irinayat-MS: Maybe, add a comment about that? (at least my brain was confused by seeing _clearly_ two gates acting on the same qubit but the depth still being 1 :D)

src/Simulation/QCTraceSimulator/SortedQubitPool.cs
```diff
@@ -0,0 +1,329 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Text;
+
+namespace Microsoft.Quantum.Simulation.QCTraceSimulatorRuntime {
+
+    /// <summary>
+    /// Either time interval or point in time on the timeline of the program.
```
irinayat-MS, line 9: Don't understand: how this class can represent an interval?

src/Simulation/QCTraceSimulator/SortedQubitPool.cs
```diff
@@ -0,0 +1,329 @@
...
+            TrailingZeroDepthGateCount = trailingZeroDepthGateCount;
+        }
+
+        public override string ToString() {
+            return $"{DepthTime}d+{TrailingZeroDepthGateCount}";
+        }
+        /// <summary>
+        /// Add gate time to availability time. If gate time is zero, advance dependency time.
+        /// </summary>
+        /// <param name="gateTime">Gate time to advance by. Assumed to be precise for comparison with 0.</param>
```
irinayat-MS, line 45: I believe, you mean here that we expect the gate time to be set from literals rather than to be calculated so the `== 0.0` comparison is safe, right? Maybe, add an assert after the `if` that gateTime is greater than some reasonable epsilon value?

src/Simulation/QCTraceSimulator/SortedQubitPool.cs
```diff
@@ -0,0 +1,329 @@
...
+        internal ComplexTime AdvanceBy(double gateTime) {
+            if (gateTime == 0.0) {
+                return new ComplexTime(DepthTime, TrailingZeroDepthGateCount + 1);
+            }
+            return new ComplexTime(DepthTime + gateTime);
+        }
+
+        /// <summary>
+        /// Subtracts argument ComplexTime from "this" ComplexTime. Returns the result.
+        /// Depth times are assumed to be precise for comparison.
```
irinayat-MS, line 56: Is it safe to assume so? If above we are likely dealing with constants here the time is result of double-precision math. Maybe, compare with tolerance or assert that when non-equal, the difference is above the tolerance threshold?

src/Simulation/QCTraceSimulator/SortedQubitPool.cs
```diff
@@ -0,0 +1,329 @@
...
+        /// </summary>
+        /// <param name="time">Time to subtract from this object.</param>
+        /// <returns>Result of subtraction.</returns>
+        internal ComplexTime Subtract(ComplexTime time) {
+            if (DepthTime == time.DepthTime) {
+                return new ComplexTime(0, TrailingZeroDepthGateCount - time.TrailingZeroDepthGateCount);
+            }
+            double result = DepthTime - time.DepthTime;
+            if (result <= 0) {
+                // This could only happen due to insufficient floating point calculation precision.
```
irinayat-MS, line 66: Or it can be a logical bug in the tracer, when it tries to subtract a later time from an earlier time...

src/Simulation/QCTraceSimulator/SortedQubitPool.cs
```diff
@@ -0,0 +1,329 @@
...
+        /// <summary>
+        /// Zero ComplexTime.
+        /// </summary>
+        internal static ComplexTime Zero { get; } = new ComplexTime(0.0, 0);
+    }
+
+    /// <summary>
+    /// Stores qubits sorted by corresponding ComplexTime values. For example, a ComplexTime value that correponds to a qubit
+    /// could be the qubit availability time: the end time of the last gate on the qubit.
+    /// Supports queries to retrieve qubit with maximum time less than given and minimum time greater than given Sample.
```
irinayat-MS, line 119: It might be useful to specify the time complexity (from perf point of view) of the supported operations on the pool.

src/Simulation/QCTraceSimulator/SortedQubitPool.cs
```diff
@@ -0,0 +1,329 @@
...
+                        MinUpperBound = ComplexTime.Min(MinUpperBound, a.Time);
+                    }
+                } else if (result < 0) {
+                    if (a == Sample && b != Sample) {
+                        MinUpperBound = ComplexTime.Min(MinUpperBound, b.Time);
+                    } else if (a != Sample && b == Sample) {
+                        MaxLowerBound = ComplexTime.Max(MaxLowerBound, a.Time);
+                    }
+                } else {
+                    // We found the value if one argument is the sample and nother is not.
```
irinayat-MS, line 218: typo: nother -> other

irinayat-MS: also, why cannot both `a` and `b` be equal to Sample?

src/Simulation/QCTraceSimulator/SortedQubitPool.cs
```diff
@@ -0,0 +1,329 @@
...
+        /// </summary>
+        private readonly SortedSet<QubitTimeNode> QubitsSortedByTime;
+
+        /// <summary>
+        /// Comparer of two nodes, which compares them by Time field and collects additional data.
+        /// </summary>
+        private readonly VisitingComparer NodeComparer;
+
+        /// <summary>
+        /// Sample which is been compared to the nodes stored in the set.
```
irinayat-MS, line 241: typo: is been -> is being

src/Simulation/QCTraceSimulator/SortedQubitPool.cs
```diff
@@ -0,0 +1,329 @@
...
+        /// Multiple comparisons are done by SortedSet to find if this Sample is present in the set.
```
irinayat-MS, line 242: Is this a statement about how SortedSet is implemented in C# CLR? Do you rely on this particular implementation in some way?

src/Simulation/QCTraceSimulator/SortedQubitPool.cs
```diff
@@ -0,0 +1,329 @@
...
+        public bool FindBound(ComplexTime requestedTime, bool getLowerBound, out ComplexTime actualTime) {
+            // We use the following approach:
+            // We call function TryGetValue on a sorted set. If value is found, we go with it.
+            // Otherwise TryGetValue must make conclusion that the value is absent.
+            // Any comparison-based algorithm without caching must inspect both
+            // maximum value less than the sample and minimum value greater than the sample.
+            // So they must be among the values seen since reset.
+            // So we just need to harvest them from VisitingComparer.
+            Sample.Time = requestedTime;
+            NodeComparer.ResetForComparison(Sample);
```
irinayat-MS, line 291: Oh! this is the line that confused me before: the value of Sample isn't used by NodeComparer and the reference to this.Sample isn't changing between invocations of FindBound... Would it make sense for the comparer to be configured with pointer(reference) to Sample during creation and ResetForComparison(Sample) changed to simply Reset()?

src/Simulation/QCTraceSimulator/SortedQubitPool.cs
```diff
@@ -0,0 +1,329 @@
...
+
+        /// <summary>
+        /// Find qubit with the specified time in the set. If such qubit is not present,
+        /// find either maximum value less than it or minimum value greater than it.
+        /// This function uses one call to SortedSet.TryGetValue() so it takes log(N) time, where N - number of nodes.
+        /// </summary>
+        /// <param name="requestedTime">Sample time to find</param>
+        /// <param name="getLowerBound">"true" to find maximum value less than sample</param>
+        /// <param name="actualTime">Time found in the set</param>
+        /// <returns>"true" if sample was found in the set.</returns>
```
irinayat-MS, line 281: probably meant `sample` -> `requestedTime`, but the comment isn't true, because the function would return `true` if it found a valid bound too, no?

src/Simulation/QCTraceSimulator/SortedQubitPool.cs
```diff
@@ -0,0 +1,329 @@
...
+            existingNode.NextNode = newNode;
+        }
+
+        /// <summary>
+        /// Find qubit with the specified time in the set. If such qubit is not present,
+        /// find either maximum value less than it or minimum value greater than it.
+        /// This function uses one call to SortedSet.TryGetValue() so it takes log(N) time, where N - number of nodes.
+        /// </summary>
+        /// <param name="requestedTime">Sample time to find</param>
+        /// <param name="getLowerBound">"true" to find maximum value less than sample</param>
```
irinayat-MS, line 279: As stated, is sounds like an opt-in: "if you want a lower bound in case the exact match cannot be found, pass in `true`", but in reality it seems to mean: "if cannot find the exact match we'll return either lower or upper bound, choose which of the two you prefer"... Maybe, name it `preferLowerBound`?

src/Simulation/QCTraceSimulator/DepthCounter.cs
irinayat-MS: All problems can be solved by another layer of indirection! :D You could add an interface DepthOptimizationPolicy with methods for each place in the code when you currently have `if (OprimizeDepth)`, then provide two concrete implementations... Your call whether it's worth doing something like this.

irinayat-MS (approved):
Don't forget to fix the braces!

src/Simulation/QCTraceSimulator/DepthCounter.cs
DmitryVasilevsky: It's on the same level as having two separate classes. And it doesn't bring clarity. So I'd like to keep it as is.

src/Simulation/QCTraceSimulator/QubitMetrics.cs
DmitryVasilevsky: It is only used in the context when StartTime is adjusted as a result of comparison so it will make this more obscure. With the updated comments I think it's clear what's going on.

src/Simulation/QCTraceSimulator/SortedQubitPool.cs
DmitryVasilevsky: I updated the comment. It's the length of a time interval (not the interval itself). And it's the same way as number 5 can represent duration or a point in time.

src/Simulation/QCTraceSimulator/SortedQubitPool.cs
DmitryVasilevsky: Yes, assumed to be set from literals. I updated the comment. I'm not sure what value would be reasonable... A very reasonable restriction would be to restrict these to integers, but for some reason this was made double in the interface.

src/Simulation/QCTraceSimulator/SortedQubitPool.cs
DmitryVasilevsky: The tracer is bug-free! :) Ha-ha, of course it could happen due to bugs in the tracer but we don't account for this in comments. I'll remove the word 'only'.

src/Simulation/QCTraceSimulator/SortedQubitPool.cs
DmitryVasilevsky: I did specify time complexity on the methods themselves. This whole class is here to have them at O( log(N) ).

src/Simulation/QCTraceSimulator/SortedQubitPool.cs
DmitryVasilevsky: Updated comment.

DmitryVasilevsky: It is possible that both a and b would be equal to sample by reference. This would be a useless comparison. And it is handled correctly. This doesn't give us any information about members of the set so such comparison is ignored (and correct result is returned).

src/Simulation/QCTraceSimulator/SortedQubitPool.cs
DmitryVasilevsky: Updated.

src/Simulation/QCTraceSimulator/SortedQubitPool.cs
DmitryVasilevsky: Yes. Old comment. Updated.

src/Simulation/QCTraceSimulator/SortedQubitPool.cs
DmitryVasilevsky: This is a very good point! I actually considered both approaches and I believe the one I settled on is better. If you do not pass Sample to reset, there's a subsequent call to TryGetValue() to which you still need to pass exactly the same object. It will be easy to forget that. Here, both calls next to each other take the same Sample object, which, in my opinion, makes it easier to understand and maintain.

src/Simulation/QCTraceSimulator/SortedQubitPool.cs
DmitryVasilevsky: This is one of the two concerns I have about this approach. User can set gate times to any floating number. I just assume that this number would be reasonable. If numbers are reasonable, such as integers, or even 1/2, or 7/8, this will work because floating point manipulations are only additions and subtractions. This may break with gate duration set to 1/3. The only real fix for this is to change gate time from floating point to integral part in the interface. I'm not sure checking against threshold is a good approach - there's not reasonable threshold here.

src/Simulation/QCTraceSimulator/SortedQubitPool.cs
DmitryVasilevsky: This is my second concern about this approach. Does this depend on CLR SortedSet implementation? Yes and No. To some extent. For any reasonable implementation should work. Of course one can specifically craft an implementation that would break this. For example, you can cache previously passed values and then compare against them (mostly for fun). I assume implementation is reasonable.

src/Simulation/Simulators.Tests/ResourcesEstimatorTests.cs
DmitryVasilevsky: QubitCount is the old value we used to have and I didn't remove it to keep compatibility. It's also more reasonable per-scope than width. Width is useful mostly for the entire program.

src/Simulation/Simulators.Tests/ResourcesEstimatorTests.cs
DmitryVasilevsky: Adding comment.

src/Simulation/QCTraceSimulator/DepthCounter.cs
DmitryVasilevsky: Updated.

src/Simulation/QCTraceSimulator/SortedQubitPool.cs
DmitryVasilevsky: I updated the comment to indicate what both "true" and "false" means.

PR #456, opened 2020-12-01
Reland "Actually enable QIR Runtime CI (and fix a build break on Linux)"

irinayat-MS:
The cmake's command: -DCMAKE_BUILD_TYPE= $Env:NATIVE_SIMULATOR_BUILD_CONFIGURATION is incorrect on two counts:
1. there should be no space after "="
2. the env variable must be enclosed in quotes

As a result, cmake ignored the build type specification and defaulted to Debug. After fixing this realized that $Env:NATIVE_SIMULATOR_BUILD_CONFIGURATION and can go back to using $Env:BUILD_CONFIGURATION.

Otherwise, this PR is pretty much the same as #445.


anpaz (approved):
lgtm

PR #458, opened 2020-12-11
Implement SWAP using conjugation

msoeken:
A `Controlled SWAP` implementation is a very typical operation in many quantum programs (also referred to as Fredkin gate). Currently, `SWAP` is implemented without conjugation, making a `Controlled SWAP` unnecessarily expensive (3 CCNOTs instead of 1 CCNOT + 2 CNOTs). Using the conjugation statement fixes this problem as `within` blocks are not being controlled.


bettinaheim (approved):

msoeken:
@bettinaheim @rmshaffer I am getting CI errors, which are due to the implementation change in this PR (see, e.g., https://dev.azure.com/ms-quantum-public/Microsoft%20Quantum%20(public)/_build/results?buildId=20889&view=logs&j=055476fc-323f-5564-dc43-0471cd9c85a2&t=82c80951-c925-591c-5223-36e3ed8e7e29). What should I do in this case?

rmshaffer:
> @bettinaheim @rmshaffer I am getting CI errors, which are due to the implementation change in this PR

I believe the recommended approach here would be to make a corresponding change to the unit test in `iqsharp` so that it passes with your changes. If you use the same branch name `msoeken/swap`, the E2E build will test the cross-repo changes together.

PR #459, opened 2020-12-12
QIR update, revamped qir tests, support for multi-level controlled callables

irinayat-MS:
- Replaced the previous QIR tests with more targeted ones, generated from new Q# with compiler from `feature/qir` branch at commit e5ecd6f.
- Implemented support for multi-level controlled callables.

@DmitryVasilevsky : Dmitry, would you mind taking a look at the unpacking code? It deals with untyped memory and I know you are good at this kind of stuff :D


src/QirRuntime/lib/QIR/qirTypes.hpp
```diff
@@ -66,18 +60,109 @@ struct QirArray
...
+    ; convert the offset to int and call __quantum__rt__tuple_create
+    %t2 = ptrtoint { %TupleHeader, %Callable*, %Array* }* %t1 to i64
+    %0 = call %TupleHeader* @__quantum__rt__tuple_create(i62 %t2)
+
+    Notice, that the TupleHeader is included into the Tuple's buffer.
+==============================================================================*/
+struct QirTupleHeader
+{
+    int refCount = 0;
+    int size = sizeof(QirTupleHeader);
```
irinayat-MS, line 96: Add a comment that the size field should be set to sizeof(header) + size_of_data (default initializing to the former might be confusing for the reader)

src/QirRuntime/lib/QIR/callables.cpp
```diff
@@ -191,37 +56,207 @@ extern "C"
...
+
+    long rc = --this->refCount;
+    if (rc == 0)
+    {
+        delete this;
+    }
+    return rc;
+}
+
+// `src` tuple header must point to a tuple with the following structure (example for depth = 2):
```
irinayat-MS, line 156: Clarify, that the inner tuple in the example can be anything, not a specific { %TupleHeader, i64, %Qubit* }

DmitryVasilevsky: Suggest replacing this { %TupleHeader, i64, %Qubit* } with QirTuple.

src/QirRuntime/lib/QIR/callables.cpp
```diff
@@ -191,37 +56,207 @@ extern "C"
...
+// { %TupleHeader, %Array*, { %TupleHeader, %Array*, { %TupleHeader, i64, %Qubit* }* }* }
+// or
+// { %TupleHeader, %Array*, { %TupleHeader, %Array*, %Qubit* }* }
+// The function will create a new tuple, where the array contains elements of all nested arrays, respectively:
+// { %TupleHeader, %Array*, { %TupleHeader, i64, %Qubit* }* }
+// { %TupleHeader, %Array*, %Qubit* }
+// The caller is responsible for releasing both the new array and the tuple.
```
DmitryVasilevsky, line 163: Suggest adding explicit mention that caller shouldn't free last element. But be careful with wording.

src/QirRuntime/lib/QIR/callables.cpp
```diff
@@ -191,37 +56,207 @@ extern "C"
...
+// The order of the elements in the array is undefined.
+QirTupleHeader* UnpackTupleWithNestedArrays(QirTupleHeader* tuple, int depth)
```
DmitryVasilevsky, line 165: Suggest FlattenControlArrays

src/QirRuntime/lib/QIR/callables.cpp
DmitryVasilevsky: Or { %TupleHeader, ... }

DmitryVasilevsky, line 156: Suggest to add wording - we cannot verify the structure so we assume the structure is correct.

src/QirRuntime/lib/QIR/callables.cpp
irinayat-MS: The caller cannot release the last element because they don't know what that is.

src/QirRuntime/lib/QIR/callables.cpp
irinayat-MS: The two cases are explicitly called out in the spec because the structure _is_ different if the inner-most argument isn't wrapped up into a tuple. We can deal with it in a uniform manner, but I think it's worth calling out that both cases do exist and the code must be correct for both of them

irinayat-MS: I've changed the wording to make it more clear that these two are _examples_ and not the fixed expected types.

src/QirRuntime/test/QIR-static/compiler/QirCore.qs
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
bettinaheim, line 1: At some point in the future (when there is no more need for manual edits in the .ll files) we should be able to replace the QirCore.qs and QirTarget.qs files with a project reference to M.Q.QSharp.Core, and generate the .ll files as part of the tests.

bettinaheim (approved):
Looks good to me - thanks for the walk-through!
I haven't looked closely at TupleWithControls, but I know Dmitry has.

bettinaheim:
@irinayat-MS Don't forget to push the change do the assert before merging.

src/QirRuntime/test/QIR-static/compiler/QirCore.qs
irinayat-MS: That would be great :)

PR #464, opened 2020-12-17
Sccarda/regenerate libraries

ScottCarda-MS:


src/Simulation/CsharpGeneration/RewriteStep.fs
```diff
@@ -53,12 +53,22 @@ type Emitter() =
...
-                CompilationLoader.GeneratedFile(source, dir, ".g.cs", content) |> ignore
-            for source in allSources |> Seq.filter (not << context.GenerateCodeForSource) do
-                let content = SimulationCode.loadedViaTestNames source context
-                if content <> null then CompilationLoader.GeneratedFile(source, dir, ".dll.g.cs", content) |> ignore
+            let isExe =
+                match context.assemblyConstants.TryGetValue AssemblyConstants.QsharpOutputType with
+                    | true, outputType -> outputType = AssemblyConstants.QsharpExe
+                    | _ -> false
+
+            if isExe then
```
bettinaheim, line 61: This distinction is handled by GenerateCodeForSource. Please adapt it there rather than adding an additional branching here.

bettinaheim:
@ScottCarda-MS I believe this is a change that will be needed to enable turning on monomorphization by default, correct? Shall we keep this PR, or shall we close it and you'll set up a new one on top of all the latest changes?

ScottCarda-MS:
> @ScottCarda-MS I believe this is a change that will be needed to enable turning on monomorphization by default, correct? Shall we keep this PR, or shall we close it and you'll set up a new one on top of all the latest changes?

@bettinaheim I'd like to keep it around if it is not too much trouble. It serves as a good reminder to me for what needs to be picked back up when I finally get back to this point in the work stack. Otherwise I will lose more context on the task.

PR #465, opened 2020-12-19
Update deprecated operation characteristics syntax

samarsha:
This is needed to unblock microsoft/qsharp-compiler#789, which removes this deprecated syntax.


ScottCarda-MS (approved):
Thanks for cleaning this up. I had thought I had done this before, but it must have been on a branch that didn't get merged in. Glad to see it is resolved with this!

PR #468, opened 2020-12-29
Update QirRuntime to reflect departure of %TupleHeader from QIR

irinayat-MS:
Updated tuples handling and the QIR tests to be compatible with qsharp-compiler 7d182f37. The main change in the QIR is that there is no %TupleHeader anymore, embedded into the concrete tuple types. However, the runtime still uses the headers to store some necessary metadata for each tuple.

### DIFF
--- a/src/QirRuntime/lib/QIR/callables.cpp
+++ b/src/QirRuntime/lib/QIR/callables.cpp
@@ -2,7 +2,7 @@
 // Licensed under the MIT License.
 
 #include <assert.h>
-#include <atomic>
+#include <limits>
 #include <memory>
 #include <stdexcept>
 #include <string.h> // for memcpy
@@ -11,22 +11,18 @@
 #include "qirTypes.hpp"
 #include "quantum__rt.hpp"
 
+/*==============================================================================
+    Implementation of quantum__rt__tuple_* and quantum__rt__callable_*
+==============================================================================*/
 extern "C"
 {
-    char* quantum__rt__tuple_create(int64_t size)
+    PTuple quantum__rt__tuple_create(int64_t size)
     {
-        assert(size >= 0);
-        char* buffer = new char[sizeof(QirTupleHeader) + size];
-
-        // at the beginning of the buffer place QirTupleHeader
-        QirTupleHeader* th = reinterpret_cast<QirTupleHeader*>(buffer);
-        th->refCount = 1;
-        th->tupleSize = size;
-
-        return th->AsTuple();
+        assert(size >= 0 && size < std::numeric_limits<int>::max());
+        return QirTupleHeader::Create(static_cast<int>(size))->AsTuple();
     }
 
-    void quantum__rt__tuple_reference(char* tuple)
+    void quantum__rt__tuple_reference(PTuple tuple)
     {
         if (tuple == nullptr)
         {
@@ -36,7 +32,7 @@ extern "C"
         (void)th->AddRef();
     }
 
-    void quantum__rt__tuple_unreference(char* tuple)
+    void quantum__rt__tuple_unreference(PTuple tuple)
     {
         if (tuple == nullptr)
         {
@@ -53,7 +49,7 @@ extern "C"
         {
             return;
         }
-        callable->AddRef();
+        (void)callable->AddRef();
     }
 
     void quantum__rt__callable_unreference(QirCallable* callable)
@@ -66,13 +62,13 @@ extern "C"
         assert(ref >= 0);
     }
 
-    QirCallable* quantum__rt__callable_create(t_CallableEntry* entries, char* capture)
+    QirCallable* quantum__rt__callable_create(t_CallableEntry* entries, PTuple capture)
     {
         assert(entries != nullptr);
         return new QirCallable(entries, capture);
     }
 
-    void quantum__rt__callable_invoke(QirCallable* callable, char* args, char* result)
+    void quantum__rt__callable_invoke(QirCallable* callable, PTuple args, PTuple result)
     {
         assert(callable != nullptr);
         callable->Invoke(args, result);
@@ -101,6 +97,60 @@ extern "C"
 }
 
 /*==============================================================================
+    Implementation of QirTupleHeader
+==============================================================================*/
+int QirTupleHeader::AddRef()
+{
+    assert(refCount > 0);
+    return ++refCount;
+}
+
+int QirTupleHeader::Release()
+{
+    --refCount;
+    if (refCount == 0)
+    {
+        char* buffer = reinterpret_cast<char*>(this);
+        delete[] buffer;
+    }
+    return refCount;
+}
+
+QirTupleHeader* QirTupleHeader::Create(int size)
+{
+    assert(size >= 0);
+    char* buffer = new char[sizeof(QirTupleHeader) + size];
+
+    // at the beginning of the buffer place QirTupleHeader, leave the buffer uninitialized
+    QirTupleHeader* th = reinterpret_cast<QirTupleHeader*>(buffer);
+    th->refCount = 1;
+    th->tupleSize = size;
+
+    return th;
+}
+
+QirTupleHeader* QirTupleHeader::CreateWithCopiedData(QirTupleHeader* other)
+{
+    const int size = other->tupleSize;
+    char* buffer = new char[sizeof(QirTupleHeader) + size];
+
+    // at the beginning of the buffer place QirTupleHeader
+    QirTupleHeader* th = reinterpret_cast<QirTupleHeader*>(buffer);
+    th->refCount = 1;
+    th->tupleSize = size;
+
+    // copy the contents of the other tuple
+    memcpy(th->AsTuple(), other->AsTuple(), size);
+
+    return th;
+}
+
+QirTupleHeader* QirTupleHeader::GetHeader(PTuple tuple)
+{
+    return reinterpret_cast<QirTupleHeader*>(tuple - sizeof(QirTupleHeader));
+}
+
+/*==============================================================================
     Implementation of QirCallable
 ==============================================================================*/
 QirCallable::~QirCallable()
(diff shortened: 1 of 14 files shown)


src/QirRuntime/lib/QIR/quantum__rt.hpp
```diff
@@ -158,8 +157,8 @@ extern "C"
 
     // Initializes the callable with the provided function table and capture tuple. The capture tuple pointer
     // should be null if there is no capture.
-    typedef void (*t_CallableEntry)(QirTupleHeader*, QirTupleHeader*, QirTupleHeader*);          // NOLINT
-    QIR_SHARED_API QirCallable* quantum__rt__callable_create(t_CallableEntry*, QirTupleHeader*); // NOLINT
+    typedef void (*t_CallableEntry)(char*, char*, char*);                      // NOLINT
```
irinayat-MS, line 160: char* -> PTuple

src/QirRuntime/lib/QIR/quantum__rt.hpp
```diff
@@ -91,13 +90,13 @@ extern "C"
     // ------------------------------------------------------------------------
 
     // Allocates space for a tuple requiring the given number of bytes and sets the reference count to 1.
-    QIR_SHARED_API QirTupleHeader* quantum__rt__tuple_create(int64_t); // NOLINT
+    QIR_SHARED_API char* quantum__rt__tuple_create(int64_t); // NOLINT
```
avasch01, line 93: This is for a generic memory buffer, right? Would it make sense to typedef a "buffer" type for this instead of using char* directly? Just makes it clearer it is not a string. You could even call that type QirTupleHeader, rather than buffer.

src/QirRuntime/lib/QIR/quantum__rt.hpp
```diff
@@ -91,13 +90,13 @@ extern "C"
     // ------------------------------------------------------------------------
 
     // Allocates space for a tuple requiring the given number of bytes and sets the reference count to 1.
-    QIR_SHARED_API QirTupleHeader* quantum__rt__tuple_create(int64_t); // NOLINT
+    QIR_SHARED_API char* quantum__rt__tuple_create(int64_t); // NOLINT
 
     // Indicates that a new reference has been added.
-    QIR_SHARED_API void quantum__rt__tuple_reference(QirTupleHeader*); // NOLINT
+    QIR_SHARED_API void quantum__rt__tuple_reference(char*); // NOLINT
```
avasch01, line 96: I would prefer "add_reference"/"remove_reference" wording to reference/unreference. Just "reference" is unclear, since it is both a verb and a noun. Yes, I realize this is not directly part of this checkin.

src/QirRuntime/lib/QIR/qirTypes.hpp
```diff
@@ -74,57 +74,87 @@ struct QirString
...
-    %t2 = ptrtoint { %TupleHeader, %Callable*, %Array* }* %t1 to i64
-    %0 = call %TupleHeader* @__quantum__rt__tuple_create(i62 %t2)
-
-    Notice, that the TupleHeader is included into the Tuple's buffer.
-==============================================================================*/
+/*======================================================================================================================
+    Tuples are opaque to the runtime and the type of the data contained in them isn't (generally) known, thus, we use
+    char* to represent the tuples QIR operates with. However, we need to manage tuples' lifetime and in case of nested
+    controlled callables we also need to peek into the tuple's content. To do this we associate with each tuple's buffer
+    a header that contains the relevant data. The header immediately precedes the tuple's buffer in memeory when the
```
avasch01, line 81: Typo.

src/QirRuntime/lib/QIR/qirTypes.hpp
```diff
@@ -74,57 +74,87 @@ struct QirString
...
+using PTuple = char*;
 struct QirTupleHeader
 {
     int refCount = 0;
-    int tupleSize = 0; // when creating the tuple, must be set to: size of this header + size of the tuple's data buffer
+    int tupleSize = 0; // when creating the tuple, must be set to the size of the tuple's data buffer
+
+    PTuple AsTuple()
+    {
+        return reinterpret_cast<PTuple>(this) + sizeof(QirTupleHeader);
```
avasch01, line 92: Is this arithmetic really correct?

src/QirRuntime/lib/QIR/quantum__rt.hpp
irinayat-MS: Yes, it would. I actually have PTuple typedef exactly for this, but apparently missed it in some places.

src/QirRuntime/lib/QIR/quantum__rt.hpp
irinayat-MS: The naming is inherited from the QIR spec. Personally, I'd prefer addref/release terminology as I'm used to it from COM era, but here I believe it's more important to be consistent with the naming of the QIR runtime methods.

src/QirRuntime/lib/QIR/qirTypes.hpp
irinayat-MS: Currently QirTupleHeader is aligned but alignment is a concern. I'd love to add more protections here. Would you suggest any specific alignment annotations and static_asserts for this?

src/QirRuntime/lib/QIR/callables.cpp
```diff
@@ -11,47 +11,36 @@
...
-        assert(size >= sizeof(QirTupleHeader));
-        char* buffer = new char[size];
-
-        // at the beginning of the buffer place QirTupleHeader
-        QirTupleHeader* th = reinterpret_cast<QirTupleHeader*>(buffer);
-        th->refCount = 1;
-        th->tupleSize = size;
-
-        return th;
+        assert(size >= 0 && size < std::numeric_limits<int>::max());
```
avasch01, line 21: Does size==0 make sense?

src/QirRuntime/lib/QIR/callables.cpp
```diff
@@ -11,47 +11,36 @@
...
-        assert(ref >= 0);
-
-        if (ref == 0)
-        {
-            char* buffer = reinterpret_cast<char*>(th);
-            delete[] buffer;
-        }
-        th = nullptr;
+        QirTupleHeader* th = QirTupleHeader::GetHeader(tuple);
+        (void)th->Release();
```
avasch01, line 43: You do not want to assert that refcount is >=0?

src/QirRuntime/lib/QIR/callables.cpp
```diff
@@ -107,6 +96,60 @@ extern "C"
...
+==============================================================================*/
+int QirTupleHeader::AddRef()
+{
+    assert(refCount > 0);
+    return ++refCount;
+}
+
+int QirTupleHeader::Release()
+{
+    --refCount;
```
avasch01, line 110: I would assert refcount is >0 here.

src/QirRuntime/lib/QIR/callables.cpp
```diff
@@ -107,6 +96,60 @@ extern "C"
...
+    {
+        char* buffer = reinterpret_cast<char*>(this);
+        delete[] buffer;
+    }
+    return refCount;
+}
+
+QirTupleHeader* QirTupleHeader::Create(int size)
+{
+    assert(size >= 0);
```
avasch01, line 121: What does size==0 mean?

src/QirRuntime/lib/QIR/callables.cpp
```diff
@@ -107,6 +96,60 @@ extern "C"
...
+    char* buffer = new char[sizeof(QirTupleHeader) + size];
```
avasch01, line 122: Ensure buffer!=0

avasch01 (approved):

src/QirRuntime/lib/QIR/callables.cpp
irinayat-MS: If refcount dropped to zero before, the memory would be released and potentially already reclaimed. The assert would be checking a random value. The similar assert in AddRef has same problem but I thought I'd have it to document that objects aren't expected to be resurrected.

src/QirRuntime/lib/QIR/callables.cpp
irinayat-MS: replied above

src/QirRuntime/lib/QIR/qirTypes.hpp
irinayat-MS: memeory -> memory :)

src/QirRuntime/lib/QIR/callables.cpp
irinayat-MS: If new[] fails to allocate it throws std::bad_alloc. There is a non-throwing version of operator new[] that would return a null pointer but I'm not sure it's worth the trouble to handle OOM situations ourselves rather than let the execution fail with the exception. Thoughts?

src/QirRuntime/lib/QIR/callables.cpp
```diff
@@ -107,6 +96,60 @@ extern "C"
...
+    th->refCount = 1;
+    th->tupleSize = size;
+
+    return th;
+}
+
+QirTupleHeader* QirTupleHeader::CreateWithCopiedData(QirTupleHeader* other)
+{
+    const int size = other->tupleSize;
+    char* buffer = new char[sizeof(QirTupleHeader) + size];
```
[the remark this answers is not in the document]
irinayat-MS: see the other comment on new[]

swernli (approved):

src/QirRuntime/lib/QIR/qirTypes.hpp
irinayat-MS: Per Dmitry's suggestion, am using the flexible array member idiom now which should take care of alignment issues.

src/QirRuntime/lib/QIR/callables.cpp
irinayat-MS: see the other comment about zero tuple size

PR #470, opened 2021-01-07
QirRuntime: Refactor to avoid duplication related to compilation of ll files

irinayat-MS:
As we use QIR input for the build more often, it has reached the point when copying the corresponding complex custom actions around doesn't make sense anymore. This change refactors the relevant bits into a CMake macro.

Also renamed one of the test files.


cesarzc (approved):
Looks good to me.

PR #471, opened 2021-01-12
Updating QirRuntime to match Q# compiler @d78fef5c (mostly user/alias/access counts)

irinayat-MS:
Q# compiler build from https://github.com/microsoft/qsharp-compiler/commit/d78fef5c0a9f4bccb72256ad9e346cda32a670cd.
QIR spec update: https://github.com/microsoft/qsharp-language/pull/67


src/QirRuntime/lib/QIR/arrays.cpp
```diff
@@ -14,35 +14,41 @@
 
 #include "CoreTypes.hpp"
 
-long QirArray::AddRef()
+int QirArray::AddRef()
 {
-    long rc = ++this->refCount;
-    assert(rc != 1); // not allowed to resurrect!
-    return rc;
+    assert(this->refCount != 0 && "Cannot resurrect released array!");
```
cesarzc, line 19: Is there a reason for this to be an `assert` rather than a `quantum__rt__fail` like in `RemoveUser`.

src/QirRuntime/lib/QIR/arrays.cpp
```diff
@@ -14,35 +14,41 @@
...
-                quantum__rt__qubit_release(qubits[i]);
-            }
-        }
         delete[] this->buffer;
         this->buffer = nullptr;
     }
     return rc;
 }
 
+void QirArray::AddUser()
```
cesarzc, line 38: Is there a place where it is documented what is the meaning of a user in the context of QIR arrays, and how it is different from references?

src/QirRuntime/lib/QIR/arrays.cpp
```diff
@@ -261,10 +290,18 @@ extern "C"
...
-    QirArray* quantum__rt__array_copy(QirArray* array)
+    QirArray* quantum__rt__array_copy(QirArray* array, bool force)
     {
-        assert(array != nullptr);
-        return new QirArray(array);
+        if (array == nullptr)
+        {
+            return nullptr;
+        }
+        if (force || array->userCount > 0)
```
cesarzc, line 299: It would be good to have a brief comment here explaining why a new instance of `QirArray` is returned only when `force` is `true` or when `userCount` is greater than zero.

src/QirRuntime/lib/QIR/arrays.cpp
```diff
@@ -261,10 +290,18 @@ extern "C"
         return array->dimensionSizes[dim];
     }
 
-    QirArray* quantum__rt__array_copy(QirArray* array)
+    QirArray* quantum__rt__array_copy(QirArray* array, bool force)
```
cesarzc, line 293: In order to make the meaning of `force` more explicit, would it make sense to rename it to `forceNewInstance` or something similar?

src/QirRuntime/lib/QIR/arrays.cpp
irinayat-MS: Yes. The user/access count doesn't affect the object's lifetime. Going below zero would represent a logical error in the generated QIR and we can _always_ catch it. Refcounts are different. If a refcount goes to zero, the object is deleted and the memory is released (and potentially reused). So, unless we are lucky and the ghost object is still in memory, the check would be done on, effectively, a random value -- we cannot base production error handling on that, but the assert is OK as far as documentation of the code goes.

I am considering adding a facility for more deterministic memory debugging (to catch leaks and double-releases reliably), but that will be a separate change.

src/QirRuntime/lib/QIR/arrays.cpp
irinayat-MS: Yes. The QIR specification: https://github.com/microsoft/qsharp-language/tree/main/Specifications/QIR (@bettinaheim has a pending PR there that explains user counts).

src/QirRuntime/lib/QIR/arrays.cpp
irinayat-MS: This particular aspect isn't driven by the QirRuntime implementation but is explicitly required by the QIR spec. I'm duplicating the spec's verbatim in the quantum_rt.hpp file. I'll add a comment in the qirTypes.hpp for the corresponding members/methods, though.

src/QirRuntime/lib/QIR/arrays.cpp
irinayat-MS: Done!

src/QirRuntime/lib/QIR/callables.cpp
```diff
@@ -101,19 +139,34 @@ extern "C"
...
+}
+
+void QirTupleHeader::AddUser()
+{
+    ++this->userCount;
+}
+
+void QirTupleHeader::RemoveUser()
+{
+    if (this->userCount == 0)
```
swernli, line 165: This may already be covered in the spec somewhere, but is the assumption that this QirRuntime is not natively thread safe? Since there is no locking or other enforced atomicity here, this check doesn't guarantee that a tuple user count won't be negative if multiple threads access the same object (same for arrays above).

swernli: Ah, I see now where we assert that the existing runtime is assumed to be single threaded. Makes sense!

swernli (approved):
Thanks for walking through these changes with us!

cesarzc (approved):

PR #474, opened 2021-01-19
'Fixing' tuple add/remove user

swernli:
This change corresponds to the ongoing work in the compiler repo as captured in https://github.com/microsoft/qsharp-compiler/tree/beheim/vqeBuild (see for example the function naming used in https://github.com/microsoft/qsharp-compiler/blob/beheim/vqeBuild/src/QsCompiler/QirGeneration/QIR/RuntimeLibrary.cs#L60).

### DIFF
--- a/src/QirRuntime/lib/QIR/bridge-rt.ll
+++ b/src/QirRuntime/lib/QIR/bridge-rt.ll
@@ -79,8 +79,8 @@ declare void @llvm.va_end(i8*)
 declare i8* @quantum__rt__tuple_create(i64)
 declare void @quantum__rt__tuple_reference(i8*)
 declare void @quantum__rt__tuple_unreference(i8*)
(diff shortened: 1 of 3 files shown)


irinayat-MS:
(for the lack of better versioning mechanism) Please add to the change description the commit SHA of the corresponding compiler's change this matches to. Also, does the compiler generate these calls now? I'm not seeing any in the current QIR tests in the runtime, which means the new functions aren't being tested right now.

irinayat-MS (approved):

swernli:
Good call, I'll add some cross links to make clear what this corresponds to. The unfortunate mismatch here is that the compiler changes are not yet merged to main but are being tested from a working branch. In that working branch, the name was changed to make it consistent with the array functions, but the outstanding PR to the spec has not been updated yet. So the unmerged compiler changes disagree with the unmerged spec changes, and runtime is trying to keep up :).

PR #475, opened 2021-01-20
QIR-based tracer

irinayat-MS:
To collect early feedback on the tracer design.

`lib/tracer` contains implementation of the tracer (the readme file in the folder outlines the requirements and design)
`test/unittests/TracerTests.cpp` contains C++ unit tests for the tracer
`test/QIR-tracer` demonstrates how a Q# program can be connected to the tracer via QIR


src/QirRuntime/lib/Tracer/tracer.cpp
```diff
@@ -0,0 +1,267 @@
...
+    //------------------------------------------------------------------------------------------------------------------
+    void CTracer::AddOperationToLayer(OpId id, LayerId layer)
+    {
+        assert(layer < this->metricsByLayer.size());
+        auto inserted = this->metricsByLayer[layer].operations.insert({id, 1});
+        if (!inserted.second)
+        {
+            assert(inserted.first->first == id);
+            inserted.first->second += 1;
+        }
```
thomashaener, line 143: Can't this be done as `this->metricsByLayer[layer].operations[id] += 1;`?

src/QirRuntime/lib/Tracer/tracer.cpp
```diff
@@ -0,0 +1,267 @@
...
+    }
+    Result CTracer::UseOne()
+    {
+        return reinterpret_cast<Result>(INVALID);
+    }
+
+    //------------------------------------------------------------------------------------------------------------------
+    // CTracer::CreateNewLayer
+    //------------------------------------------------------------------------------------------------------------------
+    LayerId CTracer::CreateNewLayer(Duration opDuration)
```
thomashaener, line 70: Maybe rename `opDuration` to `minLayerDuration`?

src/QirRuntime/lib/Tracer/tracer.hpp
```diff
@@ -0,0 +1,185 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include <limits>
+#include <memory>
+#include <unordered_map>
+#include <vector>
+
+#include "CoreTypes.hpp"
+#include "QuantumApi_I.hpp"
```
thomashaener, line 10: Reorder includes?

src/QirRuntime/lib/Tracer/tracer.cpp
```diff
@@ -0,0 +1,267 @@
...
+    void CTracer::ReleaseQubit(Qubit /*qubit*/)
+    {
+        // nothing for now
+    }
+    std::string CTracer::QubitToString(Qubit q)
+    {
+        size_t qubitIndex = reinterpret_cast<size_t>(q);
+        const QubitState& qstate = this->UseQubit(q);
+
+        stringstream str(qubitIndex);
```
thomashaener, line 48: This is causing the build to fail on linux.
```suggestion
        stringstream str(std::to_string(qubitIndex));
```

thomashaener (commented):
Thanks a lot, @irinayat-MS, looks good to me!
I've added a few suggestions and comments.

src/QirRuntime/lib/Tracer/tracer.hpp
irinayat-MS: Personally, I like grouping std includes and the project includes separately, and inside each group they are sorted alphabetically by clang-format, which improves readability and is usually sufficient to catch the header dependencies. My preference to list std includes first comes from the desire to keep "meaningful" code together. In my previous projects, std includes were seldom the source of missing header dependencies and they are easy to fix, if found. All in all, I'd like to keep the current format/order of includes as is, and consistent across QirRuntime project.

msoeken (commented):
I left only a few comments for now.

src/QirRuntime/lib/Tracer/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
...
+set(includes
+  "${public_includes}"
+  "${PROJECT_SOURCE_DIR}/lib/QIR"
+)
+
+add_library(${component_name} STATIC ${source_files})
+
+target_include_directories(${component_name} PUBLIC ${includes})
+
+MESSAGE(INFO "*** ${QIR_UTILITY_LIB}")
```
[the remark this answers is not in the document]
irinayat-MS: removed the line (it was an artefact from debugging the build)

src/QirRuntime/lib/Tracer/tracer-trc.cpp
```diff
@@ -0,0 +1,66 @@
+// Copyright (c) // NOLINT{} Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include <assert.h>
+
+#include "CoreTypes.hpp"
+#include "qirTypes.hpp"
+#include "tracer.hpp"
```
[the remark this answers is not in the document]
irinayat-MS: See my response about include grouping/order in another comment. If we decide to change it, the change would need to be applied across QirRuntime project.

src/QirRuntime/lib/Tracer/tracer-trc.cpp
```diff
@@ -0,0 +1,66 @@
...
+    RESULT* quantum__trc__single_qubit_measure(int32_t id, int32_t duration, QUBIT* q) // NOLINT
+    {
+        return tracer->TraceSingleQubitMeasurement(id, duration, q);
+    }
+
+    RESULT* quantum__trc__multi_qubit_measure(int32_t id, int32_t duration, QirArray* qs) // NOLINT
+    {
+        return tracer->TraceMultiQubitMeasurement(id, duration, qs->count, reinterpret_cast<Qubit*>(qs->buffer));
+    }
+}
```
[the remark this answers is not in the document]
irinayat-MS: We are inconsistent in QirRuntime regarding the black line at the end of the cpp/hpp files. Why is it important to enforce one? If it is, we should add the rule to clang-format if possible.

src/QirRuntime/lib/Tracer/tracer.cpp
```diff
@@ -0,0 +1,267 @@
...
+        return nullptr;
+    }
+    IDiagnostics* CTracer::AsDiagnostics()
+    {
+        return nullptr;
+    }
+    Qubit CTracer::AllocateQubit()
+    {
+        size_t qubit = qubits.size();
+        qubits.push_back({});
```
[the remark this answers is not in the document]
irinayat-MS: Yes, it should :)

src/QirRuntime/lib/Tracer/tracer.cpp
irinayat-MS: Thank you! (Why, oh why, cannot Windows and Linux do the same thing here?)

src/QirRuntime/lib/Tracer/tracer.cpp
irinayat-MS: Renamed to `minRequiredDuration`.

src/QirRuntime/lib/Tracer/tracer-trc.cpp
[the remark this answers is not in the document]
msoeken: Some older tools such as UNIX command line tools can lead to strange effects without the `\n` at the end of the last line of code.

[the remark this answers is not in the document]
msoeken: I don't think there's a `clang-format` option for that. It may be done automatically though (see [here](https://reviews.llvm.org/D37732))

src/QirRuntime/lib/Tracer/tracer.cpp
```diff
@@ -0,0 +1,267 @@
...
+    // CTracer::FindLayerToInsertOperationInto
+    //------------------------------------------------------------------------------------------------------------------
+    LayerId CTracer::FindLayerToInsertOperationInto(Qubit q, Duration opDuration) const
+    {
+        const QubitState& qstate = this->UseQubit(q);
+
+        LayerId layerToInsertInto = INVALID;
+        if (qstate.layer != INVALID)
+        {
+            const Layer& lastUsedIn = this->metricsByLayer[qstate.layer];
```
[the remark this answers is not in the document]
irinayat-MS: Thanks! As I was adding comments in the function, found a bug. Hopefully, the naming of the locals and the overall logic is clearer now.

src/QirRuntime/lib/Tracer/tracer.cpp
irinayat-MS: It can, because integer values in the map are default-initialized to zero :)

src/QirRuntime/lib/Tracer/tracer-trc.cpp
```diff
@@ -0,0 +1,66 @@
...
+namespace Quantum
+{
+    extern thread_local std::shared_ptr<CTracer> tracer;
+}
+} // namespace Microsoft
+
+using namespace Microsoft::Quantum;
+extern "C"
+{
+    void quantum__trc__on_operation_start(int64_t id) // NOLINT
```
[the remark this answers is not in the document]
irinayat-MS: Why would it be useful to do this? (if so, we should do it across QirRuntime as a whole)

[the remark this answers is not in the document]
irinayat-MS: Added to the list in https://github.com/microsoft/qsharp-runtime/issues/482

src/QirRuntime/lib/Tracer/tracer-trc.cpp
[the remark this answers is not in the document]
irinayat-MS: VSCode and VS do insert EOL marker at the end of each line, including the last one, so I don't think there is a need to require a blank last line. We should get consistent line feeds though (added that to the list in https://github.com/microsoft/qsharp-runtime/issues/482).

src/QirRuntime/lib/Tracer/tracer.hpp
irinayat-MS: If you feel strongly about it, please add to https://github.com/microsoft/qsharp-runtime/issues/482 (we'll need to update our clang.format and make sure we use consistent grouping/ordering of includes across the project).

src/QirRuntime/lib/Tracer/README.md
```diff
@@ -0,0 +1,177 @@
...
+6. (_lower priority_) the scheduling component to optimize depth and/or width of the circuit.
+
+## List of `__quantum__qis__*` methods ##
+
+___WIP___
+
+| Signature                                             | Description                                                  |
+| :---------------------------------------------------- | :----------------------------------------------------------- |
+| `void __quantum__qis__inject_global_barrier(i32 %id, i32 %duration)` | Function to insert a global barrier. It will be inserted into QIR based on a user defined intrinsic. See [Layering](#layering) section for details. |
+| `void __quantum__qis__on_operation_start(i64 %id)`    | Function to identify the start of a quantum module. The argument is a unique _id_ of the module. The tracer will have an option to treat module boundaries as barriers between layers and (_lower priority_) and option to cache estimates for a module, executed multiple times. The call to the function will be inserted into QIR by the Q# compiler when Tracer is specified as the compilation target. |
```
kuzminrobin, line 37: Typo: 
> **and** (_lower priority_) **and** option 

2 **and**s in a row.

src/QirRuntime/lib/Tracer/README.md
```diff
@@ -0,0 +1,177 @@
...
+| `void __quantum__qis__on_operation_end(i64 %id)`      | Function to identify the end of a quantum module. The argument is a unique _id_ of the module and must match the _id_ supplied on start of the module. The call to the function will be inserted into QIR by the Q# compiler when Tracer is specified as the compilation target. |
```
kuzminrobin, line 38: > void __quantum__qis__on_operation_end(**i64 %id**)

If the modules strictly follow each other or are strictly nested to each other, then the parameter **i64 %id** may be redundant (and can cause questions "_what to do if the parameter does not correspond to the expected one?_").

src/QirRuntime/lib/Tracer/README.md
```diff
@@ -0,0 +1,177 @@
...
+| `void __quantum__qis__single_qubit_op(i32 %id, i32 %duration, %Qubit* %q)` | Function for counting operations that involve a single qubit. The first argument is the id of the operation, as assigned by the client. Multiple intrinsics can be assigned the same id, in which case they will be counted together. The second argument is duration to be assigned to the particular invocation of the operation. |
```
kuzminrobin, line 39: > The second argument is **duration** to be assigned to the particular invocation of the operation

(May be this my question is premature and will be clarified later, but during the first-most reading the answer is not obvious)
The duration is in which units? Seconds? Milliseconds? CPU clock cycles?

And what does it mean "_to assign a duration to the particular invocation of the operation_"?

src/QirRuntime/lib/Tracer/README.md
```diff
@@ -0,0 +1,177 @@
+# Resource Tracer Design Document #
+
+The purpose of the Resource Tracer is to provide efficient and flexible way to estimate resources of a quantum program
+ in QIR representation. The estimates are calculated by simulating execution of the program (as opposed to the static
+ analysis).
```
kuzminrobin, line 5: The first-most reading would be easier if more details are added to the first paragraph. In particular:
> The purpose ... is ... to **estimate resources** of a ... program ...

I (as a first-time reader) would be thankful if you could provide more details about why we need this estimation (the description of the currently existing problem that implies that we need the estimation).
You could also provide the examples of the resources we plan to estimate (e.g. how much memory we use, how many qubits we allocate, how large stack is required, how many times certain functions are called). And for each item a few words about which problem this particular estimation solves. Such an info would prepare me for the expectations from the tracer (in the text below) and be ready to understand the implementation details of the tracer.

src/QirRuntime/lib/Tracer/README.md
```diff
@@ -0,0 +1,177 @@
+# Resource Tracer Design Document #
+
+The purpose of the Resource Tracer is to provide efficient and flexible way to estimate resources of a quantum program
+ in QIR representation. The estimates are calculated by simulating execution of the program (as opposed to the static
+ analysis).
+
+In addition to the standard QIR runtime functions, the quantum program will have to:
```
kuzminrobin, line 7: > **In addition to the** ... **runtime functions**, the **quantum program** will have to...

a) The beginning of the sentence sounds incomplete. In addition to **doing what** with runtime functions? In addition to ... **calling? converting?** ... the runtime functions.
b)

src/QirRuntime/lib/Tracer/README.md
```diff
@@ -0,0 +1,177 @@
+# Resource Tracer Design Document #
+
+The purpose of the Resource Tracer is to provide efficient and flexible way to estimate resources of a quantum program
+ in QIR representation. The estimates are calculated by simulating execution of the program (as opposed to the static
+ analysis).
+
+In addition to the standard QIR runtime functions, the quantum program will have to:
+
+1. convert _all_ used intrinsic operations into one of the _qis_ operations supported by the tracer (see the list below);
```
kuzminrobin, line 9: > convert **_all_** ... into **one** ...

I assume you meant "convert **_each_** ... into **a corresponding** ..."

src/QirRuntime/lib/Tracer/README.md
```diff
@@ -0,0 +1,177 @@
...
+1. (_optional_) provide description of mapping for frame tracking;
+1. (_optional_) provide names of operations for user friendly output (in the form of `tracer-config.hpp|cpp` files that,
+ ideally, can be generated by the Q# compiler from the mappings described in target.qs).
+
+The last provisions  
+
+The Resource Tracer will consist of:
+
+1. the bridge for the `__quantum__qis__*` extension methods;
+2. the native implementation to back the extensions;
```
kuzminrobin, line 24: (During the first-most reading it is not evident)
What are _extension methods_ and _extensions_? The callbacks mentioned earlier?

src/QirRuntime/lib/Tracer/README.md
```diff
@@ -0,0 +1,177 @@
...
+3. the logic for partitioning gates into layers;
+4. the logic for frame tracking;
```
kuzminrobin, line 26: What is _frame_? Is it the _stack frame_?

src/QirRuntime/lib/Tracer/README.md
```diff
@@ -0,0 +1,177 @@
...
+5. output of the collected statistics;
+6. (_lower priority_) the scheduling component to optimize depth and/or width of the circuit.
+
+## List of `__quantum__qis__*` methods ##
+
+___WIP___
```
kuzminrobin, line 32: What is **WIP**?

src/QirRuntime/lib/Tracer/README.md
```diff
@@ -0,0 +1,177 @@
...
+The Resource Tracer will consist of:
+
+1. the bridge for the `__quantum__qis__*` extension methods;
+2. the native implementation to back the extensions;
+3. the logic for partitioning gates into layers;
+4. the logic for frame tracking;
+5. output of the collected statistics;
+6. (_lower priority_) the scheduling component to optimize depth and/or width of the circuit.
+
+## List of `__quantum__qis__*` methods ##
```
kuzminrobin, line 30: > List of ... methods

Is it a list of callbacks provided by the tracer? (and the corresponding calls are inserted by the compiler to the quantum program?)

src/QirRuntime/lib/Tracer/README.md
```diff
@@ -0,0 +1,177 @@
...
+
+___WIP___
+
+| Signature                                             | Description                                                  |
+| :---------------------------------------------------- | :----------------------------------------------------------- |
+| `void __quantum__qis__inject_global_barrier(i32 %id, i32 %duration)` | Function to insert a global barrier. It will be inserted into QIR based on a user defined intrinsic. See [Layering](#layering) section for details. |
```
kuzminrobin, line 36: > Function to insert a global barrier.

(I'm a newbie and my questions may be not applicable to your readers)
I would be very pleased if somewhere in the beginning there was some sort of a glossary telling a few words about what the _barrier_ is, how it is different from the _global barrier_, and what it means _to insert a barrier_.
> **It** will be inserted ...

The _function call_ will be inserted or _the global barrier_ will be inserted?
(for my own technical writing I try to avoid pronouns (it, he, me, they, there) and use actual nouns or expressions. At times it results in a repeated words, but technically is clear during the first-most reading, and helps to avoid questions and uncertainty)

src/QirRuntime/lib/Tracer/README.md
kuzminrobin, line 37: Why the function name \_\_quantum__qis__on_**operation**_start() talks about the _operation_, but the description talks about some _module_ (rather than _operation_)? ("... _to identify the start of a quantum **module**_ ...")

kuzminrobin, line 37: > barriers between **layers** and

I would like to see _layers_ clarified in the glossary in the beginning

src/QirRuntime/lib/Tracer/README.md
kuzminrobin, line 38: > The call to the function will be inserted into QIR by the Q# compiler when Tracer is specified as the compilation target.

This repeats at least one time. As I understand this is applicable to all the functions in this table. If yes then it makes sense to mention that one time in the beginning, before the table (and to remove from the table).

src/QirRuntime/lib/Tracer/README.md
kuzminrobin, line 39: > as assigned by the **client**

Who/what is _client_?

src/QirRuntime/lib/Tracer/README.md
```diff
@@ -0,0 +1,177 @@
...
+_Note on operation ids_: The client is responsible for using operation ids in a consistent manner. Operations with the
+ same id will be counted by the tracer as the _same_ operation, even accross invocations with different number of target
+ qubits or when different functors are applied.
+
+## Native backing of the extension methods ##
+
+The Resource Tracer will reuse qir-rt library as much as possible while extending it with the callbacks specified above.
+
+__Conditionals on measurements__: The Resource Tracer will execute LLVM IR's branching structures "as is", depending on
+ the values of the corresponding variables at runtime. To enable estimation of branches that depend on a measurement
```
kuzminrobin, line 56: > To enable estimation of branches that depend on a measurement result, the source Q# program will have to be authored in such a way that the Q# compiler will translate the conditionals into __quantum__qis__apply_if* calls.

Would be nice to see an example in the form of a code fragment (with detailed comments or/and explanation).

src/QirRuntime/lib/Tracer/README.md
```diff
@@ -0,0 +1,177 @@
...
+ result, the source Q# program will have to be authored in such a way that the Q# compiler will translate the
+ conditionals into `__quantum__qis__apply_if*` calls. The tracer will add operations from _both branches_ into the
```
kuzminrobin, line 58: > The tracer will add operations from both branches into the layers it creates to compute the upper bound estimate.

I don't understand this sentence until I know what the _layer_ is.

src/QirRuntime/lib/Tracer/README.md
```diff
@@ -0,0 +1,177 @@
...
+
+Nested conditionals, conditional measurements and conditional tracked operations will _not_ be supported.
+
+__Caching__ (lower priority): It might be a huge perf win if the Resource Tracer could cache statistics for repeated
+ computations. The Tracer will have an option to cache layering results per quantum module if the boundaries of modules
+ are treated as layering barriers.
+
+## Layering ##
+
+_Definition_: ___Time___ is an integer-valued function on all quantum operations in a program (gates, measurements,
```
kuzminrobin, line 69: > is an **integer-valued function**

I assume that the _integer-valued function_ is [this](https://en.wikipedia.org/wiki/Integer-valued_function). If yes, then you may want to make it a link. You may also want to specify in _which units_ the values of the function are.

src/QirRuntime/lib/Tracer/README.md
```diff
@@ -0,0 +1,177 @@
...
+
+A sequentially executed quantum program can be assigned a trivial time function, when all quantum operations have
+ duration of 1 and unique start times, ordered to match the flow of the program. Layering compresses the timeline by
+ assuming that some operations might be executed simultaneously while allowing for different operations to have various
+ durations.
+
+_Definition_: Provided a valid _time_ function for the program a ___layer of duration N at time T, denoted as L(T,N),___
+ is a subset of operations in the program such that all of these operations have start time greater or equal _T_ and
+ finish time less than _T + N_. The program is ___layered___ if all gates in it are partitioned into layers, that don't
+ overlap in time. The union of all qubits that are involved in operations of a given layer, will be denoted _Qubits(T,N)_.
```
kuzminrobin, line 83: :-) I feel that a mathematicians' language is used here that often takes me some time to decipher (and the same concept typically can be explained simpler).
If I interpret the definition right, then 
_Layer_ is a set of operations that are executed (from start to completion) within certain period of time (within certain time frame/slot/window). That period of time starts at moment T (inclusive) and ends at moment T + N (exclusive).
If the interpretation is right, then for the term (_layer_, _layering_) to be more informative (for the newbies like me during the first reading) I would recommend to consider a term containing the word _time_, e.g. _time partition_ (_time group_), _time partitioning_ (_time grouping_). For instance, 
_Time partition_ is a set of operations that are executed (from start to completion) within certain period of time...
The program is _time-partitioned_ if all gates in it are grouped into non-overlapping time periods.

src/QirRuntime/lib/Tracer/README.md
```diff
@@ -0,0 +1,177 @@
...
+
+A sequential program can be trivially layered such that each layer contains exactly one operation. Notice, that the
+ definition of layer doesn't require the gates to be executed _in parallel_. For example, all gates in a fully sequential
+ program can be also placed into a single layer L(0, infinity). Some gates might be considered to be very cheap and take
+ zero time to execute, those gates can be added to a layer even if they act on the same qubit another gate in this layer
+ is acting on and have to be executed sequentially within the layer.
+
+_Definition_: A ___global barrier___ is any operation that acts on _all_ currently allocated qubits. We'll provide means
+ for the clients to inject global barriers equivalent to an identity operator. This will allow the clients to enforce a
+ particular layering structure (because no later operation can sink below the barrier).
```
kuzminrobin, line 93: > can sink below the barrier

What does that mean _to sink below the barrier_? (to sink below an operation that acts on all currently allocated qubits)
Based on the context of the preceding paragraphs I imagine a horizontal time line (like OX axis) where the time stamps increase when moving left to right. The fragment _to sink_ makes an impression of _vertical_ movement _across_ the horizontal time line.
> This will allow the clients to enforce a particular layering structure

By this moment would be nice to understand why one may need to enforce a particular layering structure.
> to inject global barriers equivalent to an identity operator

I assume that with _identity operator_ you mean [this](https://en.wikipedia.org/wiki/Identity_function). If that's the case then you may want to make it a link.
Does the expression "_global barriers equivalent to an identity operator_" mean "operations that acts on all currently allocated qubits such that the qubits stay unchanged"? If yes then you may want to add a clarification, e.g.
> We'll provide means for the clients to inject global barriers equivalent to an identity operator (**i.e. to insert operations that acts on all currently allocated qubits, such that the qubits stay unchanged**).

> A **global barrier** is any operation that acts on _all_ currently allocated qubits.

Why is such an operation called _global barrier_? Is it a widely known term? Why isn't it called _all-qubit operation_? (the expression _all-qubit operation_ seems to me clear during the first-most reading even without a definition)

src/QirRuntime/lib/Tracer/README.md
```diff
@@ -0,0 +1,177 @@
...
+
+The global barriers will be implemented as calls to `__quantum__qis__global_barrier` function. No additional support from
+ the compiler should be needed, as the user can define their own intrinsic to represent the barrier and map it to the above
```
kuzminrobin, line 96: > to represent the **barrier** 

I assume that with _barrier_ you mean the _global barrier_. If that's the case then you may want to consider using the expression _global barrier_ always (and not using _barrier_, since _global barrier_ is a term with a strict meaning). Otherwise the redundant questions (and uncertainty) arise: "What's _barrier_ (as opposed to _global barrier_, and what's the difference)?".

src/QirRuntime/lib/Tracer/README.md
```diff
@@ -0,0 +1,177 @@
...
+ definition of layer doesn't require the gates to be executed _in parallel_. For example, all gates in a fully sequential
+ program can be also placed into a single layer L(0, infinity). Some gates might be considered to be very cheap and take
+ zero time to execute, those gates can be added to a layer even if they act on the same qubit another gate in this layer
+ is acting on and have to be executed sequentially within the layer.
+
+_Definition_: A ___global barrier___ is any operation that acts on _all_ currently allocated qubits. We'll provide means
+ for the clients to inject global barriers equivalent to an identity operator. This will allow the clients to enforce a
+ particular layering structure (because no later operation can sink below the barrier).
+
+The global barriers will be implemented as calls to `__quantum__qis__global_barrier` function. No additional support from
```
kuzminrobin, line 95: > The global barriers will be implemented as calls to __quantum__qis__global_barrier function.

Does the signature of this function matter? If yes then you might want to add the signature.
Intuitively I feel that you meant to say:
> The global barrier will be implemented **inside of the** __quantum__qis__global_barrier **callback**. **And the _injection_ of a global barrier will be done by inserting the call (to this callback) into ...**

src/QirRuntime/lib/Tracer/README.md
```diff
@@ -0,0 +1,177 @@
...
+ the compiler should be needed, as the user can define their own intrinsic to represent the barrier and map it to the above
+ runtime function via targets.qs file. The user can choose duration of a barrier which would affect start time of the
```
kuzminrobin, line 97: > as the user can define their own intrinsic to represent the barrier and map it to the above runtime function via targets.qs file.

I feel a strong desire of an illustrative example.

kuzminrobin, line 97: > The user can choose duration of a barrier 

Would be nice to understand why one might need to choose duration of a barrier.

src/QirRuntime/lib/Tracer/README.md
```diff
@@ -0,0 +1,177 @@
...
+ following layers but no operations will be added to a barrier, independent of its width.
+
+### The Resource Tracer's Layering Algorithm ###
+
+As the tracer is executing a sequential quantum program, it will compute a time function and corresponding layering
+ using the _conceptual_ algorithm, described below (aka "tetris algorithm"). The actual implementation of layering might
+ be done differently, as long as the resulting layering is the same as if running the conceptual algorithm.
+
+1. The tracer will have a setting for preferred layer duration: P.
```
kuzminrobin, line 106: By this moment would be nice to understand why the tracer may need a setting for preferred layer duration.

src/QirRuntime/lib/Tracer/README.md
```diff
@@ -0,0 +1,177 @@
...
+1. The first encountered operation of __non-zero__ duration N is added into layer L(0, max(P,N)). The value
+ for _conditional barrier_ is set to 0.
```
kuzminrobin, line 108: > The first encountered operation of __non-zero__ duration N is added into layer L(0, max(P,N)). 

Intuitively I feel that you meant to say
> The moment of time, when the first encountered _operation_ starts, is assigned to be 0 (whereas the _functions_ and other possible _actions_/_operators_/_statements_ (if such exist in Q#) that precede (go before) the first encountered _operation_, are ignored). The first encountered _operation_ has duration N. Thus the first encountered _operation_ executes within the time period from 0 to N. A layer L(0, max(P,N)) is created and the first encountered _operation_ is added to this layer.

Next. 
> The value for _conditional barrier_ is set to 0.

I feel that a term (_conditional barrier_) is introduced without a definition.

src/QirRuntime/lib/Tracer/README.md
```diff
@@ -0,0 +1,177 @@
...
+1. When conditional callback is encountered, the layer L(t,N) of the measurement that produced the result the conditional
```
kuzminrobin, line 109: > When conditional callback is encountered

:-\ I don't know what the _conditional callback_ is. Any clarification/example/link?

src/QirRuntime/lib/Tracer/README.md
irinayat-MS: Reworded the lead sentence. The "quantum program" here means the package, consisting of QIR and C++ code. It's not Q# specific and it's up to the clients how they create such a program. I'm open to improving the terminology but I'd like to avoid creating the impression that the tracer depends on Q# compiler (unless we make it so).

src/QirRuntime/lib/Tracer/README.md
irinayat-MS: You are right, "extension" isn't adding any clarity - eliminated it.

src/QirRuntime/lib/Tracer/README.md
irinayat-MS: No, this is (as far as I understand) tracking of state change of a qubit in a quantum system, that sometimes is benefitial not to apply immediately but store as a classical bit on the qubit. I'm not aware of any article that would explain it clearly.

src/QirRuntime/lib/Tracer/README.md
irinayat-MS: I'm not sure I like the term "callback" here. The QIR provides a bunch of methods to do stuff (e.g. to allocate an array -- which hardly anybody would view as a "callback"). Similarly, the tracer provides a bunch of methods to collect statistics about the running program.

irinayat-MS: I've updated the header to "List of `__quantum__qis__*` methods, provided by the Tracer"

src/QirRuntime/lib/Tracer/README.md
irinayat-MS: WIP = Work In Progress
I've added to the table a todo about handling conditionals (because this is the only qis piece still missing) and removed the acronym.

src/QirRuntime/lib/Tracer/README.md
irinayat-MS: I moved this list to be the last section of the readme.

src/QirRuntime/lib/Tracer/README.md
```diff
@@ -0,0 +1,177 @@
...
+ is dependent on, is looked up and the _conditional barrier_ is set to _t + N_. At the end of the conditional scope the
```
kuzminrobin, line 110: > When conditional callback is encountered, the layer L(t,N) of the measurement that produced the result the conditional is dependent on, is looked up and the _conditional barrier_ is set to _t + N_. At the end of the conditional scope ...

Here is how I interpret it:
> The quantum program can contain conditionals. Before executing a conditional, the _conditional callback_ (provided by the Tracer, described earlier in <where?>) is invoked. The _conditional callback_ knows that the conditional depends on the result produced during the measurement done earlier in layer L(t,N). The _conditional callback_ finds that layer L(t,N) and places the _conditional barrier_ at a time stamp t + N (in the end of that layer). Then the _conditional callback_ returns the control back to the quantum program, the conditional is executed, and the block of code inside of the conditional is affected by the _conditional barrier_ that has just been set (i.e. the set of layers, available for the block of code inside of the conditional, is limited by the latest placed _conditional barrier_). At the end of the conditional scope the _conditional barrier_ is placed back to the time stamp of the outer conditional (or to 0 if there is no outer conditional), i.e. the result of the previous invocation of _conditional callback_ is restored (in case of nested conditionals).

src/QirRuntime/lib/Tracer/README.md
```diff
@@ -0,0 +1,177 @@
...
+ barrier is reset to 0. (Effectively, no operations, conditioned on the result of a measurement, can happen before or in
+ the same layer as the measurement, even if they don't involve the measured qubits.)
+1. Suppose, there are already layers L(0,N0), ... , L(k,Nk) and the operation being executed is a single-qubit _op_ of
+ duration __0__ (controlled and multi-qubit operations of duration 0 are treated the same as non-zero operations).
+ Starting at L(k, Nk) and scanning backwards to L(conditional barrier, Nb) find the _first_ layer that contains an
+ operation that acts on at least one of the qubits _op_ is acting on. Add _op_ into this layer. If no such layer found,
+ remember that the qubits have pending 0-duration _op_ and commit it to the layer that first uses any of the qubits in a
+ an operation of non-zero duration. At the end of the program commit all pending operations of duration zero into a new
+ layer.
```
kuzminrobin, line 119: > If no such layer found, **remember** that the qubits

(English is a second language for me, I can be wrong)
The word [_remember_](https://www.lingvolive.com/en-us/translate/en-ru/remember) has at least 2 meanings that are somewhat unrelated or opposite/antonymous:
a) _keep in memory_ (retain, neither write to memory, nor read from memory);
b) _recall_ (read from memory).
I intuitively feel that you meant a different thing - _write to memory_ - _memorize_.

kuzminrobin, line 119: > that the **qubits have** pending 0-duration _op_ and **commit it** to the layer that first uses **any of the qubits** in **a** an operation of non-zero duration.

I can interpret this in two ways:
* that **this qubit has** a pending 0-duration operation and **add the operation** to the **later** layer that first uses **this qubit** in an operation of non-zero duration **(until then all the zero-duration operations on this qubit are memorized)**.
* that there is a pending (0-duration single-qubit) operation and **add the operation** to the **later** layer that first uses a non-zero-duration operation **on any qubit** **(until then \<I need to know more to word this part>)**.

src/QirRuntime/lib/Tracer/README.md
```diff
@@ -0,0 +1,177 @@
...
+1. Suppose, there are already layers L(0,N0), ... , L(k,Nk) and the operation being executed is _op_ of duration _N > 0_.
+ Starting at L(k, Nk) and scanning backwards to L(conditional barrier, Nb) find the _last_ layer L(t, Nt) such that
+ Qubits(t, Nt) don't contain any of the _op_'s qubits and find the _first_ layer L(w, Nw) such that Qubits(w, Nw) contains
+ some of _op_'s qubits but Nw + N <= P. Add _op_ into one of the two layer with later time. If neither such layers found,
+ add _op_ into a new layer L(k+1, max(P, N)).
```
kuzminrobin, line 124: > layer L(t, Nt) such that Qubits(t, Nt) don't contain any of the _op_'s qubits

I simplify this for myself like this: "layer that doesn't touch any qubits the operation touches"
> layer L(w, Nw) such that Qubits(w, Nw) contains some of op's qubits

"layer that touches any qubit the operation touches"
> Nw + N <= P

The operation will fit into the layer L(w, Nw), if the layer can be stretched to get a layer L(w, Nw+N), and such a stretched layer's duration does not exceed the preferred layer duration P (i.e. Nw + N <= P).

src/QirRuntime/lib/Tracer/README.md
```diff
@@ -0,0 +1,177 @@
...
+
+## Special handling of SWAP ##
```
kuzminrobin, line 126: If SWAP is not a special term or abbreviation the I would recommend "swap" (lower case).

src/QirRuntime/lib/Tracer/README.md
```diff
@@ -0,0 +1,177 @@
...
+The tracer will provide a way to handle SWAP as, effectively, renaming of the involved qubits. The client will have the
+ choice of using the special handling versus treating the gate as a standard counted intrinsic.
+
+## Frame tracking ##
+
+The client might want to count differently operations that are applied in a different state. For example, if Hadamard gate
+ is applied to a qubit and then Rz and Mx gates, the client might want to count the sequence as if Rz as Mz were executed.
+ The frame is closed when the state of the qubit is reset (in Hadamard's case, another Hadamard operator is applied to
+ the qubit). The client will be able to register the required frame tracking with the tracer via a C++ registration
+ callback.
```
kuzminrobin, line 137: > For example, if Hadamard gate is applied to a qubit and then Rz and Mx gates, the client might want to count the sequence as if Rz as Mz were executed.

I don't quite get the point of this sentence. Do you mean counting Rz and Mx _only after Hadamard gate_?

src/QirRuntime/lib/Tracer/README.md
```diff
@@ -0,0 +1,177 @@
...
+
+The descriptor of the frame will contain the following information and will be provided to the Tracer when initializing
+ it in C++.
+
+- openingOp: the operation id that opens the frame on the qubits this operation is applied to
+- closingOp: the operation is that closes the the frame on the qubits this operation is applied to
```
kuzminrobin, line 143: Typo? 
> the operation ~~is~~ **id** that closes ~~the~~ the frame

src/QirRuntime/lib/Tracer/tracer.hpp
```diff
@@ -0,0 +1,214 @@
...
+        QubitState
+    ==================================================================================================================*/
+    struct QubitState
+    {
+        // The last layer this qubit was used in, `INVALID` means the qubit haven't been used yet in any
+        // operations of non-zero duration.
+        LayerId layer = INVALID;
+
+        // For layers with duration greater than one, multiple operations might fit on the same qubit, if the operations
+        // are short. `lastUsedTime` is the end time of the last operation, the qubit participated it.
```
kuzminrobin, line 59: multiple operations **on the same qubit** might fit **in the layer**?
the qubit participated ~~it~~ **in** ?

src/QirRuntime/lib/Tracer/tracer.hpp
```diff
@@ -0,0 +1,214 @@
...
+        // -------------------------------------------------------------------------------------------------------------
+        // Temporary method for initial testing
+        // TODO: replace with a safer accessor
+        const std::vector<Layer>& UseLayers()
+        {
+            return this->metricsByLayer;
+        }
+
+        void PrintLayerMetrics(std::ostream& out, const std::string& separator, bool printZeroMetrics) const;
+    };
```
kuzminrobin, line 206: (At least in the future, feel free to ignore in this PR) 
* I would recommend to start separating the declaration and definition of the member functions, see example below. 
The code base is growing, and later it will be not very convenient to quickly get an idea of the class if the definition and declaration of the member functions is in the same place. 
* Also having modern (multi-pass) compilers it is typical to declare `public`s first, then `protected`s, then `private`s (even if this results in _use_ (of private member vars declared later, in a public member func defined earlier) _before declaration_).
* I would also recommend to keep the class definition _tight_, such that the whole class definition fits in one screen page (if it doesn't fit then the author may want to consider breaking the class up into smaller classes).
* It is typical to declare the (private) member vars after the private member funcs.

Example.
.hpp:
```c++
class CTracer : public ISimulator
{
  public:
    explicit CTracer(int preferredLayerDuration);
    CTracer(int preferredLayerDuration, const std::unordered_map<OpId, std::string>& opNames);

    // ISimulator:
    Qubit AllocateQubit() override;
    void ReleaseQubit(Qubit qubit) override;
    std::string QubitToString(Qubit qubit) override;
    void ReleaseResult(Result result) override;

    IQuantumGateSet* AsQuantumGateSet() override;
    IDiagnostics* AsDiagnostics() override;
    Result M(Qubit target) override;
    Result Measure(long numBases, PauliId bases[], long numTargets, Qubit targets[]) override;
    bool AreEqualResults(Result r1, Result r2) override;
    ResultValue GetResultValue(Result result) override;
    Result UseZero() override;
    Result UseOne() override;
    
    LayerId TraceSingleQubitOp(OpId id, Duration duration, Qubit target);   // -------------------------------------------------------------------------------------------------------------
    LayerId TraceMultiQubitOp(                                              // Instead of implementing IQuantumGateSet, the tracer provides 'tracing-by-id' methods. The QIR generation     
        OpId id,                                                            // should translate all intrinsics to invoke these methods.                                                     
        Duration duration,                                                  // The tracer doesn't differentiate between control and target qubits. However, While it could provide a single 
        long nFirstGroup,                                                   // generic tracing method for and array of qubits, that would require the clients to copy control and target    
        Qubit* firstGroup,                                                  // qubits into the same array. To avoid the copy, the tracer provides a method that takes two groups of qubits, 
        long nSecondGroup,                                                  // where the first one can be empty or can be viewed as the set of controls.                                    
        Qubit* secondGroup);                                                // -------------------------------------------------------------------------------------------------------------

    Result TraceSingleQubitMeasurement(OpId id, Duration duration, Qubit target);
    Result TraceMultiQubitMeasurement(OpId id, Duration duration, long nTargets, Qubit* targets);
    LayerId GetLayerIdOfSourceMeasurement(Result r) const;
    LayerId InjectGlobalBarrier(OpId id, Duration duration);    // Backing of the rest of the bridge methods.
    const std::vector<Layer>& UseLayers();  // Configuring the tracer and getting data back from it.
                                            // Temporary method for initial testing
                                            // TODO: replace with a safer accessor
    void PrintLayerMetrics(std::ostream& out, const std::string& separator, bool printZeroMetrics) const;

private:
    QubitState& UseQubit(Qubit q);
    const QubitState& UseQubit(Qubit q) const;
    LayerId FindLayerToInsertOperationInto(Qubit q, Duration opDuration) const;     // If no appropriate layer found, return `INVALID`
    LayerId CreateNewLayer(Duration minRequiredDuration);   // Returns the index of the created layer.
    void AddOperationToLayer(OpId id, LayerId layer);       // Adds operation with given id into the given layer. Assumes that duration contraints have been satisfied.
    void UpdateQubitState(Qubit q, LayerId layer, Duration opDuration); // Update the qubit state with the new layer information

private:
    std::vector<QubitState> qubits;     // For now the tracer assumes no reuse of qubits.
    const int preferredLayerDuration = 0;   // The preferred duration of a layer. An operation with longer duration will make the containing layer longer.
    std::vector<Layer> metricsByLayer;      // The index into the vector is treated as implicit id of the layer.
    LayerId globalBarrier = INVALID;        // The last global barrier, injected by the user. No new operations can be added to the barrier or to any of the
                                            // layer that preceeded it, even if the new operations involve completely new qubits.                           
    std::unordered_map<OpId, std::string> opNames;  // Mapping of operation ids to user-chosen names, for operations that user didn't name, the output will use
                                            // operation ids.
    std::unordered_set<OpId> seenOps;       // Operations we've seen so far (to be able to trim output to include only those that were encounted).
};
```

src/QirRuntime/lib/Tracer/tracer.hpp
```diff
@@ -0,0 +1,214 @@
...
+#include <unordered_set>
+#include <vector>
+
+#include "CoreTypes.hpp"
+#include "QuantumApi_I.hpp"
+
+namespace Microsoft
+{
+namespace Quantum
+{
```
kuzminrobin, line 17: Saves some space:
```c++
namespace Microsoft::Quantum
{
```

src/QirRuntime/test/QIR-tracer/generate.py
```diff
@@ -0,0 +1,39 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+import sys, os, platform, subprocess, datetime, shutil
+
+# =============================================================================
+#  Generates QIR files for all *.qs files in this folder
+#  Accepts arguments:
+#    path to qsc.exe (absolute or rely on Path env)
```
kuzminrobin, line 9: Is not applicble?
> or rely on Path env

src/QirRuntime/test/QIR-tracer/tracer-config.hpp
```diff
@@ -0,0 +1,30 @@
...
+#include <unordered_map>
+
+namespace Microsoft
+{
+namespace Quantum
+{
+    using OpId = int;
+    using Time = int;
+    using Duration = int;
+    using LayerId = size_t;
```
kuzminrobin, line 18: I saw this earlier. You may want to extract into a header and include that header where needed.
Otherwise in one place this can be changed, in the other place forgotten to update - discrepancy (and bugs hard to catch).

src/QirRuntime/test/QIR-tracer/tracer-config.hpp
```diff
@@ -0,0 +1,30 @@
...
+}
+}
+
+namespace TracerUser
+{
+using OpId = int;
```
kuzminrobin, line 24: If this is the same as `Microsoft::Quantum::OpId` then I would recommend to reuse that. It is better to avoid giving different names/definitions to the same entity.

src/QirRuntime/lib/Tracer/tracer-qis.cpp
```diff
@@ -0,0 +1,66 @@
+// Copyright (c) // NOLINT{} Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include <assert.h>
```
kuzminrobin, line 4: `#include <cassert>` ?
The C++ wrappers (like `<cassert>`) put some of the C's identifiers into the `std::` namespace.
If the C's headers (like `<assert.h>`) are included directly to C++, then this results in the identifiers outside of `std::` namespace, which results in a non-portable C++ code.
[Example](https://en.cppreference.com/mwiki/index.php?title=Special%3ASearch&search=atan2).

src/QirRuntime/lib/Tracer/tracer.cpp
```diff
@@ -0,0 +1,305 @@
...
+
+#include "tracer.hpp"
+
+using namespace std;
+
+namespace Microsoft
+{
+namespace Quantum
+{
+    thread_local std::shared_ptr<CTracer> tracer = nullptr;
```
kuzminrobin, line 16: I would even add `static` (if that works) such that nobody else can directly access `tracer` outside of this .cpp (even those inside of `Microsoft::Quantum` namespace).
The `static` will also help to avoid the identifier conflict if someone outside of this .cpp (but still inside of `Microsoft::Quantum` namespace) adds the identifier `tracer` for a different purpose.

src/QirRuntime/lib/Tracer/tracer.cpp
```diff
@@ -0,0 +1,305 @@
...
+    std::shared_ptr<CTracer> CreateTracer(int preferredLayerDuration)
+    {
+        tracer = std::make_shared<CTracer>(preferredLayerDuration);
+        return tracer;
+    }
+    std::shared_ptr<CTracer> CreateTracer(int preferredLayerDuration, const std::unordered_map<OpId, std::string>& opNames)
+    {
+        tracer = std::make_shared<CTracer>(preferredLayerDuration, opNames);
+        return tracer;
+    }
```
kuzminrobin, line 26: Optional:
These 2 factory funcs (and the 2 constructors of CTracer) can be merged together. If you like, then I can tell you how (but the code will become a bit more tricky).

src/QirRuntime/lib/Tracer/tracer.cpp
```diff
@@ -0,0 +1,305 @@
...
+
+    //------------------------------------------------------------------------------------------------------------------
+    // CTracer's ISimulator implementation
+    //------------------------------------------------------------------------------------------------------------------
+    Qubit CTracer::AllocateQubit()
+    {
+        size_t qubit = qubits.size();
+        qubits.emplace_back(QubitState{});
+        return reinterpret_cast<Qubit>(qubit);
+    }
```
kuzminrobin, line 36: May be in the future we'll consider using an integer type of fixed size (like `int32_t`) as a `Qubit` (and its `.ll` counterpart). 
E.g. in `CoreTypes.hpp`:
`using Qubit = int32_t;`
It will be assumed that `Qubit` type is a 32-bit integer descriptor/handle of the actual allocated qubit (and documented like that).

Such a change will let us avoid all the casts between the pointer type and integer type (including in `.ll` bridge). Everywhere it will be passed natively as `Qubit` or `i32`.

Why of _fixed size_? Because some day we will need to pass this `Qubit` descriptor between the machines/architectures/clusters/datacenters.

src/QirRuntime/lib/Tracer/README.md
irinayat-MS: Added link to https://docs.microsoft.com/en-us/azure/quantum/user-guide/machines/resources-estimator to provide more context/background.

src/QirRuntime/lib/Tracer/tracer.cpp
```diff
@@ -0,0 +1,305 @@
...
+    {
+        // nothing for now
+    }
+    std::string CTracer::QubitToString(Qubit q)
+    {
+        size_t qubitIndex = reinterpret_cast<size_t>(q);
+        const QubitState& qstate = this->UseQubit(q);
+
+        stringstream str(std::to_string(qubitIndex));
+        str << " last used in layer " << qstate.layer << "(pending zero ops: " << qstate.pendingZeroOps.size() << ")";
```
kuzminrobin, line 47: This feels like a functionality alien for `QubitToString()`. I would expect from `QubitToString()` just to convert the qubit descriptor to string, and nothing else.
I would expect the other info to be added by some decorator outside of `QubitToString()`.

src/QirRuntime/lib/Tracer/README.md
irinayat-MS: The name of the function was "inherited" from the current managed API but in this readme I'm using "operation" when talking about intrinsics so ended up with the confusion. I'll rename the function to on_module_start/end

irinayat-MS: Moved the section, listing the qis methods to the end of the design doc.

src/QirRuntime/lib/Tracer/README.md
irinayat-MS: Yes, the modules must be nested, but it should be trivial for the compiler to inject the id into both methods and it will give us an additional invariant to check. I think it's OK to `fail` if the ids don't match.

I don't have strong feelings about this one, though. Could also remove the id from the module_end.

irinayat-MS: reworded

src/QirRuntime/lib/Tracer/README.md
irinayat-MS: Does the section on layering clarify the concept of time and durations?

src/QirRuntime/test/QIR-tracer/tracer-config.cpp
```diff
@@ -0,0 +1,29 @@
...
+namespace Quantum
+{
+
+    std::unique_ptr<ISimulator> CreateFullstateSimulator()
+    {
+        throw std::logic_error("Tracer should not instantiate full state simulator");
+    }
+
+} // namespace Quantum
+} // namespace Microsoft
```
kuzminrobin, line 29: What will happen if this block is removed?
(we are creating an opportunity to make a mistake (we provide a function) and then we punish for making a mistake (for calling that function))

src/QirRuntime/test/QIR-tracer/qir-tracer-driver.cpp
```diff
@@ -0,0 +1,48 @@
...
+    vector<Layer> layers = tr->UseLayers();
+
+    // TestCoreIntrinsics happens to produce 24 layers right now and we are not checking whether that's expected -- as 
+    // testing of layering logic is better done by unit tests.
+    CHECK(layers.size() == 24);
+
+    std::ofstream out;
+    out.open("qir-tracer-test.txt");
+    tr->PrintLayerMetrics(out, "\t", false /*printZeroMetrics*/);
+    out.close();
```
kuzminrobin, line 35: Any comments about what this output is needed for?

src/QirRuntime/test/QIR-tracer/tracer-target.qs
```diff
@@ -0,0 +1,297 @@
...
+            Phys.single_qubit_op_ctl(10, 1, ctls, qb);
+        }
+        controlled adjoint (ctls, ...)
+        {
+            Phys.single_qubit_op_ctl(10, 1, ctls, qb);
+        }
+    }
+
+    @Inline()
+    operation Tz(qb : Qubit) : Unit
```
cgranade, line 119: This is defined in the Microsoft.Quantum.Intrinsic namespace, but doesn't match any existing intrinsic operations that are already a part of the Q# standard library; is this operation exposed anywhere publicly?

src/QirRuntime/test/QIR-tracer/tracer-target.qs
```diff
@@ -0,0 +1,297 @@
...
+            elif (paulis[0] == PauliZ and paulis[1] == PauliZ) { set res = Phys.multi_qubit_measure(111, 1, qubits); }
+
+            //shouldn't get here
+            return res;
+        }
+    }
+
+
+    @TargetInstruction("inject_global_barrier")
+    operation Barrier(id : Int, duration : Int) : Unit
```
cgranade, line 285: Similarly here, this looks like a proposal for a new Q# intrinsic operation? If so, would be willing to write an issue on QuantumLibraries so that we can get it scheduled for the next API review? Thanks!

src/QirRuntime/lib/Tracer/README.md
irinayat-MS: Replaced "client" with "user" throughout the readme

src/QirRuntime/lib/Tracer/README.md
irinayat-MS: The support for conditionals is TBD. I'll add more details into the section when it's more clear what we are actually doing.

src/QirRuntime/lib/Tracer/README.md
irinayat-MS: Resolved now?

src/QirRuntime/lib/Tracer/tracer.hpp
```diff
@@ -0,0 +1,214 @@
...
+        // Start time of the layer.
+        const Time startTime;
+
+        // Quantum operations, assigned to this layer.
+        std::unordered_map<OpId, int /*count of the op with this id*/> operations;
+
+        // Optional id, if the layer represents a global barrier.
+        OpId barrierId = -1;
+
+        Layer(Duration duration, Time startTime)
```
kuzminrobin, line 42: Everywhere in the ReadMe.md you refer to the layer ("L(T, N)") such that the time goes first, and then the duration.
Here in the constructor - vice versa. 
The user of this code can easily make a mistake. I recommend the same order everywhere.

src/QirRuntime/lib/Tracer/tracer.cpp
```diff
@@ -0,0 +1,305 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include <cassert>
+#include <set>
+#include <sstream>
+
+#include "tracer.hpp"
+
+using namespace std;
```
kuzminrobin, line 10: You write `using namespace std` but then you still write `std::` explicitly. Why?

src/QirRuntime/lib/Tracer/README.md
irinayat-MS: Yes, it is a function whose values are integers. No units. I don't think the link would be helpful here, but I'll look into adding a couple diagrams to illustrate the concepts.

kuzminrobin (commented):
Still reviewing...

src/QirRuntime/lib/Tracer/README.md
irinayat-MS: Personally, I strongly favor single-word terminology ("layer") over multi-word ("time partitioning"). I understand that the definitions are dry: my goal was to describe the concepts as accurately as possible. Simplifying the language might feel good at first, but then you run into ambiguities, etc. To aid with comprehension I'll add diagrams.

src/QirRuntime/lib/Tracer/README.md
irinayat-MS: The full signatures are defined in the table that describes all functions, supported by the tracer. The reader can search the document to see it. I'll be reworking the definition.

src/QirRuntime/lib/Tracer/README.md
irinayat-MS: When I wrote this I thought I could define the barrier in terms of quantum operations, but in reality it's not. For example, it won't let through the qubits, that are allocated _after_ the barrier was injected, so the current definition doesn't hold.

Researches want to exercise some control over layering and barriers is one of the means to do so. The question "why they want to do this" is out of scope of this design document.

src/QirRuntime/lib/Tracer/README.md
irinayat-MS: This is exactly why I don't like multi-word terminology! I've inherited the term "global barrier" from some previous prototypes of the tracer, but now thinking about it, I'm inclined to go with just "barrier".

src/QirRuntime/lib/Tracer/README.md
irinayat-MS: Request from researches. My understanding is that the barrier might act as a placeholder for external work that is estimated separately and later combined with the results of the tracer.

src/QirRuntime/lib/Tracer/README.md
irinayat-MS: How else would you calculate the layers?

src/QirRuntime/lib/Tracer/README.md
irinayat-MS: tested on a native speaker, they also found it ambiguous :)
reworded

src/QirRuntime/lib/Tracer/README.md
irinayat-MS: https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.intrinsic.swap?view=qsharp-preview

src/QirRuntime/lib/Tracer/README.md
kuzminrobin: Up to you...

src/QirRuntime/lib/Tracer/tracer.cpp
```diff
@@ -0,0 +1,305 @@
...
+        Qubit* firstGroup,
+        long nSecondGroup,
+        Qubit* secondGroup)
+    {
+        assert(nFirstGroup >= 0);
+        assert(nSecondGroup > 0);
+
+        this->seenOps.insert(id);
+
+        // Operations that involve a single qubit can special case duration zero.
```
kuzminrobin, line 200: Something is skipped in the comment...
> // Operations that involve a single qubit **can special case** duration zero.

src/QirRuntime/lib/Tracer/tracer.cpp
```diff
@@ -0,0 +1,305 @@
...
+            return "";
+        }
+
+        auto nameIt = opNames.find(opId);
+        return nameIt == opNames.end() ? std::to_string(opId) : nameIt->second;
+    }
+    void CTracer::PrintLayerMetrics(std::ostream& out, const std::string& separator, bool printZeroMetrics) const
+    {
+        // Sort the operations by id so the output is deterministic.
+        std::set<OpId> seenOpsOrederedById(this->seenOps.begin(), this->seenOps.end());
```
kuzminrobin, line 278: Minor optimization:
I see that `this->seenOps` is only appended to, then iterated through in the sorted order of unique elements. I'm afraid the `unordered_set` can be expensive. Consider 
```c++
std::vector<OpId> seenOps;   // Instead of unordered_set
this->seenOps.push_back(id);   // Instead of this->seenOps.insert(id);

// Instead of `std::set<OpId> seenOpsOrederedById()`:
std::sort(this->seenOps.begin(), this->seenOps.end());
.. = std::unique(this->seenOps.begin(), this->seenOps.end());
```

kuzminrobin, line 278: Typo?
> seenOpsOr**~~e~~**deredById

kuzminrobin (approved):
Can you please show the updated `README.md`?

src/QirRuntime/lib/Tracer/tracer.cpp
irinayat-MS: I don't have a rule of thumb. Mostly, if it's long type that stands out, like chrono::tah-dah-dah::tah-dah-dah, I omit 'std::' but if the typename is short or there is plenty of space on the line, I retain 'std::'... Just a personal readability adaptation, I guess. Don't know if it's part of our coding style or if it matters enough to be codifying in the style.

src/QirRuntime/lib/Tracer/tracer.cpp
irinayat-MS: The `tracer` is also accessed from tracer-qis.cpp

src/QirRuntime/lib/Tracer/tracer.cpp
irinayat-MS: The factories currently are only relevant when running QIR, but not if someone wants to consume/test the tracer directly from C++ (thus the constructors aren't private). I don't know whether running directly from C++ is a scenario we'd want to support in the future, though. For example, we could create the tracer on the stack for unit tests, but I'm not doing it... Don't have a strong opinion on this one, maybe, revisit later?

src/QirRuntime/lib/Tracer/tracer.cpp
irinayat-MS: That's feedback on the QIR spec. The qubit handle should probably be i64 to allow implementers to stuff a pointer into it, if they'd wish

src/QirRuntime/lib/Tracer/tracer.cpp
irinayat-MS: What do you mean by "qubit descriptor"? The qubit's id? `QubitToString` is a public API on ISimulator -- the _only_ way for the client to get information about the qubit. I suppose, we could do effectively nothing here and provide a more expressive printout functionality as public API on CTracer (e.g. different verbosity levels, output to json, etc).

src/QirRuntime/lib/Tracer/tracer.cpp
irinayat-MS: No, it's me taking liberty with English and using "special case" as a verb. Moved the comments around to make it clear (hopefully) _why_ we bother to redirect to single qubit case.

src/QirRuntime/test/QIR-tracer/tracer-config.cpp
irinayat-MS: It would fail to link.

This is an artefact of how QIR support is wired in if the clients consume the runtime using qdk.dll ad the qdk creates the default (fullstate) simulator for them. This is a sign of bad dependencies between the components and we should look into it, but not as part of this PR. Logged as https://github.com/microsoft/qsharp-runtime/issues/519.

src/QirRuntime/test/QIR-tracer/qir-tracer-driver.cpp
irinayat-MS: I used it for a manual validation and forgot to remove... Actually, converted to an INFO on output to string

src/QirRuntime/test/QIR-tracer/qir-tracer-driver.cpp
```diff
@@ -0,0 +1,48 @@
...
+#include "tracer-config.hpp"
+#include "tracer.hpp"
+
+using namespace std;
+using namespace Microsoft::Quantum;
+
+namespace TracerUser
+{
+
+TEST_CASE("Invoke each operator from Q# core once", "[qir-tracer]")
```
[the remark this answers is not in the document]
irinayat-MS: "intrinsic"

src/QirRuntime/test/QIR-tracer/tracer-target.qs
```diff
@@ -0,0 +1,297 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.Quantum.Instructions {
+
+    open Microsoft.Quantum.Targeting;
+
+    @TargetInstruction("single_qubit_op")
+    operation single_qubit_op(op_id: Int, duration: Int, qb : Qubit) : Unit {
```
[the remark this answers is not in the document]
irinayat-MS: I'm following here the naming convention of QIR, e.g. `__quantum__rt__string_concatenate` (https://github.com/microsoft/qsharp-language/blob/main/Specifications/QIR/Data-Types.md#strings), etc.

src/QirRuntime/test/QIR-tracer/tracer-target.qs
```diff
@@ -0,0 +1,297 @@
...
+            else { Phys.single_qubit_op_ctl(2, 1, ctls, qb); }
+        }
+        controlled adjoint (ctls, ...)
+        {
+            if (Length(ctls) == 1) { Phys.single_qubit_op_ctl(1, 1, ctls, qb); }
+            else { Phys.single_qubit_op_ctl(2, 1, ctls, qb); }
+        }
+    }
+
+    operation CNOT(control : Qubit, target : Qubit) : Unit
```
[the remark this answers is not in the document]
irinayat-MS: copy-pasted from some other place that it was defined like this... fixed

src/QirRuntime/lib/Tracer/tracer.cpp
kuzminrobin: Sorry, with "qubit descriptor" I meant the instance of the `Qubit` type - the parameter `q`.

src/QirRuntime/lib/Tracer/README.md
```diff
@@ -0,0 +1,186 @@
...
+
+As the tracer is executing a sequential quantum program, it will compute a time function and corresponding layering
+ using the _conceptual_ algorithm, described below (aka "tetris algorithm"). The actual implementation of layering might
+ be done differently, as long as the resulting layering is the same as if running the conceptual algorithm.
+
+_Definition_: A ___barrier___ is a layer that no more operations can be added into.
+
+A user will be able to inject global barriers by calling `__quantum__qis__global_barrier` function. The user can choose
+ duration of a barrier which would affect start time of the following layers but no operations will be added to a barrier,
+ independent of its width.
```
kuzminrobin, line 65: > The user can choose duration of a barrier which would affect start time of the following layers but no operations will be added to a barrier, independent of its **width**

I assume that with **width** you express the same as with **duration** mentioned earlier in this sentence. If that's the case then I would recommend to stick to the existing terms (duration), and avoid adding new ones (width) for the same concept.

src/QirRuntime/test/QIR-tracer/generate.py
```diff
@@ -0,0 +1,39 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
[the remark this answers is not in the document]
irinayat-MS: This is not automation but a local tool to create test input that is then checked in. I cannot switch to using Q# project for the tracer yet, because it relies on custom targeting that we still need to figure out how to integrate with the compiler.

src/QirRuntime/test/QIR-tracer/generate.py
irinayat-MS: why not? if you add the folder where qsc.exe is located to %PATH%, it will be found, no?

src/QirRuntime/test/QIR-tracer/generate.py
```diff
@@ -0,0 +1,39 @@
...
+
+# find all qs files in this folder
+files_to_process = ""
+output_file = "tracer-qir"
+for file in os.listdir(root_dir):
+  (file_name, ext) = os.path.splitext(file)
+  if ext == ".qs":
+    files_to_process = files_to_process + " " + file
+
+# Compile as a lib so all functions are retained and don't have to workaround the current limitations of
```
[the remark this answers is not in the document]
irinayat-MS: Yes, there is no pruning of unreachable code when compiling as a lib and not as an exe.

src/QirRuntime/lib/Tracer/README.md
```diff
@@ -0,0 +1,186 @@
...
+ duration __0__ (controlled and multi-qubit operations of duration 0 are treated the same as non-zero operations).
+ Starting at L(k, Nk) and scanning backwards to L(_conditional barrier_, Nb) find the _first_ layer that contains an
+ operation that acts on the qubit of _op_. Add _op_ into this layer. If no such layer is found, add _op_ to the list of
+ pending operations on the qubit. At the end of the program still pending operations are ignored.
+1. Suppose, there are already layers L(0,N0), ... , L(k,Nk) and the operation being executed is _op_ of duration _N > 0_
+ or it involves more than one qubit. Starting at L(k, Nk) and scanning backwards to L(_conditional barrier_, Nb) find the
+ _last_ layer L(t, Nt) such that Qubits(t, Nt) don't contain any of the _op_'s qubits and find the _first_ layer L(w, Nw)
+ such that Qubits(w, Nw) contains some of _op_'s qubits but Nw + N <= P. Add _op_ into one of the two layer with later
+ time. If neither such layers is found, add _op_ into a new layer L(k+1, max(P, N)). Add the pending operations of all
+ involved qubits into the same layer and clear the pending lists.
```
kuzminrobin, line 87: > add _op_ into a new layer L(k+**1**, max(P, N)).

Did you mean "add _op_ into a new layer L(k+**Nk**, max(P, N))"?

src/QirRuntime/lib/Tracer/README.md
```diff
@@ -0,0 +1,186 @@
...
+
+_Definition_: A ___barrier___ is a layer that no more operations can be added into.
+
+A user will be able to inject global barriers by calling `__quantum__qis__global_barrier` function. The user can choose
+ duration of a barrier which would affect start time of the following layers but no operations will be added to a barrier,
+ independent of its width.
+
+#### The conceptual algorithm ####
+
+1. The tracer must be set the preferred layer duration: P.
```
kuzminrobin, line 69: > The tracer **must** be set the preferred layer duration: P

The word "**must**" tells me that there cannot be _default_ preferred layer duration.
Did you means something like "**can be**" (instead of "must"), which would tell me that if it is not set then the default will be used.

src/QirRuntime/test/QIR-tracer/tracer-target.qs
irinayat-MS: These are needed for Hadamard tracking so moved into `Microsoft.Quantum.Instructions` namespace

src/QirRuntime/lib/Tracer/tracer.hpp
irinayat-MS: After we move to C++17?

src/QirRuntime/lib/Tracer/tracer.hpp
irinayat-MS: Totally agree. Not sure why my brain flipped to the reverse order when implementing layering...

src/QirRuntime/lib/Tracer/tracer.hpp
irinayat-MS: Reworded.

src/QirRuntime/lib/Tracer/tracer.hpp
irinayat-MS: My personal preference is to have all data members (private and public) to be declared at the top of the class. I don't know what's typical here because I worked on large codebases that had different takes as far as the order of declarations is concerned.

The readme.md in the root of QirRuntime attempts to capture the style guidelines and, while I did collect feedback from a few other people, it's biased towards my preferences. It's probably a good time now to review the guidelines, codify as many of them as possible into clang format and ensure consistency across the project.

src/QirRuntime/test/QIR-tracer/tracer-config.hpp
irinayat-MS: Yes. Moved the typedefs into a header in `public` folder.

src/QirRuntime/test/QIR-tracer/tracer-config.hpp
irinayat-MS: A piece of construction debris left behind... cleaned up :)

src/QirRuntime/lib/Tracer/tracer.cpp
irinayat-MS: The Qubit type is meaningless, it's just a handle to QubitState. I actually don't know what a user might expect to be printed here... I think we'd need to see end-to-end usage scenarios before deciding on this.

src/QirRuntime/lib/Tracer/README.md
irinayat-MS: => "Note: The tracer assumes that the preferred layer duration is _P_." I need to introduce notation for the default layer duration so I can use it in the following steps of the algorithm.

src/QirRuntime/lib/Tracer/README.md
irinayat-MS: Moved the corresponding section up.

src/QirRuntime/lib/Tracer/README.md
irinayat-MS: within the frame, Rx is counted as Rz. added "instead" at the end of the sentence

src/QirRuntime/lib/Tracer/README.md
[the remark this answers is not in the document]
irinayat-MS: Moved it into Tracer namespace for now. We can decide later how it should be promoted into the public interface.

src/QirRuntime/test/QIR-tracer/tracer-target.qs
irinayat-MS: @bettinaheim, I suspect that eventually we'll need barriers to be in the public interface but I'm not sure of the timeline. For now, I've moved it into the Tracer namespace to allow for experimentation and testing but should we kick off the motions to add it to the libraries as well?

cgranade: I think that would make sense, yeah. No harm in getting the Q# API reviewed in advance so that the approval is ready and in place for whatever timelines work on the C&R side.

kuzminrobin:
Reapproving.

PR #477, opened 2021-01-21
Add space between -D and name in CMake command

samarsha:
After updating from CMake 3.16 to 3.19.3, the bootstrap.ps1/build-native-simulator.ps1 scripts stopped working:

```
PS C:\Users\samarsha\Source\Repos\microsoft\qsharp-runtime> .\bootstrap.ps1
Setting up build environment variables

...

On Windows build native simulator using the default C/C++ compiler (should be MSVC)
CMake Error: Parse error in command line argument: -DBUILD_SHARED_LIBS:
Should be: VAR:type=value

CMake Error: Problem processing arguments. Aborting.

Error: could not load cache
##vso[task.logissue type=error;]Failed to build Native simulator.
```

Adding a space between `-D` and the variable name seems to fix the problem. That's also the format that `cmake --help` prefers:

```
> cmake --help
Usage
...

Options
...
  -D <var>[:<type>]=<value>    = Create or update a cmake cache entry.
```

### DIFF
--- a/src/QirRuntime/build-qir-runtime.ps1
+++ b/src/QirRuntime/build-qir-runtime.ps1
@@ -37,7 +37,7 @@ if ($Env:ENABLE_QIRRUNTIME -eq "true") {
 
     Push-Location $qirRuntimeBuildFolder
 
-    cmake -G Ninja -DCMAKE_BUILD_TYPE="$Env:BUILD_CONFIGURATION" ../..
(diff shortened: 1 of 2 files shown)


irinayat-MS:
A similar change should probably be done in `build-qir-runtime.ps1`. Thanks for taking care of it!

irinayat-MS (approved):
LGTM. Thanks!

PR #479, opened 2021-01-22
Add execution context that can be configured to track object allocations

irinayat-MS:
Two qir-static tests detect memory leaks so am not enabling allocation tracking in them.

Note: the update in the readme is mostly just line splits to make it more reviewable in the future (the meaningful change in it is about the execution context)


src/QirRuntime/README.md
```diff
@@ -65,65 +72,88 @@ To execute all tests locally run `test.py` from the project's root folder:
...
+ allows the client to access the target simulator directly, if so desired.
+1. Link against the *shared qdk* library. The example of this approach can be found in `test/QIR-dynamic` folder. In the
+ future we'll provide fully self-contained packages of the runtime to enable this workflow completely outside of the
+ current project. For now, this way of consuming QIR only supports running against the native full state simulator.
 
-QIR's architecture assumes a single target, whether that be hardware or a particular simulator. As a result, there is no provision in the QIR specifications to choose a target dynamically. To connect QIR to the simulators from this runtime, we provide `SetSimulatorForQIR(IQuantumGateSet*)` method. Switching simulators while executing QIR isn't supported and would yield undefined behavior.
+QIR's architecture assumes a single target, whether that be hardware or a particular simulator. As a result, there is no
+ provision in the QIR specifications to choose a target dynamically. To connect QIR to the simulators from this runtime,
+ we provide `InitializeQirContext` and `ReleaseQirContext` methods. Switching contexts while executing QIR isn't
```
irinayat-MS, line 117: The only meaningful change in the readme is here about the context.

src/QirRuntime/lib/QIR/CMakeLists.txt
```diff
@@ -22,8 +22,10 @@ endif()
 # Step 1: create qir-rt-support lib from the C++ sources
 #
 set(rt_sup_source_files
+  "allocationsTracker.cpp"
   "arrays.cpp"
   "callables.cpp"
+  "context.cpp"
```
kuzminrobin, line 28: Minor (feel free to ignore):
As I understand, here the alphabetic ordering of the files is used (files are compiled in the alphabetic order).
There is also another point of view about the compilation order, that you may find worth knowing. See [Build the Most Failing Files First](https://github.com/kuzminrobin/code_review_notes/blob/master/misc_posts.md#build-the-most-failing-files-first) section.

src/QirRuntime/lib/QIR/allocationsTracker.cpp
```diff
@@ -0,0 +1,85 @@
...
+    void AllocationsTracker::OnRelease(void* object)
+    {
+        auto tracked = this->allocatedObjects.find(object);
+        if (tracked == this->allocatedObjects.end())
+        {
+            quantum__rt__fail(quantum__rt__string_create("Attempting to release an object that isn't tracked!"));
+        }
+        else
+        {
+            if (tracked->second == 0)
```
kuzminrobin, line 63: Minor:
As I understand the `this->allocatedObjects` is declared like this: 
std::unordered_map<void\*, **int**> allocatedObjects;
that is the _signed_ type (`int`) is used for the ref count (rather than an _unsigned_ type). Which means that theoretically the ref count can become below 0 (negative), either as a result of the memory corruption or in a hypothetical multithreading environment where 2 threads can call `AllocationsTracker::OnRelease()` in parallel for an object with ref count 1 (in which case the lines 67 - 70 below can get executed in parallel and in the line 69 the ref count can become `-1` for the slightly-later thread). If the ref count goes below 0 then during the subsequent `AllocationsTracker::OnRelease()` the line 63 will not catch that.
You may want to consider replacing the line 63 with `if (tracked->second <= 0)`
or to use the unsigned type (e.g. `unsigned int`) instead of `int` in the fragment 
std::unordered_map<void\*, **int**> allocatedObjects;

src/QirRuntime/lib/QIR/arrays.cpp
```diff
@@ -9,13 +9,29 @@
...
+using namespace Microsoft::Quantum;
 
 int QirArray::AddRef()
 {
+    if (g_context != nullptr && g_context->trackAllocatedObjects)
+    {
+        g_context->allocationsTracker->OnAddRef(this);
+    }
+
     assert(this->refCount != 0 && "Cannot resurrect released array!");
```
kuzminrobin, line 35: Minor:
It probably makes sense to execute the (green) line 35 before lines 30 - 34.

irinayat-MS: I want to be able to test the tracking error handling and if I move the assert up, it would fail the tests in debug builds. The thinking here is that the tracking, when enabled, is guaranteed to detect (and handle in some well-defined way) the reference counting bugs, and the assert is a "best effort" debug-only diagnostics (and documentation).

src/QirRuntime/lib/QIR/arrays.cpp
```diff
@@ -9,13 +9,29 @@
...
 #include "quantum__rt.hpp"
 
-#include "CoreTypes.hpp"
+namespace Microsoft
+{
+namespace Quantum
+{
+    extern thread_local std::unique_ptr<QirExecutionContext> g_context;
+}
+} // namespace Microsoft
```
kuzminrobin, line 24: Minor:
I see this in 4 places. You may want to consider placing this in a header (`context.hpp`?).

src/QirRuntime/lib/QIR/arrays.cpp
```diff
@@ -55,6 +76,11 @@ QirArray::QirArray(int64_t qubits_count)
     , ownsQubits(true)
     , refCount(1)
 {
+    if (g_context != nullptr && g_context->trackAllocatedObjects)
+    {
+        g_context->allocationsTracker->OnAllocate(this);
+    }
+
```
kuzminrobin, line 83: Minor:
Just in case you may want to consider placing this to the _very end_ of the constructors (such that only the _successfully_ and _fully_ constructed instances are added to the `g_context->allocationsTracker`). 

Otherwise the remaining part of the constructor can fail (e.g. during the allocation `QUBIT** qbuffer = new QUBIT*[count]` throw an exception), and the `this->buffer` will stay *un*initialized. And if you break in the debugger upon such a failure and analyze the latest instance in `g_context->allocationsTracker`, then that instance will have a misleading `->buffer`.

src/QirRuntime/lib/QIR/context.cpp
```diff
@@ -0,0 +1,68 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include <assert.h>
```
kuzminrobin, line 4: Minor:
Typically the C++ files do not include the C standard headers _directly_ (`#include <assert.h>`) but instead the C++ files include the C++-wrappers of those (`#include <cassert>`). I don't know exactly why.

src/QirRuntime/lib/QIR/context.cpp
```diff
@@ -0,0 +1,68 @@
...
+#include "allocationsTracker.hpp"
+
+#ifdef _WIN32
+#define QIR_SHARED_API __declspec(dllexport)
+#else
+#define QIR_SHARED_API
+#endif
+
+extern "C" QIR_SHARED_API Result ResultOne = nullptr;
+extern "C" QIR_SHARED_API Result ResultZero = nullptr;
```
kuzminrobin, line 19: Minor:
I don't have experience in writing DLLs, but these two variables seem to me a part of the API exported/exposed by the DLL. In which case it probably makes sense to mention them in the corresponding .hpp file, and providing comments about what they are.

src/QirRuntime/lib/QIR/CMakeLists.txt
irinayat-MS: Right now the code churns pretty much across the board so I feel it's too early to optimize the build system yet. Currently, the longest steps of the build are catch2 tests (because I'm using a single header include which is known to be sub-optimal as far as build-time is concerned -- this one will be worth looking into at some point) and compilation of *.ll files. For the latter I don't know how to speed things up but I also don't expect to grow the body of these tests much.

src/QirRuntime/lib/QIR/context.cpp
irinayat-MS: Would you mind filing a bug in this repo, something along the lines of "QirRuntime coding style improvements" for this and maybe other issues you can think of while looking at the code? I'm absolutely open to doing the cleanups and standardizing on better conventions but those changes are better to be done across the project and not mixed into the feature PRs. Thank you!

src/QirRuntime/lib/QIR/context.cpp
irinayat-MS: The constants are defined in generated QIR and are troublesome. I remember bringing it up in the past but cannot find any bugs on this...

src/QirRuntime/lib/QIR/context.cpp
```diff
@@ -0,0 +1,68 @@
...
+    QirExecutionContext::QirExecutionContext(ISimulator* sim, bool trackAllocatedObjects)
+        : simulator(sim)
+        , trackAllocatedObjects(trackAllocatedObjects)
+    {
+        if (this->trackAllocatedObjects)
+        {
+            this->allocationsTracker = std::make_unique<AllocationsTracker>();
+        }
+    }
+    QirExecutionContext::~QirExecutionContext() {}
```
kuzminrobin, line 65: Minor:
What is the purpose for the empty destructor in this case?
FYI:
[Know All the Effects of the Empty Destructor](https://github.com/kuzminrobin/code_review_notes/blob/master/cpp_design_bookmarks.md#know-all-the-effects-of-the-empty-destructor)
[Know What Happens If Rule of Three Is Violated](https://github.com/kuzminrobin/code_review_notes/blob/master/cpp_design_bookmarks.md#know-what-happens-if-rule-of-three-is-violated)

irinayat-MS: I had to move the destructor into the cpp file to avoid including extra headers into context.hpp but I should replace it with `= default`. Thanks for pointing this out!

kuzminrobin (approved):

src/QirRuntime/test/unittests/QirRuntimeTests.cpp
```diff
@@ -841,4 +865,64 @@ TEST_CASE("Unpacking input tuples of nested callables (case1)", "[qir_support]")
...
+
+    QirArray* releaseTwice = quantum__rt__array_create_1d(1, 1);
+    quantum__rt__array_unreference(releaseTwice);
+    CHECK_THROWS(quantum__rt__array_unreference(releaseTwice));
+
+    QirArray* maybeLeaked = quantum__rt__array_create_1d(1, 1);
+    CHECK_THROWS(ReleaseQirContext());
+
+    quantum__rt__array_unreference(maybeLeaked);
+    ReleaseQirContext();
```
kuzminrobin, line 886: `CHECK_NO_THROW(ReleaseQirContext())`?

src/QirRuntime/test/unittests/QirRuntimeTests.cpp
```diff
@@ -841,4 +865,64 @@ TEST_CASE("Unpacking input tuples of nested callables (case1)", "[qir_support]")
...
+    CHECK_THROWS(quantum__rt__tuple_unreference(releaseTwice));
+
+    PTuple maybeLeaked = quantum__rt__tuple_create(1);
+    CHECK_THROWS(ReleaseQirContext());
+
+    quantum__rt__tuple_unreference(maybeLeaked);
+    ReleaseQirContext();
+}
+
+void CallableEntry(PTuple, PTuple, PTuple) {}
```
kuzminrobin, line 908: Minor:
**static** void CallableEntry(PTuple, PTuple, PTuple) {} ?

You may also want to consider making the function `CallableEntry()` a local function (move line 908 to between the lines 910 and 911). Such that if at some point in the future somebody decides to remove (or move/shift) the whole test case, then the function `CallableEntry()` will (re)moved together with the test case.
See "The static function of the local class" in [Nested (Local) Functions](https://github.com/kuzminrobin/code_review_notes/blob/master/cpp_design_bookmarks.md#nested-local-functions)

kuzminrobin (approved):

src/QirRuntime/lib/QIR/context.cpp
bettinaheim: Do you have any suggestions for how to best expose the results constants that the QIR relies on?

bettinaheim (approved):

src/QirRuntime/lib/QIR/allocationsTracker.cpp
irinayat-MS: QIR runtime is single threaded but it won't hurt to make the check broader here.

src/QirRuntime/lib/QIR/arrays.cpp
```diff
@@ -9,13 +9,29 @@
...
+#include "allocationsTracker.hpp"
+#include "CoreTypes.hpp"
+#include "context.hpp"
 #include "qirTypes.hpp"
 #include "quantum__rt.hpp"
 
-#include "CoreTypes.hpp"
+namespace Microsoft
+{
+namespace Quantum
```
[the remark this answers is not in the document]
irinayat-MS: Added to https://github.com/microsoft/qsharp-runtime/issues/482. This should be done across the project.

src/QirRuntime/lib/QIR/context.cpp
irinayat-MS: My preference would be to wrap those in function calls, something like: `%Result* __quantum__rt__result_zero()`.

src/QirRuntime/test/unittests/QirRuntimeTests.cpp
irinayat-MS: The check_nothrow could, in theory, allow other tests to run rather than terminate on exception, however, if an exception is thrown, the context won't be cleaned up and other tests would run into asserts about that... I suppose, it's still good to have check_nothrow for documentation purposes here.

src/QirRuntime/test/unittests/QirRuntimeTests.cpp
irinayat-MS: Marked it static and renamed to `DummyCallableEntry`

src/QirRuntime/test/unittests/QirRuntimeTests.cpp
kuzminrobin: If terminating upon unexpected exception provides a better diagnostic then it probably makes sense to leave as is. I saw such situations earlier and agree, check_nothrow may be a worse idea.

PR #480, opened 2021-01-24
Expose quota information from Workspace.

ricardo-espinoza:
This pull request exposes the Quota information through the IWorkspace interface so it can be consumed by clients like IQSharp. It is mostly a wrapper for generated code under /src/Azure/Azure.Quantum.Client/generated/.


vxfield (approved):
Looks good to me.

PR #481, opened 2021-01-26
Update C# generation for access modifier changes in Q# compiler

samarsha:
This updates C# generation to work with the changes in microsoft/qsharp-compiler#835.


irinayat-MS (approved):
lgtm

Issue #156, opened 2020-03-03
Make printing stack trace optional and configurable


tcNickolas:
** Please describe what you would like the feature to accomplish.**
I'm updating the Quantum Katas to QDK version 0.10.2002.2610, and it looks like the cell output in case of an error became a lot bulkier. As of PR https://github.com/microsoft/qsharp-runtime/pull/79/, they include a mandatory Q# stack trace, which (for the use case of the Katas) doesn't really help the user figure out their issue, and might scare them away instead (this was the feedback we got when we had long stack trace in new Python tutorials).

![image](https://user-images.githubusercontent.com/10113024/75731034-6a3c9000-5ca3-11ea-90da-ecaa87b517db.png)

I considered turning off the test output, but that won't work, since the katas rely on the output of `Message` and `DumpMachine` for conveying useful information about the test outcomes (such as the first line of task 1.1) and for allowing the user to debug their solution. 

Ideally we would have an option to turn off the extra stack tracing, and then use that option in the `TestTarget` method of `TestSuiteRunner` in the katas.

anpaz-msft:
I think #162 takes care of this, right @tcNickolas?

tcNickolas:
I'm not sure about #162, #157 looks more relevant to it?

bettinaheim:
@tcNickolas Could you confirm whether that is still relevant? If so, I'll assign it to an iteration shortly.

tcNickolas:
This specific behavior in the Kata notebooks was fixed, yes. I'm not sure how configurable stack trace printing is now.

DmitryVasilevsky:
@tcNickolas : SimulatorBase has public methods EnableExceptionPrinting() and DisableExceptionPrinting() which control printout in case of exception. Disabling exception printing will disable all exeption printing including words 'Unhandled exception.' and the message, I assume you want the message, but don't want stack trace. I can make printing of stack trace configurable.

tcNickolas:
To provide a bit of background from the offline discussion: currently the Q# Jupyter notebooks do not print the exception stack trace, so I'm happy with their behavior.

However, the stack trace is still present in the katas when they are executed as Q# standalone projects; here is what the overall test output looks like:
![vs-stacktrace-screenshot](https://user-images.githubusercontent.com/10113024/106205018-b376b780-6172-11eb-9e17-997fe00175f6.jpg)

[...]

PR #487, opened 2021-01-28
Making stack trace printout optional

DmitryVasilevsky:
This change allows control over stack trace printout. Added EnableStackTracePrinting property on SimulatorBase. Set to true (default) to enable printout, set to false to disable printout. Only stack trace printout is disabled, exception message is still printed.
This partially addresses issue https://github.com/microsoft/qsharp-runtime/issues/156


src/Simulation/Common/SimulatorBase.cs
```diff
@@ -43,6 +43,7 @@ public abstract class SimulatorBase : AbstractFactory<AbstractCallable>, IOperat
 
         public abstract string Name { get; }
 
+        public bool EnableStackTracePrinting { get; set; } = true;
```
cgranade, line 46: I'd suggest adding an API documentation comment here. In particular, since this is only effective if `DisableExceptionPrinting()` has not been called, that would be good to note in a `<remark>` here to prevent confusion.

cgranade (commented):
This looks like a reasonable addition to me, thanks for tagging me in.

It would be good to test manually that this works as expected in IQ#, however, as IQ# uses the `DisableExceptionPrinting` method to control stack traces:

```C#
        /// <summary>
        ///     Adds functionality to a given quantum simulator to display
        ///     diagnostic output and stack traces for exceptions.
        /// </summary>
        /// <param name="simulator">
        ///     The simulator to be augmented with stack trace display functionality.
        /// </param>
        /// <param name="channel">
        ///     The Jupyter display channel to be used to display stack traces.
        /// </param>
        /// <returns>
        ///     The value of <paramref name="simulator" />.
        /// </returns>
        public static T WithStackTraceDisplay<T>(this T simulator, IChannel channel)
        where T: SimulatorBase
        {
            simulator.DisableExceptionPrinting();
            simulator.OnException += (exception, stackTrace) =>
            {
                channel.Display(new DisplayableException
                {
                    Exception = exception,
                    StackTrace = stackTrace
                });
            };
            return simulator;
        }
```

https://github.com/microsoft/iqsharp/blob/1f65d7a57523c2a3e8b2977000e976826a2d9c28/src/Jupyter/Extensions.cs#L178

In particular, the change here may not address @tcNickolas' report in #156, since as per the screenshot in that issue, the stack traces are being generated by the `Run` method in https://github.com/microsoft/QuantumKatas/blob/eb9a70e150051a87beb04c90a4ef95f073a09b36/utilities/Microsoft.Quantum.Katas/KataMagic.cs#L110:

```C#
        /// <summary>
        /// What this Magic does when triggered. It will:
        /// - find the Test to execute based on the given name,
        /// - compile the code after found after the name as the user's answer.
        /// - run (simulate) the test and report its result.
        /// </summary>
        public virtual async Task<ExecutionResult> Run(string input, IChannel channel)
        {
            channel = channel.WithNewLines();

            // Expect exactly two arguments, the name of the Kata and the user's answer (code).
            var args = input?.Split(new char[] { ' ', '\n', '\t' }, 2);
            if (args == null || args.Length != 2)
            {
                channel.Stderr("Invalid parameters. Usage: `%kata Test \"Q# operation\"`");
                return ExecuteStatus.Error.ToExecutionResult();
            }

            var name = args[0];
            var code = args[1];

            var test = FindTest(name);
            if (test == null)
            {
                channel.Stderr($"Invalid test name: {name}");
                return ExecuteStatus.Error.ToExecutionResult();
            }

            var userAnswer = Compile(code, channel);
            if (userAnswer == null) { return ExecuteStatus.Error.ToExecutionResult(); }

            return Simulate(test, userAnswer, channel)
                ? "Success!".ToExecutionResult()
                : ExecuteStatus.Error.ToExecutionResult();
        }
```

Modifying that magic command to use redirect stack traces through [`DisplayableExceptionEncoders.cs`](https://github.com/microsoft/iqsharp/blob/cdb3f8211f12a0836848845c8f7729899209f366/src/Jupyter/Visualization/DisplayableExceptionEncoders.cs#L49) and then making those encoders more configurable (likely using IQ#'s `IConfigurationSource` functionality) would allow for fine-tuning how much stack trace information is shown from the `%kata` magic command.

tcNickolas:
Just to clarify (I need to post the offline discussion to that issue for the most up-to-date status): I'm happy with the current behavior in Q# Jupyter notebooks, it's the stack trace in the Katas executed as Q# standalone projects that is excessive for me.

cgranade:
> Just to clarify (I need to post the offline discussion to that issue for the most up-to-date status): I'm happy with the current behavior in Q# Jupyter notebooks, it's the stack trace in the Katas executed as Q# standalone projects that is excessive for me.

Ah, got it; I had misunderstood from the other issue, then. In that case, I'm fine with this PR as-is, other than that an API comment would help prevent confusion about when the new property applies.

src/Simulation/Common/SimulatorBase.cs
DmitryVasilevsky: Added comment.

bettinaheim (approved):

Issue #44, opened 2019-08-26
Use accelerated SWAP kernel to implement Intrinsic.SWAP.


cgranade:
**Please describe what you would like the feature to accomplish.**
Currently, the Microsoft.Quantum.Intrinsic.SWAP operation is provided along with a pure-Q# implementation in terms of three `CNOT` calls:

```Q#
    operation SWAP (qubit1 : Qubit, qubit2 : Qubit) : Unit
    is Adj + Ctl {
        body (...)
        {
            CNOT(qubit1, qubit2);
            CNOT(qubit2, qubit1);
            CNOT(qubit1, qubit2);
        }
        
        adjoint self;
    }
```

The C++ runtime for the `QuantumSimulator` target machine supports, however, an accelerated kernel for applying SWAP gates. It would help with simulator performance to expose this as an implementation of the `SWAP` intrinsic operation, similarly to how [`X` is exposed by `QuantumSimulator`](https://github.com/microsoft/qsharp-runtime/blob/master/src/Simulation/Simulators/QuantumSimulator/X.cs).

swernli:
With the addition of support for target packages, we now have an internal `ApplyUncontrolledSWAP` intrinsic (see https://github.com/microsoft/qsharp-runtime/blob/main/src/Simulation/TargetDefinitions/Intrinsic/ApplyUncontrolledSWAP.qs). That might be a good place to plug into the simulator SWAP and then use the decomposition over the uncontrolled SWAP.

PR #490, opened 2021-01-30
Add Type3 target package

swernli:
This adds the Type3 target package, which is very similar to Type1 but includes support for an intrinsinc native SWAP, and uses a NOOP when handing post-measurement preparation (reset and re-prepare of qubit state does not need to be explicit). This includes the package itself, simlator support, tests, and updates to the solution and packing script.


src/Simulation/Type3Core/Microsoft.Quantum.Type3.Core.csproj
```diff
@@ -0,0 +1,11 @@
+﻿<Project Sdk="Microsoft.Quantum.Sdk/0.14.2012122063-beta">
+
+  <Import Project="..\TargetDefinitions\TargetPackages\Type3.Package.props" />
+
+  <PropertyGroup>
+    <Description>Type2 Targeting support for the Q# programming language.</Description>
```
irinayat-MS, line 6: Did you mean Type3 here?

src/Simulation/TargetDefinitions/Decompositions/SWAPFromSinglyControlled.qs
```diff
@@ -0,0 +1,56 @@
...
+        body (...) {
+            ApplyUncontrolledSWAP(qubit1, qubit2);
+        }
+        adjoint self;
+        controlled (ctls, ...) {
+            if (Length(ctls) == 0) {
+                ApplyUncontrolledSWAP(qubit1, qubit2);
+            }
+            else {
+                within {
```
irinayat-MS, line 47: Nit: is it worth using `within` here? I'd expect folks to be used to decomposing SWAP into three 'independent' CNOT operations rather than thinking about the outer CNOTs as body/adjoint pair.

src/Simulation/Type3Core/Microsoft.Quantum.Type3.Core.csproj
swernli: Whoops! I sure did. Thanks for catching!

src/Simulation/Simulators/QuantumSimulator/ApplyUncontrolledSWAP.cs
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Runtime.InteropServices;
+using Microsoft.Quantum.Simulation.Core;
+
+namespace Microsoft.Quantum.Simulation.Simulators
+{
+    public partial class QuantumSimulator
```
irinayat-MS, line 10: Maybe add a link to https://github.com/microsoft/qsharp-runtime/issues/44?

src/Simulation/TargetDefinitions/Decompositions/SWAPFromSinglyControlled.qs
swernli: I tend to agree, and I originally had the three independent `CNOT` version both here and in the normal decomposition of `SWAP` used by QSharp.Core, but ran into a merge conflict with https://github.com/microsoft/qsharp-runtime/pull/458 done by @msoeken. Given that this change was merged before mine, I didn't want to undo it, so I used that formulation for both versions of the `SWAP` decomposition.

irinayat-MS (approved):
lgtm!

src/Simulation/TargetDefinitions/Decompositions/SWAPFromSinglyControlled.qs
swernli: but now looking again at @msoeken 's change, I think I might be able to make this decomposition more efficient and still correct if I remove the controls from `CNOT` in the within. @msoeken is that right?

src/Simulation/Simulators/QuantumSimulator/ApplyUncontrolledSWAP.cs
swernli: That's a great idea! I'll also add a comment on the issue that points to this part of the implementation to make it easier to find for whoever picks up that work.

src/Simulation/TargetDefinitions/Decompositions/SWAPFromSinglyControlled.qs
swernli: I confirmed with @bettinaheim that the controls for the conjugation form are not needed, so went ahead and removed them.

PR #495, opened 2021-02-02
Attributes for QIR emission

bettinaheim:
This PR also cleans up some issues with the .sln file.


cgranade:
This looks fine to me, but formally should be scheduled for API review; I'm happy to add it to the scheduling board on libraries if that's OK with you?

src/Simulation/QSharpFoundation/Core.qs
```diff
@@ -19,6 +19,11 @@ namespace Microsoft.Quantum.Core {
     @Attribute()
     newtype EntryPoint = Unit;
 
+    /// # Summary
+    /// Compiler-recognized attribute used to indicate that a callable should always be inlined.
+    @Attribute()
+    newtype Inline = Unit;
```
samarsha, line 25: I don't understand how you can guarantee that it is always inlined. What code is generated for:

```qsharp
// @Inline() And
// @Inline() Or

function Foo(f : Bool -> Bool -> Bool) : Bool {
    return f(true, false);
}

function Bar() : Unit {
    let _ = Foo(And);
    let _ = Foo(Or);
}
```

bettinaheim:
> This looks fine to me, but formally should be scheduled for API review; I'm happy to add it to the scheduling board on libraries if that's OK with you?

Sounds good. These two attributes are currently needed for QIR emission (meaning I need the feature branch for the compiler e2e to succeed).

src/Simulation/QSharpFoundation/Core.qs
bettinaheim: Ah, I should clarify in the doc comment: It is always inlined during QIR emission.

samarsha: I'm still not sure I follow. :) I assume the emitted QIR for `Foo` will take in an argument that is a pointer to a `%Callable`. That pointer forces the function call to be made and not inlined, unless copies of `Foo` are generated for each inlinable argument.

cgranade:
> > This looks fine to me, but formally should be scheduled for API review; I'm happy to add it to the scheduling board on libraries if that's OK with you?
> 
> Sounds good. These two attributes are currently needed for QIR emission (meaning I need the feature branch for the compiler e2e to succeed).

Cool, I'll go on and approve now then, so as to unblock. Following the result of the review, if we need to make another PR then that should be doable before release. Thank you!

cgranade (approved):

src/Simulation/QSharpFoundation/Core.qs
bettinaheim: I updated the doc comment to be more precise and say that it is used during QIR emission to inline a callable if possible (admittedly "possible" is a bit relative). I don't think it is worth going into the detail of callable values indeed never being inlined (which callable it is is runtime information and we "can't" inline, as you point out).

PR #496, opened 2021-02-02
Added all standard Q# gates to qis-bridge and removed non-standard ones

irinayat-MS:
Also:
Refactored tests that depend on the fullstate simulator into a separate component.
Renamed a few gates, such as ControledSAdjoint, to avoid having gate name in the middle of the string.


src/QirRuntime/README.md
```diff
@@ -98,12 +98,42 @@ For tests that depend on the native simulator and qdk shared libraries, you migh
...
 This project contains an implementation of the QIR runtime per the
  [QIR specifications](https://github.com/microsoft/qsharp-language/tree/main/Specifications/QIR) and the translation
- layer between the QIR and the IR, generated by Clang from the native code. We call the translation layer "QIR Bridge".
- This project also provides an optional implementation of intrinsics from `quantum__qis*` namespace that are used in QIR,
- generated from Q#.
+ layer between the QIR and the IR, generated by Clang from the native code. Translation layer is called the "QIR Bridge".
 
 ![QIR Bridge architecture diagram](qir.png?raw=true "QIR Bridge architecture diagram")
 
+This project also provides an implementation of the quantum instruction set, used by Q# for simulation against the full
```
kuzminrobin, line 105: From time to time I forget what QIS stands for. And I don't remember any doc explicitly telling what QIS is.
This README.md seems to me a good place to provide that info, and if the QIS stands for "quantum instruction set", then we could mention like this
> This project also provides an implementation of the quantum instruction set **(QIS)**,

irinayat-MS: This naming convention depends on the compiler. For example, the math functions will likely use the same naming scheme and they aren't exactly part of the quantum instruction set... The subset of `__quantum__qis*` functions listed here is exactly the QIS supported by our simulators, though.

src/QirRuntime/lib/QIR/intrinsics.cpp
```diff
@@ -15,38 +15,79 @@
...
+    void quantum__qis__exp__ctladj(QirArray* ctls, QirArray* paulis, double angle, QirArray* qubits)
     {
-        Microsoft::Quantum::g_context->simulator->AsQuantumGateSet()->ControlledX(1, &control, qubit);
+        assert(paulis->count == qubits->count);
+
+        std::vector<PauliId> pauliIds = ExtractPauliIds(paulis);
+        return Microsoft::Quantum::g_context->simulator->AsQuantumGateSet()->ControlledExp(
+            ctls->count, reinterpret_cast<Qubit*>(ctls->buffer), paulis->count,
+            reinterpret_cast<PauliId*>(pauliIds.data()), reinterpret_cast<Qubit*>(qubits->buffer), angle);
     }
```
kuzminrobin, line 72: I don't see the difference between 
`quantum__qis__exp__ctl()` and 
`quantum__qis__exp__ctladj()`. 
May be the latter can call the former?
```c++
void quantum__qis__exp__ctladj(QirArray* ctls, QirArray* paulis, double angle, QirArray* qubits)
{
    quantum__qis__exp__ctl(ctls, paulis, angle, qubits);
}
```

cgranade: That may be a bug; the sign of the angle passed to `ControlledExp` should be inverted when calling the controlled adjoint variant.

src/QirRuntime/lib/QIR/intrinsics.cpp
```diff
@@ -56,55 +97,102 @@ extern "C"
...
+    {
+        return Microsoft::Quantum::g_context->simulator->AsQuantumGateSet()->ControlledR(
+            ctls->count, reinterpret_cast<Qubit*>(ctls->buffer), axis, qubit, angle);
+    }
+
+    void quantum__qis__r__ctladj(QirArray* ctls, PauliId axis, double angle, QUBIT* qubit)
+    {
+        return Microsoft::Quantum::g_context->simulator->AsQuantumGateSet()->ControlledR(
+            ctls->count, reinterpret_cast<Qubit*>(ctls->buffer), axis, qubit, angle);
+    }
```
kuzminrobin, line 120: Minor:
I see that 
`quantum__qis__r__ctl()` and 
`quantum__qis__r__ctladj()` are identical.
May be it is possible to use a single function instead of two? Or can one of them just call the other of them, in such a way showing explicitly that they have the same implementation?

src/QirRuntime/lib/QIR/intrinsics.cpp
```diff
@@ -15,38 +15,79 @@
...
+
+    void quantum__qis__exp__adj(QirArray* paulis, double angle, QirArray* qubits)
+    {
+        assert(paulis->count == qubits->count);
+
+        std::vector<PauliId> pauliIds = ExtractPauliIds(paulis);
+        return Microsoft::Quantum::g_context->simulator->AsQuantumGateSet()->Exp(
+            paulis->count, reinterpret_cast<PauliId*>(pauliIds.data()), reinterpret_cast<Qubit*>(qubits->buffer),
+            -angle);
+    }
```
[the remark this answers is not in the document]
irinayat-MS: Good idea! And would help to avoid the bugs like not flipping the angle :)

src/QirRuntime/lib/QIR/bridge-qis.ll
```diff
@@ -29,122 +29,264 @@
 %struct.QirCallable = type opaque
 %struct.QirRange = type { i64, i64, i64 }
 %struct.QirString = type opaque
+;%PauliId = type {i32}
```
kuzminrobin, line 32: Forgot to remove or forgot to uncomment or is it just a clarification?

src/QirRuntime/lib/QIR/intrinsics.cpp
irinayat-MS: You are both correct. The angle should be inverted, but to make it easier to see the logic, it helps to implement the adjoint in terms of body.

src/QirRuntime/lib/QIR/bridge-qis.ll
irinayat-MS: The intent was to clarify... Does it help?

src/QirRuntime/lib/QIR/intrinsics.cpp
irinayat-MS: Yeap, another bug :(

src/QirRuntime/lib/QIR/bridge-qis.ll
```diff
@@ -29,122 +29,264 @@
...
+  call void @quantum__qis__exp__body(%struct.QirArray* %paulis, double %angle, %struct.QirArray* %qubits)
+  ret void
+}
+
+define void @__quantum__qis__exp__adj(%Array* %.paulis, double %angle, %Array* %.qubits) {
+  %paulis = bitcast %Array* %.paulis to %struct.QirArray*
+  %qubits = bitcast %Array* %.qubits to %struct.QirArray*
+  call void @quantum__qis__exp__adj(%struct.QirArray* %paulis, double %angle, %struct.QirArray* %qubits)
+  ret void
 }
```
kuzminrobin, line 81: I feel something that I don't know yet behind the observation that all the parameters start with dot `.` (after `%`), but the parameter `double %angle` does not (it has no `.` after `%`).

kuzminrobin, line 81: Just curiosity, feel free to ignore.
What will happen if `ret void` is removed?

irinayat-MS: Ah! That's my own sneaky naming convention! :D If you use temporary variables (like `%0` or `%42`) in LLVM's IR, Clang insists on them being in order and having no gaps which makes it a pain to update the code. So I went with naming the locals explicitly, using leading dot (which is _not_ a special char in LLVM's IR) for types from QIR and no dot for native types.

irinayat-MS: It fails to compile :)

kuzminrobin: I guess I have an intuitive guess of what that means... No need to reply.

irinayat-MS: To be less glib with the response, I believe LLVM requires explicit ends for basic blocks, which `ret` instruction provides. And it also requires explicit type specifications for everything, including `void` so you end up with `ret void`

src/QirRuntime/test/FullstateSimulator/qir-test-simulator.ll
```diff
@@ -0,0 +1,1092 @@
...
+%Array = type opaque
+%Qubit = type opaque
+%Callable = type opaque
+
+@ResultZero = external global %Result*
+@ResultOne = external global %Result*
+@PauliI = constant i2 0
+@PauliX = constant i2 1
+@PauliY = constant i2 -1
+@PauliZ = constant i2 -2
```
kuzminrobin, line 14: Funny :-)
Looks like the `i*` types (like `i64`, `i2`) are always _signed_, 
and the 2-bit unsigned value of `3` (having binary representation of `11`) 
is displayed as the signed 2-bit value of `-1` (also having binary representation of `11`).

And the same for 2-bit unsigned `2` displayed as 2-bit signed `-2` (both `10`).

src/QirRuntime/test/FullstateSimulator/qir-test-simulator.ll
```diff
@@ -0,0 +1,1092 @@
+
```
kuzminrobin, line 1: Is this file an expected output to be compared to the actual output generated from the `qir-test-simulator.qs`?
If yes then I fail to understand in which fragments this expected output file is compared to the actual output file. Please advise.

irinayat-MS: Right now the foo.ll files in this project are generated manually from the corresponding foo.qs files. There is no validation that the QIR itself is correct or matches a particular expected pattern (that probably should be done as part of compiler tests for QIR generation). The purpose of these tests is to check that runtime handles the QIR correctly. For example, the tests in qir-static use various mock simulators to validate expectations. Unfortunately, the fullstate simulator would produce random outcomes for most sensible operations. I suppose, I can update the test to handle some situations that guarantee non-random outcomes (like ensuring that R and Adjoint R do cancel each other). Do you think that would be useful?

kuzminrobin (approved):

src/QirRuntime/lib/QIR/bridge-qis.ll
kuzminrobin: May be 
```llvm
; Clarifications:
;%PauliId = type {i32}
```
?
(to avoid making an impression that it is forgotten to be either removed or uncommented)

src/QirRuntime/test/FullstateSimulator/qir-test-simulator.ll
kuzminrobin: Looks like I still need to get a better understanding of the terminology and mechanisms of the build-and-test process, in order to fully understand your answer. That is why I don't want this my topic to be treated as an obstacle for moving forward with completing this PR. Feel free to go ahead with the PR, no need to answer to the text below.

Let me summarize what I understand so far.
Some `.ll` files are written manually (e.g. `qsharp-runtime\src\QirRuntime\lib\QIR\bridge-*.ll`. They don't have the corresponding `.qs` files). 
But some `.ll` files can be generated from the corresponding `.qs` files. For such `.ll` files we don't need to store their commit history in Git, we can generate them upon need with the build scripts (CMake files). But there are reasons why it is better to generate them manually from time to time and store them in the Git history.

irinayat-MS: Ah! I see what you mean. Yes, in the future we want to be able to generate the test *.ll files from the corresponding *.qs files as part of the build. There are a few obstacles to that at the moment:
- QIR generation pipeline isn't multi-platform yet (Stefan is working on it)
- Some minor manual edits are needed to the generated QIR to make it compile, some of these are because of bugs that Bettina is looking into and some are caused by using multiple *.ll files in a single project -- we might need to restructure the tests because of that.

The bottom line is, that _for now_ the test *.ll files are being generated manually, doctored a bit and then checked in.

src/QirRuntime/lib/QIR/bridge-qis.ll
```diff
@@ -29,122 +29,264 @@
...
   %ctls = bitcast %Array* %.ctls to %struct.QirArray*
-  call void @quantum__qis__crz(%struct.QirArray* %ctls, double %.theta, %class.QUBIT* %q)
+  call void @quantum__qis__z__ctl(%struct.QirArray* %ctls, %class.QUBIT* %q)
   ret void
 }
+
+;
+; non-standard gates
+;
+define %Result* @__quantum__qis__mz(%Qubit* %.q) {
```
swernli, line 286: Just curious, why is this still needed? When does this particular instruction get included in the code?

src/QirRuntime/lib/QIR/intrinsics.cpp
```diff
@@ -15,38 +15,79 @@
...
-    Result quantum__qis__measure(QirArray* paulis, QirArray* qubits)
+    void quantum__qis__h__ctl(QirArray* ctls, Qubit qubit)
+    {
+        Microsoft::Quantum::g_context->simulator->AsQuantumGateSet()->ControlledH(
+            ctls->count, reinterpret_cast<Qubit*>(ctls->buffer), qubit);
+    }
+
+    Result quantum__qis__measure__body(QirArray* paulis, QirArray* qubits)
     {
         const long count = qubits->count;
```
swernli, line 87: Very minor: this is the one function where you save the qubit count into a local variable to avoid calling it twice (once in the assert and once in the simulator invocation). Is it worth removing this to keep it looking like the other methods?

src/QirRuntime/test/FullstateSimulator/FullstateSimulatorTests.cpp
```diff
@@ -197,7 +199,7 @@ TEST_CASE("Fullstate simulator: TTSAdj=Id", "[fullstate_simulator]")
         iqa->H(q);
```
swernli, line 199: I like these tests! Perhaps it's worth adding tests for adjoint R and adjoint Exp to verify the angle flipping now that those operations are present?

src/QirRuntime/test/QIR-dynamic/qir-driver.cpp
```diff
@@ -26,7 +26,10 @@ TEST_CASE("QIR: SimpleVQE with full state simulator", "[qir]")
...
     const int ret2 = Microsoft__Quantum__Testing__QIR__QuantumRandomNumberGenerator__body();
-    INFO(std::string("two random numbers: ") + std::to_string(ret1) + " " + std::to_string(ret2));
-    REQUIRE(ret1 != ret2);
-
+    const int ret3 = Microsoft__Quantum__Testing__QIR__QuantumRandomNumberGenerator__body();
+    INFO(
+        std::string("Three random numbers: ") + std::to_string(ret1) + ", " + std::to_string(ret2) + ", " +
+        std::to_string(ret3));
+    CHECK(ret1 != ret2);
+    CHECK(ret1 != ret3);
```
swernli, line 34: maybe a bit silly, but is it also worth adding a check for `ret2 != ret3`?

src/QirRuntime/test/QIR-static/compiler/QirTarget.qs
```diff
@@ -1,135 +1,63 @@
...
-        }
-        controlled adjoint (ctls, ...)
-        {
-            Phys.Rz(-theta / 2.0, qb);
-            CNOT(ctls[0], qb);
-            Phys.Rz(theta / 2.0, qb);
-            CNOT(ctls[0], qb);
-        }
-	}
+    // Non-standard operation. Should it exist?
```
swernli, line 58: In my opinion, no, it can be removed. But that may be better handled in a separate clean-up PR later if there are test changes that need to be made too?

src/QirRuntime/test/QIR-static/compiler/QirTarget.qs
```diff
@@ -1,135 +1,63 @@
...
-    operation Rx (theta : Double, qb : Qubit) : Unit {
-        body intrinsic;
-    }
-
-    operation Rz (theta : Double, qb : Qubit) : Unit {
-        body intrinsic;
-    }
-}
-
 namespace Microsoft.Quantum.Intrinsic {
```
swernli, line 4: This file is almost the same as the QSharp.Core target package, particularly the list of files defined in this props file: https://github.com/microsoft/qsharp-runtime/blob/main/src/Simulation/TargetDefinitions/TargetPackages/QSharpCore.Package.props#L6. I wonder if there is a way to pull that in here without bringing all the other dependencies such that you don't have to repeat them here. Perhaps in a later PR there will be a chance to remove this file in favor of using the target package?

The biggest difference is the presence of the `open Microsoft.Quantum.Targeting` but I'm not sure that is still needed?

src/QirRuntime/test/FullstateSimulator/qir-test-simulator.ll
irinayat-MS: Logged https://github.com/microsoft/qsharp-runtime/issues/498 to track this

src/QirRuntime/test/QIR-dynamic/qir-driver.cpp
irinayat-MS: Sure, no harm in that :D

src/QirRuntime/test/QIR-static/compiler/QirTarget.qs
irinayat-MS: Yep, after talking to Guen and seeing the confusion this callback creates, I went ahead and removed it.

src/QirRuntime/test/QIR-static/compiler/QirTarget.qs
irinayat-MS: I've logged https://github.com/microsoft/qsharp-runtime/issues/498 about automating qir files generation. Part of that work will be to get rid of the custom core/target.qs files in the runtime.

src/QirRuntime/lib/QIR/bridge-qis.ll
irinayat-MS: On its way out...

src/QirRuntime/lib/QIR/intrinsics.cpp
irinayat-MS: I have a mild preference for how it is now as it keeps the line of arguments from wrapping one more time...

src/QirRuntime/test/FullstateSimulator/FullstateSimulatorTests.cpp
irinayat-MS: The angles bugs in this PR were at a different level and wouldn't be caught by these tests, but it does add coverage to provide these tests.

src/QirRuntime/test/QIR-dynamic/qir-driver.cpp
irinayat-MS: Though, it this iteration fails on Linux I will convert this component from a test to a sample with command line output...

PR #497, opened 2021-02-03
Reduce the number of intrinsic callables.

cgranade:
This PR replaces a number of intrinsic callables with pure-Q# implementations, reducing the size of the interface between the runtime and Q# applications. This PR is still in draft, as additional unit tests need to be added to ensure that the pure Q# implementations work as expected.

**Update:** added a bunch of tests, and updated to 0.15.


bettinaheim:
@cgranade Would it make sense to add an implementation for the following: ModPowL, Round, PowD, and ExpD?

src/Simulation/QSharpFoundation/Math/Math.qs
```diff
@@ -97,8 +97,9 @@ namespace Microsoft.Quantum.Math {
     ///
     /// # Remarks
     /// See [System.Math.Log10](https://docs.microsoft.com/dotnet/api/system.math.log10) for more details.
-    function Log10 (input : Double) : Double {
-        body intrinsic;
+    function Log10(input : Double) : Double {
```
irinayat-MS, line 100: LLVM has intrinsic for log base 10 (https://llvm.org/docs/LangRef.html#id2078). Calling it _might_ be faster and have better precision than converting the bases ourselves.

src/Simulation/Simulators.Tests/TestProjects/IntrinsicTests/Bitwise/Tests.qs
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.Quantum.Tests {
+    open Microsoft.Quantum.Bitwise;
+    open Microsoft.Quantum.Diagnostics;
+
+    @Test("QuantumSimulator")
+    function XorIsCorrect() : Unit {
+        EqualityFactI(199, Xor(248, 63), "Xor was incorrect.");
```
irinayat-MS, line 10: Nit: because the test results are often reviewed by humans for both passing and failing tests on my previous projects we used to word the assert messages in terms of what is expected, then passing test ("green") would read as the expectation that was met and failing test ("red") would read as the expectation that wasn't met. It removed the cognitive dissonance when reviewing test logs with "Xor was incorrect." messages while still passing :) Instead, the message would be more like "Bitwise Xor of 11111000 and 00111111 should be 11000111"

cgranade: Thanks for raising that! In this case, `EqualityFactI` uses `FormattedFailure` to print the expected and actual values for each failing test, such that I think I the best way to incorporate that feedback is to add the arguments to the message:

```suggestion
        EqualityFactI(199, Xor(248, 63), "Xor(248, 63) was incorrect.");
```

src/Simulation/QSharpFoundation/Math/Math.qs
cgranade: I'm fine either way; @bettinaheim had highlighted this as something that could be made concrete instead of intrinsic, but I'm happy to undo that change if that would be preferred.

src/Simulation/Simulators.Tests/TestProjects/IntrinsicTests/Bitwise/Tests.qs
irinayat-MS: My concern was mostly with having _negative_ message printed when the test is actually passing. Your call.

src/Simulation/QSharpFoundation/Math/Math.qs
bettinaheim: There are also intrinsics for exp, sqrt and the like, but we first need to expose them via the .NET wrapper. I'd suggest to handle that here for now so that we can unblock e.g. the VQE example and start setting up more extensive testing, and then we can look into leveraging the llvm intrinsics directly.

src/Simulation/QSharpFoundation/Convert/Convert.qs
```diff
@@ -61,48 +60,47 @@ namespace Microsoft.Quantum.Convert {
...
+
+
     /// # Summary
     /// Converts a given double-precision floating-point number to an equivalent string representation.
     ///
     /// # Remarks
     /// See [C# Convert.ToString](https://docs.microsoft.com/dotnet/api/system.convert.tostring?view=netframework-4.7.1#System_Convert_ToString_System_Double_) for more details.
     function DoubleAsString(a : Double) : String {
-        body intrinsic;
+        return $"{a}";
```
irinayat-MS, line 80: @bettinaheim: Bettina, does this change mean that `%String* @__quantum__rt__double_to_string(double %val)` and other to string conversion functions can be removed from the QIR spec and from the runtime?

irinayat-MS (approved):

src/Simulation/QSharpFoundation/Convert/Convert.qs
bettinaheim: Unfortunately, at least not yet. I don't have anything convenient to call in llvm directly and I would need to look in detail into how to do this using string initialization only. For now, this interpolated string will precisely result in a call to __quantum__rt__double_to_string and this change merely removed the redundancy in the QIR generation that would otherwise have required to handle the function call and the string interpolation separately.

src/Simulation/QSharpFoundation/Math/Math.qs
cgranade: Sounds good, let me see if I can make some progress on `ModPowL`, `Round`, `PowD`, and `ExpD` as highlighted above, then. Thanks!

src/Simulation/Simulators.Tests/TestProjects/IntrinsicTests/Bitwise/Tests.qs
cgranade: Ah, got it; sorry for misunderstanding. Here, the string argument is the message to be printed in the case of failure, while the name of each test is taken from the name of the callable implementing the test. Thus here, `"Xor(248, 63) was incorrect."` only gets printed in the negative case, while `☑ XorIsCorrect` should be printed in the positive case.

src/Simulation/QSharpFoundation/Math/Math.qs
cgranade: Those should now be added, resolving. Thank you!

cgranade:
Since microsoft/QuantumLibraries#405 concerns the same part of the runtime, I also added an implementation of the `Microsoft.Quantum.Convert.DoubleAsInt` function from that proposal as a part of this PR. Thank you!

cgranade:
It looks like there's a build failure when making `PI` a pure Q# function, so I backed out that change for now and will defer to a future PR once @swernli and I have had a chance to look at why only `PI` has to be intrinsic. That should allow for the build to complete and unblock you, @guenp.

bettinaheim:
> Since [microsoft/QuantumLibraries#405](https://github.com/microsoft/QuantumLibraries/issues/405) concerns the same part of the runtime, I also added an implementation of the `Microsoft.Quantum.Convert.DoubleAsInt` function from that proposal as a part of this PR. Thank you!

I'll add support for that in the QIR emission with the next PR.

swernli:
@cgranade I filed https://github.com/microsoft/qsharp-runtime/issues/502 to capture the issue with making `PI` intrinsic, and I intend to fix it as part of refactoring the test projects.

PR #500, opened 2021-02-05
Added some math funcs implementation

### LINKED ISSUE #508: Replace the exclusive term
1. In [src/QirRuntime/test/QIR-static/qir-driver.cpp](src/QirRuntime/test/QIR-static/qir-driver.cpp) see comment for line 60 (`bool dummy);`).
2. See also [this](https://github.com/microsoft/qsharp-runtime/pull/500#pullrequestreview-586077737).

1 
> In general, style guide is to avoid the term "dummy," as that can be an exclusive term; something like "placeholder" or even "_" is strongly preferred.
> Our own style guide on this isn't public, unfortunately, but it's worth noting both that our code of conduct explicitly bans discriminatory language, and that the term "dummy" is widely considered to fall under that definition. In particular, both Google and Twitter's style guides exclude "dummy" explicitly.
To be clear, I don't object to the functionality of having a placeholder argument like this, but I would strongly suggest that name be replaced as soon as is reasonably possible.

2 "Ensure exclusive language is not used in codebase"

kuzminrobin:
The first step for #501 
Resolves #508

### DIFF
--- a/src/QirRuntime/lib/QIR/bridge-qis.ll
+++ b/src/QirRuntime/lib/QIR/bridge-qis.ll
@@ -281,3 +281,15 @@ define void @__quantum__qis__z__ctl(%Array* %.ctls, %Qubit* %.q) {
 }
 
 
+;===============================================================================
+; quantum.qis math functions
+;
+
+; LLVM intrinsics:
+; https://llvm.org/docs/LangRef.html
+declare double @llvm.sqrt.f64(double %.val)
+
+define double @__quantum__qis__sqrt__body(double %.val) {
+  %value = call double @llvm.sqrt.f64(double %.val)
+  ret double %value
+}
--- a/src/QirRuntime/lib/QIR/bridge-qis.ll
+++ b/src/QirRuntime/lib/QIR/bridge-qis.ll
@@ -287,7 +287,6 @@ define void @__quantum__qis__z__ctl(%Array* %.ctls, %Qubit* %.q) {
 
 ; LLVM intrinsics (https://llvm.org/docs/LangRef.html):
 declare double      @llvm.sqrt.f64(double %.val)
-;declare double      @llvm.ceil.f64(double %Val)
 
 ; Native implementations:
 declare i1          @quantum__qis__isnan__body(double %d)
@@ -319,10 +318,3 @@ define double @__quantum__qis__sqrt__body(double %d) {      ; https://en.cpprefe
   %result = call double @llvm.sqrt.f64(double %d)           
   ret double %result
 }
-
-;define i64 @__quantum__qis__ceiling__body(double %d) {
-;  %doubleResult = call double @llvm.ceil.f64(double %d)
-;  %result = call i64 @quantum__qis__doubleasint__body(double $doubleResult)  ; Should be the same as DoubleAsInt in https://github.com/microsoft/qsharp-runtime/pull/497/files
-;  ret i64 %result
-;}
-
--- a/src/QirRuntime/lib/QIR/bridge-qis.ll
+++ b/src/QirRuntime/lib/QIR/bridge-qis.ll
@@ -285,11 +285,44 @@ define void @__quantum__qis__z__ctl(%Array* %.ctls, %Qubit* %.q) {
 ; quantum.qis math functions
 ;
 
-; LLVM intrinsics:
-; https://llvm.org/docs/LangRef.html
-declare double @llvm.sqrt.f64(double %.val)
+; LLVM intrinsics (https://llvm.org/docs/LangRef.html):
+declare double      @llvm.sqrt.f64(double %.val)
+;declare double      @llvm.ceil.f64(double %Val)
 
-define double @__quantum__qis__sqrt__body(double %.val) {
-  %value = call double @llvm.sqrt.f64(double %.val)
-  ret double %value
+; Native implementations:
+declare i1          @quantum__qis__isnan__body(double %d)
+declare double      @quantum__qis__infinity__body()
+declare i1          @quantum__qis__isinf__body(double %d)
+
+; API for the user code:
+define double @__quantum__qis__nan__body() {                ; http://www.cplusplus.com/reference/cmath/nan-function/
+  %result = call double @llvm.sqrt.f64(double -1.0)         ; sqrt(<negative>) -> NaN
+  ret double %result
 }
+
+define i1 @__quantum__qis__isnan__body(double %d) {         ; http://www.cplusplus.com/reference/cmath/isnan/
+  %result = call i1 @quantum__qis__isnan__body(double %d)
+  ret i1 %result
+}
+
+define double @__quantum__qis__infinity__body() {           ; https://en.cppreference.com/w/c/numeric/math/INFINITY
+  %result = call double @quantum__qis__infinity__body()
+  ret double %result
+}
+
+define i1 @__quantum__qis__isinf__body(double %d) {         ; https://en.cppreference.com/w/cpp/numeric/math/isinf
+  %result = call i1 @quantum__qis__isinf__body(double %d)   
+  ret i1 %result
+}
+
+define double @__quantum__qis__sqrt__body(double %d) {      ; https://en.cppreference.com/w/cpp/numeric/math/sqrt
+  %result = call double @llvm.sqrt.f64(double %d)           
+  ret double %result
+}
+
+;define i64 @__quantum__qis__ceiling__body(double %d) {
+;  %doubleResult = call double @llvm.ceil.f64(double %d)
+;  %result = call i64 @quantum__qis__doubleasint__body(double $doubleResult)  ; Should be the same as DoubleAsInt in https://github.com/microsoft/qsharp-runtime/pull/497/files
+;  ret i64 %result
+;}
+
--- a/src/QirRuntime/lib/QIR/bridge-qis.ll
+++ b/src/QirRuntime/lib/QIR/bridge-qis.ll
@@ -13,7 +13,7 @@
 %Range = type { i64, i64, i64 }
 %Result = type opaque
 %String = type opaque
-%Pauli = type {i2}
+%Pauli = type i2
 
 ;=======================================================================================================================
 ; Native types
@@ -29,9 +29,7 @@
 %struct.QirCallable = type opaque
 %struct.QirRange = type { i64, i64, i64 }
 %struct.QirString = type opaque
-
-; Assumptions:
-;  %PauliId = type {i32}
+%PauliId = type i32
 
 ;===============================================================================
 ; declarations of the native methods this bridge delegates to
@@ -139,53 +137,53 @@ define %Result* @__quantum__qis__measure__body(%Array* %.paulis, %Array* %.qubit
   ret %Result* %.r
 }
 
-define void @__quantum__qis__r__body(i2 %.pauli, double %theta, %Qubit* %.q) {
+define void @__quantum__qis__r__body(%Pauli %.pauli, double %theta, %Qubit* %.q) {
   %q = bitcast %Qubit* %.q to %class.QUBIT*
-  %pauli = zext i2 %.pauli to i32
-  call void @quantum__qis__r__body(i32 %pauli, double %theta, %class.QUBIT* %q)
+  %pauli = zext %Pauli %.pauli to %PauliId
+  call void @quantum__qis__r__body(%PauliId %pauli, double %theta, %class.QUBIT* %q)
   ret void
 }
 
-define void @__quantum__qis__r__adj(i2 %.pauli, double %theta, %Qubit* %.q) {
+define void @__quantum__qis__r__adj(%Pauli %.pauli, double %theta, %Qubit* %.q) {
   %q = bitcast %Qubit* %.q to %class.QUBIT*
-  %pauli = zext i2 %.pauli to i32
-  call void @quantum__qis__r__adj(i32 %pauli, double %theta, %class.QUBIT* %q)
+  %pauli = zext %Pauli %.pauli to %PauliId
+  call void @quantum__qis__r__adj(%PauliId %pauli, double %theta, %class.QUBIT* %q)
   ret void
 }
 
-define void @__quantum__qis__r__ctl(%Array* %.ctls, {i2, double, %Qubit*}* %.args) {
+define void @__quantum__qis__r__ctl(%Array* %.ctls, {%Pauli, double, %Qubit*}* %.args) {
   %ctls = bitcast %Array* %.ctls to %struct.QirArray*
 
-  %.ppauli = getelementptr inbounds {i2, double, %Qubit*}, {i2, double, %Qubit*}* %.args, i32 0, i32 0
-  %.pauli = load i2, i2* %.ppauli
-  %pauli = zext i2 %.pauli to i32
+  %.ppauli = getelementptr inbounds {%Pauli, double, %Qubit*}, {%Pauli, double, %Qubit*}* %.args, i32 0, i32 0
+  %.pauli = load %Pauli, %Pauli* %.ppauli
+  %pauli = zext %Pauli %.pauli to %PauliId
 
-  %.ptheta = getelementptr inbounds {i2, double, %Qubit*}, {i2, double, %Qubit*}* %.args, i32 0, i32 1
+  %.ptheta = getelementptr inbounds {%Pauli, double, %Qubit*}, {%Pauli, double, %Qubit*}* %.args, i32 0, i32 1
   %theta = load double, double* %.ptheta
 
-  %.pq = getelementptr inbounds {i2, double, %Qubit*}, {i2, double, %Qubit*}* %.args, i32 0, i32 2
+  %.pq = getelementptr inbounds {%Pauli, double, %Qubit*}, {%Pauli, double, %Qubit*}* %.args, i32 0, i32 2
   %.q = load %Qubit*, %Qubit** %.pq
   %q = bitcast %Qubit* %.q to %class.QUBIT*
 
-  call void @quantum__qis__r__ctl(%struct.QirArray* %ctls, i32 %pauli, double %theta, %class.QUBIT* %q)
+  call void @quantum__qis__r__ctl(%struct.QirArray* %ctls, %PauliId %pauli, double %theta, %class.QUBIT* %q)
   ret void
 }
 
-define void @__quantum__qis__r__ctladj(%Array* %.ctls, {i2, double, %Qubit*}* %.args) {
+define void @__quantum__qis__r__ctladj(%Array* %.ctls, {%Pauli, double, %Qubit*}* %.args) {
   %ctls = bitcast %Array* %.ctls to %struct.QirArray*
 
-  %.ppauli = getelementptr inbounds {i2, double, %Qubit*}, {i2, double, %Qubit*}* %.args, i32 0, i32 0
-  %.pauli = load i2, i2* %.ppauli
-  %pauli = zext i2 %.pauli to i32
+  %.ppauli = getelementptr inbounds {%Pauli, double, %Qubit*}, {%Pauli, double, %Qubit*}* %.args, i32 0, i32 0
+  %.pauli = load %Pauli, %Pauli* %.ppauli
+  %pauli = zext %Pauli %.pauli to %PauliId
 
-  %.ptheta = getelementptr inbounds {i2, double, %Qubit*}, {i2, double, %Qubit*}* %.args, i32 0, i32 1
+  %.ptheta = getelementptr inbounds {%Pauli, double, %Qubit*}, {%Pauli, double, %Qubit*}* %.args, i32 0, i32 1
   %theta = load double, double* %.ptheta
 
-  %.pq = getelementptr inbounds {i2, double, %Qubit*}, {i2, double, %Qubit*}* %.args, i32 0, i32 2
+  %.pq = getelementptr inbounds {%Pauli, double, %Qubit*}, {%Pauli, double, %Qubit*}* %.args, i32 0, i32 2
   %.q = load %Qubit*, %Qubit** %.pq
   %q = bitcast %Qubit* %.q to %class.QUBIT*
 
-  call void @quantum__qis__r__ctladj(%struct.QirArray* %ctls, i32 %pauli, double %theta, %class.QUBIT* %q)
+  call void @quantum__qis__r__ctladj(%struct.QirArray* %ctls, %PauliId %pauli, double %theta, %class.QUBIT* %q)
   ret void
 }
 
--- a/src/QirRuntime/lib/QIR/bridge-qis.ll
+++ b/src/QirRuntime/lib/QIR/bridge-qis.ll
@@ -287,15 +287,17 @@ define void @__quantum__qis__z__ctl(%Array* %.ctls, %Qubit* %.q) {
 
 ; LLVM intrinsics (https://llvm.org/docs/LangRef.html):
 declare double      @llvm.sqrt.f64(double %.val)
+declare double      @llvm.log.f64(double %Val)
 
 ; Native implementations:
 declare i1          @quantum__qis__isnan__body(double %d)
 declare double      @quantum__qis__infinity__body()
 declare i1          @quantum__qis__isinf__body(double %d)
+declare double      @quantum__qis__arctan2__body(double %y, double %x)
 
 ; API for the user code:
-define double @__quantum__qis__nan__body() {                ; http://www.cplusplus.com/reference/cmath/nan-function/
-  %result = call double @llvm.sqrt.f64(double -1.0)         ; sqrt(<negative>) -> NaN
+define double @__quantum__qis__nan__body() {                ; Q#: function NAN() : Double       http://www.cplusplus.com/reference/cmath/nan-function/
+  %result = call double @llvm.sqrt.f64(double -1.0)         ; sqrt(<negative>) -> NaN   
   ret double %result
 }
 
@@ -316,5 +318,23 @@ define i1 @__quantum__qis__isinf__body(double %d) {         ; https://en.cpprefe
 
 define double @__quantum__qis__sqrt__body(double %d) {      ; https://en.cppreference.com/w/cpp/numeric/math/sqrt
   %result = call double @llvm.sqrt.f64(double %d)           
+  ret double %result
+}
+
+define double @__quantum__qis__log__body(double %d) {       ; https://en.cppreference.com/w/cpp/numeric/math/log
+  %result = call double @llvm.log.f64(double %d)           
+  ret double %result
+}
+
+define i1 @__quantum__qis__isnegativeinfinity__body(double %d) {    ; Q#: function IsNegativeInfinity(d : Double) : Bool
+                                                                    ; https://en.cppreference.com/w/cpp/numeric/math/log    https://llvm.org/docs/LangRef.html#llvm-log-intrinsic
+  %negInf = call double @llvm.log.f64(double 0.0)                   ; ln(0) -> (-infinity)
+  %result = fcmp oeq double %negInf, %d                             ; %result = (%negInf == %d)
+  ret i1 %result
+}
+
+define double @__quantum__qis__arctan2__body(double %y, double %x) {  ; Q#: function ArcTan2 (y : Double, x : Double) : Double
+                                                                    ; https://en.cppreference.com/w/cpp/numeric/math/atan2
+  %result = call double @quantum__qis__arctan2__body(double %y, double %x)
   ret double %result
 }
--- a/src/QirRuntime/lib/QIR/intrinsicsMath.cpp
+++ b/src/QirRuntime/lib/QIR/intrinsicsMath.cpp
@@ -23,4 +23,9 @@ bool quantum__qis__isinf__body(double d)
     return std::isinf(d);           // https://en.cppreference.com/w/cpp/numeric/math/isinf
 }
 
+double quantum__qis__arctan2__body(double y, double x)
+{
+    return std::atan2(y, x);        // https://en.cppreference.com/w/cpp/numeric/math/atan2
+}
+
 }   // extern "C"
--- a/src/QirRuntime/lib/QIR/intrinsicsMath.cpp
+++ b/src/QirRuntime/lib/QIR/intrinsicsMath.cpp
@@ -10,7 +10,7 @@ extern "C"
 // Implementations:
 bool quantum__qis__isnan__body(double d)
 {
-    return isnan(d);                // https://en.cppreference.com/w/cpp/numeric/math/isnan
+    return std::isnan(d);           // https://en.cppreference.com/w/cpp/numeric/math/isnan
 }
 
 double quantum__qis__infinity__body()
@@ -20,7 +20,7 @@ double quantum__qis__infinity__body()
 
 bool quantum__qis__isinf__body(double d)
 {
-    return isinf(d);                // https://en.cppreference.com/w/cpp/numeric/math/isinf
+    return std::isinf(d);           // https://en.cppreference.com/w/cpp/numeric/math/isinf
 }
 
 }   // extern "C"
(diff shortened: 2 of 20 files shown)


src/QirRuntime/lib/QIR/bridge-qis.ll
```diff
@@ -281,3 +281,15 @@ define void @__quantum__qis__z__ctl(%Array* %.ctls, %Qubit* %.q) {
 }
 
 
+;===============================================================================
+; quantum.qis math functions
```
irinayat-MS, line 285: As we discussed, it might make sense have a separate .ll file for math functions, but we can look into this later as it's basically a build structure refactor.

cgranade (commented):
This looks reasonable to me, but I'm not an LLVM expert per se so I'll leave it to others to approve. In any case, it looks like the tests are all at the C level, but it would be good to have integration tests similar to https://github.com/microsoft/qsharp-runtime/blob/0cf4a5e168e0ab9b9d6ecdb176e25548c67bea7f/src/QirRuntime/test/QIR-dynamic/qir-test-random-lnx.ll#L90 that ensure that the new intrinsic works correctly when called from QIR files generated by the Q# compiler.

src/QirRuntime/test/QIR-static/qir-test-math.cpp
```diff
@@ -0,0 +1,16 @@
+#include <cmath>
+
+#include "catch.hpp"
+
+extern "C" double __quantum__qis__sqrt__body(double);   // NOLINT
```
irinayat-MS, line 5: I'm not sure I like this direct testing of the bridge. While it has some value, I'd prefer to add a qs file that we'd use to generate QIR and then test that. This way we'll also check that our expectation of the generated signatures matches the compiler's output.

src/QirRuntime/lib/QIR/bridge-qis.ll
kuzminrobin: #505
(for now I would like to postpone that refactoring/maintenance)

src/QirRuntime/test/QIR-static/qir-test-qubits-results.qs
```diff
@@ -1,3 +1,4 @@
+//
```
kuzminrobin, line 1: Clarification: This change enables the (C/C++) syntax highlighting in my editor for this .qs file.
Let me know if you would like me to remove this change.

kuzminrobin:
> This looks reasonable to me, but I'm not an LLVM expert per se so I'll leave it to others to approve. In any case, it looks like the tests are all at the C level, but it would be good to have integration tests similar to
> 
> https://github.com/microsoft/qsharp-runtime/blob/0cf4a5e168e0ab9b9d6ecdb176e25548c67bea7f/src/QirRuntime/test/QIR-dynamic/qir-test-random-lnx.ll#L90
> 
> that ensure that the new intrinsic works correctly when called from QIR files generated by the Q# compiler.

Moved the tests to .qs file.

src/QirRuntime/test/QIR-static/qir-test-math.cpp
kuzminrobin: Moved the test to the .qs file.

src/QirRuntime/test/QIR-static/qir-test-arrays.qs
```diff
@@ -33,9 +35,9 @@ namespace Microsoft.Quantum.Testing.QIR
         if (dummy)
         {
             let res1 = TestControlled();
-            //Q# compiler crashes if both TestControlled and TestPartials are enabled
-            //let res2 = TestPartials(17, 42);
+            let res2 = TestPartials(17, 42);
```
kuzminrobin, line 38: FYI: For me the compiler didn't crash.

src/QirRuntime/lib/QIR/intrinsicsMath.cpp
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include <cmath>
+
+extern "C"
+{
+
+// Declarations:
+bool        quantum__qis__isnan__body(double d);
```
irinayat-MS, line 10: The declarations are unnecessary here. Maybe, extract into a header file so
a) it's easier to find them (people often start exploration of a codebase from the headers)
b) when/if more complicated implementations are added, the header can be included into the files that test them

src/QirRuntime/test/QIR-dynamic/CMakeLists.txt
```diff
@@ -36,6 +36,7 @@ add_executable(qir-dynamic-tests
 target_link_libraries(qir-dynamic-tests PUBLIC
   ${QIR_TESTS_LIBS}
   qdk
+  qir-qis-support
```
irinayat-MS, line 39: The idea of the "dynamic" tests was to validate that can link against the dynamic lib (qdk) without any knowledge about the static libs that go into it. If it's failing to link after your changes, it means that we aren't building the qdk component correctly. Ideally, should fix the build. Or log a bug and disable the component for now.

src/QirRuntime/test/QIR-static/CMakeLists.txt
```diff
@@ -20,7 +20,8 @@ add_custom_target(qir_static_test_lib DEPENDS ${QIR_TESTS_LIBS})
 # The executable target for QIR tests triggers the custom actions to compile ll files
 #
 add_executable(qir-static-tests
-  qir-driver.cpp)
+  qir-driver.cpp
+  qir-test-math.cpp)
```
irinayat-MS, line 24: Nit: maybe, move the closing parenthesis to a new line, so future additions look cleaner in diffs?

src/QirRuntime/test/QIR-static/compiler/QirTarget.qs
```diff
@@ -5,6 +5,26 @@ namespace Microsoft.Quantum.Intrinsic {
 
     open Microsoft.Quantum.Targeting;
 
+    function NAN() : Double {
```
irinayat-MS, line 8: Should these be marked `@Inline`? (Not a big deal, probably, because eventually we want to use compiler's target-specific qs files anyway.)

src/QirRuntime/test/QIR-static/qir-driver.cpp
```diff
@@ -58,7 +58,7 @@ extern "C" int64_t Microsoft__Quantum__Testing__QIR__Test_Arrays( // NOLINT
     int64_t index,
     int64_t val,
     bool dummy);
-TEST_CASE("QIR: Using 1D arrays", "[qir]")
+TEST_CASE("QIR: Using 1D arrays", "[qir][qir.arr1d]")
```
irinayat-MS, line 61: Will catch's filters still work with category specified like this?

src/QirRuntime/test/QIR-static/qir-test-arrays.qs
irinayat-MS: You are probably on a newer version of the compiler, which means whatever bug it was, it's fixed. Good :)

src/QirRuntime/test/QIR-static/qir-test-math.qs
```diff
@@ -0,0 +1,24 @@
...
+
+namespace Microsoft.Quantum.Testing.QIR.Math
+{
+    open Microsoft.Quantum.Intrinsic;
+
+    function TestSqrt() : Int
+    {
+        mutable testResult = 0;
+
+        if(           2.0 != Sqrt(  4.0) )    { set testResult = 1; }
```
irinayat-MS, line 12: You shouldn't need nested `ifs` here because you are not allocating qubits, early returns should compile just fine:

```
if (2.0 != Sqrt(  4.0) )    { return 1; }
if (3.0 != Sqrt(  9.0) )    { return 2; }
// etc.
```

src/QirRuntime/test/QIR-static/qir-test-qubits-results.qs
irinayat-MS: Would you mind adding the licensing boilerplate instead (all our source files are required to have it, but apparently I've missed this one)?

```
```

cgranade: As a slight nit, the most up to date version of the header does not have "All rights reserved." It's not essential to go fix in existing headers, but is recommended for all new headers.

src/QirRuntime/test/QIR-static/qir-test-math.qs
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.Quantum.Testing.QIR.Math
+{
+    open Microsoft.Quantum.Intrinsic;
+
+    function TestSqrt() : Int
+    {
```
cgranade, line 9: Q# style is to have braces in the same line, as opposed to C#'s next-line style. I'd also suggest that the name "TestSqrt" sounds like a verb, whereas Q# style is for functions to be named as nouns.

src/QirRuntime/test/QIR-static/qir-test-math.qs
cgranade: That would also have the beneficial effect of eliminating the `mutable` variable.

src/QirRuntime/test/QIR-static/qir-driver.cpp
```diff
@@ -58,7 +58,7 @@ extern "C" int64_t Microsoft__Quantum__Testing__QIR__Test_Arrays( // NOLINT
     int64_t index,
     int64_t val,
     bool dummy);
-TEST_CASE("QIR: Using 1D arrays", "[qir]")
```
cgranade, line 61: In general, style guide is to avoid the term "dummy," as that can be an exclusive term; something like "placeholder" or even "_" is strongly preferred.

src/QirRuntime/test/QIR-static/qir-test-arrays.qs
```diff
@@ -33,9 +35,9 @@ namespace Microsoft.Quantum.Testing.QIR
         if (dummy)
         {
```
cgranade, line 36: Similarly here.

src/QirRuntime/lib/QIR/bridge-qis.ll
```diff
@@ -281,3 +281,48 @@ define void @__quantum__qis__z__ctl(%Array* %.ctls, %Qubit* %.q) {
 }
 
 
+;===============================================================================
+; quantum.qis math functions
+;
+
+; LLVM intrinsics (https://llvm.org/docs/LangRef.html):
+declare double      @llvm.sqrt.f64(double %.val)
+;declare double      @llvm.ceil.f64(double %Val)
```
guenp, line 290: Can we drop this from the PR or is this commented to be added back in later? If so, would you mind adding a note on that with optionally a link to a GH ticket?

src/QirRuntime/lib/QIR/bridge-qis.ll
```diff
@@ -281,3 +281,48 @@ define void @__quantum__qis__z__ctl(%Array* %.ctls, %Qubit* %.q) {
...
+  %result = call i1 @quantum__qis__isinf__body(double %d)   
+  ret i1 %result
+}
+
+define double @__quantum__qis__sqrt__body(double %d) {      ; https://en.cppreference.com/w/cpp/numeric/math/sqrt
+  %result = call double @llvm.sqrt.f64(double %d)           
+  ret double %result
+}
+
+;define i64 @__quantum__qis__ceiling__body(double %d) {
```
guenp, line 323: Same comment here; can we drop this and if not, could we add a note and GH issue #?

guenp (commented):
Just some comments; thanks for implementing this!

src/QirRuntime/test/QIR-static/qir-driver.cpp
kuzminrobin: For me it does work like this.
```
<test.exe> 
<test.exe> [qir]
<test.exe> [qir.arr1d]
```
And I had a test failure that was showing up on exactly one testcase (coincidentally exactly this one ;-) ). So I wanted to fix-and-run multiple times one particular testcase. So I added this finer granularity.

src/QirRuntime/test/QIR-dynamic/CMakeLists.txt
irinayat-MS: as discussed offline: add the export decls on the new APIs

src/QirRuntime/lib/QIR/bridge-qis.ll
kuzminrobin: This is work in progress to be added in the next commit/PR.
(I was working on it when I got the first feedback about Sqrt, so I postponed Ceil until I resolve all for Sqrt)

src/QirRuntime/lib/QIR/bridge-qis.ll
kuzminrobin: This is work in progress to be added in the next commit/PR.
(I was working on it when I got the first feedback about Sqrt, so I postponed Ceil until I resolve all for Sqrt)

src/QirRuntime/test/QIR-static/qir-driver.cpp
irinayat-MS: FYI, to run just a single test, you can use its description string as the filter

src/QirRuntime/lib/QIR/bridge-qis.ll
cgranade: Ceiling shouldn't be needed at all post-#497, though? Ceiling, Floor, and Round are all expressed in terms of Truncate, such that that's the only intrinsic needed to support those.

src/QirRuntime/test/QIR-static/compiler/QirTarget.qs
```diff
@@ -5,6 +5,26 @@ namespace Microsoft.Quantum.Intrinsic {
...
+        body intrinsic;
+    }
+
+    function IsNan(d: Double) : Bool {
```
cgranade, line 12: Are these intended to be new functions in the public Q# API? If so, they will need to go through API review first.

src/QirRuntime/test/QIR-static/qir-test-qubits-results.qs
kuzminrobin: Fixed.

src/QirRuntime/lib/QIR/intrinsicsMath.cpp
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
cgranade, line 1:
```suggestion
// Copyright (c) Microsoft Corporation.
```

src/QirRuntime/test/QIR-static/qir-driver.cpp
kuzminrobin: I would prefer to refrain dealing with this case of "dummy" in this PR since "dummy" is out of this PR's scope. This case of "dummy" has important functionality behind it (not knowing which can result in a loss of a few hours for a new-comer), and if I touch this "dummy" I feel obliged to document in details that important functionality.

However I'm inclined to agree about the use of "dummy" in general, and would like to know more about it and other similar identifiers/names. Any direct link about more info? (preferably reachable from [Coding Style](https://ms-quantum.visualstudio.com/Quantum%20Program/_wiki/wikis/Quantum-OKRs.wiki/487/Coding-Style))

src/QirRuntime/test/QIR-static/qir-test-arrays.qs
kuzminrobin: I will focus on _my_ changes for now.
However I will keep your code style recommendations in mind for the lines/functions/operations I touch. Thank you for sharing your knowledge. Please keep sharing, I appreciate that.

src/QirRuntime/lib/QIR/bridge-qis.ll
kuzminrobin: Thanks for letting me know on time, Chris. Removing Ceil from my plate, adding Truncate.

src/QirRuntime/test/QIR-static/compiler/QirTarget.qs
kuzminrobin: (FYI: @bettinaheim, @irinayat-MS, @guenp)
Bettina,
What do I need to do to get those through API review?

I assume that Q#'s `Sqrt()` (inspired by C#'s [Math.Sqrt](https://docs.microsoft.com/en-us/dotnet/api/system.math.sqrt?view=net-5.0)) is expected to work with `NaN` and `PositiveInfinity` also (i.e. the Q#'s `Sqrt()` requires them in Q#). Thus it seems to me reasonable for the functionality of NaN, PositiveInfinity/Infinity, `IsNan()`, `IsInf()` to _become new functions in the public Q# API_ (in simulators can be implemented through the following C++'s <cmath> calls/constants - [`NAN`/`nan()`/`quiet_NaN()`](https://en.cppreference.com/w/cpp/numeric/math/NAN), [`INFINITY`/`infinity`](https://en.cppreference.com/w/cpp/numeric/math/INFINITY), [`isnan()`](https://en.cppreference.com/w/cpp/numeric/math/isnan), [`isinf()`](https://en.cppreference.com/w/cpp/numeric/math/isinf)).

Chris, thanks!

src/QirRuntime/lib/QIR/intrinsicsMath.cpp
kuzminrobin: Done (including for my files added to tests).
Where can I get more info about this? (to know in the future what is recommended or not recommended)

src/QirRuntime/lib/QIR/intrinsicsMath.cpp
kuzminrobin: Fixed.

src/QirRuntime/test/QIR-dynamic/CMakeLists.txt
kuzminrobin: Fixed.

src/QirRuntime/test/QIR-static/CMakeLists.txt
kuzminrobin: Fixed.

src/QirRuntime/test/QIR-static/compiler/QirTarget.qs
kuzminrobin: Fixed.

src/QirRuntime/test/QIR-static/qir-test-math.qs
kuzminrobin: Fixed.

src/QirRuntime/test/QIR-static/qir-test-math.qs
kuzminrobin: Fixed.

src/QirRuntime/test/QIR-static/compiler/QirTarget.qs
cgranade: Please see the [Q# API Design Process](https://github.com/microsoft/QuantumLibraries/wiki/Q%23-API-Design-Process) documentation on the Q# libraries wiki for more details, but roughly the Q# API Review is a monthly meeting that I run with the Q# API review board (currently @guenp, @msoeken, @bettinaheim, and @efratshabtai) to make sure that all new or modified public Q# APIs meet our [style guide](https://docs.microsoft.com/quantum/resources/contributing/style-guide) and our [API design principles](https://docs.microsoft.com/azure/quantum/contributing-api-design-principles).

Following that process, if we want to add these functions to the public Q# API, my suggestion would be to open a proposal on the [microsoft/quantumlibraries repo](https://github.com/microsoft/QuantumLibraries), so that it can be scheduled for inclusion in the March review. Should the Q# API review approve the new API, and should @bettinaheim sign off on the corresponding impact on the runtime interface, we can then implement those new functions as `body intrinsic` callables in the QSharpFoundation project on the qsharp-runtime repo, with the changes here in QirTarget.qs and bridge-qis.ll being used to implement those new functions.

kuzminrobin: See [413](https://github.com/microsoft/QuantumLibraries/issues/413) in QuantumLibraries.
I propose those for the Q# API Design Review.
@cgranade, @bettinaheim.

src/QirRuntime/lib/QIR/intrinsicsMath.cpp
```diff
@@ -0,0 +1,31 @@
...
+}
+
+bool quantum__qis__isinf__body(double d)
+{
+    return std::isinf(d);           // https://en.cppreference.com/w/cpp/numeric/math/isinf
+}
+
+double quantum__qis__arctan2__body(double y, double x)
+{
+    return std::atan2(y, x);        // https://en.cppreference.com/w/cpp/numeric/math/atan2
```
cgranade, line 28: As per your note below, does this introduce a discrepancy with existing semantics for Q#'s `ArcTan2` function?

src/QirRuntime/test/QIR-static/qir-driver.cpp
irinayat-MS: What does the negative filter do? (test.exe ~[qir]) Does it still filter our the whole category?

irinayat-MS (approved):
Looks good to me on the QirRuntime's side of things. Signing off modulo Chris' comments on the semantics of arctan2.

src/QirRuntime/test/QIR-static/compiler/Constants.qs
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+// these are all the static methods and const fields  form System.Math class of .NET CLR
+// that are not exposed as language operators and are relevant within type System.
+// If there are two versions of the function for Int and Double types, the corresponding
+// functions have suffix I or D. ExpD also has a suffix to avoid name clash with Primitives.Exp.
+
+namespace Microsoft.Quantum.Math {
```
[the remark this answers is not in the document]
kuzminrobin: Yes. That's how I interpreted the offline response.
I wrote:
> I want to add a test case to the .qs file in qsharp-runtime/src/QirRuntime/test/QIR-static and I want my test case to access the E() constant defined in Microsoft.Quantum.Math namespace in qsharp-runtime\src\Simulation\QSharpFoundation\Math\Constants.qs.
However, in order to generate the .ll files from .qs tests, the qsharp-runtime/src/QirRuntime/test/QIR-static/generate.py does not use libraries of the Q# compiler, it uses just a minimal set of a few .qs files (in qsharp-runtime/src/QirRuntime/test/QIR-static/compiler).
What is the best/right way to build the .qs tests that use the Q# libraries?

The reply was:
> Assuming you are asking for the purpose of testing the functions implemented in the qir runtime (opposed to setting up integration tests), you could simple compile [the file where it is declared](https://github.com/microsoft/qsharp-runtime/blob/main/src/Simulation/QSharpFoundation/Math/Constants.qs) along with the .qs file for your test.

src/QirRuntime/test/QIR-static/qir-test-math.qs
```diff
@@ -0,0 +1,64 @@
...
+
+        // Fails because of lack of precision:
+        // if(  PI()/6.0  != ArcTan2( 1.0, Sqrt(3.0) ) )     { return 10; }      // tg(Pi/6) = sin(Pi/6) / cos(Pi/6) = (1/2) / (Sqrt(3)/2) = 1/Sqrt(3) = y/x.  ArcTan2(1.0, Sqrt(3)) -> Pi/6
+
+        if( not IsNan(ArcTan2(             NAN(),                0.0) )) { return 11; }
+        if( not IsNan(ArcTan2(               0.0,              NAN()) )) { return 12; }
+        if( not IsNan(ArcTan2(             NAN(),              NAN()) )) { return 13; }
+
+        // The infinity cases show discrepancy between  
+        // https://docs.microsoft.com/en-us/dotnet/api/system.math.atan2?view=net-5.0
```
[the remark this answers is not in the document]
kuzminrobin: Looks like the [C#'s (.NET's) semantics](https://docs.microsoft.com/en-us/dotnet/api/system.math.atan2?view=net-5.0) somewhat contradicts to itself:
On the one hand it says: 
> if x and y are either PositiveInfinity or NegativeInfinity, the method returns NaN

On the other hand it remarks:
> This method calls into the underlying C runtime, and the exact result or valid input range may differ between different operating systems or architectures.

The [C](https://en.cppreference.com/w/c/numeric/math/atan2) (and [C++](https://en.cppreference.com/w/cpp/numeric/math/atan2)) runtime returns NaN in one case only - "If either x is NaN or y is NaN".

For now I suspect that the phrase 
> if x and y are either PositiveInfinity or NegativeInfinity, the method returns NaN

is out-of-date even for C# (and for the whole .NET), and even in C# the behavior corresponds to the C (and C++) runtime.
So for now I would leave the implementation and documentation for Q# as is (since Guen is blocked by this missing functionality), and get back to the discussion later if needed.

I filed a documentation issue [77](https://github.com/microsoft/qsharp-language/issues/77) in qsharp-language.

src/QirRuntime/lib/QIR/intrinsicsMath.cpp
kuzminrobin: See [this my reply](https://github.com/microsoft/qsharp-runtime/pull/500#discussion_r572467750).

src/QirRuntime/test/QIR-static/qir-test-math.qs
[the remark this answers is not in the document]
cgranade: Confirming, it looks like the current behavior of Q#'s `ArcTan2` follows what I would expect from the cppreference docs that you shared:

image

This would be good to test for explicitly in the C#-based runtime, but given that the two implementations seem to agree, I agree that it makes sense to defer adding those tests until the next PR.

src/QirRuntime/test/QIR-static/qir-driver.cpp
kuzminrobin: Yes, it filters our the whole [qir] category.

src/QirRuntime/test/QIR-static/qir-test-math.qs
```diff
@@ -0,0 +1,64 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.Quantum.Testing.QIR.Math
+{
```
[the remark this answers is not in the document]
kuzminrobin: Fixed.

src/QirRuntime/test/QIR-static/qir-test-math.qs
```diff
@@ -0,0 +1,64 @@
...
+    open Microsoft.Quantum.Intrinsic;
+    open Microsoft.Quantum.Math;        // E()
+
+    function SqrtTest() : Int {
+        if( 2.0 != Sqrt(  4.0) )    { return 1; }
```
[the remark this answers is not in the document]
kuzminrobin: Yes. it indicates which test failed.
Added a comment.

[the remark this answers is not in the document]
kuzminrobin: Fixed.

src/QirRuntime/test/QIR-static/qir-test-math.qs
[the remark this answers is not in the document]
kuzminrobin: Thanks for confirming! Resolving this conversation for now (in this PR).

src/QirRuntime/test/QIR-static/qir-test-strings.qs
```diff
@@ -0,0 +1,30 @@
...
+
+namespace Microsoft.Quantum.Testing.QIR.Str {
+
+    open Microsoft.Quantum.Intrinsic;
+
+    function PauliToStringTest() : Int {
+
+        mutable pauli   = PauliI;
+        mutable str     = $"Pauli value: {pauli}";
+        if "Pauli value: PauliI" != str      { return 1; }     // The return value indicates which test case has failed.
```
cgranade, line 12: You may be able to avoid mutability here, since `pauli` and `str` are each only used once:

```suggestion
        if "Pauli: PauliI" != $"Pauli: {PauliI}" { return 1; }     // The return value indicates which test case has failed.
```

cgranade (approved):
Approving to unblock @guenp, but we should as quickly as possible:

- [x] Ensure exclusive language is not used in codebase
- [ ] Extend QSharpFoundation unit tests to cover what you noticed about `ArcTan2`, so that we can prevent regressions that would cause QIR and C# runtimes to disagree

kuzminrobin:
> Approving to unblock @guenp, but we should as quickly as possible:
> 
> * [x]  Ensure exclusive language is not used in codebase
> * [x]  Extend QSharpFoundation unit tests to cover what you noticed about `ArcTan2`, so that we can prevent regressions that would cause QIR and C# runtimes to disagree

Filed #508.
Added to #506:
> Extend QSharpFoundation unit tests to cover what has been noticed about ArcTan2, so that we can prevent regressions that would cause QIR and C# runtimes to disagree

src/QirRuntime/test/QIR-static/qir-test-strings.qs
kuzminrobin: Fixed.

src/QirRuntime/lib/QIR/bridge-qis.ll
```diff
@@ -13,7 +13,7 @@
 %Range = type { i64, i64, i64 }
 %Result = type opaque
 %String = type opaque
-%Pauli = type {i2}
+%Pauli = type i2
```
irinayat-MS, line 16: Nit: Notice, that generated QIR doesn't actually use this type but uses i2 directly. My choice of not introducing type aliases was to match the generated code verbatim, in hope that it will be easier to find bugs like the wrong parameter in string creation. At some point I even had a script that attempted to automatically extract declared signatures from generated QIR and compare them to defined here but I couldn't get it reliable enough...

src/QirRuntime/lib/QIR/strings.cpp
```diff
@@ -135,7 +135,10 @@ extern "C"
             return quantum__rt__string_create("PauliY");
         case PauliId_Z:
             return quantum__rt__string_create("PauliZ");
+        default:
+            break;
         }
+        return quantum__rt__string_create("<Unexpected Pauli Value>");
```
irinayat-MS, line 141: What does the QIR spec says for this? Shouldn't the call fail rather than return a random string?

kuzminrobin: My impression is that the [QIR Spec](https://github.com/microsoft/qsharp-language/blob/main/Specifications/QIR/Data-Types.md) does not cover that case (I searched through all the occurrences of "pauli").
Thus I needed to invent something.

src/QirRuntime/lib/QIR/bridge-qis.ll
kuzminrobin: OK, I'll keep that in mind.

I'll admit that now the types `i2` and `%Pauli` are the same (`%Pauli` being an alias/typedef of `i2`). Before this change, the were, strictly speaking, _different types_ (`%Pauli` was a struct with one field of type `i2`), and casting to `i32` was causing compilation error.

src/QirRuntime/lib/QIR/strings.cpp
irinayat-MS: Log a bug against the spec? Imho, it should cover error cases like this.

src/QirRuntime/lib/QIR/bridge-qis.ll
irinayat-MS: This is probably how %Pauli was defined in the earlier versions of generated QIR when I copied it over into the bridge. But because it was never actually used, the outdated type didn't come to my attention.

PR #503, opened 2021-02-05
Make targeted exe tests independent

### LINKED ISSUE #502: UnitTests.csproj should be split into separate projects by execution target
Currently the [UnitTests.csproj](https://github.com/microsoft/qsharp-runtime/blob/main/src/Simulation/Simulators.Tests/TestProjects/UnitTests/UnitTests.csproj) includes dependent projects that have different execution targets, namely HoneywellExe.csproj, IonQExe.csproj, and QCIExe.csproj. Pulling in projects with different execution targets as part of a single compilation can cause failures due to collisions, as we found when updating `body instrinsic` math functions to be concrete Q# implementations (see https://github.com/microsoft/qsharp-runtime/pull/497#issuecomment-773728087). To resolve this, these projects should be split out into independent unit tests so that it's always zero or one execution target in any given compilation.

swernli:
This separates the unit tests that use execution targets into their own projects to avoid the issues that come from multiple generated C# projects being compiled together.

Fixes #502


src/Simulation/Simulators.Tests/TestProjects/HoneywellExe/HoneywellExe.csproj
```diff
@@ -22,4 +22,11 @@
     </ItemGroup>
```
ScottCarda-MS, line 22: Should these Q# projects have their name changed so they don't have "Exe" in them, considering their output type is "Library"?

ScottCarda-MS (approved):
lgmt, just had a question about the naming of the Q# projects.

src/Simulation/Simulators.Tests/TestProjects/HoneywellExe/HoneywellExe.csproj
swernli: That's a good question... they were originally Exe projects that got changed to Library so they could all be linked together, but now that they aren't linked together they can be changed back to Exe.

ScottCarda-MS: Actually, if you want them to output exe, I think you have to specify `<OutputType>Exe</OutputType>`. I think without this, it defaults to the 'library' behavior.

swernli: Oh, whoops... I thought it was the other way around. Fixing now...

PR #507, opened 2021-02-09
Add C# generation and runtime support for "[value, size = n]" array constructor

samarsha:
This adds C# generation and runtime support for microsoft/qsharp-compiler#856.


src/Simulation/CSharpGeneration/SimulationCode.fs
```diff
@@ -591,6 +592,9 @@ module SimulationCode =
 // TODO: diagnostics.
             | _ -> failwith ""
 
+        and buildSizedArray value size =
+            ident "QArray" <.> (ident "Repeat", [ buildExpression value; buildExpression size ])
```
bettinaheim, line 596: I think this may cause issues; if the given item to initialize the array with is an array itself, then you would need to formally copy it n times (meaning the reference count needs to be updated to reflect its use in all places).

osostest:
Test

osostest (commented):
Test

src/Simulation/CSharpGeneration/SimulationCode.fs
samarsha: You're right that I needed to increment the ref count, but I think it's sound if the entire array owns one reference to the value. Since array indices are immutable, I don't think there is a way to mutate `xs[x][y]` in place. For example,

```qsharp
mutable items = [[1], size = 2];
set items w/= 0 <- items[0] w/ 0 <- 2;
```

will make a copy of `items[0]` before assigning it in-place to index 0 of `items`. Can you think of an example where this isn't true?

samarsha: Discussed offline: overwriting an array index can decrement the reference count, causing references in other indices to have reference count 0, so the count needs to be incremented initially for each index.

samarsha:
@bettinaheim @cesarzc @ScottCarda-MS @swernli ping

src/Simulation/Simulators.Tests/CoreTests.cs
```diff
@@ -324,5 +324,11 @@ public void CatchFail()
         [Fact]
         public void InternalCallables() =>
             OperationsTestHelper.RunWithMultipleSimulators(s => Circuits.InternalCallablesTest.Run(s).Wait());
+
+        [Fact]
+        public void CreateArrayWithNegativeSize() =>
+            Assert.ThrowsAny<Exception>(() =>
```
swernli, line 330: nit: I usually like to have the specific exception type in here rather than the base class, but that's not a big deal.

src/Simulation/Simulators.Tests/Circuits/Arrays.qs
```diff
@@ -0,0 +1,77 @@
...
+        mutable result = new 'U[Length(source)];
+        for (i in 0 .. Length(source) - 1) {
+            let m = mapper(source[i]);
+            set result = result w/ i <- m;
+        }
+        
+        return result;
+    }
+
+    operation LengthTest() : Unit {
```
swernli, line 18: Is `LengthTest` used anywhere?

swernli: Oh, nvm. I see it was originally in the removed file and has been updated to use the new syntax here. Cool!

swernli (approved):
LGTM!

src/Simulation/Simulators.Tests/CoreTests.cs
samarsha: I wanted to use the specific exception type too, but bug #536 causes the wrong exception type to be thrown, and I didn't want to depend on the buggy behavior in this test case.

samarsha: Added a comment to the test about the exception type bug.

Issue #501, opened 2021-02-05
Implement bridge and native backing for math functions Q# depends on


kuzminrobin:
Implement the following math functions
```
function ArcCos (x : Double) : Double   arccos__body
function ArcSin (y : Double) : Double   arcsin__body
function ArcTan (d : Double) : Double   arctan__body
function ArcTan2 (y : Double, x : Double) : 
                        Double          arctan2__body
function Cos (theta : Double) : Double  cos__body
function Cosh (d : Double) : Double     cosh__body
function Sin (d : Double) : Double      sin__body
function Sinh (d : Double) : Double     sinh__body
function Tan (d : Double) : Double      tan__body
function Tanh (d : Double) : Double     tanh__body
function Ceiling (value : Double) : Int ceiling__body
function Floor (value : Double) : Int   floor__body
function DivRemL(dividend : BigInt, divisor : BigInt) : 
                    (BigInt, BigInt)    divreml__body
function IEEERemainder (x : Double, y : Double) : Double    
                                        ieeeremainder_body
function Log (input : Double) : Double  log__body
function Sqrt (d : Double) : Double     sqrt__body
```
and other functions in the following requested order:
```
__quantum__qis__sqrt__body
__quantum__qis__ceiling__body
__quantum__qis__log__body
__quantum__qis__powd__body
__quantum__qis__arctan2__body
__quantum__qis__mini__body
__quantum__qis__absi__body
__quantum__qis__absd__body

__quantum__rt__int_power
__quantum__rt__pauli_to_string

__quantum__qis__drawrandomint__body
__quantum__qis__message__body
```
Refer the internal work item 25149 for more details.

cgranade:
Following https://github.com/microsoft/qsharp-runtime/pull/497, a few of those functions will no longer be intrinsic. In particular, `Ceiling`, `Floor`, `MinI`, `AbsI`, and `AbsD` should have pure Q# implementations such that providing explicit implementations for those should no longer be needed once #497 completes. The per-repo build completed, but that PR is currently awaiting an end-to-end build.

kuzminrobin:
Implementing 
```
function DivRemL(dividend : BigInt, divisor : BigInt) : 
                    (BigInt, BigInt)    divreml__body
```
has been postponed until BigInt is imeplemented. See #525.

Issue #482, opened 2021-01-27
QirRuntime coding style improvement. #include


kuzminrobin:
https://github.com/microsoft/qsharp-runtime/pull/479/files#r564862295
```c++
#include <assert.h>
```
> Typically the C++ files do not include the C standard headers _directly_ (`#include <assert.h>`) but instead the C++ files include the C++-wrappers of those (`#include <cassert>`).

bettinaheim:
Related/similar: https://github.com/microsoft/qsharp-runtime/issues/484

PR #511, opened 2021-02-09
Added DrawRandomInt() and Message()

kuzminrobin:
Next step for #501


src/QirRuntime/test/QIR-static/qir-test-output.qs
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.Quantum.Testing.QIR.Out {
+
+    open Microsoft.Quantum.Intrinsic;
+
+    function MessageTest(msg: String) : Unit
+    {
```
cgranade, line 9:
```suggestion
    function MessageTest(msg : String) : Unit {
```

src/QirRuntime/test/QIR-static/compiler/QirTarget.qs
```diff
@@ -101,4 +105,8 @@ namespace Microsoft.Quantum.Intrinsic {
             return Measure([PauliZ], [qb]);
         }
     }
+
+    function Message (msg : String) : Unit {
```
cgranade, line 109:
```suggestion
    function Message(msg : String) : Unit {
```

src/QirRuntime/test/QIR-static/qir-test-math.qs
```diff
@@ -60,5 +60,9 @@ namespace Microsoft.Quantum.Testing.QIR.Math {
         return 0;
     }
 
+    operation DrawRandomIntTest(min : Int, max : Int) : Int {
+        return DrawRandomInt(min, max);
+    }
```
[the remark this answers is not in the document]
kuzminrobin: Thanks for catching `DrawRandomIntTest` name mismatch. Fixed.

:-) The motivation is that _I'm a newbie in Q#_ and 
(a) I'm not sure if logic of that complexity is implementable with the current means available in Q#, and 
(b) it will take me a few times more time to write that in Q# rather than in C++.
Having said that, and keeping in mind that @irinayat-MS is waiting for my CR of her PR (and I have a number of other items on my plate) I decided to do that in C++. Let me know if that needs to be re-written in Q#.

src/QirRuntime/test/QIR-static/qir-test-math.cpp
```diff
@@ -23,3 +27,52 @@ TEST_CASE("QIR: Math.ArcTan2", "[qir.math][qir.Math.ArcTan2]")
...
+
+        // Increment the counter for this number:
+        const int64_t index = randomNumber - minimum;
+        REQUIRE(0 <= index);
+        REQUIRE(index < counters.size());
+        ++(counters[index]);
+    }
+
+    // Make sure that each number has been generated, and the distribution is approximately uniform:
+    for(auto counter : counters)
```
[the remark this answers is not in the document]
kuzminrobin: ;-) No, I don't know with what probability this is expected to fail, really. I was studying the probability theory about 25 years ago, and math was not my major (SW dev was the major). I'll take a deeper look into your formulas after I commit fixes (and get the build process running for this PR), and am done with Irina's CR.  
If you have good links for me to look at (in order to refresh my memory in this particular part of probability) then please feel free to provide.

src/QirRuntime/test/QIR-static/qir-test-output.qs
kuzminrobin: Thanks for catching. Sorry for the repeated flaw. I'll try to be more attentive.

src/QirRuntime/test/QIR-static/compiler/QirTarget.qs
kuzminrobin: :-) thanks.

src/QirRuntime/test/QIR-static/qir-test-math.cpp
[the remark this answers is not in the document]
samarsha: I think we should avoid adding any tests that have a nonzero probability of failure. To test random distributions, it would make more sense to me to use a fixed random seed. Even if the probability of failure is very low, we run a lot of CI builds, and having nondeterministic CI failures is not fun. :)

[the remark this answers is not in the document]
kuzminrobin: Oh... Thank you for the insight, @samarsha.
As I understand, at least for the test purposes, the sequence of the numbers generated by the DrawRandomInt() should be 100% predictable (fully controlled/customized pseudo-random). And the test should compare the actual sequence with the expected one.
Otherwise, the intermittent test run failures will be bothering the CI/PR builds/runs for the unrelated changes.
I will push my current changes (to make sure the build is fixed and earlier feedback is addressed), and then will fix this test. For now I will comment out the logic that can fail intermittently.

src/QirRuntime/lib/QIR/bridge-qis.ll
```diff
@@ -336,3 +347,8 @@ define double @__quantum__qis__arctan2__body(double %y, double %x) {  ; Q#: func
   %result = call double @quantum__qis__arctan2__body(double %y, double %x)
   ret double %result
 }
+
+define i64 @__quantum__qis__drawrandomint__body(i64 %min, i64 %max) {  ; operation DrawRandomInt (min : Int, max : Int) : Int      https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.random.drawrandomint
```
irinayat-MS, line 351: Nit: maybe, put the comment above and split across two lines? (we are trying to keep the files in this project under 120 char wide)

src/QirRuntime/lib/QIR/intrinsicsMath.cpp
```diff
@@ -28,4 +33,19 @@ double quantum__qis__arctan2__body(double y, double x)
     return std::atan2(y, x);        // https://en.cppreference.com/w/cpp/numeric/math/atan2
 }
 
+int64_t quantum__qis__drawrandomint__body(int64_t minimum, int64_t maximum)
+{
+    if(minimum > maximum)
+    {
+        throw std::invalid_argument( excStrDrawRandomInt );
```
irinayat-MS, line 40: We should probably go through quantum__rt__fail here as the bad arguments would indicate either user or QIR generation error.

Formatting nit: we are not using spaces inside parenthesis of function calls.

src/QirRuntime/lib/QIR/intrinsicsOut.cpp
```diff
@@ -0,0 +1,15 @@
...
+#include <iostream>
+
+#include "qirTypes.hpp"
+#include "quantum__qis.hpp"
+
+extern "C"
+{
+    void quantum__qis__message__body(QirString* qstr)   // NOLINT
+    {
+        std::cout << qstr->str << std::endl;
```
irinayat-MS, line 13: Dumping to cout is OK short term, but going forward it might not be sufficient and should probably be configurable through QirExecutionContext. Check with Cesar re requirements for output in Azure environment (IQ# also might have special needs here).

src/QirRuntime/lib/QIR/quantum__qis_internal.hpp
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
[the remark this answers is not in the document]
kuzminrobin: Fixed.

src/QirRuntime/test/QIR-static/qir-test-output.qs
kuzminrobin: Fixed.

src/QirRuntime/test/QIR-static/compiler/QirTarget.qs
kuzminrobin: Fixed.

src/QirRuntime/test/QIR-static/qir-driver.cpp
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
+// Copyright (c) Microsoft Corporation.
```
irinayat-MS, line 1: Nit: It would be better to updated the licensing comments in one PR across the repo that doesn't have any meaningful changes...

src/QirRuntime/test/QIR-static/qir-test-math.cpp
```diff
@@ -23,3 +27,56 @@ TEST_CASE("QIR: Math.ArcTan2", "[qir.math][qir.Math.ArcTan2]")
...
+    while(--times)
+    {
+        const int64_t randomNumber = 
+            Microsoft__Quantum__Testing__QIR__Math__TestDrawRandomInt__body(minimum, maximum);
+        REQUIRE(minimum <= randomNumber);
+        REQUIRE(randomNumber <= maximum);
+
+        // Increment the counter for this number:
+        const int64_t index = randomNumber - minimum;
+        REQUIRE(0 <= index);
```
irinayat-MS, line 56: Isn't it the same as `REQUIRE(minimum <= randomNumber)` above?

src/QirRuntime/test/QIR-static/qir-test-math.cpp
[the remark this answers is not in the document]
irinayat-MS: I am confused, are you saying that repeated runs of a Q# program that calls DrawRandomInt() would produce the same sequence of integers?

src/QirRuntime/test/QIR-static/qir-test-ouput.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+#include "catch.hpp"
+
+#include "qirTypes.hpp"
+
+extern "C" void Microsoft__Quantum__Testing__QIR__Out__MessageTest__body(QirString*);        // NOLINT
+
+
+// https://stackoverflow.com/a/5419388/6362941
+struct CoutRedirector
```
irinayat-MS, line 14: Yeah, testability is another reason why implementing qis__message as std::cout might not be the right way to go. For example, if the restore of the buffer redirect fails, catch will continue running other tests but what will happen to catch's output to cout?

src/QirRuntime/test/QIR-static/qir-test-output.qs
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.Quantum.Testing.QIR.Out {
+
+    open Microsoft.Quantum.Intrinsic;
+
+    function MessageTest(msg: String) : Unit {
+        Message( msg );
```
irinayat-MS, line 9: Nit: should be no spaces around params in a call

irinayat-MS: BTW, if you run clang formatter fixes this?

src/QirRuntime/test/QIR-static/qir-test-math.cpp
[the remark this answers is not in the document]
samarsha: If the random seed is fixed, then the sequence from DrawRandomInt should be the same every time.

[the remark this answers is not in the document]
irinayat-MS: There is no means on Q#'s side to provide the seed, right? So, the Q# code will always generated the same sequence, right? Is that the expected behaviour?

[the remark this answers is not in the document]
samarsha: No, I don't think Q# has any way to set the seed. The default behavior is to randomize the seed, so each program execution is different. I'm suggesting that we should fix the seed only for tests, so that tests that rely on randomness pass deterministically.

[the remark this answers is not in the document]
cgranade: In general, there may not even be a seed. E.g.: if a QRNG is used to back `DrawRandomInt()`, there's no pseudorandom RNG to seed in that case.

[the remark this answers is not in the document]
samarsha: That's true, so I don't think a QRNG should be used as a source of randomness for CI tests.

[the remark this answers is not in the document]
cgranade: Very much so agreed! Was only speaking to why it's intentional that there's no Q# API for setting the seed.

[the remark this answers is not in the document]
cgranade: As for probability of failure, a back of the envelope calculation puts that at around 10⁻²² for the number of samples that are taken so far, such that we would need around 10¹¹ builds to have any nontrivial support over the number of failures being at least one. We may still prefer deterministic for all the reasons discussed above, but I think the rate of failure from randomness in this test is comparable to the rate of failure caused by cosmic rays hitting build agent RAM.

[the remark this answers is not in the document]
kuzminrobin: I would say, ideally I need to make the random number generator customizable, such that by default the generated random numbers are as random as possible, but _for the test purposes_ we could change the settings minimally, such that the generated random numbers are 100% predictable, and as much as possible of the functionality to be tested.
To summarize, I will be targeting at different random number sequences _by default_ and _during the test_. 
I'll get back to this.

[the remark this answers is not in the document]
cgranade: @kuzminrobin: I don't disagree with customizability, but I think it is critical for that customization to **not** be exposed at the Q# level. In particular, Q# as a language and as a platform is designed to be transparent across devices (including simulators), so that a Q# developer has confidence that their quantum applications are not inadvertently specialized to a particular simulator or device — this is important enough that it's even our first [language design principle](https://github.com/microsoft/qsharp-language/#design-principles).

From that perspective, a developer looking to control how a particular target provides random numbers should do so in a way that makes it explicit that they're configuring the target and not at the Q# level.

src/QirRuntime/lib/QIR/bridge-qis.ll
kuzminrobin: Fixing.

(Explanation: I try to keep the code dense such that more code fits in one screen page. In that regard I assume comments are not code, and it is ok if the comments go out of the screen page. The readers mostly read comments much more rare than the code, at least if the code is simple).

src/QirRuntime/test/QIR-static/qir-driver.cpp
kuzminrobin: Agreed.

Explanation: Since this PR was the second time when Chris had to remind me to remove "All rights reserved" (which was the result of copy-paste of the code in this directory), I decided to secure myself from repeating the mistaken copy-paste pattern and removed all the "All rights reserved" from all the files where I likely to copy it again from during subsequent code changes.

src/QirRuntime/test/QIR-static/qir-test-math.cpp
kuzminrobin: Yes, it is (currently).

However in the future the values of the `minimum` and `maximum` can change, the mechanism of mapping them to the `index` can also change.  
By double-checking the `index` I want to make sure that the container is indexed correctly, no matter what.

src/QirRuntime/test/QIR-static/qir-test-ouput.cpp
kuzminrobin: Agreed.

I have written an email to Cesar (copy to you). I'll wait for what he says. Any recommendations about a better outputting and testing approach are welcome. 
For now I'm thinking about outputting not to C++'s `std::cout` and not to C's `stdout`, but to introduce an abstraction in front of those in the form of a logger (that buffers the output and then flushes to `std::cout` or `stdout`), such that for the test purposes I substitute the logger until the end of the scope. And the test will analyze the logger's buffer after the `Message()` function writes to it.

src/QirRuntime/test/QIR-static/qir-test-output.qs
kuzminrobin: Copy-paste cost... ;-) fixed.
I'll try clang-format and see how it works.

src/QirRuntime/lib/QIR/intrinsicsMath.cpp
kuzminrobin: Fixed.

src/QirRuntime/test/QIR-static/qir-test-output.qs
kuzminrobin: Here is the output of 
> qsharp-runtime\src\QirRuntime\test\QIR-static>**clang-format --style=file qir-test-output.qs** 

:-)
```

namespace Microsoft.Quantum.Testing.QIR.Out
{

    function MessageTest(msg
                         : String)
        : Unit
    {
        Message(msg);
    }

} // namespace Out
```
So yes, _it does remove the spaces inside of parentheses_, but based on our "qsharp-runtime/src/QirRuntime/**.clang-format** ", which is C++-oriented I believe, it _does a number of undesirable changes_ to the .qs file, including the change of a comment in the last line. ;-)

cgranade: @samarsha has been working on developing a Q# formatter, we should consider using that has been packaged as a redeployable tool (https://github.com/microsoft/qsharp-compiler/issues/864).

src/QirRuntime/lib/QIR/intrinsicsOut.cpp
cgranade: Tagging in @anjbur for IQ# discussion.

src/QirRuntime/test/QIR-static/qir-test-math.cpp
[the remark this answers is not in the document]
kuzminrobin: Before I posted my previous message I didn't refresh the page for a number of hours, and didn't notice a number of messages ;-). I was basically answering the Irina's message 
> I am confused, are you saying that repeated runs of a Q# program that calls DrawRandomInt() would produce the same sequence of integers?

Before I continue working on this I'll push the changes addressing the other CR notes.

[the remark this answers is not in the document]
kuzminrobin: I have updated my branch, please review the changes relating to this discussion.

src/QirRuntime/lib/QIR/intrinsicsOut.cpp
kuzminrobin: I have updated my branch and addressed all the concerns except the last Chris' email about `Extensions.cs#L96`. I'll take a look at the email in more details tomorrow.

src/QirRuntime/test/QIR-static/qir-test-ouput.cpp
kuzminrobin: I have updated my branch and addressed all the concerns except the last Chris' email about `Extensions.cs#L96`. I'll take a look at the email in more details tomorrow.

kuzminrobin: To unblock Guen, I decided to postpone the changes (if any) inspired by Chris' email. 
I filed #518 to keep track of that. Marking this discussion as resolved in this PR.

src/QirRuntime/lib/QIR/intrinsicsOut.cpp
kuzminrobin: To unblock Guen, I decided to postpone the changes (if any) inspired by Chris' email. 
I filed #518 to keep track of that. Marking this discussion as resolved in this PR.

src/QirRuntime/lib/QIR/CMakeLists.txt
```diff
@@ -51,6 +51,7 @@ compile_from_qir(bridge-qis ${bridge_qis_target})
 set(qis_sup_source_files
   "intrinsics.cpp"
   "intrinsicsMath.cpp"
+  intrinsicsOut.cpp
```
cgranade, line 54: Is it significant that this isn't in quotes like the other files?

src/QirRuntime/lib/QIR/bridge-qis.ll
```diff
@@ -336,3 +347,10 @@ define double @__quantum__qis__arctan2__body(double %y, double %x) {  ; Q#: func
   %result = call double @quantum__qis__arctan2__body(double %y, double %x)
   ret double %result
 }
+
+; operation DrawRandomInt (min : Int, max : Int) : Int      
+; https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.random.drawrandomint
```
cgranade, line 352: [nit] In general, en-us should be dropped from documentation URLs as that has the side effect of overwriting users' language preferences.

```suggestion
; https://docs.microsoft.com/qsharp/api/qsharp/microsoft.quantum.random.drawrandomint

```

src/QirRuntime/test/QIR-static/qsharp/Math.qs
```diff
@@ -28,4 +28,8 @@ namespace Microsoft.Quantum.Intrinsic {
         body intrinsic;
     }
 
-}
\ No newline at end of file
+    operation DrawRandomInt(min : Int, max : Int) : Int {
```
cgranade, line 31: I would have expected this operation to be in the Microsoft.Quantum.Random namespace, but it looks like this PR adds it to the Microsoft.Quantum.Intrinsic namespace?

src/QirRuntime/test/QIR-static/qsharp/qir-test-arrays.qs
```diff
@@ -40,6 +41,8 @@ namespace Microsoft.Quantum.Testing.QIR {
             let res5 = LogTest();
             let res6 = ArcTan2Test();
             let res7 = PauliToStringTest();
+            let res8 = TestDrawRandomInt(0, 1);
```
cgranade, line 44: Out of curiosity, why are the math and random tests added to the `qir-test-arrays.qs` file? It feels a bit surprising for an operation named `Test_Arrays` to test things unrelated to arrays.

src/QirRuntime/lib/QIR/CMakeLists.txt
kuzminrobin: I suspect the quotes started when (or in a project where) the file names contained the chars (like spaces) that were not ok without the quotes. My added file name does not contain such chars so the quotes would be redundant (I may be wrong since I'm new to `cmake`, but I suspect that this part of functionality is the same as for `make` which I dealt with). 
So I skipped the quotes. Somebody please correct me if I'm wrong.

src/QirRuntime/lib/QIR/bridge-qis.ll
kuzminrobin: Wow, what a curious functionality!.. Thank you for sharing the knowledge.
Fixed this in the subsequent added changes I'm working on.

src/QirRuntime/test/QIR-static/qsharp/Math.qs
kuzminrobin: Thanks for catching this. Fixed.

src/QirRuntime/lib/QIR/quantum__qis_internal.hpp
```diff
@@ -0,0 +1,22 @@
...
+#else
+#define QIR_SHARED_API
+#endif
+
+// For test purposes only:
+namespace Quantum { namespace Qis { namespace Internal  // Replace with `namespace Quantum::Qis::Internal` after migration to C++17.
+{
+    QIR_SHARED_API extern char const excStrDrawRandomInt[]; // = "Invalid Argument: minimum > maximum for DrawRandomInt()";
+
+    QIR_SHARED_API extern std::ostream& SetQOstream(std::ostream & newOStream);
```
cesarzc, line 20: An alternative name for this API could be `SetOutputStream`, which I think is a bit more intuitive.

src/QirRuntime/test/QIR-static/qsharp/qir-test-arrays.qs
kuzminrobin: This is a trick to cheat the compiler.

We are building the _executable_ from the .qs files. And the compiler includes into the executable only _those functions_, that _are reachable_ from the `@Entrypoint` function. 
```
    @EntryPoint()
    operation Test_Arrays(array : Int[], index : Int, val : Int, compilerDecoy : Bool) : Int {
```
And for the `@Entrypoint` function the compiler generates a special wrapper that can be called from our C++ test code, see in `qir-gen.ll`
```
define i64 @Microsoft__Quantum__Testing__QIR__Test_Arrays(i64 %array__count, i64* %array, i64 %index, i64 %val, i1 %compilerDecoy) #0 {
```
Pay attention to the first 2 parameters as opposed to the 1st parameter in the Q# declaration for the `@Entrypoint` function shown earlier. And compare to the 1st parameter in the actual implementation of the `@Entrypoint` function in `qir-gen.ll`:
```
define i64 @Microsoft__Quantum__Testing__QIR__Test_Arrays__body(%Array* %array, i64 %index, i64 %val, i1 %compilerDecoy) {
```

The other option would be for the compiler to generate the _library_, rather than _executable_. In that case _all_ the functions will be included into the library (not only those reachable from the `@EntryPoint`). However the compiler will _not_ generate the C++-callable wrapper, and we will have to work around.
I'm filing #521 to track that. @irinayat-MS, please let me know if we already have an issue for tracking this. I'll resolve one of them as a duplicate of the other of them and merge the info from both of them into one.

src/QirRuntime/lib/QIR/intrinsicsOut.cpp
```diff
@@ -0,0 +1,43 @@
...
+{
+    void quantum__qis__message__body(QirString* qstr)   // NOLINT
+    {
+        GetQOstream() << qstr->str << std::endl;
+    }
+}   // extern "C"
+
+
+// Internal API:
+static  std::ostream * currOStream = &std::cout;    // Log to std::cout by default.
```
cesarzc, line 23: Is the convention to write pointers' declarations like `static std::ostream * currOStream = &std::cout;` instead of `static std::ostream* currOStream = &std::cout;` (the `*` next to the type with no spacing).

kuzminrobin: As far as I know (AFAIK) there is no such a convention in our C++ code. What I know is that there are at least 2 significant groups of C/C++ devs with different preferences.

One group prefers writing `int* i;`, reasoning it with separation of the type and variable name.

Another group prefers writing `int *i;` reasoning it that `*` changes the variable rather than a type, and if we write `int* i, j;` then only the first variable `i` will be the pointer, but the second variable `j` will not be a pointer (it will be `int`). And such a fragment can mislead those who are new to C/C++. But if we write `int *i, j;` then it will mislead less (by being closer to the reality).

Knowing that I personally prefer writing something _intermediate_ ;-) Both for pointers `*` and references `&`.

src/QirRuntime/lib/QIR/quantum__qis_internal.hpp
kuzminrobin: Fixed.

src/QirRuntime/lib/QIR/CMakeLists.txt
irinayat-MS: I believe you are right. Because I'm also new to cmake, a considerable portion of the initial coding of the CMakeLists.txt files happened by copy-paste from various examples so there are inconsistencies in formatting wrt optional things like quotes. Please feel free to standardize the files you touch (or add to the list of cleanups in https://github.com/microsoft/qsharp-runtime/issues/482)

src/QirRuntime/lib/QIR/bridge-qis.ll
```diff
@@ -60,6 +60,8 @@ declare void @quantum__qis__y__ctl(%struct.QirArray*, %class.QUBIT*)
 declare void @quantum__qis__z__body(%class.QUBIT*)
 declare void @quantum__qis__z__ctl(%struct.QirArray*, %class.QUBIT*)
 
+declare void @quantum__qis__message__body(%"struct.QirString"* %str)
```
irinayat-MS, line 63: FYI: this is another situation when quotes aren't necessary

src/QirRuntime/lib/QIR/intrinsicsMath.cpp
```diff
@@ -28,4 +39,38 @@ double quantum__qis__arctan2__body(double y, double x)
...
+
 }   // extern "C"
+
+namespace   // Visible in this translation unit only.
+{
+    thread_local bool seedIsRandom = true;
+}
+
+// For test purposes only:
+namespace Quantum { namespace Qis { namespace Internal
```
irinayat-MS, line 66: Please run clang formatter to have same line formatting across the lines (I know, you prefer more compact namespace declarations and we probably can change the formatter to do this rather than one namespace per line, but for now the automatically applied formatting will split the lines)

src/QirRuntime/lib/QIR/intrinsicsMath.cpp
```diff
@@ -28,4 +39,38 @@ double quantum__qis__arctan2__body(double y, double x)
...
+    thread_local bool seedIsRandom = true;
+}
+
+// For test purposes only:
+namespace Quantum { namespace Qis { namespace Internal 
+{
+
+char const excStrDrawRandomInt[] = "Invalid Argument: minimum > maximum for DrawRandomInt()";
+
+void UseRandomSeed(bool random)
```
irinayat-MS, line 71: `SetRandomSeed`?

src/QirRuntime/lib/QIR/intrinsicsOut.cpp
irinayat-MS: Please run clang formatter -- it will do automated formatting that at some point we've agreed on for this project...

src/QirRuntime/lib/QIR/quantum__qis_internal.hpp
```diff
@@ -0,0 +1,22 @@
...
+#ifdef _WIN32
+#define QIR_SHARED_API __declspec(dllexport)
+#else
+#define QIR_SHARED_API
+#endif
+
+// For test purposes only:
+namespace Quantum { namespace Qis { namespace Internal  // Replace with `namespace Quantum::Qis::Internal` after migration to C++17.
+{
+    QIR_SHARED_API extern char const excStrDrawRandomInt[]; // = "Invalid Argument: minimum > maximum for DrawRandomInt()";
```
irinayat-MS, line 18: Do these three things need to be exported?

src/QirRuntime/test/QIR-static/qir-test-math.cpp
```diff
@@ -23,3 +28,96 @@ TEST_CASE("QIR: Math.ArcTan2", "[qir.math][qir.Math.ArcTan2]")
     REQUIRE(0 == Microsoft__Quantum__Testing__QIR__Math__ArcTan2Test__body());
 }
 
+TEST_CASE("QIR: Math.DrawRandomInt", "[qir.math][qir.Math.DrawRandomInt]")
+{
+#ifdef __APPLE__
+    std::vector<int64_t> expectedLargeNumbers( {
```
irinayat-MS, line 34: I'm not convinced testing the specific output from the random generator, even if we seed it, is a good idea. We are already seeing differences between platforms. What if there are differences between clang versions? Will we have to update the tests after moving to 11.0?

Our implementation is a thin wrapper around std random generator, why are we testing the specifics of that generator?

src/QirRuntime/test/QIR-static/qir-test-ouput.cpp
```diff
@@ -0,0 +1,52 @@
...
+#include "qirTypes.hpp"
+#include "quantum__qis_internal.hpp"
+
+extern "C" void Microsoft__Quantum__Testing__QIR__Out__MessageTest__body(QirString*);        // NOLINT
+
+
+// https://stackoverflow.com/a/5419388/6362941
+// https://github.com/microsoft/qsharp-runtime/pull/511#discussion_r574170031
+// https://github.com/microsoft/qsharp-runtime/pull/511#discussion_r574194191
+struct QOstreamRedirector
```
irinayat-MS, line 17: Nit: personally, I like to indicate in the name of such types that they are meant to be scoped (e.g. `QirContextScope`), because then it's easier to see why you might want a block of {} in the code that uses them.

src/QirRuntime/test/QIR-static/qir-test-ouput.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+        // Log something (to the redirected output):
+        QirString qstr{std::string(testStr1)};
+        Microsoft__Quantum__Testing__QIR__Out__MessageTest__body(&qstr);
+        qstr.str = testStr2;
+        Microsoft__Quantum__Testing__QIR__Out__MessageTest__body(&qstr);
+
+    } // Recover the output stream.
+
+    REQUIRE(outStrStream.str() == (testStr1 + "\n" + testStr2 + "\n"));
```
irinayat-MS, line 51: Is your intent here to test for multiple instances of output to be separated by new line? Which is valid, but you might also consider using SECTIONs to run a few tests with same setup.

src/QirRuntime/test/QIR-static/qsharp/qir-test-arrays.qs
irinayat-MS: And, btw, the block has the following comment: "// The purpose of this block is to keep the Q# compiler from optimizing away other tests when generating QIR", but GitHub's PR review UI doesn't show it by default...

src/QirRuntime/test/QIR-static/qir-test-ouput.cpp
```diff
@@ -0,0 +1,52 @@
...
+// Licensed under the MIT License.
+#include <iostream>
+#include <sstream>
+
+#include "catch.hpp"
+
+#include "qirTypes.hpp"
+#include "quantum__qis_internal.hpp"
+
+extern "C" void Microsoft__Quantum__Testing__QIR__Out__MessageTest__body(QirString*);        // NOLINT
```
irinayat-MS, line 11: Would it help to add a comment that we are aware of the signature mismatch between this decl and the generated function in QIR but it's safe in this case?

irinayat-MS (approved):
Could you please run clang format on the files you touched? As far as the substance of the change is concerned, my only big concern is the tests for random (see my in-file comment), which I'm leaving to your discretion.

src/QirRuntime/lib/QIR/intrinsicsOut.cpp
kuzminrobin: Fixed.

src/QirRuntime/lib/QIR/bridge-qis.ll
kuzminrobin: OK, when touching the lines I will keep that in mind.

src/QirRuntime/lib/QIR/intrinsicsMath.cpp
kuzminrobin: Fixed.

src/QirRuntime/lib/QIR/intrinsicsMath.cpp
kuzminrobin: Oh... ;-) That's what is standing behind UseFoo() that I saw earlier... OK. Nice to know.
Renaming `UseRandomSeed()` to `RandomizeSeed()`.

src/QirRuntime/test/QIR-static/qir-test-ouput.cpp
kuzminrobin: The intent was to test that 
* the output actually takes place (not ignored),
* the output corresponds to the expected one (is not random or unrelated),
* the multiple outputs are separated with a new line char (not merged/concatenated into a single line).

If with SECTION you mean `TEST_CASE( .. ) { .. }` then I'm not sure about the value it adds if we run SECTIONs with the output redirected. Can you please elaborate?
(I was assuming that I need to test that a single Q# call to `Message()` and a few consecutive Q# calls to `Message()` work as expected)

src/QirRuntime/test/QIR-static/qir-test-ouput.cpp
kuzminrobin: Hm... Very good question... I didn't even notice that (strictly speaking) the parameter type is different between here and in the generated `.ll`. And such a function call (strictly speaking) may be unsafe.
To stay safe, we can do the following:
* assume that `%String*` and `%struct.QirString*` are the same as (are just aliases of) `void*`;
* for that we can replace the current type definitions to `%String = type void`, `%struct.QirString = type void`, or, even simpler, use `void*` everywhere instead of `%String*` and `%struct.QirString*`.

This will let us avoid all those casts (in `.ll` files) between `%String*` and `%struct.QirString*`, and to stay safe when making calls between `.cpp` and `.ll` files.
What do you think?

For now in this line I'm changing the parameter type to `void*` for the call to be safe.

src/QirRuntime/lib/QIR/quantum__qis_internal.hpp
kuzminrobin: I assume, you mean "Why these three have QIR_SHARED_API?".
Removed.
Thanks for catching.

src/QirRuntime/test/QIR-static/qir-test-ouput.cpp
kuzminrobin: Sounds very reasonable to me...
Renaming to `OstreamRedirectorScoped`.

src/QirRuntime/lib/QIR/intrinsicsOut.cpp
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+#include <iostream>
+
+#include "qirTypes.hpp"
+#include "quantum__qis.hpp"
+
+// Forward declarations:
+static std::ostream& GetQOstream();
```
cesarzc, line 10: One more small request. Since the public API is named `SetOutputStream` now, it would be great to match the names in the internal implementation with the public API name (i.e. `GetOutputStream`, `currOutputStream`).

kuzminrobin: Fixed

src/QirRuntime/test/QIR-static/qir-test-math.cpp
kuzminrobin: Agreed. Changing.

cesarzc (approved):

src/QirRuntime/lib/QIR/quantum__qis_internal.hpp
```diff
@@ -13,10 +13,17 @@
 #endif
 
 // For test purposes only:
-namespace Quantum { namespace Qis { namespace Internal  // Replace with `namespace Quantum::Qis::Internal` after migration to C++17.
+namespace Quantum // Replace with `namespace Quantum::Qis::Internal` after migration to C++17.
```
irinayat-MS, line 16: Would you mind filing a bug for migrating to C++17? Hopefully, it should be as simple as changing 14 to 17 in `set(CMAKE_CXX_STANDARD 14)` in the root CMakeLists.txt.

src/QirRuntime/test/QIR-static/qir-test-math.cpp
```diff
@@ -30,94 +30,114 @@ TEST_CASE("QIR: Math.ArcTan2", "[qir.math][qir.Math.ArcTan2]")
...
+    // Test that the Q# random number generator is a wrapper around the C++ generator:
+    size_t times = 1000;
+    while(--times)
     {
-        actualNumbers.emplace_back( 
-            Microsoft__Quantum__Testing__QIR__Math__TestDrawRandomInt__body(-10, 10));
+        const uint64_t qsRndNum = 
+            Microsoft__Quantum__Testing__QIR__Math__TestDrawRandomInt__body(std::numeric_limits<int64_t>::min(),
+                                                                            std::numeric_limits<int64_t>::max());
+        const uint64_t cppRndNum = Quantum::Qis::Internal::GetLastGeneratedRandomNumber();  // This call must be done
```
irinayat-MS, line 40: (not blocking) Why not create a C++ random generator here with the same seed as ours and test that the two produce the same random number? Also, why do you run it multiple times? With randomized seed the first produced number will be sufficiently random, no?

src/QirRuntime/lib/QIR/intrinsicsMath.cpp
```diff
@@ -48,29 +49,40 @@ int64_t quantum__qis__drawrandomint__body(int64_t minimum, int64_t maximum)
...
 {
-    seedIsRandom = random;
-}
-
-}}} // namespace Quantum::Qis::Internal
+    namespace Internal
+    {
+        char const excStrDrawRandomInt[] = "Invalid Argument: minimum > maximum for DrawRandomInt()";
+
+        void RandomizeSeed(bool randomize)
```
irinayat-MS, line 77: Is it still used?

src/QirRuntime/lib/QIR/quantum__qis_internal.hpp
kuzminrobin: ;-)
I tried, it cases a compiler error in (at least) one of the headers in MS VC++ (and that error is reported for dozens of translation units (.cpp files)).
Filed #523

src/QirRuntime/test/QIR-static/qir-test-math.cpp
kuzminrobin: I'm afraid I have already spent too much time for this PR. To unblock Guen I will prefer to move forward and push what I have now.

src/QirRuntime/lib/QIR/intrinsicsMath.cpp
kuzminrobin: No.  ;-)
But the code that uses it is _commented out_ (not _removed_). So I left this functionality for now.

PR #515, opened 2021-02-11
Better error reporting on failure to load the fullstate simulator

irinayat-MS:


swernli (approved):
LGTM

src/QirRuntime/lib/Simulators/FullstateSimulator.cpp
```diff
@@ -21,14 +22,22 @@ typedef HMODULE QUANTUM_SIMULATOR;
...
-    return ::LoadLibraryA("Microsoft.Quantum.Simulator.Runtime.dll");
+    return ::LoadLibraryA(FULLSTATESIMULATORLIB);
 #elif __APPLE__
-    return ::dlopen("libMicrosoft.Quantum.Simulator.Runtime.dylib", RTLD_LAZY);
+    return ::dlopen(FULLSTATESIMULATORLIB, RTLD_LAZY);
 #else
-    return ::dlopen("libMicrosoft.Quantum.Simulator.Runtime.so", RTLD_LAZY);
+    return ::dlopen(FULLSTATESIMULATORLIB, RTLD_LAZY);
 #endif
 }
```
kuzminrobin, line 42: May be something like this?
```c++
#define LIBNAME = "Microsoft.Quantum.Simulator.Runtime"
#ifdef _WIN32
    static const char* FULLSTATESIMULATORLIB = LIBNAME ".dll";
    static inline void LoadLibrary()  { ::LoadLibraryA(FULLSTATESIMULATORLIB) }
#else
    #if __APPLE__
        static const char* FULLSTATESIMULATORLIB = LIBNAME ".dylib";
    #else
        static const char* FULLSTATESIMULATORLIB = LIBNAME ".so";
    #endif

    static inline void LoadLibrary()  { ::dlopen(FULLSTATESIMULATORLIB, RTLD_LAZY) }
#endif

QUANTUM_SIMULATOR LoadQuantumSimulator()
{
    LoadLibrary();
}
```

kuzminrobin (approved):

src/QirRuntime/lib/Simulators/FullstateSimulator.cpp
kuzminrobin: or instead of all the occurrences of `static` you can enclose the whole block into _anonymous_ namespace
```c++
namespace
{
    . . .
}
```

irinayat-MS: There is also custom prefix... I prefer specifying the full name in one string because then you can search for it.

irinayat-MS: As far as having only a couple static functions I don't see much value in using an anonymous namespace. There are no types to hide.

irinayat-MS: Oh, you meant to _add_ `static` or anonymous namespace! :D Will do.

PR #516, opened 2021-02-11
Clean up target package and simulator files.

swernli:
This gets rid of several unused  files that had mistakenly been merged from the feature branch, as well as cleaning up the implementations to match the new Q# style and removing unusing `using` statements from C#.


irinayat-MS (approved):
lgtm (as far a skim + spot check strategy allows, please let me know if there is anything specific in this PR that deserves more scrutiny)

PR #517, opened 2021-02-12
Generate ll file for QIR-static tests from Q# project

irinayat-MS:
The Q# build is failing on my WSL Ubuntu 20.4 so for now I'm still checking in the generated ll file so Linux/Mac can use this checked in file and the pipelines don't have to change.

To avoid generating QIR test file, run build|test.py with `noqirgen` parameter.


src/QirRuntime/README.md
```diff
@@ -36,26 +36,29 @@ CI builds and tests are enabled for this project. The build has no external depe
...
-2. In the Ubuntu's terminal:
+1. On the host Windows machine [enable WSL](https://docs.microsoft.com/en-us/windows/wsl/install-win10) and install
+ Ubuntu 20.04 LTS.
+1. In the Ubuntu's terminal:
     1. `$ sudo apt install cmake` (`$ cmake --version` should return 3.16.3)
-    2. `$ sudo apt-get install ninja-build` (`$ ninja --version` should return 1.10.0)
-    3. `$ sudo apt install clang` (`$ clang++ --version` should return 10.0.0)
-    4. Set Clang as the preferred C/C++ compiler:
+    1. `$ sudo apt-get install ninja-build` (`$ ninja --version` should return 1.10.0)
+    1. `$ sudo apt install clang` (`$ clang++ --version` should return 10.0.0)
```
kuzminrobin, line 55: > sudo **apt-get**
sudo **apt**

Why different?

src/QirRuntime/README.md
```diff
@@ -36,26 +36,29 @@ CI builds and tests are enabled for this project. The build has no external depe
...
+    1. Set Clang as the preferred C/C++ compiler:
         - $ export CC=/usr/bin/clang
         - $ export CXX=/usr/bin/clang++
```
kuzminrobin, line 58: I would place these `$ export ...` commands _last_, after everything is installed (these `$ export` commands are not required to install `clang-tidy` below, right?).
The users typically install everything needed first, and then customize their *profile* scripts, to where these `$ export` commands will go.

src/QirRuntime/README.md
irinayat-MS: Probably, because I've changed from manually maintained numbered points (which are a pain) to the automatic ones?

src/QirRuntime/README.md
irinayat-MS: Well, I might not be a typical user, but I like dealing with groups of related tasks: install clang + set it to be the default, move on to installing other stuff... The thing I care much more about in this case is whether this export is at all necessary. When I wrote the instructions last spring I was running into issues with CMake refusing to pick up clang even if I passed it via the CMake's cmd compiler arg. We should try this again and, ideally, eliminate the requirement for export, which is too invasive... but this is outside of the scope of this PR.

src/QirRuntime/README.md
```diff
@@ -65,6 +68,11 @@ Some of the tests depend on Microsoft.Quantum.Simulator.Runtime library. To run
  from this repository or provide your own version of the library in a folder the OS would search during dynamic library
  lookup.
 
+Some of the tests use generated QIR (*.ll) files as build input. Currently the files are checked-in as part of the project
+ but in the future they will be replaced by automatic generation during build. To regenerate the files, run generateqir.py
+ or build/test scripts without specifying `noqirgen`. To use the checked-in files without regenerating them, run build/test
+ scripts with `noqirgen` argument.
+
```
kuzminrobin, line 75: Minor:
The name `noqirgen` tells me that if `noqirgen` is specified, then no QIR generation (`.ll` files generation) happens. But if it is _not_ specified, then the QIR will be generated. To summarize, _the QIR is **generated by default**_.
This seems to me a perfect approach for the future, when the generation is implemented and stable, or if you know for sure when exactly the generation will be implemented.
For now, I assume, the generation is not complete (or is unstable). And if it is _not_ known for sure when the generation will be implemented, probably it makes sense to _not generate by default_. I.e. to rename the flag to `qirgen` (and if it is not specified then to _not_ generate).

src/QirRuntime/lib/QIR/bridge-rt.ll
```diff
@@ -381,7 +381,7 @@ define void @__quantum__rt__callable_memory_management(i32 %index, %Callable* %.
 ; NYI:
 ;define %String* @__quantum__rt__bigint_to_string(%BigInt*)
 
-define %String* @__quantum__rt__string_create(i32 %ignoredStrLength, i8* %null_terminated_buffer) {
+define %String* @__quantum__rt__string_create(i32 %length_ignored, i8* %null_terminated_buffer) {
```
kuzminrobin, line 384: Thanks for fixing this.

src/QirRuntime/test/QIR-static/qsharp/qir-gen.csproj
```diff
@@ -0,0 +1,9 @@
+<Project Sdk="Microsoft.Quantum.Sdk/0.15.2102128445-alpha">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <QirGeneration>True</QirGeneration>
+  </PropertyGroup>
+
+</Project>
```
kuzminrobin, line 9: File extension is `.csproj`. _C#_ rather than _Q#_ project?

kuzminrobin (approved):

src/QirRuntime/README.md
irinayat-MS: QIR generation is working on Windows and can/should be done by default. The only reason to provide this parameter is to allow suppression of QIR generation on Linux that seems to have issues at least for some people.

src/QirRuntime/test/QIR-static/qsharp/qir-gen.csproj
irinayat-MS: Yes, currently, Q# projects are built using dotnet.

PR #520, opened 2021-02-16
QIR Entry Point Driver Samples

cesarzc:
This change implements multiple entry point drivers for different QIR programs.


src/QirRuntime/test/CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
 add_subdirectory(FullstateSimulator)
 add_subdirectory(QIR-dynamic)
 add_subdirectory(QIR-static)
+add_subdirectory(Standalone)
```
irinayat-MS, line 4: Should we, maybe, create a new higher level folder, something like `samples`, rather than add it to `test`?

src/QirRuntime/test/Standalone/InputTypes/.clang-tidy
```diff
@@ -0,0 +1,3 @@
+Checks:
```
irinayat-MS, line 1: why do you need a special .clang-tidy file in this folder? why inheriting it from the parent folders isn't OK?

src/QirRuntime/test/Standalone/InputTypes/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+#  qir-test-arrays
+#  qir-test-functors
+#  qir-test-noqsharp
+#  qir-test-qsharp
+#  qir-test-partials
+#  qir-test-qubits-results
+  qir-input-types
+)
+
+foreach(file ${INPUT_TYPES_FILES})
```
irinayat-MS, line 13: I presume, this sample will integrate with Q# projects fully and, thus, will always have a single input QIR file, correct? In that's the case, you could simplify this a bit.

src/QirRuntime/test/Standalone/InputTypes/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+  ${QIR_INPUT_TYPES_LIBS}
+  ${QIR_BRIDGE_UTILITY_LIB}
+  ${QIR_BRIDGE_QIS_UTILITY_LIB}
+  qir-rt-support
+  qir-qis-support
+  simulators
+)
+
+target_include_directories(qir-input-types PUBLIC
+  "${test_includes}"
```
irinayat-MS, line 36: if this sample is meant to be "user code", it probably shouldn't depend on test-specific headers (another reason to have this new code outside of `test` so it's hard to pick up unwanted dependencies)

src/QirRuntime/test/Standalone/InputTypes/compiler/QirCore.qs
```diff
@@ -0,0 +1,30 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
irinayat-MS, line 1: If you generate the QIR from Q# project, you shouldn't need this file.

src/QirRuntime/test/Standalone/InputTypes/input-types.qs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.Quantum.Testing.QIR
+{
+    open Microsoft.Quantum.Intrinsic;
+
+    @EntryPoint()
+    operation InputTypes(anInt : Int, aDouble : Double, aBool : Bool, aString : String, anArray : Int[]) : Unit {
+
```
irinayat-MS, line 10: Maybe, add a comment on why there is no code here? What exactly are we testing/concept-proving?

src/QirRuntime/test/Standalone/InputTypes/qir-input-types.ll
```diff
@@ -0,0 +1,49 @@
...
+
+@ResultZero = external global %Result*
+@ResultOne = external global %Result*
+@PauliI = constant i2 0
+@PauliX = constant i2 1
+@PauliY = constant i2 -1
+@PauliZ = constant i2 -2
+@EmptyRange = internal constant %Range { i64 0, i64 1, i64 -1 }
+
+define void @Microsoft__Quantum__Testing__QIR__InputTypes__body(i64 %anInt, double %aDouble, i1 %aBool, %String* %aString, %Array* %anArray) {
```
irinayat-MS, line 15: FYI: there is currently a bug in QIR generation that the array, created by the wrapper, isn't properly released (https://github.com/microsoft/qsharp-compiler/issues/844)

src/QirRuntime/test/Standalone/Shared/CLI11.hpp
```diff
@@ -0,0 +1,8258 @@
+#pragma once
```
irinayat-MS, line 1: Why this file is needed?

src/QirRuntime/test/Standalone/InputTypes/input-types.qs
irinayat-MS: Also, why do we need two samples: InputTypes _and_ RandomBit?

src/QirRuntime/test/Standalone/InputTypes/qir-driver.cpp
```diff
@@ -0,0 +1,92 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#include <iostream>
+#include "../Shared/CLI11.hpp"
```
irinayat-MS, line 5: where is this file coming from? do we have to take this dependency?

src/QirRuntime/test/Standalone/InputTypes/qir-driver.cpp
```diff
@@ -0,0 +1,92 @@
...
+extern "C" void DebugLog(int64_t value)
+{
+    std::cout << value << std::endl;
+}
+extern "C" void DebugLogPtr(char* value)
+{
+    std::cout << (const void*)value << std::endl;
+}
+
+extern "C" void SetupQirToRunOnFullStateSimulator();
```
irinayat-MS, line 17: Please don't use this factory. It's not exposed in header files for a reason :) Instead, include context.hpp, QuantumApi_I.hpp, and SimFactory.hpp and use the following pattern in the code:
```
    std::unique_ptr<ISimulator> sim = CreateFullstateSimulator();
    QirContextScope qirctx(sim.get(), false /*trackAllocatedObjects*/);
```

src/QirRuntime/test/CMakeLists.txt
cesarzc: Yes, that sounds like a good idea. I'll move this to a higher level folder before publishing this PR for merge into `main`.

src/QirRuntime/test/Standalone/InputTypes/.clang-tidy
cesarzc: The main reason was to change `HeaderFilterRegex` to exclude `CLI11.hpp` since it's a third-party header that makes clang-tidy unhappy. I haven't gotten to do this specifically yet but do you have any pointer on how to do that? So far, I haven't found examples on how to do this in the clang-tidy documentation.

src/QirRuntime/test/Standalone/InputTypes/CMakeLists.txt
cesarzc: I'll remove this.

src/QirRuntime/test/Standalone/InputTypes/compiler/QirCore.qs
cesarzc: Yes, I started doing some of this work before it was possible to do that but now that it is available I'll merge main and delete these files that are no longer needed :)

src/QirRuntime/test/Standalone/InputTypes/.clang-tidy
irinayat-MS: I don't like this. I don't like pulling into production code a foreign header, I like it even less that this header forces us to suppress static analysis... If we _must_ use the header, we might be able to suppress clang-tidy for it selectively with:
```
  --header-filter=<string>       -
                                   Regular expression matching the names of the
                                   headers to output diagnostics from. Diagnostics
                                   from the main file of each translation unit are
                                   always displayed.
                                   Can be used together with -line-filter.
                                   This option overrides the 'HeaderFilterRegex'
                                   option in .clang-tidy file, if any.
  --line-filter=<string>         -
                                   List of files with line ranges to filter the
                                   warnings. Can be used together with
                                   -header-filter. The format of the list is a
                                   JSON array of objects:
                                     [
                                       {"name":"file1.cpp","lines":[[1,3],[5,7]]},
                                       {"name":"file2.h"}
                                     ]
```
per https://clang.llvm.org/extra/clang-tidy/ (but I haven't tried it)

src/QirRuntime/test/Standalone/InputTypes/generate.py
```diff
@@ -0,0 +1,45 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
cgranade, line 1: Is this script used anywhere in CI builds? If so, I'd suggest migrating it to PowerShell; we've consolidated the rest of our build scripts to use PowerShell rather than having a mix of bash, cmd, and PowerShell.

src/QirRuntime/test/Standalone/InputTypes/input-types.qs
cesarzc: The concept is to have a few examples to know how the generated C++ driver should look like. _RandomBit_ was just the first example that worked as a standalone executable, then _InputTypes_ is meant to look like a program that receives all the supported entry-point input types.

I also have to add an _OutputTypes_ that returns all the supported output types. I suppose all of that can be done in one example but I think that for reference is better to have two programs because it is easier to see what is going on.

Regarding why there is no code here, since this is just to prove the concept of the C++ driver, programs don't really have to do anything. Once that Robin's PR supporting `Message` is merged I'll add a few `Message` calls with the values of the input arguments to make it easy to verify that the driver is doing its job.

I'll add a readme.md to the root of Standalone root explaining this.

src/QirRuntime/test/Standalone/InputTypes/generate.py
```diff
@@ -0,0 +1,45 @@
...
+  current_time = now.strftime("%H:%M:%S")
+  print(current_time + ": " + message)
+# =============================================================================
+
+root_dir = os.path.dirname(os.path.abspath(__file__))
+
+# parameters
+qsc = sys.argv[1] # argv[0] is the name of this script file
+
+# if no file name argument, process all qs files in this folder, otherwise treat all additional args as input files
```
cgranade, line 26: The command line handling logic here is fairly complex; it may be worth using the built-in [`argparse` module](https://docs.python.org/3/library/argparse.html) or a third-party module like `click` to help provide help to the user in running this script.

src/QirRuntime/test/Standalone/InputTypes/generate.py
```diff
@@ -0,0 +1,45 @@
...
+  for i in range(2, len(sys.argv)):
+    (file_name, ext) = os.path.splitext(sys.argv[i])
+    files_to_process = files_to_process + " " + file_name + ".qs"
+    if i == 2 and len(sys.argv) == 3:
+      output_file = file_name
+
+command = (qsc + " build --qir s --build-exe --input " + files_to_process +
+           " compiler\\qircore.qs compiler\\qirtarget.qs --proj " + output_file)
+log("Executing: " + command)
+subprocess.run(command, shell = True)
```
cgranade, line 44: It may be more robust here to use `shell=False` and pass a list as the command. In particular, that avoids issues if `files_to_process` includes files with spaces in their names, as `shell=True` causes the current shell to handle splitting the command into `argv`.

src/QirRuntime/test/Standalone/InputTypes/input-types.qs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.Quantum.Testing.QIR
+{
```
cgranade, line 5:
```suggestion
namespace Microsoft.Quantum.Testing.QIR {
```

src/QirRuntime/test/Standalone/InputTypes/input-types.qs
```diff
@@ -0,0 +1,12 @@
...
+    open Microsoft.Quantum.Intrinsic;
+
+    @EntryPoint()
+    operation InputTypes(anInt : Int, aDouble : Double, aBool : Bool, aString : String, anArray : Int[]) : Unit {
```
cgranade, line 9: This is named like a function, but is declared as an operation; I'd suggest a verb as the name here.

src/QirRuntime/test/Standalone/InputTypes/qir-driver.cpp
cgranade: Is this what provides `CLI::App`?

src/QirRuntime/test/Standalone/InputTypes/qir-driver.cpp
```diff
@@ -0,0 +1,92 @@
...
+    int64_t anInt,
+    double_t aDouble,
+    bool aBool,
+    char* aString,
+    int64_t anArrayLength,
+    int64_t* anArray);
+
+int main(int argc, char *argv[])
+{
+    try{
```
cgranade, line 28: It looks like you have a mix of same-line and next-line brace styles; which is used in this codebase?

cgranade (commented):
Are these samples intended to be user-facing or contributor-facing? If so, we should make sure they are located in repos that onboard to our samples publication pipeline, and that they have all required metadata for said onboarding (see internal wiki for details).

src/QirRuntime/test/Standalone/Shared/CLI11.hpp
cesarzc: It is a [command line parser utility](https://github.com/CLIUtils/CLI11) used for getting the arguments that need to be passed to the QIR entry-point (work in progress).

src/QirRuntime/test/Standalone/InputTypes/qir-driver.cpp
cesarzc: This file is coming from [CLI11](https://github.com/CLIUtils/CLI11).

It is not required in the sense that we could implement our own parser to not take this dependency but I don't think that would be a good decision since we can use something that already exists, it is in actively maintained, and has all the features that we would need.

The [install](https://github.com/CLIUtils/CLI11#install) section of the _readme_ describes that one of the ways to "install" the library is by copying the header in your include folder, which I found the most convenient for experimentation. We can add it as a git submodule as an alternative.

src/QirRuntime/test/Standalone/InputTypes/qir-driver.cpp
cesarzc: I'll change it to the pattern you are suggesting.

src/QirRuntime/test/Standalone/Shared/CLI11.hpp
irinayat-MS: When incorporating external code we probably need to:
- confirm the licensing terms and acknowledge it in the licensing manifest of our repo
- do a security review
- provide some way of versioning for maintenance and being able to pick up the updates
- consider: if this is a sample for end users, is it OK to force _them_ to take the dependency?
- - ???

The bottom line is, we shouldn't take external dependencies lightly. Do we have a precedent that involves CLI? Who can review this from the point of view of component governance?

cesarzc:
> 
> 
> Are these samples intended to be user-facing or contributor-facing? If so, we should make sure they are located in repos that onboard to our samples publication pipeline, and that they have all required metadata for said onboarding (see internal wiki for details).

These samples are contributor-facing. The idea is to have a few examples to know how the generated C++ driver for QIR should look like, so these samples might get replaced or removed at some point when the compiler generates a QIR executable.

src/QirRuntime/test/Standalone/InputTypes/generate.py
cesarzc: This file is no longer needed since Irina recently added support for [generating .ll files from a Q# project](https://github.com/microsoft/qsharp-runtime).

I'll remove this file.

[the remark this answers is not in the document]
cesarzc: Per this [message](https://github.com/microsoft/qsharp-runtime/pull/520#discussion_r577835359), this file is no longer needed and I'll remove it.

src/QirRuntime/test/Standalone/InputTypes/generate.py
cesarzc: I think this file has been consolidated [here](https://github.com/microsoft/qsharp-runtime/blob/main/src/QirRuntime/generateqir.py) and no longer receives any arguments. I'll remove this file from this change to leverage the latest changes in QIR generation.

src/QirRuntime/test/Standalone/InputTypes/generate.py
cesarzc: It seems like in the [consolidated version](https://github.com/microsoft/qsharp-runtime/blob/19039f61c3c88115297d2c81a612de58e37048c9/src/QirRuntime/generateqir.py#L19) still uses `shell=True` but it now calls `dotnet build` instead of calling the compiler command line tool directly. Do you think with that changes this suggestion still applies? If so, we can make the change as a separate PR.

src/QirRuntime/test/Standalone/InputTypes/input-types.qs
cesarzc: I'll change the name to `ExerciseInputTypes`.

src/QirRuntime/test/Standalone/InputTypes/qir-driver.cpp
cesarzc: The convention is next-line brace styles. I'll change it to make it consistent.

src/QirRuntime/test/Standalone/InputTypes/qir-input-types.ll
cesarzc: Thanks for the heads up!

cesarzc:
@cgranade for some reason, the reply box is not shown for [this message](https://github.com/microsoft/qsharp-runtime/pull/520#discussion_r577811514) but yes, CLI11.hpp is where `CLI::App` comes from.

cesarzc:
Thank you @irinayat-MS and @cgranade for your comments! I'll address them shortly.

I just wanted to mention that this PR is still work in progress so some things might change considerably.

src/QirRuntime/test/Standalone/InputTypes/qir-driver.cpp
irinayat-MS: FYI: if you run clang formatter, it will apply the style automatically

src/QirRuntime/test/Standalone/Shared/CLI11.hpp
kuzminrobin: @cesarzc, FYI: We have a `qsharp-runtime\src\QirRuntime\externals` folder for external components.

src/QirRuntime/test/Standalone/RandomBit/compiler/QirCore.qs
```diff
@@ -0,0 +1,30 @@
...
+    function RangeEnd (range : Range) : Int { body intrinsic; }
+
+    function RangeReverse (range : Range) : Range { body intrinsic; }
+}
+
+namespace Microsoft.Quantum.Targeting {
+
+    @Attribute()
+    newtype TargetInstruction = String;
+}
```
kuzminrobin, line 30: Are you sure you need _all these_? 
My understanding was that the `.qs` files inside of this `compiler` directory define _those Q# entities only_ (functions, operations, etc.) that your `.qs` test (e.g. `random-bit.qs`) _uses_, but those are not implemented in the compiler's libraries. I see that your .qs test uses H() and M(), so likely you only need the H() and M() definitions in `compiler/QirTarget.qs`.

src/QirRuntime/test/Standalone/InputTypes/qir-driver.cpp
```diff
@@ -0,0 +1,145 @@
...
+        if (!simulatorOutputFileOpt->empty()){
+            simulatorOutputFileBuffer.close();
+        }
+
+        if (!operationOutputFileOpt->empty()){
+            operationOutputFileBuffer.close();
+        }
+    } catch(...) {
+        return 1;
+    }
```
kuzminrobin, line 142: It is typically easier to catch the unexpected exception bugs if the exceptions are _not caught_. In that case the app either immediately crashes upon the unexpected exception (if the debugger is not attached), or breaks in the debugger at the moment the exception is thrown (if the debugger is attached).
Currently your code catches all the exceptions, prints _nothing to console_, and just returns 1. This can make an impression that the test has run successfully (while it has terminated upon unexpected exception).
You may also find useful the macros of Catch2 test library, see `REQUIRE(..)`, `REQUIRE_THROWS_AS(..)` in the other tests and [here](https://github.com/catchorg/Catch2/blob/devel/docs/assertions.md#top).

src/QirRuntime/test/Standalone/InputTypes/qir-driver.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+        // TODO: Remove this after the Message Q# function is integrated into the QIR runtime.
+        (*simulatorOutputStream) << anInt << std::endl;
+        (*simulatorOutputStream) << aDouble << std::endl;
+        (*simulatorOutputStream) << aBool << std::endl;
+        for (int64_t n : anIntegerArray) {
+            (*simulatorOutputStream) << n << " ";
+        }
+
+        (*simulatorOutputStream) << std::endl;
```
kuzminrobin, line 121: FYI: `Message()` has been pushed to `main` in #511 ([PR](https://github.com/microsoft/qsharp-runtime/pull/511)).

kuzminrobin:
I'm done with the first pass of this PR. I didn't dive deep into the code that uses CLI11 library.

I have a single GitHub account tied with _my own_ email at gmail.com (which I check rarely, up to couple of times per day). So if you need my attention to a PR or an issue/bug open on GitHub, then please notify me over Teams or MS email.

src/QirRuntime/test/Standalone/InputTypes/qir-driver.cpp
irinayat-MS: If we want this sample to look as close as possible to the code a user might produce and, thus, would rather avoid testing it with Catch2, we can still test the final binary through ctest (and we probably should).

src/QirRuntime/test/CMakeLists.txt
cesarzc: Created a new top level folder called `samples` and added a sub-folder called `StandaloneInputReference` for the input reference.

src/QirRuntime/test/Standalone/InputTypes/CMakeLists.txt
cesarzc: Simplified.

src/QirRuntime/test/Standalone/InputTypes/CMakeLists.txt
cesarzc: Removed test dependency.

src/QirRuntime/test/Standalone/InputTypes/compiler/QirCore.qs
cesarzc: Using a Q# project and the `QirGeneration` property now.

src/QirRuntime/test/Standalone/InputTypes/qir-driver.cpp
cesarzc: Not using the not-exposed factory anymore.

src/QirRuntime/test/Standalone/InputTypes/generate.py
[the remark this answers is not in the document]
cesarzc: Removed.

cesarzc: Removed.

src/QirRuntime/test/Standalone/InputTypes/generate.py
cesarzc: File removed.

src/QirRuntime/test/Standalone/InputTypes/input-types.qs
cesarzc: Renamed in the new file location.

src/QirRuntime/test/Standalone/RandomBit/compiler/QirCore.qs
cesarzc: They are not needed anymore. I used them in the past to generate QIR (before the change to produce them using a Q# project was merged to `main`. I removed them in the latest iteration of changes.

src/QirRuntime/test/Standalone/InputTypes/qir-driver.cpp
cesarzc: Thanks! I added that to the most recent changes and it works great :)

src/QirRuntime/.clang-tidy
```diff
@@ -26,4 +26,3 @@ CheckOptions:
     value:  'CamelCase'
   - key:    readability-identifier-naming.NamespaceCase
     value:  'CamelCase'
-
```
irinayat-MS, line 29: Nit: if you are not making intentional changes to the files, maybe, revert them, so no changes at all?

src/QirRuntime/samples/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+set(standalone_includes ""${PROJECT_SOURCE_DIR}/externals/CLI11"")
```
irinayat-MS, line 1: Do we expect this path to be used by multiple samples?

src/QirRuntime/samples/StandaloneInputReference/.clang-tidy
```diff
@@ -0,0 +1,3 @@
+Checks:
```
irinayat-MS, line 1: As commented previously, would it be possible to avoid introducing a different clang-tidy file but instead filter out the header with --header-filter and/or --line-filter settings?

src/QirRuntime/samples/StandaloneInputReference/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+  ${QIR_BRIDGE_QIS_UTILITY_LIB}
+  qir-rt-support
+  qir-qis-support
+  simulators
+)
+
+target_include_directories(qir-input-reference-standalone PUBLIC
+  "${standalone_includes}"
+  "${public_includes}"
+  "${PROJECT_SOURCE_DIR}/lib/QIR"
```
irinayat-MS, line 24: You shouldn't need this one, unless you call quantum__rt* or quantum__qis* methods directly, because I've recently moved QirTypes.hpp into public.

src/QirRuntime/samples/StandaloneInputReference/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+)
+
+add_dependencies(qir-input-reference-standalone qir_standalone_input_reference_target)
+
+install(TARGETS qir-input-reference-standalone RUNTIME DESTINATION "${CMAKE_BINARY_DIR}/bin")
```
irinayat-MS, line 29: We should probably have some tests for the sample, even if not using catch2. Something that at least runs the built binary...
```
    add_test(NAME ${target} COMMAND ${target}
    )
```
But I don't know what kind of output it should produce to be visible in CI...

src/QirRuntime/samples/StandaloneInputReference/qir-driver.cpp
```diff
@@ -0,0 +1,126 @@
...
+        CLI::Option *operationOutputFileOpt = app.add_option(
+            "-o,--operation-output",
+            operationOutputFile,
+            "File to write the output of the Q# operation to");
+
+        // Add the options that correspond to the parameters that the QIR entry-point needs.
+        // Option for a Q# Int type.
+        int64_t anInt = 0;
+        CLI::Option *anIntOpt = app.add_option(
+            "--anint",
```
irinayat-MS, line 39: Just curious, why include article in the name of the argument? It doesn't make sense for non-English speakers and makes the names harder to parse...

src/QirRuntime/samples/StandaloneInputReference/qir-driver.cpp
```diff
@@ -0,0 +1,126 @@
...
+        std::ostream* simulatorOutputStream = &std::cout;
+        std::filebuf simulatorOutputFileBuffer;
+        if (!simulatorOutputFileOpt->empty()){
+            simulatorOutputFileBuffer.open(simulatorOutputFile, std::ios::out);
+            std::ostream simulatorOutputFileStream(&simulatorOutputFileBuffer);
+            // TODO: Call into the QIR runtime API to redirect the output of the simulator.
+            simulatorOutputStream = &simulatorOutputFileStream;
+        }
+
+        // Redirect the Q# opertion output from std::cout if the --operation-output option is present.
```
irinayat-MS, line 91: typo: opertion -> operation

src/QirRuntime/samples/StandaloneInputReference/qir-driver.cpp
```diff
@@ -0,0 +1,126 @@
...
+extern "C" void Quantum__StandaloneSupportedInputs__ExerciseInputs__body( // NOLINT
+    int64_t anInt,
+    double aDouble);
+
+int main(int argc, char *argv[])
+{
+    try{
+        CLI::App app("QIR Standalone Entry Point Inputs Reference");
+
+        // Add the --simulator-output and --operation-output options.
```
irinayat-MS, line 21: What is the difference between the simulator and operation output?

src/QirRuntime/samples/StandaloneInputReference/qir-driver.cpp
```diff
@@ -0,0 +1,126 @@
...
+        std::ostream* operationOutputStream = &std::cout;
+        std::filebuf operationOutputFileBuffer;
+        if (!operationOutputFileOpt->empty()){
+            operationOutputFileBuffer.open(operationOutputFile, std::ios::out);
+            std::ostream operationOutputFileStream(&operationOutputFileBuffer);
+            operationOutputStream = &operationOutputFileStream;
+        }
+
+        // TODO: Remove this after the Message Q# function is integrated into the QIR runtime.
+        (*operationOutputStream) << "SIMULATOR OUTPUT\n----------------" << std::endl;
```
irinayat-MS, line 101: It's been integrated now, right?

src/QirRuntime/samples/StandaloneInputReference/qir-driver.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+        // Close opened file buffers;
+        if (!simulatorOutputFileOpt->empty()){
+            simulatorOutputFileBuffer.close();
+        }
+
+        if (!operationOutputFileOpt->empty()){
+            operationOutputFileBuffer.close();
+        }
+    } catch(...) {
```
irinayat-MS, line 122: Why not let it throw and terminate with, hopefully, a more specific message?

src/QirRuntime/samples/StandaloneInputReference/qir-standalone-input-reference.ll
```diff
@@ -0,0 +1,59 @@
+
+%Result = type opaque
```
irinayat-MS, line 2: Is it a manually authored LL file? Why does it need to be checked in?

src/QirRuntime/test/Standalone/InputTypes/.clang-tidy
cesarzc: I added `InheritParentConfig: true` to the `.clang-tidy` file for the standalone samples, it is an opt-in check for that directory.

Ideally, `HeaderFilterRegex` should have a negative lookahead regex to exclude particular files and include the rest but this is not currently supported. There has been multiple attempts to do this ([here](https://reviews.llvm.org/D34654) and [here](http://clang-developers.42468.n3.nabble.com/clang-tidy-Negative-Lookahead-Support-td4048783.html)) but its seems like they got no traction.

The other option was `--line-filter` but that is not supported as a configuration file option either.

src/QirRuntime/test/Standalone/Shared/CLI11.hpp
cesarzc: Followed up on this offline, added it to the manifest, moved it to the `externals` folder, and we should be good to go now. We have to add it to NOTICE.txt once we start distributing it as part of the QDK.

src/QirRuntime/test/Standalone/InputTypes/qir-driver.cpp
cesarzc: Moved the _CLI11.hpp_ file to the _externals_ folder and added the component to _cgmanifest.json_. We don't have to add it to the Third-Party notice since we are not distributing it through any of our packages yet. Regarding the copy of the file, I still think the header-only distribution of this library is the most convenient one so I kept a copy of the most recent release of the header.

src/QirRuntime/test/Standalone/InputTypes/qir-driver.cpp
cesarzc: Ran clang formatter :)

src/QirRuntime/test/Standalone/InputTypes/input-types.qs
cesarzc: Added readme.md.

src/QirRuntime/test/Standalone/InputTypes/generate.py
cesarzc: Since this has been consolidated, and not directly related to the contents of this PR, I'll resolve this comment.

src/QirRuntime/.clang-tidy
cesarzc: Reverted the changes to .clang-tidy.

src/QirRuntime/samples/StandaloneInputReference/.clang-tidy
cesarzc: Used `InheritParentConfig: true` in the new file.

Unfortunately `--line-filter` is not supported in the configuration file, so including headers in clang-tidy checks is opt-in through `HeaderFilterRegex`.

src/QirRuntime/samples/StandaloneInputReference/CMakeLists.txt
cesarzc: I am using the _quantum__qis_internal.hpp_ header to use the `Quantum::Qis::Internal::SetOutputStream` API so I need _lib/QIR_ as include directory.

src/QirRuntime/samples/StandaloneInputReference/qir-driver.cpp
cesarzc: Yes, this code has been updated to call into the `Quantum::Qis::Internal::SetOutputStream` API.

src/QirRuntime/samples/StandaloneInputReference/qir-driver.cpp
cesarzc: Fixed it.

src/QirRuntime/samples/StandaloneInputReference/qir-driver.cpp
cesarzc: Fixed.

src/QirRuntime/test/Standalone/InputTypes/qir-driver.cpp
cesarzc: As Irina mentioned, the goal is to make this sample as close as possible to the use that will be produced automatically produced so we shouldn't use Catch2 in this case. I removed the global try-catch and will add a test as suggested by Irina in this [comment](https://github.com/microsoft/qsharp-runtime/pull/520#discussion_r581438121).

src/QirRuntime/samples/StandaloneInputReference/qir-driver.cpp
cesarzc: I removed the global try-catch.

src/QirRuntime/samples/StandaloneInputReference/qir-driver.cpp
```diff
@@ -0,0 +1,141 @@
...
+#include "QirContext.hpp"
+#include "QirTypes.hpp"
+#include "QuantumApi_I.hpp"
+#include "SimFactory.hpp"
+#include "quantum__qis_internal.hpp"
+#include "quantum__rt.hpp"
+#include <cctype>
+#include <iostream>
+#include <map>
+#include <vector>
```
kuzminrobin, line 15: I know that clang-format by default recommends the order of the headers inclusion _opposite_ to the order of headers currently in our files. I know the reasons behind both options.
As I remember, this topic is not adjusted explicitly in our [.clang-format](https://github.com/microsoft/qsharp-runtime/blob/main/src/QirRuntime/.clang-format), at least the option [`SortIncludes`](https://clang.llvm.org/docs/ClangFormatStyleOptions.html) is not specified explicitly.

For the files added by me I preferred to stay with the same order that's in other files.

For your case you may consider 
* either the same or 
* listing the most specific headers first, and then the more general ones (rather than sorting "CLI11.hpp" in the same group as other ".." files). That is you might consider placing "CLI11.hpp" after the other ".." files but before the <.. > files.

src/QirRuntime/samples/StandaloneInputReference/qir-driver.cpp
```diff
@@ -0,0 +1,141 @@
...
+        "-o,--operation-output", operationOutputFile, "File where the output of the Q# operation is written");
+
+    // Add the options that correspond to the parameters that the QIR entry-point needs.
+    // Option for a Q# Int type.
+    int64_t intValue = 0;
+    app.add_option("--int-value", intValue, "An integer value")->required();
+
+    // Option for a Q# Double type.
+    double_t doubleValue = 0.0;
+    app.add_option("--double-value", doubleValue, "A double value")->required();
```
kuzminrobin, line 63: The fragment `->required()` makes me think that the test will not run if the `--double-value` is not specified in the command line. Is that the case? Does the test really require that value to run? Some default value like `0.0` will not work?
If that `--double-value` is not specified then what will the test return to command line? Will that be interpreted as the _failure_ of the test?

src/QirRuntime/samples/StandaloneInputReference/qir-driver.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+    // Redirect the Q# operation output from std::cout if the --operation-output option is present.
+    ostream* operationOutputStream = &cout;
+    filebuf operationOutputFileBuffer;
+    if (!operationOutputFileOpt->empty())
+    {
+        operationOutputFileBuffer.open(operationOutputFile, ios::out);
+        ostream operationOutputFileStream(&operationOutputFileBuffer);
+        operationOutputStream = &operationOutputFileStream;
+    }
```
kuzminrobin, line 123: I'm not strong at streams, but at a glance I have an impression that
* the `operationOutputFileStream` is a local variable on the stack that will survive until the nearest `}`
* the `operationOutputStream` starts pointing to that local variable on the stack (in line 122);
* upon `}` the `operationOutputFileStream` goes out of scope and gets destroyed, after which the `operationOutputStream` becomes a dangling pointer (pointing to the non-existent variable on the stack).

If I'm mistaken then at least a comment clarifying the situation, would help.

src/QirRuntime/samples/StandaloneInputReference/qir-driver.cpp
kuzminrobin: Sorry. Looks like this is not a test, this is a sample.

kuzminrobin:
[This item](https://github.com/microsoft/qsharp-runtime/pull/520#discussion_r587925655) seems to me risky, requires resolution. Otherwise seem to me all right to merge.

src/QirRuntime/samples/StandaloneInputReference/qir-driver.cpp
cesarzc: Great catch! I changed it, and now it uses an `ofstream` that can be declared in the scope of the `main` function and opened inside the `if`.

src/QirRuntime/samples/StandaloneInputReference/qir-driver.cpp
cesarzc: I changed the order to be consistent with the order of headers currently in our other cpp files.

src/QirRuntime/samples/StandaloneInputReference/CMakeLists.txt
cesarzc: Added a test for the sample in the cmake file and also in the test.py file.

src/QirRuntime/externals/cgmanifest.json
```diff
@@ -17,6 +17,14 @@
...
                     "CommitHash": "87074da73ecb1c2e7c35fd14f50ca21c4c002adc" }
             }
+        },
+        {
+            "Component": {
+                "Type": "git",
+                "Git": { 
+                    "RepositoryUrl": "https://github.com/CLIUtils/CLI11", 
+                    "CommitHash": "5cb3efabce007c3a0230e4cc2e27da491c646b6c" }
+            }
```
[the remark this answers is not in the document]
cesarzc: This is used by component governance which can send us alerts in case security vulnerabilities are found in any of the components we declare.

src/QirRuntime/samples/StandaloneInputReference/qir-driver.cpp
cesarzc: The simulation output is the one produced by calls to the `Message` Q# API. The operation output is what the Q# entry-point operation returns. In this case it is an `Int` but it can be a `Result`, an array of `Results` or whatever the a Q# entry-point operation can return.

src/QirRuntime/samples/StandaloneInputReference/qir-standalone-input-reference.ll
cesarzc: Removed the auto-generated file.

cesarzc:
> 
> 
> [This item](https://github.com/microsoft/qsharp-runtime/pull/520#discussion_r587925655) seems to me risky, requires resolution. Otherwise seem to me all right to merge.

Thanks for the review Robin! Fixed the issue and it should be ready to merge now.

src/QirRuntime/samples/StandaloneInputReference/qir-driver.cpp
```diff
@@ -0,0 +1,269 @@
...
+#include "CoreTypes.hpp"
+#include "QirContext.hpp"
+#include "QirTypes.hpp"
+#include "QuantumApi_I.hpp"
+#include "SimFactory.hpp"
+
+#include "quantum__qis_internal.hpp"
+#include "quantum__rt.hpp"
+
+#include "CLI11.hpp"
```
kuzminrobin, line 19: Consider placing this \#include between the <> includes and "" includes (between `#include <vector>` and `#include "CoreTypes.hpp"`). That is if we are ordering the headers from the most general to the least general.

src/QirRuntime/samples/StandaloneInputReference/qir-driver.cpp
```diff
@@ -0,0 +1,269 @@
...
+
+// This is the function corresponding to the QIR entry-point.
+extern "C" int64_t Quantum__StandaloneSupportedInputs__ExerciseInputs__body( // NOLINT
+    int64_t intValue,
+    double doubleValue,
+    Result resultValue,
+    QirString* stringValue,
+    int64_t rangeStart,
+    int64_t rangeStep,
+    int64_t rangeEnd);
```
kuzminrobin, line 32: (Up to you)
I'm comparing this C/C++ signature, the _range_ part
> Quantum__StandaloneSupportedInputs__ExerciseInputs__body(
    . . .
    **int64_t rangeStart,
    int64_t rangeStep,
    int64_t rangeEnd**);

with the `.qs` signature 
> operation ExerciseInputs (. . ., **rangeValue : Range**)

and with the generated `.ll` output
> %Range = type **{ i64, i64, i64 }**
define i64 @Quantum__StandaloneSupportedInputs__ExerciseInputs__body(. . ., **%Range %rangeValue**)

The C/C++ signature looks somewhat _hacky_ ;-)
Would be nice to see the comments in the code explaining that the C/C++ signature is a conscious choice and is compatible with `.ll` (I didn't dive deep but in those comments I would welcome some analysis and mentioning of liknks like these - 
https://llvm.org/docs/LangRef.html#structure-types , 
https://llvm.org/docs/LangRef.html#preallocated-operand-bundles
https://llvm.org/docs/LangRef.html#structure-type
).

The following C/C++ signature seems to me somewhat safer (if it doesn't work then would be nice to see in the comments the explanation why it does not work)
> Quantum__StandaloneSupportedInputs__ExerciseInputs__body(
    . . .
    **QirRange range**);

src/QirRuntime/samples/StandaloneInputReference/qir-driver.cpp
```diff
@@ -0,0 +1,269 @@
...
+
+template<typename D, typename S>
+void TranslateVectorToBuffer(D** destinationBuffer, vector<S>sourceVector, function<D(S)> translationFunction)
+{
+    *destinationBuffer = new D[sourceVector.size()];
+    for (int index = 0; index < sourceVector.size(); index++)
+    {
+        (*destinationBuffer)[index] = translationFunction(sourceVector[index]);
+    }
+}
```
kuzminrobin, line 73: (If you have time and desire)

To make a step to more memory-safe code, you may want to consider returning a `unique_ptr<D>` instead of updating the `*destinationBuffer`.
Otherwise currently in the fragments like this
```c++
    bool* boolArray = nullptr;
    TranslateVectorToBuffer<bool, char>(&boolArray . . .
```
the `boolArray` is not deallocated. 
If it was happening not in `main()` but in some other function then we would be leaking that mem.

The change could be something like this:
> template<typename D, typename S>
**unique_ptr\<D>** TranslateVectorToBuffer(~~D** destinationBuffer,~~ vector\<S>sourceVector,
{
    **unique_ptr\<D> retVal {** new D[sourceVector.size()] };

Use:
> **unique_ptr\<bool>**  boolArray = ~~nullptr;~~
TranslateVectorToBuffer<bool, char>( ~~\&boolArray,~~ boolAsCharVector, TranslateCharToBool);
QirArray* qirboolArray = CreateQirArray(boolArray **.get()**, boolAsCharVector.size());

src/QirRuntime/samples/StandaloneInputReference/qir-driver.cpp
```diff
@@ -0,0 +1,269 @@
...
+    // Close opened file buffers;
+    if (simulationOutputFileStream.is_open())
+    {
+        simulationOutputFileStream.close();
+    }
+
+    if (operationOutputFileStream.is_open())
+    {
+        operationOutputFileStream.close();
+    }
```
kuzminrobin, line 268: (Up to you)
Just in case I would `close()` in the reverse order of `open()`. 

Some slightly related info is [here](https://github.com/kuzminrobin/code_review_notes/blob/master/cpp_design_bookmarks.md#know-about-the-compilers-resource-allocation-and-deallocation-order).

kuzminrobin (approved):
Approving with minor recommendations.

src/QirRuntime/samples/StandaloneInputReference/qir-driver.cpp
cesarzc: Changed it to the placing that you suggested.

src/QirRuntime/samples/StandaloneInputReference/qir-driver.cpp
cesarzc: Changed the ordering.

src/QirRuntime/samples/StandaloneInputReference/qir-driver.cpp
cesarzc: I looked more into this and provided some detailed analysis offline. In summary, it seems like the expectations for a QIR signature like the following:
```
define i64 @Quantum__StandaloneSupportedInputs__ExerciseInputs__body(i64 %intValue, double %doubleValue, %Result* %resultValue, %String* %stringValue, %Range %rangeValue)
```
Where `%Range` is defined as:
```
%Range = type { i64, i64, i64 }
```
Align with the expectations of a C signature like the following:
```
extern "C" int64_t Quantum__StandaloneSupportedInputs__ExerciseInputs__body( // NOLINT
 int64_t intValue,
 double doubleValue,
 Result resultValue,
 QirString* stringValue,
 int64_t rangeStart,
 int64_t rangeStep,
 int64_t rangeEnd);
```
But do not align with the expectations of a C signature like the following:
```
extern "C" int64_t Quantum__StandaloneSupportedInputs__ExerciseInputs__body( // NOLINT
 int64_t intValue,
 double doubleValue,
 Result resultValue,
 QirString* stringValue,
 QirRange rangeValue);
```
I changed the entry-point operation to not receive a Range for now, and I'll suggest to change QIR generation in this case to receive a pointer to QirRange rather than passing it by value since pointers are maybe the most reliable way to interop between C++ and QIR for struct values.

cesarzc: We need to expand the arguments received by the QIR operation anyway when the latest compiler entry-point mapping is merged so we can remove this hacky part for now to not confuse people.

kuzminrobin: OK, thank you for the offline analysis, Cesar.
Looks like compiler operates not exactly the way I expected.

PR #522, opened 2021-02-17
Generate QIR during pipeline

swernli:
This updates the QirRuntime pipeline to expect the qir-static.ll file to be generated at runtime instead of checked in, now that we can support cross-platform QIR generation properly.

This also updates the pipeline to use pwsh instead of powershell to get uniform behavior on Windows.


swernli:
@irinayat-MS Good suggestions! Adding the .ll to the artifacts and building such that it matches the configuration in the environment should be easy updates.

src/QirRuntime/build-qir-runtime.ps1
```diff
@@ -2,6 +2,17 @@
...
 
 if ($Env:ENABLE_QIRRUNTIME -eq "true") {
+    Write-Host "##[info]Compile Q# Projects into QIR"
+    $qirStaticPath = Join-Path $PSScriptRoot test QIR-static qsharp
+    dotnet build $qirStaticPath -c $Env:BUILD_CONFIGURATION -v $Env:BUILD_VERBOSITY
+    if ($LastExitCode -ne 0) {
+        Write-Host "##vso[task.logissue type=error;]Failed to compile Q# project at '$qirStaticPath' into QIR."
+        return
+    }
+    Copy-Item -Path (Join-Path $qirStaticPath qir *.ll) -Destination (Split-Path $qirStaticPath -Parent)
```
guenp, line 12: Should we perhaps delete the currently generated QIR files that are in the `/src/QirRuntime/test/QIR-static` folder to avoid confusion?

guenp (approved):

src/QirRuntime/build-qir-runtime.ps1
irinayat-MS: Yes, it would be good to delete the _one_ file that is now being generated (the second file was manually authored and covers functionality not supported by Q# but present in QIR spec)

build/steps.yml
```diff
@@ -5,18 +5,18 @@ steps:
 - template: steps-init.yml
 
 
-- powershell: ./build.ps1
+- pwsh: ./build.ps1
```
irinayat-MS, line 8: for my education: what's the difference and why to use pwsh over powershell?

irinayat-MS (approved):
Awesome! Thanks for doing this!

src/QirRuntime/build-qir-runtime.ps1
swernli: This change does remove that file.

build/steps.yml
cgranade: Up to and including version 5.1, PowerShell was based on .NET Framework and called "Windows PowerShell." With version 6.0, they renamed to "PowerShell Core" and based on .NET Core instead. With version 7.0, they renamed again to "PowerShell" with no other qualifications, mirroring the ".NET Framework" → ".NET Core" → ".NET 5" naming path. In Azure Pipelines, the `powershell` task gives PowerShell 5.1 on Windows agents, even though that version is now long deprecated and has some compatibility bugs that impede cross-platform support. Whenever possible, using `pwsh` is suggested to ensure that 6.0 or later is used.

src/QirRuntime/build-qir-runtime.ps1
```diff
@@ -2,6 +2,17 @@
 # Licensed under the MIT License.
 
 if ($Env:ENABLE_QIRRUNTIME -eq "true") {
+    Write-Host "##[info]Compile Q# Projects into QIR"
+    $qirStaticPath = Join-Path $PSScriptRoot test QIR-static qsharp
```
cgranade, line 6: Beyond the scope of this PR, but we may want to consider in the future making this rely on PowerShell cmdlet functions that wrap `dotnet build` and ensure that arguments such as build configuration and assembly versions are set from environment variables in a uniform way.

build/steps.yml
irinayat-MS: Thanks for the explanations, Chris!

cgranade: No worries, happy to help!

Issue #519, opened 2021-02-16
QirRuntime: qdk.dll has hardcoded dependency on fullstate simulator


irinayat-MS:
To enable running QIR against qdk.dll I added `CreateFullstateSimulator`, but it might cause other clients to define the function, even if they are not using it (e.g. see https://github.com/microsoft/qsharp-runtime/pull/475#discussion_r574224017). Should review the direction of dependencies and/or how the qdk.dll is built.

PR #524, opened 2021-02-18
Cleanup qdk.dll build

irinayat-MS:
Fixes https://github.com/microsoft/qsharp-runtime/issues/519

Also moved a couple headers from lib/QIR into public and took the opportunity to replace <assert.h> and <string.h> includes with <cassert> and <cstring> respectively.


src/QirRuntime/lib/qdk/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
+#===============================================================================
+# Produce the qdk dynamic library
+#
+add_library(qdk SHARED
+  qdk.cpp
+  $<TARGET_OBJECTS:qir-rt-support-obj>
+  $<TARGET_OBJECTS:qir-qis-support-obj>
+  $<TARGET_OBJECTS:simulators-obj>
```
cgranade, line 8: Just as a side note to validate my own understanding, at this point the `add_subdirectory(Simulators)` directive from earlier has run such that simulators have been built by this point in the make? d496b03a25428e91c10e14450c34208823bd3fc6 from the **experimental** open systems simulator / QIR integration branch depends on that, so just wanted to make sure I understood. Thanks for the help!

src/QirRuntime/test/QIR-static/qir-gen.ll
```diff
@@ -1,1690 +1,1691 @@
-%Result = type opaque
```
swernli, line 1: I thought this file was deleted by my change in #522 which should have updated the gitignore too. How is it ending up in the diff here?

swernli (approved):
Clean up looks good, just one question about why the .ll file seems to have reappeared.

src/QirRuntime/lib/qdk/CMakeLists.txt
irinayat-MS: Yes, I believe, you understanding is correct.

src/QirRuntime/test/QIR-static/qir-gen.ll
irinayat-MS: Good question. Don't know. I deleted the file and rebuilt locally again, and it was correctly filtered out by .gitignore this time... Thanks for catching this!

Oh, I see! @kuzminrobin change https://github.com/microsoft/qsharp-runtime/commit/ec763fbeb513b85beebc9985ae8e0d40162f573c added the file back. Likely, because it was updated compared to the pre-delete state and merge decided to resurrect it... I guess, we'll need to keep an eye on this for a while.

kuzminrobin: Looks like during rebasing my branch, I resolved the conflict incorrectly such that the file was recovered.
I assume that now this file needs to be always generated (if we build the tests in a clean directory).

kuzminrobin:
Approving.
Sorry for being late with reviewing this PR.
(I have a single GitHub account tied up with my personal email, and I get notifications to my personal email which I check rarely. Feel free to ping me any time (over Teams or MS email) and remind me about the PR waiting for me)

PR #527, opened 2021-02-23
QirRuntime: support ApplyIf* and ApplyConditionally* intrinsics

irinayat-MS:
Also enabled conditional rewrites for the QIR-static tests.


src/QirRuntime/test/QIR-static/qsharp/qir-test-conditionals.qs
```diff
@@ -0,0 +1,72 @@
...
+        // Other variants
+        Adjoint Microsoft.Quantum.Simulation.QuantumProcessor.Extensions.ApplyConditionallyA(
+            [r1], [r2], (Y, q1), (X, q1));
+        Controlled Microsoft.Quantum.Simulation.QuantumProcessor.Extensions.ApplyConditionallyC(
+            [q2], ([r1], [r2], (Y, q1), (X, q1)));
+        Adjoint Controlled Microsoft.Quantum.Simulation.QuantumProcessor.Extensions.ApplyConditionallyCA(
+            [q2], ([r1], [r2], (Y, q1), (X, q1)));
+    }
+
+    operation TestConditionalRewrite() : Unit {
```
swernli, line 60: Because the rest of the tests call into the conditional functions directly, this is the only test that depends on the rewrite step. But by changing the settings on the project in QIR-static/qsharp, now all if-statements will undergo a rewrite. I think it would be better to leave the existing project unchanged and include the test cases that explicitly call the ApplyIf variants, and then have a separate project that includes this test case that uses the execution target and suppressed C# generation as the place to test target dependent transformations.

src/QirRuntime/lib/QIR/conditionals.cpp
```diff
@@ -0,0 +1,112 @@
...
+    if (clb != clbFunc)
+    {
+        quantum__rt__callable_update_reference_count(clbFunc, -1);
+    }
+    quantum__rt__tuple_update_reference_count(argsTuple, -1);
+}
+
+static bool ArraysContainEqualResults(QirArray* rs1, QirArray* rs2)
+{
+    assert(rs1 != nullptr && rs2 != nullptr && rs1->count == rs2->count);
```
kuzminrobin, line 41: As for `assert( .. , rs1->count == rs2->count)` (comparing the array counts). 
I assume this fragment is only called by _our code_ (compiler generated or written manually by us, etc.), the _user_ never calls this fragment (directly or indirectly). And I assume that if the comparison/assertion fails then it definitely means a bug in our code. That is why it is the right thing to assert if the arrays have different count.
(But if the _user_ can call this fragment (directly or indirectly) and pass the arrays with different count then I'm not sure if assertion is the right thing to do)

src/QirRuntime/test/QIR-static/qir-test-conditionals.cpp
```diff
@@ -0,0 +1,111 @@
...
+// TestConditionalOnResult() is authored in a way that the expected path through the function only applies X operator
+// for the chosen sequence of measurement results, and all other paths apply Y. Thus, the correct execution must get the
+// expected maximum number of X and ControlledX callbacks.
+struct ConditionalsTestSimulator : public Microsoft::Quantum::SimulatorStub
+{
+    int cX = 0;
+    int ccX = 0;
+
+    vector<ResultValue> mockMeasurements;
+    int nMeasure = -1;
```
kuzminrobin, line 27: FYI: I suspect you meant `0` instead of `-1`.

src/QirRuntime/test/QIR-static/qsharp/qir-gen.csproj
```diff
@@ -3,7 +3,9 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>netcoreapp3.1</TargetFramework>
+    <CSharpGeneration>false</CSharpGeneration>
     <QirGeneration>True</QirGeneration>
+    <ExecutionTarget>honeywell.qpu</ExecutionTarget>
```
kuzminrobin, line 8: Can you please provide some more details about this line?

kuzminrobin (approved):

src/QirRuntime/test/QIR-static/qsharp/qir-test-conditionals.qs
irinayat-MS: Agree. Removed the rewrite setting from the project and will only keep the direct tests for ApplyIf* and ApplyConditionally*

src/QirRuntime/test/QIR-static/qir-test-conditionals.cpp
irinayat-MS: No, I do mean -1 here as I increment first thing the Measure callback. Given that you've found it confusing, I'll change the code to start from 0.

src/QirRuntime/test/QIR-static/qsharp/qir-gen.csproj
irinayat-MS: This is the target that triggers conditional rewrites in the compiler. As discussed in https://github.com/microsoft/qsharp-runtime/pull/527#discussion_r581454320 I'm removing it.

src/QirRuntime/lib/QIR/conditionals.cpp
irinayat-MS: It's a good question of at what layer of the code (Q#, QIR, QirRuntime) checks like this should be done. I can easily replace the assert with fail, but I don't think that "defensive" coding is _always_ the right answer. We should think through where we can draw the boundaries that would allow us to assume the inputs has been already checked... For now, I'll keep the checks as asserts.

swernli:
@irinayat-MS The change at #529 has merged, so you can incorporate that here now if you are ready.

Issue #451, opened 2020-11-25
Add a default simulator option to the EntryPoint attribute


Strilanc:
Some quantum methods have to run under the Toffoli simulator. E.g. running a 1000 qubit adder under the quantum simulator is a recipe for freezing your computer as it tries to allocate the universe. It would be convenient if the entry point attribute was able to indicate the default simulator to use, to avoid that particular gotcha.

** Describe the solution you'd like **

```
    @EntryPointWithDefaultSimulator("ToffoliSimulator")
    operation ExampleRequiringToffoliSimulator() : Unit {
        ...
    }
```

samarsha:
There's a [`DefaultSimulator` property](https://github.com/microsoft/qsharp-compiler/blob/f21165084cd2a85924febe9f268435c76bec15cb/src/QuantumSdk/README.md#defined-project-properties) that you can put in your .csproj file:

```xml
<PropertyGroup>
  <DefaultSimulator>ToffoliSimulator</DefaultSimulator>
</PropertyGroup>
```

This would apply to the whole Q# project, but projects can only have one entry point, so it should be equivalent. Would this work for you, or do you still prefer to have it on the `EntryPoint` attribute instead?

bettinaheim:
I think this should be covered by the project property as @samarsha mentions. I'll assign it to myself until we get confirmation.

Strilanc:
I would explicitly prefer it to be in the code in this case, same as for
tests. It's pretty rare to include your project file in example code blocks
on the internet.

samarsha:
Another thing to consider is that there has been some discussion about possibly making Toffoli a separate execution target with its own target packages, where certain gates would not be supported at compile time instead of being runtime errors like they are now. That would work better if Toffoli was set in the project config instead of at the entry point, although it could be possible to support both...

@swernli @bettinaheim Do you have any thoughts about this?

swernli:
[...]

Issue #357, opened 2020-08-25
CSharpGeneration can't handle QsharpCompile of files with same filename


swernli:
When using `QsharpCompile` in a csproj to explicitly pull in files not in the current folder tree, the resulting C# generated files will end up all in the same folder, namely `obj\qsharp\src\*.g.cs`. However, if two files in different folders have the same filename, such as `..\lib1\utils.qs` and `..\lib2\utils.qs` both try to generate to `obj\qsharp\src\utils.g.cs` and whichever is compiled second ends up overwriting the first, causing Not Found errors during C# compilation for anything expected to be generated from the first file.

I'm not really sure how to resolve this, as the behavior of preserving folder hierarchy for generated files doesn't make sense when considering files that are not in the same hierarchy. But if it can't be supported it should at least generate an explicit error instead of the cryptic side effects.

ScottCarda-MS:
Fixed with #588.

PR #528, opened 2021-02-24
Multiple Entry Points For Entry Point Driver 

ScottCarda-MS:
Refactors the Entry Point Driver and C# generation to accommodate Q# code having multiple entry points.

Major Changes:
 - Updated C# generation:
   - Now generate a unique entry point class for each entry point. These classes will no longer contain a `Main` method.
   - The `Main` method is generated in its own namespace and class: `__QsEntryPoint__.__QsEntryPoint__`.
   - Added a public function `generateMainSource` to the C# gen's `EntryPoint` module. This function parallels `generateSource` in the same module and generates the content for the C# gen's Main file.
   - The outputted files for C# gen are now:
     - the `.g.cs` for the C# of the Q# operations (same as before)
     - `.g.EntryPoint.cs` instead of `EntryPoint.g.cs` for generated entry point classes
     - `EntryPoint.g.Main.cs` for the name of the new file containing the `Main` method.
 - `Microsoft.Quantum.EntryPointDriver.Azure.Submit` was made public as it is used in the generated C# for entry points.
 - `Microsoft.Quantum.EntryPointDriver.Simulation.Simulate` was made public as it is used in the generated C# for entry points.
 - The `IEntryPoint` interface was changed considerably.
 - And more changes under the Entry Point Driver project. Too many to list here 😄.

Dependent PR on compiler repo: https://github.com/microsoft/qsharp-compiler/pull/881


src/Simulation/EntryPointDriver/Driver.cs
```diff
@@ -241,146 +291,220 @@ private void AddOptionIfAvailable<T>(Command command, OptionInfo<T> option)
...
-            return descriptor.Length > 30 ? argument.Name : descriptor;
+            protected override string ArgumentDescriptor(IArgument argument)
+            {
+                // Hide long argument descriptors.
+                var descriptor = base.ArgumentDescriptor(argument);
+                return descriptor.Length > 30 ? argument.Name : descriptor;
+            }
         }
     }
+
```
samarsha, line 509: nit:
```suggestion
```

src/Simulation/EntryPointDriver/Driver.cs
```diff
@@ -241,146 +291,220 @@ private void AddOptionIfAvailable<T>(Command command, OptionInfo<T> option)
...
         /// <summary>
-        /// The subscription option.
+        /// Creates the simulate command.
         /// </summary>
-        internal static readonly OptionInfo<string> SubscriptionOption = new OptionInfo<string>(
-            ImmutableList.Create("--subscription"), "The subscription ID.");
+        /// <param name="entryPointCommands">The entry point commands that will be the sub commands to the created command.</param>
+        /// <returns>Tuple with the created simulate command and the validators for that command.</returns>
+        private static (Command, IEnumerable<ValidateSymbol<CommandResult>>) CreateSimulateCommand(
+            IEnumerable<(Command, IEnumerable<ValidateSymbol<CommandResult>>)> entryPointCommands)
```
samarsha, line 306: Since this pair of command and validators is used in several places (and the list of validators on a command is unfortunately internal to System.CommandLine), I recommend creating a new type to hold them instead of using a tuple.

src/Simulation/EntryPointDriver/Driver.cs
```diff
@@ -241,146 +291,220 @@ private void AddOptionIfAvailable<T>(Command command, OptionInfo<T> option)
...
-        /// </summary>
-        internal static readonly OptionInfo<string> ResourceGroupOption = new OptionInfo<string>(
-            ImmutableList.Create("--resource-group"), "The resource group name.");
+            var simulate = new Command("simulate", "(default) Run the program using a local simulator.");
+            if (entryPointCommands.Count() == 1)
+            {
+                var (epCommand, valids) = entryPointCommands.First();
+                simulate.AddCommand(epCommand);
+                SetSubCommandAsDefault(simulate, epCommand, valids);
+                validators = valids;
```
samarsha, line 316: nit: Same as above - you can return from both branches instead of using a mutable variable.

src/Simulation/EntryPointDriver/Driver.cs
```diff
@@ -241,146 +291,220 @@ private void AddOptionIfAvailable<T>(Command command, OptionInfo<T> option)
...
+            var submit = new Command("submit", "Submit the program to Azure Quantum.")
+            {
+                IsHidden = true,
+            };
+            if (entryPointCommands.Count() == 1)
+            {
+                var (epCommand, valids) = entryPointCommands.First();
+                submit.AddCommand(epCommand);
+                SetSubCommandAsDefault(submit, epCommand, valids);
+                validators = valids;
```
samarsha, line 347: nit: Same as above - you can return from both branches instead of using a mutable variable.

src/Simulation/EntryPointDriver/Driver.cs
samarsha, line 306: Have you considered swapping the command nesting, so that `simulate` and `submit` are sub-commands of the entry point command? I'm not sure if that would be better but it would mean that `simulate` and `submit` wouldn't both need the list of entry points.

src/Simulation/EntryPointDriver/Driver.cs
```diff
@@ -241,146 +291,220 @@ private void AddOptionIfAvailable<T>(Command command, OptionInfo<T> option)
...
+            validators = validators.Concat(AddOptionIfAvailable(command, BaseUriOption));
+            validators = validators.Concat(AddOptionIfAvailable(command, LocationOption));
+            validators = validators.Concat(AddOptionIfAvailable(command, JobNameOption));
+            validators = validators.Concat(AddOptionIfAvailable(command, ShotsOption));
+            validators = validators.Concat(AddOptionIfAvailable(command, OutputOption));
+            validators = validators.Concat(AddOptionIfAvailable(command, DryRunOption));
+            validators = validators.Concat(AddOptionIfAvailable(command, VerboseOption));
+            validators = validators.Concat(MarkOptionsAsMutuallyExclusive(
+                command,
+                new[] { BaseUriOption.Aliases.First(), LocationOption.Aliases.First() }));
```
samarsha, line 416: nit: You can chain the `.Concat`s instead of reassigning `validators` each time.

src/Simulation/EntryPointDriver/Driver.cs
```diff
@@ -241,146 +291,220 @@ private void AddOptionIfAvailable<T>(Command command, OptionInfo<T> option)
...
-    /// A modification of the command-line <see cref="HelpBuilder"/> class.
-    /// </summary>
-    internal sealed class QsHelpBuilder : HelpBuilder
-    {
+        /// <param name="parseResult">The command-line parsing result.</param>
+        /// <param name="azureSettings">The Azure submission settings.</param>
+        /// <param name="entryPoint">The entry point to submit.</param>
+        /// <returns>The exit code.</returns>
+        private async Task<int> Submit(ParseResult parseResult, AzureSettings azureSettings, IEntryPoint entryPoint) =>
+            await entryPoint.Submit(parseResult, new AzureSettings
```
samarsha, line 471: nit: (I know this was in the original code, but) the `async` and `await` are unnecessary here.

src/Simulation/EntryPointDriver/Driver.cs
```diff
@@ -241,146 +291,220 @@ private void AddOptionIfAvailable<T>(Command command, OptionInfo<T> option)
...
-        internal static readonly OptionInfo<OutputFormat> OutputOption = new OptionInfo<OutputFormat>(
-            ImmutableList.Create("--output"),
-            OutputFormat.FriendlyUri,
-            "The information to show in the output after the job is submitted.");
+        /// <param name="entryPoint">The entry point.</param>
+        /// <returns>The OptionInfor for the simulator option.</returns>
+        private OptionInfo<string> CreateSimulatorOption(IEntryPoint entryPoint) =>
+            new OptionInfo<string>(
+                this.settings.SimulatorOptionAliases,
+                entryPoint.DefaultSimulatorName,
```
samarsha, line 429: The default simulator option applies to the entire project, rather than for a specific entry point. Maybe it could be moved out of `IEntryPoint` into `DriverSettings`?

src/Simulation/EntryPointDriver/Driver.cs
```diff
@@ -241,146 +291,220 @@ private void AddOptionIfAvailable<T>(Command command, OptionInfo<T> option)
...
-            ImmutableList.Create("--dry-run"),
-            false,
-            "Validate the program and options, but do not submit to Azure Quantum.");
+        /// <param name="entryPoint">The entry point.</param>
+        /// <returns>The OptionInfo for the --target option.</returns>
+        private OptionInfo<string?> CreateTargetOption(IEntryPoint entryPoint)
+        {
+            var targetAliases = ImmutableList.Create("--target");
+            const string targetDescription = "The target device ID.";
+            return string.IsNullOrWhiteSpace(entryPoint.DefaultExecutionTarget)
```
samarsha, line 448: Same as above: the default execution target is a project setting instead of an entry-point-specific setting.

src/Simulation/EntryPointDriver/IEntryPoint.cs
```diff
@@ -16,10 +17,13 @@ namespace Microsoft.Quantum.EntryPointDriver
     /// Contains entry point properties needed by the command-line interface and allows the entry point to use
     /// command-line arguments. The implementation of this interface is code-generated.
     /// </remarks>
-    /// <typeparam name="TIn">The entry point's argument type.</typeparam>
-    /// <typeparam name="TOut">The entry point's return type.</typeparam>
-    public interface IEntryPoint<TIn, TOut>
+    public interface IEntryPoint
     {
+        /// <summary>
+        /// The name of the entry point operation.
```
samarsha, line 23: nit: The entry point can technically be a function as well.

```suggestion
        /// The name of the entry point.
```

samarsha (changes requested):

Simulation.sln
```diff
@@ -1,688 +1,688 @@
...
-		{7E24885B-D86D-477E-A840-06FA53C33FE1} = {34D419E9-CCF1-4E48-9FA4-3AD4B86BEEB4}
-		{7F80466B-A6B5-4EF1-A9E9-22ABAE3C20C1} = {34D419E9-CCF1-4E48-9FA4-3AD4B86BEEB4}
-	EndGlobalSection
-	GlobalSection(ExtensibilityGlobals) = postSolution
-		SolutionGuid = {929C0464-86D8-4F70-8835-0A5EAF930821}
-	EndGlobalSection
-EndGlobal
\ No newline at end of file
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
```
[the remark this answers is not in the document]
ScottCarda-MS: I'm not sure I can manually change that on my end without breaking my ability to work with the solution.

src/Simulation/CSharpGeneration/EntryPoint.fs
```diff
@@ -119,22 +119,96 @@ let private callableTypeNames context (callable : QsCallable) =
     let returnTypeName = SimulationCode.roslynTypeName context callable.Signature.ReturnType
     callableName, argTypeName, returnTypeName
 
-/// The main method for the standalone executable.
-let private mainMethod context entryPoint =
+let private generateEntryPointClassName (entryPoint : QsCallable) =
```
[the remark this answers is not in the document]
ScottCarda-MS: Added

src/Simulation/CSharpGeneration/EntryPoint.fs
```diff
@@ -119,22 +119,96 @@ let private callableTypeNames context (callable : QsCallable) =
...
+    (entryPoint.FullName.Namespace, entryPointClassName +  entryPoint.FullName.Name)
```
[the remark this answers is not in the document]
ScottCarda-MS: done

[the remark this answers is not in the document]
ScottCarda-MS: changed to returning QsQualifiedName

src/Simulation/CSharpGeneration/EntryPoint.fs
```diff
@@ -149,35 +223,43 @@ let private entryPointClass context entryPoint =
...
 /// Generates C# source code for a standalone executable that runs the Q# entry point.
-let generate context (entryPoint : QsCallable) =
-    let ns =
-        ``namespace`` entryPoint.FullName.Namespace
-            ``{``
-                (Seq.map using SimulationCode.autoNamespaces)
-                [entryPointClass context entryPoint]
-            ``}``
-    ``compilation unit`` [] [] [ns]
+let generateEntryPointSource context (entryPoints : seq<QsCallable>) (mainNamespace : NamespaceDeclarationSyntax option) =
```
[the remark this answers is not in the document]
ScottCarda-MS: done

src/Simulation/EntryPointDriver/Azure.cs
```diff
@@ -202,7 +202,7 @@ internal enum OutputFormat
     /// <summary>
     /// Settings for a submission to Azure Quantum.
     /// </summary>
-    internal sealed class AzureSettings
+    public sealed class AzureSettings
```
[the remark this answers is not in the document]
ScottCarda-MS: done

src/Simulation/EntryPointDriver/Driver.cs
```diff
@@ -13,66 +13,130 @@
...
+
+        /// <summary>
+        /// The driver settings.
+        /// </summary>
+        private readonly DriverSettings settings;
+
+        /// <summary>
+        /// All the registered entry points of the program.
+        /// </summary>
+        private readonly IEnumerable<IEntryPoint> entryPoints;
```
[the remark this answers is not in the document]
ScottCarda-MS: done

src/Simulation/EntryPointDriver/Driver.cs
```diff
@@ -82,46 +146,18 @@ public Driver(DriverSettings settings, IEntryPoint<TIn, TOut> entryPoint)
...
-                submit, 
-                new[] { BaseUriOption.Aliases.First(), LocationOption.Aliases.First() });
-
-            var root = new RootCommand(entryPoint.Summary) { simulate, submit };
-            foreach (var option in entryPoint.Options)
-            {
-                root.AddGlobalOption(option);
-            }
+            var (simulate, simulateValids) = CreateSimulateCommand(simulateSubCommands);
+            var (submit, submitValids) = CreateSubmitCommand(submitSubCommands);
```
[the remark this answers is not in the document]
ScottCarda-MS: done

src/Simulation/EntryPointDriver.Tests/Tests.qs
```diff
@@ -360,3 +360,22 @@ namespace EntryPointTest {
...
+
+namespace EntryPointTest {
+    
+    @EntryPoint()
+    operation  MultipleEntryPoints1() : String {
+        return "Hello from Entry Point 1!";
+    }
+
+    @EntryPoint()
+    operation  MultipleEntryPoints2() : String {
```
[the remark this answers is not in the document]
ScottCarda-MS: done

src/Simulation/EntryPointDriver/Driver.cs
```diff
@@ -82,46 +146,18 @@ public Driver(DriverSettings settings, IEntryPoint<TIn, TOut> entryPoint)
...
 
-            // Set the simulate command as the default.
-            foreach (var option in simulate.Options)
+            var root = new RootCommand() { simulate, submit };
+            if (this.entryPoints.Count() == 1)
             {
-                root.AddOption(option);
+                SetSubCommandAsDefault(root, simulate, simulateValids);
```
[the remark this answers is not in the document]
ScottCarda-MS: Correct, this will be a future feature, but for now, if there are multiple entry points specified, no default is set.

src/Simulation/EntryPointDriver/Driver.cs
```diff
@@ -199,29 +237,41 @@ private T DefaultIfShadowed<T>(OptionInfo<T> option, T value)
...
             }
             else if (option.Required)
             {
-                command.AddValidator(commandResult =>
-                    $"The required option {option.Aliases.First()} conflicts with an entry point parameter name.");
+                ValidateSymbol<CommandResult> validator = commandResult =>
+                    $"The required option {option.Aliases.First()} conflicts with an entry point parameter name.";
+
+                command.AddValidator(validator);
+                validators =  new[] { validator };
```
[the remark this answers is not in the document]
ScottCarda-MS: I've made this change for the sake of expediency, but please keep in mind that mutable variables are the norm in C# and having multiple return statements in C# code is, to the thinking of many, to be avoided.

src/Simulation/EntryPointDriver/Driver.cs
ScottCarda-MS: done

src/Simulation/EntryPointDriver/Driver.cs
ScottCarda-MS: done

src/Simulation/EntryPointDriver/Driver.cs
ScottCarda-MS: done

src/Simulation/EntryPointDriver/Driver.cs
ScottCarda-MS: done

src/Simulation/EntryPointDriver/Driver.cs
ScottCarda-MS: done

src/Simulation/EntryPointDriver/IEntryPoint.cs
ScottCarda-MS: done

src/Simulation/EntryPointDriver/Azure.cs
```diff
@@ -2,33 +2,34 @@
...
         /// <typeparam name="TOut">The entry point's return type.</typeparam>
+        /// <param name="entryPoint">The entry point.</param>
+        /// <param name="info">The information about the entry point.</param>
+        /// <param name="input">The input argument tuple to the entry point.</param>
+        /// <param name="settings">The submission settings.</param>
         /// <returns>The exit code.</returns>
-        internal static async Task<int> Submit<TIn, TOut>(
-            IEntryPoint<TIn, TOut> entryPoint, ParseResult parseResult, AzureSettings settings)
+        public static async Task<int> Submit<TIn, TOut>(
+            IEntryPoint entryPoint, EntryPointInfo<TIn, TOut> info, TIn input, AzureSettings settings)
```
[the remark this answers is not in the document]
ScottCarda-MS: Currently this is just used by our generate C#. Do we generally support 3rd party developers to use these like an API? If so the `entryPoint` parameter would be useful to have as it allows a more complete picture of information and creates symmetry  with the Simulate method. I used the parameter myself when I was working with this to print out useful information in this handle. If we do not intend anyone to use these outside of our own generated C#, then I agree it makes sense to remove it, as it is unused in our implementation.

src/Simulation/CSharpGeneration/RewriteStep.fs
```diff
@@ -61,9 +61,24 @@ type Emitter() =
...
+
+                let main = EntryPoint.mainNamespace entryPointCallables
+
+                let entryPointSources =
+                    entryPointCallables
+                    |> Seq.groupBy (fun ep -> ep.SourceFile)
+
+                let (sourceFile, callables) = Seq.head entryPointSources
+                let content = EntryPoint.generateEntryPointSource context callables (Some main)
+                CompilationLoader.GeneratedFile(sourceFile, dir, ".EntryPoint.g.cs", content) |> ignore
```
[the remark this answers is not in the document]
ScottCarda-MS: I gave this one quite a lot of thought, and I believe it creates much more work and special casing to handle the possibility that, if we have a special file for the main, we may have a collision with another file of the same name being used as an entry point file. My conclusion was to avoid having to deal with this by simply putting the main in the first (possibly only) entry point file.

src/Simulation/EntryPointDriver.Tests/Tests.fs
```diff
@@ -878,9 +885,17 @@ let ``Shows help text for submit command with default target`` () =
...
-                      --my-cool-bool (REQUIRED)                           A neat bit.
-                      -?, -h, --help                                      Show help and usage information"
+                      -?, -h, --help                                      Show help and usage information
+
+                      Commands:
+                        EntryPointTest.Help    This test checks that the entry point documentation appears correctly in the command line help message."
     let given = testWithTarget "foo.target" "Help"
     given ["submit"; "--help"] |> yields message
+
+[<Fact(Skip="Multiple Entry Points not yet supported in compiler.")>]
```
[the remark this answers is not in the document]
ScottCarda-MS: This PR works well as a rewrite of the C# generation and its interactions with the Entry Point Driver without the need of multiple entry points enabled, and therefor doesn't depend on them. I don't think a feature branch is necessary here.

src/Simulation/EntryPointDriver/Driver.cs
```diff
@@ -199,29 +237,41 @@ private T DefaultIfShadowed<T>(OptionInfo<T> option, T value)
...
         /// Adds a validator to the command such that the specified options cannot be simultaneously included.
         /// </summary>
         /// <param name="command">The command to add the validator to.</param>
         /// <param name="primaryAliases">The primary aliases of the options to be marked as mutually exclusive.</param>
-        private void MarkOptionsAsMutuallyExclusive(Command command, string[] primaryAliases) =>
-            command.AddValidator(result =>
+        /// <returns>The list of validators added to the command during this function.</returns>
+        private static IEnumerable<ValidateSymbol<CommandResult>> MarkOptionsAsMutuallyExclusive(Command command, string[] primaryAliases)
+        {
+            var validators = Enumerable.Empty<ValidateSymbol<CommandResult>>();
```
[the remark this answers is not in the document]
ScottCarda-MS: removed

src/Simulation/EntryPointDriver/Driver.cs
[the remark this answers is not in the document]
samarsha: Yes, in C# all local variables are mutable, but using mutability doesn't always make code easier to read and understand. Here, I think it hinders readability. I'm not aware of any style guideline against multiple returns in C#; I don't think single-entry single-exit is commonly accepted these days.

src/Simulation/EntryPointDriver/Driver.cs
ScottCarda-MS: Do we not support passing the desired simulator into the Entry Point attribute in Q# anymore? I thought that was what set the default simulator per entry point.

samarsha: No, the entry point attribute has `Unit` as its underlying type - see #451.

ScottCarda-MS: Oh okay, I believe I was mistakenly thinking of the Unit Test attribute.

Simulation.sln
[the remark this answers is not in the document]
samarsha: If Git `core.autocrlf` is true (the default on Windows), then line endings are automatically converted to CRLF in your working directory and back to LF in the repository.

src/Simulation/EntryPointDriver/Driver.cs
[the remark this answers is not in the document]
samarsha: Also, to clarify, all of my "nit" comments are optional suggestions.

src/Simulation/EntryPointDriver/Driver.cs
```diff
@@ -82,46 +146,18 @@ public Driver(DriverSettings settings, IEntryPoint<TIn, TOut> entryPoint)
...
+            var (submit, submitValids) = CreateSubmitCommand(submitSubCommands);
 
-            // Set the simulate command as the default.
-            foreach (var option in simulate.Options)
+            var root = new RootCommand() { simulate, submit };
+            if (this.entryPoints.Count() == 1)
             {
-                root.AddOption(option);
+                SetSubCommandAsDefault(root, simulate, simulateValids);
+                root.Description = this.entryPoints.First().Summary;
```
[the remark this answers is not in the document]
ScottCarda-MS: We only want to set the description in the case of the root command. The other two places where the `SetSubCommandAsDefault` are the `submit` and `simulate` commands, and they keep their original descriptions.

src/Simulation/EntryPointDriver.Tests/Tests.fs
```diff
@@ -878,9 +885,9 @@ let ``Shows help text for submit command with default target`` () =
...
                       --dry-run                                           Validate the program and options, but do not submit to Azure Quantum.
                       --verbose                                           Show additional information about the submission.
-                      -n <n> (REQUIRED)                                   A number.
-                      --pauli <PauliI|PauliX|PauliY|PauliZ> (REQUIRED)    The name of a Pauli matrix.
-                      --my-cool-bool (REQUIRED)                           A neat bit.
-                      -?, -h, --help                                      Show help and usage information"
+                      -?, -h, --help                                      Show help and usage information
+
+                      Commands:
+                        EntryPointTest.Help    This test checks that the entry point documentation appears correctly in the command line help message."
```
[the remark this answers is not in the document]
ScottCarda-MS: I can't seem to figure out how to edit the System.CommandLine's behavior for outputting Help. I can see there is a HelpBuilder class at the bottom of the Driver.cs class, but that doesn't seem to be very extensible. Do you know how I can accomplish what you suggest?

src/Simulation/EntryPointDriver/Driver.cs
ScottCarda-MS: done

src/Simulation/EntryPointDriver.Tests/Tests.fs
```diff
@@ -878,9 +885,17 @@ let ``Shows help text for submit command with default target`` () =
...
     let given = testWithTarget "foo.target" "Help"
     given ["submit"; "--help"] |> yields message
+
+[<Fact(Skip="Multiple Entry Points not yet supported in compiler.")>]
+let ``Supports multiple entry points`` () =
+    let given = test "Multiple entry points"
+    given ["simulate"; "EntryPointTest.MultipleEntryPoints1"] |> yields "Hello from Entry Point 1!"
+    given ["simulate"; "EntryPointTest.MultipleEntryPoints2"] |> yields "Hello from Entry Point 2!"
```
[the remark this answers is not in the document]
ScottCarda-MS: done

src/Simulation/EntryPointDriver.Tests/Tests.fs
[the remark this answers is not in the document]
ScottCarda-MS: I've opted to hide the default commands, like what is done elsewhere.

src/Simulation/EntryPointDriver/Driver.cs
ScottCarda-MS: I've spent the day trying this out, and the code complexity wasn't noticeably different. From a user standpoint, my preference is to have it as the entry point command last, as that will match more closely to and English reading of the command line: `MyApp simulate MyEntryPoint` as opposed to `MyApp MyEntryPoint`. Subject being the name of the app, the verb being 'simulate' or 'submit' and the direct object being the entry point being simulated or submitted. Hence I'm going to keep it as the subject-verb-object ordering for the command hierarchy.

Simulation.sln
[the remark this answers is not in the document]
ScottCarda-MS: The line endings should now be LF.

src/Simulation/EntryPointDriver/Driver.cs
ScottCarda-MS: I was able to do this, along with moving the CreateDefaultCustomSimulator function to the driver settings. The two had to be moved together. I think this change helps keep the entry point classes tightly defined. Thanks!

src/Simulation/EntryPointDriver/Driver.cs
ScottCarda-MS: done

src/Simulation/EntryPointDriver/Azure.cs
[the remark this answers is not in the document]
ScottCarda-MS: Turns out these need to be public for the System.CommandLine to properly parse the arguments. I've reverted this back to the properties being public.

src/Simulation/EntryPointDriver/Azure.cs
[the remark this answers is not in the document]
ScottCarda-MS: After a conversation with Bettina, I learned that we don't intend for these to be used like public API's, so I'll remove the parameter.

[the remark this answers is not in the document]
ScottCarda-MS: removed unused param

src/Simulation/CSharpGeneration/RewriteStep.fs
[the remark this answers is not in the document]
samarsha: Hmm, now that you mention file collisions, I think there's already a collision if you have two Q# files `Foo.qs` and `Foo.EntryPoint.qs`, where `Foo.qs` contains the `@EntryPoint()`. `Foo.qs` generates `Foo.g.cs` and `Foo.EntryPoint.g.cs`, and `Foo.EntryPoint.qs` also generates `Foo.EntryPoint.g.cs`.

I think it would be fixed if the suffix was `.g.EntryPoint.cs` instead of `.EntryPoint.g.cs`. In that case, you could also generate a separate file for the Main method named `EntryPoint.g.Main.cs` (or something with a unique `.g.???.cs` suffix), and I don't think it could conflict with any user file.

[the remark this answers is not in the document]
ScottCarda-MS: So for my understanding, the problem you are suggesting is that, for the two existing suffixes `.g.cs` and `.Entrypoint.g.cs`, one suffix is a substring of the other. And the solution is to have a suffix set so that no suffix is the substring of another. For example `.g.cs`, `.g.EntryPoint.cs`, and (newly) `.g.Main.cs`. Is that right?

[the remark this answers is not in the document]
samarsha: Yes, I think so.

src/Simulation/EntryPointDriver/Driver.cs
```diff
@@ -13,66 +13,159 @@
...
 
             var targetAliases = ImmutableList.Create("--target");
             const string targetDescription = "The target device ID.";
-            TargetOption = string.IsNullOrWhiteSpace(entryPoint.DefaultExecutionTarget)
+            this.TargetOption = string.IsNullOrWhiteSpace(settings.DefaultExecutionTarget)
                 ? new OptionInfo<string?>(targetAliases, targetDescription)
-                : new OptionInfo<string?>(targetAliases, entryPoint.DefaultExecutionTarget, targetDescription);
+                : new OptionInfo<string?>(targetAliases, this.settings.DefaultExecutionTarget, targetDescription);
+
+            this.entryPoints = entryPoints.ToList().AsReadOnly();
```
samarsha, line 168: The array passed in by the generated code should already implement `IReadOnlyCollection<IEntryPoint>`. I think you can change the constructor parameter type to that?

src/Simulation/EntryPointDriver/Driver.cs
```diff
@@ -241,146 +316,241 @@ private void AddOptionIfAvailable<T>(Command command, OptionInfo<T> option)
...
-            ImmutableList.Create("--workspace"), "The workspace name.");
+        /// <param name="entryPointCommands">The entry point commands that will be the sub commands to the created command.</param>
+        /// <returns>The created simulate command with the validators for that command.</returns>
+        private static CommandWithValidators CreateSimulateCommand(
+            IEnumerable<CommandWithValidators> entryPointCommands)
+        {
+            var simulate = new Command("simulate", "(default) Run the program using a local simulator.");
+            if (entryPointCommands.Count() == 1)
+            {
+                var (epCommand, validators) = entryPointCommands.First();
```
samarsha, line 336: `CreateSimulateCommand` and `CreateSubmitCommand` should probably also take in `IReadOnlyCollection<CommandWithValidators>`.

src/Simulation/EntryPointDriver/Driver.cs
```diff
@@ -82,46 +175,18 @@ public Driver(DriverSettings settings, IEntryPoint<TIn, TOut> entryPoint)
...
-                submit, 
-                new[] { BaseUriOption.Aliases.First(), LocationOption.Aliases.First() });
-
-            var root = new RootCommand(entryPoint.Summary) { simulate, submit };
-            foreach (var option in entryPoint.Options)
-            {
-                root.AddGlobalOption(option);
-            }
+            var (simulate, simulateValidators) = CreateSimulateCommand(simulateSubCommands);
+            var (submit, _) = CreateSubmitCommand(submitSubCommands);
```
samarsha, line 182: If the `submit` validators aren't used, could `CreateSubmitCommand` return only a `Command`?

src/Simulation/EntryPointDriver/Driver.cs
```diff
@@ -241,146 +316,241 @@ private void AddOptionIfAvailable<T>(Command command, OptionInfo<T> option)
...
+                .Concat(AddOptionIfAvailable(command, JobNameOption))
+                .Concat(AddOptionIfAvailable(command, ShotsOption))
+                .Concat(AddOptionIfAvailable(command, OutputOption))
+                .Concat(AddOptionIfAvailable(command, DryRunOption))
+                .Concat(AddOptionIfAvailable(command, VerboseOption))
+                .Concat(MarkOptionsAsMutuallyExclusive(
+                    command,
+                    new[] { BaseUriOption.Aliases.First(), LocationOption.Aliases.First() }));
+
+            return (command, validators.ToList());
```
samarsha, line 440: If someone forgets the `ToList()`, this will still compile, but could have unintended consequences since evaluating `validators` has side effects. Can you change `CommandWithValidators` to contain an `IReadOnlyCollection` instead of an `IEnumerable`?

src/Simulation/EntryPointDriver/Driver.cs
```diff
@@ -241,146 +316,241 @@ private void AddOptionIfAvailable<T>(Command command, OptionInfo<T> option)
...
+                return (value.Command, value.Validators);
+            }
+
+            /// <summary>
+            /// Allows the type to be converted from a tuple.
+            /// </summary>
+            public static implicit operator CommandWithValidators((Command Command, IEnumerable<ValidateSymbol<CommandResult>> Validators) value)
+            {
+                return new CommandWithValidators(value.Command, value.Validators);
+            }
```
samarsha, line 553: This is a lot of code for such a small type, and the implicit conversions to and from the tuple type don't seem necessary. Wouldn't something like

```csharp
private struct CommandWithValidators
{
    public Command Command;
    public IReadOnlyCollection<ValidateSymbol<CommandResult>> Validators;
}
```

work? (Maybe with a constructor as well.)

src/Simulation/EntryPointDriver.Tests/Tests.fs
```diff
@@ -878,9 +882,77 @@ let ``Shows help text for submit command with default target`` () =
...
                       --dry-run                                           Validate the program and options, but do not submit to Azure Quantum.
                       --verbose                                           Show additional information about the submission.
-                      -n <n> (REQUIRED)                                   A number.
-                      --pauli <PauliI|PauliX|PauliY|PauliZ> (REQUIRED)    The name of a Pauli matrix.
-                      --my-cool-bool (REQUIRED)                           A neat bit.
                       -?, -h, --help                                      Show help and usage information"
     let given = testWithTarget "foo.target" "Help"
     given ["submit"; "--help"] |> yields message
+
+[<Fact(Skip="Multiple Entry Points not yet supported in compiler.")>]
```
samarsha, line 889: Are you planning to enable these tests? Is the compiler package working?

samarsha (dismissed):

src/Simulation/EntryPointDriver.Tests/Tests.fs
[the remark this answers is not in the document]
samarsha: Before I would approve this PR to merge into main, I would like to see the tests with multiple entry points enabled here, as well as checking if IQ# needs any updates to handle multiple entry points. @cgranade Do you know if IQ# is relying on there being only one entry point?

src/Simulation/EntryPointDriver/Driver.cs
ScottCarda-MS: I just used the the intellisense suggestion to generate the equivalent type for the tuple. The tuple conversions make it easy to do the deconstructing.

src/Simulation/EntryPointDriver/Driver.cs
ScottCarda-MS: I'm not sure what you mean by 'evaluating `validators`' Since each validator is a Func, putting them in a list shouldn't have any side effects. I'm not sure why the preference for `IReadOnlyCollection` or IEnumerable?

src/Simulation/CSharpGeneration/EntryPoint.fs
```diff
@@ -119,65 +97,169 @@ let private callableTypeNames context (callable : QsCallable) =
     let returnTypeName = SimulationCode.roslynTypeName context callable.Signature.ReturnType
     callableName, argTypeName, returnTypeName
 
-/// The main method for the standalone executable.
-let private mainMethod context entryPoint =
+/// Generates the class name for an entry point class.
+let private generateEntryPointClassName (entryPoint : QsCallable) =
```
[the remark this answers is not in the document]
ScottCarda-MS: This conflicts with the existing `entryPointClassName` field.

src/Simulation/CSharpGeneration/EntryPoint.fs
```diff
@@ -119,65 +97,169 @@ let private callableTypeNames context (callable : QsCallable) =
...
+
+    let driver = ``new`` (``type`` (driverNamespace + ".Driver")) ``(`` [driverSettings context; entryPointArray] ``)``
+    let commandLineArgsName = "args"
+    arrow_method "System.Threading.Tasks.Task<int>" "Main" ``<<`` [] ``>>``
+        ``(`` [param commandLineArgsName ``of`` (``type`` "string[]")] ``)``
+        [``private``; ``static``; async]
+        (Some (``=>`` (await (driver <.> (ident "Run", [ident commandLineArgsName])))))
+
+/// Generates a namespace for the main function
+let mainNamespace context (entryPoints : seq<QsCallable>) =
```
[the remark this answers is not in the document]
ScottCarda-MS: Can `generateSource` be internal as well? These two should probably have the same access.

src/Simulation/CSharpGeneration/EntryPoint.fs
```diff
@@ -119,65 +97,169 @@ let private callableTypeNames context (callable : QsCallable) =
...
+
+/// Generates C# source code for a standalone executable that runs the Q# entry point.
+let generateSource context (entryPoints : seq<QsCallable>) (mainNamespace : NamespaceDeclarationSyntax option) =
+    let entryPointNamespaces = entryPoints |> Seq.groupBy (fun ep -> ep.FullName.Namespace)
+
+    let epNamespaces = [for ns, eps in entryPointNamespaces -> entryPointNamespace context ns eps]
+
+    let namespaces =
+        match mainNamespace with
+        | Some mainNamespace -> (mainNamespace :> MemberDeclarationSyntax) :: epNamespaces
```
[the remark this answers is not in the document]
ScottCarda-MS: The parameter's type is `NamespaceDeclarationSyntax` to enforce that what is passed in should be a namespace, instead of any of the other possible kinds of `MemberDeclarationSyntax` types. However, the `compilation unit` function expects a list of `MemberDeclarationSyntax` objects for the namespaces. This is why `entryPointNamespace` returns a list of `MemberDeclarationSyntax` objects instead of `NamespaceDeclarationSyntax` objects, even though that's what they really are. That's why this casting occurs here.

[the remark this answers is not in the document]
ScottCarda-MS: Alternatively, I could return a list of `NamespaceDeclarationSyntax` objects from `entryPointNamespace` (not doing the cast at the end of that function) and create a list of `NamespaceDeclarationSyntax` objects. Then do a typecast across the list when I go to pass it to the `compilation unit` function. That would unify where the casting happens. Yah I think I'll do that, and that will make things a bit clearer.

src/Simulation/CSharpGeneration/RewriteStep.fs
```diff
@@ -27,31 +24,22 @@ type Emitter() =
...
-        
-        member this.ImplementsPreconditionVerification = true
+
+        member this.ImplementsPreconditionVerification = false
         member this.ImplementsPostconditionVerification = false
         member this.ImplementsTransformation = true
 
-        member this.PreconditionVerification compilation =
-            if compilation.EntryPoints.Length > 1 then
-                _Diagnostics <- IRewriteStep.Diagnostic
```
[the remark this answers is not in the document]
ScottCarda-MS: Yah this should go, but I don't know what the F# equivalent is to `Enumerable.Empty<IRewriteStep.Diagnostic>()`.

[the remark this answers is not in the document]
ScottCarda-MS: It looks like `upcast []` might work, but I don't really know what that does. Let me know if that is not the right solution.

src/Simulation/EntryPointDriver/Driver.cs
ScottCarda-MS: Looks like that will work.

src/Simulation/EntryPointDriver/Driver.cs
ScottCarda-MS: Is the idea to just replace all the instances of `IEnumerable` with `IReadOnlyCollection`?

src/Simulation/EntryPointDriver/Driver.cs
ScottCarda-MS: Tailoring the function to its specific use here by removing the validator part of the returned tuple undermines the function's purpose and degrades modularity of the function. The function may add validators to the command it creates, so those validators should be returned with the command, even if they aren't used in this particular case.

src/Simulation/EntryPointDriver.Tests/Tests.fs
ScottCarda-MS: Yes, the test will be enabled.

ScottCarda-MS: done

src/Simulation/EntryPointDriver/Driver.cs
ScottCarda-MS: done

src/Simulation/EntryPointDriver/DriverSettings.cs
```diff
@@ -27,23 +29,51 @@ public sealed class DriverSettings
...
         /// Creates a new driver settings instance.
         /// </summary>
         /// <param name="simulatorOptionAliases">The aliases for the simulator command-line option.</param>
         /// <param name="quantumSimulatorName">The name of the quantum simulator.</param>
         /// <param name="toffoliSimulatorName">The name of the Toffoli simulator.</param>
-        /// <param name="resourcesEstimatorName">The name of the resources estimator.</param>
+        /// <param name="resourcesEstimatorName">The name of the resources estimator.</param>
+        /// <param name="defaultSimulatorName">The name of the default simulator to use.</param>
+        /// <param name="defaultExecutionTarget">The name of the default execution target to use.</param>
+        /// <param name="createDefaultCustomSimulator">The function for crating a new instance of the default simulator if it is a custom simulator.</param>
```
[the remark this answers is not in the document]
ScottCarda-MS: done

src/Simulation/EntryPointDriver.Tests/Tests.qs
```diff
@@ -360,3 +360,35 @@ namespace EntryPointTest {
...
+
+//
+// Multiple Entry Points
+//
+
+// --- Multiple entry points
+
+namespace EntryPointTest {
+    @EntryPoint()
+    operation  MultipleEntryPoints1() : String {
```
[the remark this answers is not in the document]
ScottCarda-MS: fixed

src/Simulation/CSharpGeneration/EntryPoint.fs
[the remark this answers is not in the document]
ScottCarda-MS: done

src/Simulation/CSharpGeneration/EntryPoint.fs
```diff
@@ -119,65 +97,169 @@ let private callableTypeNames context (callable : QsCallable) =
...
         ``{``
             members
         ``}``
 
-/// Generates C# source code for a standalone executable that runs the Q# entry point.
-let generate context (entryPoint : QsCallable) =
-    let ns =
-        ``namespace`` entryPoint.FullName.Namespace
+/// Generates a namespace for a set of entry points that share the namespace
+let private entryPointNamespace context name (entryPoints : seq<QsCallable>) =
```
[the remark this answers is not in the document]
ScottCarda-MS: done

src/Simulation/CSharpGeneration/EntryPoint.fs
```diff
@@ -87,11 +69,7 @@ let private customSimulatorFactory name =
         if isCustomSimulator
         then ``new`` (``type`` name) ``(`` [] ``)``
         else upcast SyntaxFactory.ThrowExpression (``new`` (``type`` "InvalidOperationException") ``(`` [] ``)``)
-        
-    arrow_method "IOperationFactory" "CreateDefaultCustomSimulator" ``<<`` [] ``>>``
-        ``(`` [] ``)``
-        [``public``]
-        (Some (``=>`` factory))
+    ``() =>`` [] factory :> ExpressionSyntax
```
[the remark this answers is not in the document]
ScottCarda-MS: done

src/Simulation/EntryPointDriver.Tests/Tests.fs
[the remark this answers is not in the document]
ScottCarda-MS: The tests have been enabled. Let's consolidate the conversation around IQ# and feature branches to this comment here: https://github.com/microsoft/qsharp-compiler/pull/881#discussion_r583265321. I will leave this this thread on this PR open until that one is resolved and won't merged either in until then, but let's keep the conversation in that thread so it is all in one place.

src/Simulation/CSharpGeneration/EntryPoint.fs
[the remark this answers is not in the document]
ScottCarda-MS: This can't be `internal` because both this and `generateSource` are used by Test.fs.

src/Simulation/CSharpGeneration/EntryPoint.fs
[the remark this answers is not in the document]
samarsha: Since this is the qualified version of `entryPointClassName`, Maybe `entryPointClassFullName` or `entryPointFullName`?

src/Simulation/CSharpGeneration/EntryPoint.fs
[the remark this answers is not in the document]
samarsha: OK - could use InternalsVisibleTo for the test project, but I'm also not a huge fan of InternalsVisibleTo, so I'm OK with leaving them public.

src/Simulation/CSharpGeneration/RewriteStep.fs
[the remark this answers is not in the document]
samarsha: I think the equivalent of `Enumerable.Empty<IRewriteStep.Diagnostic>()` is `Seq.empty`.

src/Simulation/EntryPointDriver/Driver.cs
samarsha: `IEnumerable` is good if you're enumerating a sequence of items once from start to finish, but enumerating it more than once can cause the sequence to be recomputed (e.g. if it came from a LINQ expression). See https://www.jetbrains.com/help/resharper/PossibleMultipleEnumeration.html.

Also, in general, `IEnumerable.Count()` has to enumerate the whole collection to count it (O(n)) - it has a special check to avoid that for `IEnumerable`s that happen to be `ICollection`s, but since `IReadOnlyCollection` comes with its own `Count` property, I think it's clearer to use that.

src/Simulation/EntryPointDriver/Driver.cs
samarsha: OK.

src/Simulation/EntryPointDriver/Driver.cs
samarsha: Sorry, you're right. I confused myself since `AddOptionIfAvailable` has side effects but the returned `IEnumerable` doesn't. Why does it call `ToList()` instead of returning the `IEnumerable` though?

src/Simulation/EntryPointDriver/Driver.cs
samarsha: The deconstructing is done by the `Deconstruct` method, not the tuple conversions.

I think the amount of code should be proportional to the scale of the problem. The problem being solved here is needing a named container with two fields. I think that should be a few lines of code (1 line of code in C# 9 if you use a positional record), rather than 50.

I also think that implicit conversions should be used sparingly, and being able to write `(command, validators)` instead of `new CommandWithValidators(command, validators)` does not seem like it justifies adding implicit conversion operators.

src/Simulation/CSharpGeneration/RewriteStep.fs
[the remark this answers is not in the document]
ScottCarda-MS: done

src/Simulation/EntryPointDriver/Driver.cs
ScottCarda-MS: The `ToList()` was to evaluate the enumerable so it is only evaluated once.

src/Simulation/EntryPointDriver/Driver.cs
ScottCarda-MS: Simplified to the struct provided, with a constructor.

src/Simulation/CSharpGeneration/RewriteStep.fs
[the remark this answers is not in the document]
ScottCarda-MS: done

src/Simulation/CSharpGeneration/EntryPoint.fs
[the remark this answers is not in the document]
ScottCarda-MS: done. Went with `entryPointClassFullName`.

src/Simulation/EntryPointDriver/Driver.cs
ScottCarda-MS: I've replace all the references of IEnumerable (except one where it is the type provided by the System.CommandLine) with List types. This addresses the issue of the repeated computations, and List also has its own dedicated Count implementation that is constant time.

src/Simulation/EntryPointDriver/Driver.cs
ScottCarda-MS: IEnumerable has been replaced.

src/Simulation/CSharpGeneration/RewriteStep.fs
```diff
@@ -61,9 +48,28 @@ type Emitter() =
...
+
+                let content = EntryPoint.generateMainSource context entryPointCallables
+                let outputFolder = Path.GetFullPath(dir.TrimEnd(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar) + Path.DirectorySeparatorChar.ToString());
+                let outputUri = Uri(outputFolder);
+                let fileDir = Path.GetDirectoryName(outputUri.LocalPath);
+                let targetFile = Path.GetFullPath(Path.Combine(fileDir, "EntryPoint.g.Main.cs"));
+                if content <> null then
+                    if not (Directory.Exists(fileDir)) then
+                        Directory.CreateDirectory(fileDir) |> ignore;
+                    File.WriteAllText(targetFile, content);
```
samarsha, line 68: Semicolons aren't used in F#.

samarsha, line 68: For consistency, this should probably go through the Q# compiler to write the file. However, `CompilationLoader.GeneratedFile` only works if the generated file corresponds to a Q# source file. Maybe that method should be changed or a new one added that works for this case?

src/Simulation/EntryPointDriver/Driver.cs
```diff
@@ -13,66 +13,161 @@
 using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
-using Microsoft.Quantum.Simulation.Core;
-using static Microsoft.Quantum.EntryPointDriver.Driver;
 
 namespace Microsoft.Quantum.EntryPointDriver
 {
+    using Validators = List<ValidateSymbol<CommandResult>>;
```
samarsha, line 19: I don't think you're actually using the mutability of `List` - `ImmutableList` might be better.

src/Simulation/CSharpGeneration/RewriteStep.fs
ScottCarda-MS: I've added a WriteFile static function to the compilation loader in the compiler repo's branch, and use that here. The other method is commented out for now until we can get the proper coordination for merging all of this into the respective main branches.

src/Simulation/EntryPointDriver/Driver.cs
ScottCarda-MS: My preference is to use `List` here as that is more in-line with the mutable-by-default way C# works.

samarsha: Regardless of C#'s original design, it does have immutable collections now, and I'd recommend using them when possible unless mutation is needed, because immutable data structures are significantly easier to reason about. That would also be consistent with the rest of the entry point driver code which uses `ImmutableList`.

ScottCarda-MS: I disagree in general with what I see as an over-use of immutables in a language like C# that has mutable variables as a default, and I don't believe using immutables makes it easier for developers such as myself to reason about the code (actually it is quite the opposite, there have been many times working in F# where I have wanted to pull my hair out over the lack of mutables 😄). I suspect the reason you may find immutables easier to reason with is because of your background with functional programming, but perhaps I am wrong?

Regardless, I can see your point that, despite my objections, immutable data types are everywhere in this code base, so it would be more consistent to have this be one as well. I will update this to an `ImmutableList`.

samarsha: There's a lot of discussion in the industry about the advantages of immutable data ([here's an example](https://onor.io/2014/05/12/what-is-immutability-and-why-should-i-care/)), so I don't think it's just me. Using immutable types rules out an entire class of hard-to-fix bugs (state being accidentally mutated by a distant part of the program). They also have some disadvantages, so I don't think they should be used everywhere, but I think using mutable data types should be a conscious choice when mutability is specifically needed, and not the default option.

src/Simulation/CSharpGeneration/RewriteStep.fs
```diff
@@ -61,9 +48,31 @@ type Emitter() =
...
+
+                let entryPointCallables =
+                    compilation.EntryPoints
+                    |> Seq.map (fun ep -> context.allCallables.[ep])
+
+                let entryPointSources =
+                    entryPointCallables
+                    |> Seq.groupBy (fun ep -> ep.Source.CodeFile)
+
+                // ToDo: remove commented out code
```
ScottCarda-MS, line 60: Yes, I am aware of this, and will remove the commented out code at the next push.

src/Simulation/CSharpGeneration/RewriteStep.fs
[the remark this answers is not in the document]
bettinaheim: Hm, on the name collision, I believe there indeed isn't a way to stick with both the files that are generated based on Q# files, and any files we generate additionally have the same .g.cs ending without the possibility that they collide. What we could do is have a convention but two endings as needed like .g1.cs and .g2.cs (etc) and edit the sdk to look for that pattern instead.
Rather than applying that strategy in a more ad-hoc manner and giving the entry point related files different names, I suggest to make it more general and track during emission which file names are already used, and increment a counter as needed. It would be a good opportunity to fix this bug: https://github.com/microsoft/qsharp-runtime/issues/357

src/Simulation/CSharpGeneration/EntryPoint.fs
```diff
@@ -119,65 +97,169 @@ let private callableTypeNames context (callable : QsCallable) =
...
         [``internal``]
         ``{``
             members
         ``}``
 
-/// Generates C# source code for a standalone executable that runs the Q# entry point.
-let generate context (entryPoint : QsCallable) =
-    let ns =
-        ``namespace`` entryPoint.FullName.Namespace
+/// Generates a namespace for a set of entry points that share the namespace
```
bettinaheim, line 178: Is the namespace name dependent on the namespace of the entry point operation, such that potentially multiple such namespaces are created?

src/Simulation/EntryPointDriver.Tests/Tests.fs
```diff
@@ -836,6 +837,9 @@ let ``Shows help text for submit command`` () =
                       %s submit [options]
 
                     Options:
+                      -n <n> (REQUIRED)                                   A number.
+                      --pauli <PauliI|PauliX|PauliY|PauliZ> (REQUIRED)    The name of a Pauli matrix.
+                      --my-cool-bool (REQUIRED)                           A neat bit.
```
bettinaheim, line 842: :)

src/Simulation/EntryPointDriver.Tests/Tests.fs
```diff
@@ -878,9 +882,77 @@ let ``Shows help text for submit command with default target`` () =
...
                       --verbose                                           Show additional information about the submission.
-                      -n <n> (REQUIRED)                                   A number.
-                      --pauli <PauliI|PauliX|PauliY|PauliZ> (REQUIRED)    The name of a Pauli matrix.
-                      --my-cool-bool (REQUIRED)                           A neat bit.
                       -?, -h, --help                                      Show help and usage information"
     let given = testWithTarget "foo.target" "Help"
     given ["submit"; "--help"] |> yields message
+
+[<Fact>]
+let ``Supports simulating multiple entry points`` () =
```
bettinaheim, line 890: Is there also test coverage for multiple entry points in multiple namespaces?
Edit: Based on the Q# file it doesn't look like it. It would be good to add.

src/Simulation/EntryPointDriver/Azure.cs
```diff
@@ -2,33 +2,32 @@
...
+using Microsoft.Quantum.Simulation.Core;
 using static Microsoft.Quantum.EntryPointDriver.Driver;
 
 namespace Microsoft.Quantum.EntryPointDriver
 {
     /// <summary>
     /// Provides entry point submission to Azure Quantum.
     /// </summary>
-    internal static class Azure
+    public static class Azure
```
bettinaheim, line 19: I saw the conversation below mentioning that System.CommandLine needs the class and all properties to be public. Could you elaborate why that is? I.e. what changed to enable multiple entry points that requires AzureSettings to be public when it didn't need to be before?

src/Simulation/EntryPointDriver/Driver.cs
bettinaheim, line 19: I second going with immutable data structures by default, unless there is a good reason not to. Languages evolve, and their usage evolves. The default choice depends on the design of the code base, and this project is set up to use immutable data structures.

src/Simulation/EntryPointDriver.Tests/Tests.fs
ScottCarda-MS: 😄 I don't think I can take credit for that. I believe that was Sarah originally; I just moved the entry point's arguments to the top of the list.

src/Simulation/CSharpGeneration/RewriteStep.fs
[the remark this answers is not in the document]
ScottCarda-MS: Discussed offline. Here is the strategy agreed to so far:
 - `*.g.cs` for regular generated C# files.
 - `*.g.EntryPoint.cs` for generated entry point files.
 - `EntryPoint.g.Main.cs` for the generated Main file.
 - If when the system goes to write a file that already exists, it will append a '1' to the 'g' token of the file name, for example `*.g1.cs`. If that also already exists, it will increment that number as needed, `*.g2.cs`, `*.g3.cs` and so on. This goes for entry points as well: `*.g1.EntryPoint.cs` `*.g2.EntryPoint.cs` and so on. This will be applicable to Main also, but it should need it. It will probably just be easier to implement it without special-casing Main.

The reason to do both methods is that the incrementing strategy will avoid name conflicts in general, and the difference in suffixes will add additional structure and clarity to the generated files.

src/Simulation/CSharpGeneration/RewriteStep.fs
ScottCarda-MS: removed

src/Simulation/EntryPointDriver/Driver.cs
ScottCarda-MS: changed to ImmutableList

ScottCarda-MS: changed to using ImmutableList

src/Simulation/EntryPointDriver/Azure.cs
ScottCarda-MS: I think this is getting confused with the AzureSettings class. That class has to have public members for the System.CommandLine to properly populate its members, and it had public members before, so no change there. What has changed is that the settings class, as well as this Azure class have had to become public so they could be directly referenced by the generated entry points. This was not needed before when the Driver class was generic, but removing the type parameter from the Driver class made it possible to have multiple entry points.

src/Simulation/CSharpGeneration/RewriteStep.fs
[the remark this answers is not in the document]
ScottCarda-MS: @bettinaheim The work to implement the enumeration in the file names (bullet 4 above) is proving to be too complicated a piece of work for it to be feasible to merge in with the rest of the work for Multiple Entry Points. Therefore I am dropping the last bullet off from the list and #357 will have to be addressed separately. That work really is unrelated to Multiple Entry Points, so it would be better to have a separate PR for that anyway.

src/Simulation/EntryPointDriver.Tests/Tests.fs
ScottCarda-MS: I've updated the existing tests to include an entry point in a separate namespace.

src/Simulation/CSharpGeneration/EntryPoint.fs
ScottCarda-MS: Yes, there may be more than one namespace with entry point classes in them.

src/Simulation/CSharpGeneration/RewriteStep.fs
```diff
@@ -61,9 +49,22 @@ type Emitter() =
...
+
+                let entryPointCallables =
+                    compilation.EntryPoints
+                    |> Seq.map (fun ep -> context.allCallables.[ep])
+
+                let entryPointSources =
+                    entryPointCallables
+                    |> Seq.groupBy (fun ep -> ep.Source.CodeFile)
+
+                let mainSourceFile = "./EntryPoint" |> Path.GetFullPath |> Uri |> CompilationUnitManager.GetFileId
```
ScottCarda-MS, line 61: @bettinaheim Would it matter if it was `"./EntryPoint"` vs `Path.Combine(outputFolder, "EntryPoint")`? Would there be any tricky problems with using the outputFolder as the path?

ScottCarda-MS: I've gone with the `Path.Combine(outputFolder, "EntryPoint")` idea as I think that will more reliably put the Main file in the root of the output folder, and not accidently in a subfolder.

src/Simulation/EntryPointDriver/Azure.cs
bettinaheim: Got it. I was referring to the AzureSettings property in the class, but that answers my question.

bettinaheim (approved):
LGTM

PR #529, opened 2021-02-24
Updating ClassicalControls to reduce intrinsics

swernli:
This updates the structure of the classical control ApplyIf* and ApplyControlled* callables in terms of each other to reduce 8 intrinsics to 2 intrinsics. Also updates the qir-gen.csproj to use the local QSharp.Core project instead of the one from the QDK for easier testing.

### DIFF
--- a/src/QirRuntime/test/QIR-static/qsharp/main.cs
+++ b/src/QirRuntime/test/QIR-static/qsharp/main.cs
@@ -1,0 +1,13 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+// Currently, compiling to QIR has to suppress C# generation but then we need to provide Main function ourselves.
+namespace CompilerWorkaround
+{
+    class Program
+    {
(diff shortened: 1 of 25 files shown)


irinayat-MS:
How much of a race is this PR with https://github.com/microsoft/qsharp-runtime/pull/527? Shall I wait with merging 527 and remove the adj/ctl bridge APIs from it?

swernli:
That's up to you, @irinayat-MS ; I think we know what updates would be need to made to the conditional support in #527 if it merges first. If you are ok waiting, then that will be two PRs instead of 3, but either way I don't think it's a problem.

irinayat-MS:
OK. I've set 527 to merge automatically. If it merges successfully I'll follow up after yours to validate that we can remove the bridge for adj/ctl APIs. If it fails for whatever reason and your PR merges first, I'll align it with your refactor in a single change.

kuzminrobin (approved):
Looks Good!

PR #531, opened 2021-02-24
Added Parity() to the Q# library, and the tests.

kuzminrobin:
These changes are quite different from the ones in my previous PRs.
Some of the changes here go to the _Q# library_ rather than to _QirRuntime_. So I wanted this PR to be reviewed separately from the subsequent changes.

### DIFF
--- a/src/QirRuntime/test/QIR-static/qsharp/qir-test-other.qs
+++ b/src/QirRuntime/test/QIR-static/qsharp/qir-test-other.qs
@@ -1,0 +1,40 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.Quantum.Bitwise {
+
+    function MyParity(param : Int) : Int {
+        mutable tmpParam = param;
+        mutable result = false;
+        while tmpParam != 0 {
+            set tmpParam = tmpParam &&& (tmpParam - 1);
+            set result = not result;
+        }
+        return (result ? 1 | 0);
+    }
+}
+
+namespace Microsoft.Quantum.Testing.QIR.Other {
+
+    open Microsoft.Quantum.Bitwise;
+
+    function ParityTest() : Int {
+        //function Parity (a : Int) : Int
+        if 0 != MyParity(0)                       { return  1; }
+        if 1 != MyParity(1)                       { return  2; }
+        if 1 != MyParity(2)                       { return  3; }
+        if 0 != MyParity(3)                       { return  4; }
+        if 0 != MyParity(0xFF)                    { return  5; }
+        if 1 != MyParity(0x100)                   { return  6; }
+        if 0 != MyParity(0xFFFF)                  { return  7; }
+        if 1 != MyParity(0x10000)                 { return  8; }
+        if 1 != MyParity(0x7F00000000000000)      { return  9; }
+        if 0 != MyParity(0x0F00000000000000)      { return 10; }
+                                                         
+        if 0 != MyParity(-1)                      { return 11; }    // 0xFFFFFFFFFFFFFFFF
+        if 1 != MyParity(-2)                      { return 12; }    // 0xFFFFFFFFFFFFFFFE
+
+        return 0;
+    }
+
+}
--- a/src/QirRuntime/test/QIR-static/qsharp/qir-test-other.qs
+++ b/src/QirRuntime/test/QIR-static/qsharp/qir-test-other.qs
@@ -1,38 +1,25 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-namespace Microsoft.Quantum.Bitwise {
-
-    function MyParity(param : Int) : Int {
-        mutable tmpParam = param;
-        mutable result = false;
-        while tmpParam != 0 {
-            set tmpParam = tmpParam &&& (tmpParam - 1);
-            set result = not result;
-        }
-        return (result ? 1 | 0);
-    }
-}
-
 namespace Microsoft.Quantum.Testing.QIR.Other {
 
     open Microsoft.Quantum.Bitwise;
 
     function ParityTest() : Int {
         //function Parity (a : Int) : Int
-        if 0 != MyParity(0)                       { return  1; }
-        if 1 != MyParity(1)                       { return  2; }
-        if 1 != MyParity(2)                       { return  3; }
-        if 0 != MyParity(3)                       { return  4; }
-        if 0 != MyParity(0xFF)                    { return  5; }
-        if 1 != MyParity(0x100)                   { return  6; }
-        if 0 != MyParity(0xFFFF)                  { return  7; }
-        if 1 != MyParity(0x10000)                 { return  8; }
-        if 1 != MyParity(0x7F00000000000000)      { return  9; }
-        if 0 != MyParity(0x0F00000000000000)      { return 10; }
-                                                         
-        if 0 != MyParity(-1)                      { return 11; }    // 0xFFFFFFFFFFFFFFFF
-        if 1 != MyParity(-2)                      { return 12; }    // 0xFFFFFFFFFFFFFFFE
+        if 0 != Parity(0)                       { return  1; }
+        if 1 != Parity(1)                       { return  2; }
+        if 1 != Parity(2)                       { return  3; }
+        if 0 != Parity(3)                       { return  4; }
+        if 0 != Parity(0xFF)                    { return  5; }
+        if 1 != Parity(0x100)                   { return  6; }
+        if 0 != Parity(0xFFFF)                  { return  7; }
+        if 1 != Parity(0x10000)                 { return  8; }
+        if 1 != Parity(0x7F00000000000000)      { return  9; }
+        if 0 != Parity(0x0F00000000000000)      { return 10; }
+                                                       
+        if 0 != Parity(-1)                      { return 11; }    // 0xFFFFFFFFFFFFFFFF
+        if 1 != Parity(-2)                      { return 12; }    // 0xFFFFFFFFFFFFFFFE
 
         return 0;
     }
--- a/src/Simulation/QSharpFoundation/Bitwise/Bitwise.qs
+++ b/src/Simulation/QSharpFoundation/Bitwise/Bitwise.qs
@@ -75,7 +75,9 @@ namespace Microsoft.Quantum.Bitwise {
 
 
     /// # Summary
-    /// Returns the bitwise PARITY of an integer (1 if its binary representation contains odd number of ones and 0 otherwise).
+    /// Returns the bitwise PARITY of an integer
+    /// (1 if its [two's complement](https://en.wikipedia.org/wiki/Signed_number_representations#Two's_complement)
+    /// representation contains odd number of ones and 0 otherwise).
     ///
     /// # Example
     /// ```qsharp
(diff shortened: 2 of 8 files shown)


src/Simulation/QSharpFoundation/Bitwise/Bitwise.qs
```diff
@@ -83,7 +83,17 @@ namespace Microsoft.Quantum.Bitwise {
     /// let x = Parity(a); // x : Int = 1.
     /// ```
     function Parity (a : Int) : Int {
-        body intrinsic;
+        mutable tmpParam = a;
+        mutable result = false;
+        while tmpParam != 0 {
+            // Clear (to `0`) the least significant `1` (not to confuse with the least significant bit). E.g.
```
cgranade, line 89:
```suggestion
            // Clear (set to `0`) the least significant `1` (not to be confused with the least significant bit). E.g.
```

src/QirRuntime/test/QIR-static/qsharp/qir-test-other.qs
```diff
@@ -0,0 +1,27 @@
...
+        if 1 != Parity(2)                       { return  3; }
+        if 0 != Parity(3)                       { return  4; }
+        if 0 != Parity(0xFF)                    { return  5; }
+        if 1 != Parity(0x100)                   { return  6; }
+        if 0 != Parity(0xFFFF)                  { return  7; }
+        if 1 != Parity(0x10000)                 { return  8; }
+        if 1 != Parity(0x7F00000000000000)      { return  9; }
+        if 0 != Parity(0x0F00000000000000)      { return 10; }
+                                                       
+        if 0 != Parity(-1)                      { return 11; }    // 0xFFFFFFFFFFFFFFFF
```
cgranade, line 21: The current behavior of `Parity` is to fail if its input is negative:

image

Your implementation generalizes that to return the parity of the twos-complement representation of negative inputs, which could be reasonable, but should at least be specified in the documentation comment for `Parity`.

Separately, it looks like [the language spec is currently a bit vague](https://github.com/microsoft/qsharp-language/blob/main/Specifications/Language/3_Expressions/BitwiseExpressions.md) on what `&&&` returns for negative integers. @bettinaheim, would it make sense to mention explicitly that the twos-complement representation is used in that case?

src/Simulation/QSharpFoundation/Bitwise/Bitwise.qs
```diff
@@ -83,7 +83,17 @@ namespace Microsoft.Quantum.Bitwise {
     /// let x = Parity(a); // x : Int = 1.
     /// ```
     function Parity (a : Int) : Int {
-        body intrinsic;
+        mutable tmpParam = a;
```
cgranade, line 86: It looks like the algorithm here is to clear bits one at a time, counting how many iterations are needed to do so? In that case, I'm not sure `tmpParam` communicates that algorithm? In particular, `tmpParam` isn't a parameter, such that that's a bit confusing. Perhaps `remaining`?

src/QirRuntime/test/QIR-static/qsharp/qir-test-other.qs
kuzminrobin: Hmm... :-/ It is rather surprising to find out about any current behavior of `Parity()`. My understanding was that `Parity()` is currently [defined `intrinsic`](https://github.com/microsoft/qsharp-runtime/blob/d9d08086d5c655afb0ad9ccf9f7c0870589ea7ef/src/Simulation/QSharpFoundation/Bitwise/Bitwise.qs#L85) and has no implementation. I assume, you might mean something related to generation of the C# code (from Q#), which I know nearly nothing about. 
It is also suspicious to see the `UInt64` in the end of your screenshot. It makes me think that currently the signature has a parameter of the _unsigned_ 64-bit integer type (as a opposed to _signed_ 64-bit integer - `Int`), which contradicts the [documentation](https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.bitwise.parity).
Another curious observation I had when I was writing a test. I got a compiler error for the fragment
`Parity(0xFFFFFFFFFFFFFFFF)` (64-bit value consisting of 1s in all bit positions). The error was telling that the value does not fit in the data type, or something like that. This made me think that the compiler treaded the value as an _unsigned_ 64-bit integer, which really does not fit in the _signed_ 64-bit type (at least in two's complement representation).
To summarize, I have a suspicion, that the failure described by you, happens _before_ the actual call to `Parity()` (and may be happens during compilation).

I'm updating the documentation comment for `Parity()` to the following:
```q#
    /// Returns the bitwise PARITY of an integer 
    /// (1 if its [two's complement](https://en.wikipedia.org/wiki/Signed_number_representations#Two's_complement) 
    /// representation contains odd number of ones and 0 otherwise).
```

src/Simulation/QSharpFoundation/Bitwise/Bitwise.qs
kuzminrobin: Yes, the algorithm is to clear bits one at a time. But instead of counting the iterations it flips the parity bit (the least significant bit in the `result` variable).
Renaming `tmpParam` to `remainingOnes`.

src/QirRuntime/test/QIR-static/qsharp/qir-test-other.qs
swernli: The implementation is here:
https://github.com/microsoft/qsharp-runtime/blob/d9d08086d5c655afb0ad9ccf9f7c0870589ea7ef/src/Simulation/QSharpFoundation/Bitwise/Bitwise.cs#L49
And indeed, the first thing it does is convert to a UInt64.

cgranade: @kuzminrobin: The current implementation is provided by the C# runtime, as you say. In particular, the implementation at [line 49 of `Bitwise.cs`](https://github.com/microsoft/qsharp-runtime/blob/cgranade/experimental/opensim/src/Simulation/QsharpFoundation/Bitwise/Bitwise.cs#L49) starts by converting the input to `Parity` (Q# `Int`, represented in the C# runtime as `long`) to a C# `ulong` with a call to `System.Convert.ToUInt64`. Thus, the error message in the screenshot is not from before `Parity`, but is an exception in the C# runtime that is converted to a Q# `fail`.

cgranade: @swernli: jinx...!

src/Simulation/QSharpFoundation/Bitwise/Bitwise.qs
kuzminrobin: I'm not opposing to the expression "set to `0`", just want to make sure that the decision is fully informed.

At my first job I was reading a microcontroller datasheet. In all cases there was written there either "clear (to '0')" or "set (to '1')". Since then it seemed to me very clear when the documentation was written like that. And if in any single case it was written just "clear" or "set", then I still was clearly understanding what they meant.
Later on the other datasheets were mentioning something like "to reset" means "to set to a power-on-reset value" (which for one entities/bits meant '0', for the others meant '1'). And it also seemed to me very clear, even if they were writing just "reset".
The expression "set to `0`" may sound slightly confusing or odd for those who have similar background as mine.

But if you have strong preference regarding that, then there is absolutely no problem for me, I'll make that change.

src/QirRuntime/test/QIR-static/qsharp/qir-test-other.qs
kuzminrobin: @swernli, @cgranade,
Thank you for the link and explanation, very informative!
As I understand, the reasons behind the cast/conversion to the _unsigned_ integer type were the same as [here](https://github.com/kuzminrobin/code_review_notes/blob/master/cpp_design_bookmarks.md#know-the-danger-of-overflowing-and-underflowing-the-signed-types) (still applicable to the implementations prior to C++20).

Grand Total:
I'm adding to the comment that the two's complement representation is used.
Changing the implementation to the one that was in `.cs`.

src/Simulation/QSharpFoundation/Bitwise/Bitwise.qs
kuzminrobin: Not applicable any more. Code fragment has been replaced.

src/Simulation/QSharpFoundation/Bitwise/Bitwise.qs
kuzminrobin: Not applicable any more.

src/Simulation/QSharpFoundation/Bitwise/Bitwise.qs
```diff
@@ -75,15 +75,38 @@ namespace Microsoft.Quantum.Bitwise {
 
 
     /// # Summary
-    /// Returns the bitwise PARITY of an integer (1 if its binary representation contains odd number of ones and 0 otherwise).
+    /// Returns the bitwise PARITY of an integer
+    /// (1 if its [two's complement](https://en.wikipedia.org/wiki/Signed_number_representations#Two's_complement)
+    /// representation contains odd number of ones and 0 otherwise).
```
cgranade, line 80: Some clients (e.g.: Visual Studio) don't support rich formatting in summaries, such that for maximum portability, it's generally best to keep the summary to plain text only and to defer links and other formatting to a `/// # Description` section:
```suggestion
    /// Returns the bitwise parity of an integer.
    ///
    /// # Description
    /// This function outputs the bitwise parity of the 
    /// [two's complement](https://en.wikipedia.org/wiki/Signed_number_representations#Two's_complement)
    /// representation of its input, returning `1` if that representation
    /// contains an odd number of ones, and returning `0` otherwise.
```

kuzminrobin: Fixed.

src/Simulation/QSharpFoundation/Bitwise/Bitwise.qs
```diff
@@ -75,15 +75,38 @@ namespace Microsoft.Quantum.Bitwise {
...
     /// ```
     function Parity (a : Int) : Int {
-        body intrinsic;
+        mutable v = a;
+        // http://graphics.stanford.edu/~seander/bithacks.html#ParityMultiply
+        // XOR the bits in every 2-bit pair, save the result in the least significant bit (LSB) of the pair:
+        set v = v ^^^ (v >>> 1);    // bit[0] = bit[0] ^ bit[1]; bit[2] = bit[2] ^ bit[3]; ..
+        // Now only the even bits contain the information.
+        // XOR the even bits in every 4-bit nibble, save the result in the LSB of the nibble:
+        set v = v ^^^ (v >>> 2);    // bit[0] = bit[0] ^ bit[2]; bit[4] = bit[4] ^ bit[6]; ..
```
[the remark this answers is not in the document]
kuzminrobin: Changed.

src/Simulation/QSharpFoundation/Bitwise/Bitwise.qs
```diff
@@ -75,15 +75,38 @@ namespace Microsoft.Quantum.Bitwise {
...
     /// ```qsharp
     /// let a = 248;
     /// let x = Parity(a); // x : Int = 1.
     /// ```
     function Parity (a : Int) : Int {
-        body intrinsic;
+        mutable v = a;
+        // http://graphics.stanford.edu/~seander/bithacks.html#ParityMultiply
+        // XOR the bits in every 2-bit pair, save the result in the least significant bit (LSB) of the pair:
+        set v = v ^^^ (v >>> 1);    // bit[0] = bit[0] ^ bit[1]; bit[2] = bit[2] ^ bit[3]; ..
```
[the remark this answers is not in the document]
kuzminrobin: Changed.

cgranade (approved):
This PR does modify the behavior of a Q# standard library function, but having read through I think that makes sense — the current behavior is really a bug in the C#-based runtime caused by a cast to `ulong`. From that perspective, I'm happy to approve and skip formal API review.

PR #533, opened 2021-02-25
Tracer: Fix the target.qs file and link in QIS lib

irinayat-MS:
In response to PR comments I made some changes to target.qs but then forgot to regenerate it. This change fixes that and it also makes the Tracer to link in QIS bridge and support libs.

### DIFF
--- a/src/QirRuntime/test/QIR-tracer/tracer-target.qs
+++ b/src/QirRuntime/test/QIR-tracer/tracer-target.qs
@@ -3,26 +3,35 @@
 
 namespace Microsoft.Quantum.Instructions {
 
+    // We'll use TargetInstruction attribute to suppress Q#'s compiler decoration of names for the generated callbacks.
+    open Microsoft.Quantum.Targeting;
+
+    @TargetInstruction("single_qubit_op")
     operation single_qubit_op(op_id: Int, duration: Int, qb : Qubit) : Unit {
         body intrinsic;
     }
 
+    @TargetInstruction("multi_qubit_op")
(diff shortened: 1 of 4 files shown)


src/QirRuntime/test/QIR-tracer/tracer-target.qs
```diff
@@ -3,26 +3,35 @@
 
 namespace Microsoft.Quantum.Instructions {
 
+    // We'll use TargetInstruction attribute to suppress Q#'s compiler decoration of names for the generated callbacks.
+    open Microsoft.Quantum.Targeting;
+
+    @TargetInstruction("single_qubit_op")
     operation single_qubit_op(op_id: Int, duration: Int, qb : Qubit) : Unit {
```
cgranade, line 10: With the addition of the target instruction attributes, would it now be possible to make these operation follow the Q# style guide?

swernli (approved):

src/QirRuntime/test/QIR-tracer/tracer-target.qs
irinayat-MS: We could, but it would mean different names for the same thing that makes it harder to search across projects and connect to the implementations. The operation is local to this target file and isn't meant to ever be called from user code directly. Could we possibly keep it as is?

irinayat-MS: I'll keep it as is in this change as need to unblock build. Can revisit any time.

Issue #534, opened 2021-02-25
Classical Control rewrite support should be refactored


swernli:
https://github.com/microsoft/qsharp-runtime/blob/main/src/Simulation/QSharpFoundation/ClassicalControl.qs defines the Q# operations used in our classically controlled rewrite step (see https://github.com/microsoft/qsharp-compiler/blob/main/src/QsCompiler/Transformations/ClassicallyControlled.cs). As mentioned in a comment on the qs file, the namespace name is out of date and should be updated, along with other clean-up and refactoring to allow it to be used correctly by the rewrite step <s>without needing to be exposed publicly</s>. This refactor will need to take several steps:

- [x] Minimize number of intrinsics in ClassicalControl.qs to reduce load on implementers (#529)
- [x] Update generated code for concrete intrinsics that use interfaces to not be gate specific. (#538)
- [x] Rename namespace, updating qsharp-compiler rewrite step with new name. (#692, https://github.com/microsoft/qsharp-compiler/pull/1037)
- [x] <s>Switch Q# foundation to use interfaces for intrinsic implementation.</s> not required for rename.
- [x] <s>Change remaining callables in ClassicalControl.qs to be internal.</s> control callables will not be internal.
- [x] <s>Change all non-intrinsic callables in ClassicalControl.qs to be internal. (#535)</s> control callables will not be internal.

PR #535, opened 2021-02-25
Make non-intrinsic ClassicalControl callables internal

swernli:
This continues the clean-up of classical control helper callables by making them internal. See #534.


swernli:
Huh, I didn't anticipate how this change would negatively interact with the QIR tests for classical control helpers. I need to see if there is a way to both make this change and keep those tests working...

swernli:
Upon further review, making these helpers internal is not a goal for our design (we want people to be able to write Q# code that calls into them, and it it turns out making them actually internal is very hard). Closing this in favor of keeping the API public.

PR #537, opened 2021-02-26
Updated Version Numbers

ScottCarda-MS:
The old version number seemed to reference data types that are not available in the main branch of the compiler repo, but rather a not-yet merged in feature branch. The new version number references a package made from the main branch of the compiler repo.

### DIFF
--- a/Simulation.sln
+++ b/Simulation.sln
@@ -63,7 +63,7 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "IonQExe", "src\Simulation\S
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "QCIExe", "src\Simulation\Simulators.Tests\TestProjects\QCIExe\QCIExe.csproj", "{C015FF41-9A51-4AF0-AEFC-2547D596B10A}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "TargetedExe", "src\Simulation\Simulators.Tests\TestProjects\TargetedExe\TargetedExe.csproj", "{D292BF18-3956-4827-820E-254C3F81EF09}"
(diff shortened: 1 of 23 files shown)


swernli (changes requested):
This won’t be able to merge. We are in an awkward state where QIR support is in main here, including tests that are part of the build pipeline, while the required changes to the compiler to make QIR generations work are in a working branch of Qsharp-compiler. We need to get those changes merged into main in compiler, but to keep runtime building we need to use those packages here.

ScottCarda-MS:
> This won’t be able to merge. We are in an awkward state where QIR support is in main here, including tests that are part of the build pipeline, while the required changes to the compiler to make QIR generations work are in a working branch of Qsharp-compiler. We need to get those changes merged into main in compiler, but to keep runtime building we need to use those packages here.

How long before the main branch on the compiler repo has the data types needed to work with this?

ScottCarda-MS:
> > This won’t be able to merge. We are in an awkward state where QIR support is in main here, including tests that are part of the build pipeline, while the required changes to the compiler to make QIR generations work are in a working branch of Qsharp-compiler. We need to get those changes merged into main in compiler, but to keep runtime building we need to use those packages here.
> 
> How long before the main branch on the compiler repo has the data types needed to work with this?

And can those tests be marked as Skip until that happens?

ScottCarda-MS:
> > > This won’t be able to merge. We are in an awkward state where QIR support is in main here, including tests that are part of the build pipeline, while the required changes to the compiler to make QIR generations work are in a working branch of Qsharp-compiler. We need to get those changes merged into main in compiler, but to keep runtime building we need to use those packages here.
> > 
> > 
> > How long before the main branch on the compiler repo has the data types needed to work with this?
> 
> And can those tests be marked as Skip until that happens?

It sounds like there will be some time put aside next week to address this and try to get the main of the compiler repo to agree with the main of the runtime repo. If so, this PR may not be needed.

swernli:
In the meantime, if you are blocked, you can try `0.15.2102129941-alpha`, which should hopefully have the changes you need in it.

ScottCarda-MS:
This was addressed in the compiler PR here (https://github.com/microsoft/qsharp-compiler/pull/893). With this the main branch of the compiler is now compatible with the main branch of this repo. This makes this PR unnecessary now.

PR #539, opened 2021-02-26
Tracer: Support of conditionals on measurement results

irinayat-MS:
@msoeken - FYI


src/QirRuntime/public/TracerTypes.hpp
```diff
@@ -10,8 +10,9 @@ namespace Quantum
     using OpId = int;
     using Time = int;
     using Duration = int;
-    using LayerId = size_t;
+    using LayerId = int64_t;
 
-    constexpr LayerId INVALID = std::numeric_limits<size_t>::max();
+    constexpr LayerId INVALID = std::numeric_limits<int64_t>::min();
+    constexpr LayerId REQUESTNEW = std::numeric_limits<int64_t>::max();
```
kuzminrobin, line 16: Consider
    constexpr LayerId INVALID = std::numeric_limits<**LayerId**>::min();
    constexpr LayerId REQUESTNEW = std::numeric_limits<**LayerId**>::max();
    
If one replaces the type (`int64_t`) behind the `LayerId`, but forgets to replace the same type for `INVALID` and/or `REQUESTNEW`, then the behavior can become unexpected.

kuzminrobin:
Any comment about what the `tracer-qir.bc` binary file is for and why a binary file is needed in the repo?

src/QirRuntime/lib/Tracer/tracer.hpp
```diff
@@ -75,6 +75,14 @@ namespace Quantum
         // layer that preceeded it, even if the new operations involve completely new qubits.
         LayerId globalBarrier = INVALID;
 
+        // The temporary barriers between layers that force operations, guarded by a condition on a measurement result,
+        // to be placed into layers _after_ the measurement.
+        std::vector<LayerId> conditionalFences;
```
kuzminrobin, line 80: Minor:
> // The temporary **barriers** between

Consider any word other than _barrier_ (delimiter, entity, ..).

irinayat-MS:
> Any comment about what the `tracer-qir.bc` binary file is for and why a binary file is needed in the repo?

Oops, not needed. Checked in by mistake. Thanks for catching it!

src/QirRuntime/lib/Tracer/tracer.hpp
```diff
@@ -105,10 +113,19 @@ namespace Quantum
         // Adds operation with given id into the given layer. Assumes that duration contraints have been satisfied.
         void AddOperationToLayer(OpId id, LayerId layer);
 
-        // Update the qubit state with the new layer information
+        // Update the qubit state with the new layer information.
         void UpdateQubitState(Qubit q, LayerId layer, Duration opDuration);
 
+        // Considers global barriers and conditional fences to find the barrier currently in effect.
+        LayerId GetEffectiveBarrier() const;
```
kuzminrobin, line 120: Minor (to be considered later after, this PR):

This fragment gives a hint that probably it makes sense to consider figuring out either 
(a) common part between 
or 
(b) a common name for 
the two concepts - _[global] barriers_ and _conditional fences_.
Something like
(a) _global barriers_ and _conditional barriers_ - the common part is _barriers_. The function name stays `GetEffectiveBarrier()`.
(b) common name - _delimiters_, _isolators_, etc. Assuming something like "there are 2 types of _delimiters_ - _the [global] barriers_ and _the conditional fences_". The function name - `GetEffectiveDelimiter()`.

src/QirRuntime/lib/Tracer/tracer.hpp
```diff
@@ -105,10 +113,19 @@ namespace Quantum
...
 
+        // Considers global barriers and conditional fences to find the barrier currently in effect.
+        LayerId GetEffectiveBarrier() const;
+
+        // For the given results finds the latest layer of the measurements that produced the results.
+        LayerId FindLatestMeasurementLayer(long count, Result* results);
+
       public:
+        // INVALID LayerId is treated as -Infinity, and REQUESTNEW -- as +Infinity
+        static LayerId LaterLayer(LayerId l1, LayerId l2);
```
kuzminrobin, line 127: Minor:
From the function name, function signature, and comment it is unclear what the function does with the params.
For a _public_ func I would prefer that to be clear without looking into the implementation.

src/QirRuntime/lib/Tracer/tracer.cpp
```diff
@@ -251,18 +273,76 @@ namespace Quantum
...
+        }
+
+        vector<LayerId>& fences = this->tracer->conditionalFences;
+        assert(!fences.empty());
+        this->tracer->conditionalFences.pop_back();
+
+        // Update the latest layer (we expect the stack of fences to be shallow so a linear search through it
+        // should be OK).
+        this->tracer->latestConditionalFence =
+            fences.empty() ? INVALID : *std::max_element(fences.begin(), fences.end());
```
kuzminrobin, line 343: Question:

The `pop_back()` in line 338 makes me _allow_ that `this->tracer->conditionalFences` _can become empty_ (and the data block is allowed to be deallocated). That is why the `fences` (introduced in line 336) can become (after line 338) a reference to an _empty vector_.
In the line 343 the `std::max_element()` will return (for an empty vector) the `fences.end()` - the iterator pointing to _after_ the data (or even after the deallocated block). The subsequent dereferencing `*` of such an iterator can cause the undefined behavior.

What does guarantee that I'm mistaken?

src/QirRuntime/lib/Tracer/tracer.cpp
```diff
@@ -81,28 +93,37 @@ namespace Quantum
...
+        LayerId layerToInsertInto = REQUESTNEW;
 
-        const LayerId firstLayerAfterBarrier =
-            this->globalBarrier == INVALID
-                ? this->metricsByLayer.empty() ? INVALID : 0
-                : this->globalBarrier + 1 == this->metricsByLayer.size() ? INVALID : this->globalBarrier + 1;
+        const LayerId barrier = this->GetEffectiveBarrier();
+        const LayerId firstLayerAfterBarrier = barrier == INVALID ? this->metricsByLayer.empty() ? REQUESTNEW : 0
+                                               : barrier + 1 == this->metricsByLayer.size() ? REQUESTNEW
+                                                                                            : barrier + 1;
```
kuzminrobin, line 121: Minor:
I believe everything is correct here, but for those who are uncertain, this fragment can encourage them to spend time for double-checking the correctness of logic here. I would recommend placing parentheses explicitly.
Also [FYI](https://www.viva64.com/en/b/0391/#ID4F03F23BDD).

kuzminrobin (approved):
I looked at the changes only and didn't dive into the existing logic.
The changes look good to me. Approving.

src/QirRuntime/lib/Tracer/tracer.hpp
irinayat-MS: I've updated the readme to explain that "being a fence" is a role that can be assigned to any layer, and that "a barrier" is a special layer, that can be injected by the user. To align with this terminology clarification, the method has been renamed to `GetEffectiveFence`.

swernli (approved):

PR #541, opened 2021-02-27
Clean up ClassicalControl.qs

swernli:
Minor clean-up for ClassicalControl.qs and associated C# implementation

Related to work in Classical Control rewrite support should be refactored #534


ScottCarda-MS (approved):
lgmt

PR #547, opened 2021-03-05
Adapt runtime to latest LSP package.

cgranade:
This PR contains changes to the runtime needed to support microsoft/qsharp-compiler#886.


src/Simulation/CSharpGeneration/Microsoft.Quantum.CSharpGeneration.fsproj
```diff
@@ -22,7 +22,7 @@
 
   <ItemGroup>
     <PackageReference Update="FSharp.Core" Version="4.7.0" />
-    <PackageReference Include="Microsoft.Quantum.Compiler" Version="0.15.2102129527-alpha" />
+    <PackageReference Include="Microsoft.Quantum.Compiler" Version="0.15.2103130749-beta" />
```
swernli, line 25: To update this version, you need to also update every QDK version through the whole repo. However, because there are changes in qsharp-runtime main that aren't yet in qsharp-compiler main, doing so to a beta package that is missing those dependent changes means it will fail the build. We are working in getting out of this bad state, but for now our best bet is to merge compiler main into one of our working branches, do an alpha build, and and then have you update to that package version across this repo.

I'll kick off a build that hopefully will resolve this issue for you.

ricardo-espinoza (approved):

Issue #544, opened 2021-03-04
QIR Runtime should support alternate or custom target packages


swernli:
Right now, the QIR runtime is built using the explicit list of gates from the QSharp.Core target package, as seen from the definition of the `IQuantumGateSet` in the public header [QuantumApi_I.h](https://github.com/microsoft/qsharp-runtime/blob/main/src/QirRuntime/public/QuantumApi_I.hpp#L11). Because of this, any gate sets that don't have the same list of intrinsics will need to either copy and modify the runtime files or provide stubs for the core set and add additional ones. This can be especially problematic if the alternate gate set defines gates of the same name but with different signatures. We need to separate the gate set from the math related intrinsics and the runtime support such that a generic, gate set agnostic runtime library can be produced and added to a gate set definition to complete the implementation.

PR #549, opened 2021-03-07
Refactor QIR runtime CMakeLists and dynamic libraries

swernli:
This change introduces a refactor and fixup of the CMake infrastructure and powershell scripts for our QirRuntime build. This is part of the work needed to address #544. In particular, it splits out what was one dynamic library target, qdk, and turns it into three targets for the main QIR runtime, the QSharp Core functionality (intrinsic gates and simulator), and the QIR tracer. It fixes the generation of that dll by including additional support for `dllexport` when compiling on Windows and updates the corresponding IR files to also properly export API. The test for dynamic loading was subtly incorrect in a way that resulted in actually statically linking the dependencies, so this adjusts the test CMake to more explicitly use dynamic linking.

Renaming any variables and refactoring tests further into their own separate CMake project were out of scope for this change and will be addressed in a later PR.


swernli:
Of note: we want to pick better names than `qir.dll`, `QSharpCore.dll` and `qirtracer.dll`, ideally something that matches our naming schema of `Microsoft.Quantum.*`. Given that these don't have any specific project file or corresponding namespace like the C# code there isn't an obvious choice. Any suggestions? So far my best thought is: `Microsoft.Quantum.Qir.Runtime.dll`, `Microsoft.Quantum.Qir.Qsharp.Core.dll`, and `Microsoft.Quantum.Qir.Tracer.dll`.

ScottCarda-MS:
> Of note: we want to pick better names than `qir.dll`, `QSharpCore.dll` and `qirtracer.dll`, ideally something that matches our naming schema of `Microsoft.Quantum.*`. Given that these don't have any specific project file or corresponding namespace like the C# code there isn't an obvious choice. Any suggestions? So far my best thought is: `Microsoft.Quantum.Qir.Runtime.dll`, `Microsoft.Quantum.Qir.Qsharp.Core.dll`, and `Microsoft.Quantum.Qir.Tracer.dll`.

`Microsoft.Quantum.` makes sense because that's who we are. `Qir` makes sense because that is the more specific area. `Runtime.dll` `Core.dll` and `Tracer.dll` make sense because those are the names of the specific dlls. What does the `Qsharp` part of the `Microsoft.Quantum.Qir.Qsharp.Core.dll` infer? Are there multiple dlls under the idea of `Qsharp`? It might me a bit redundant because `Qir` is already specific to Q#, right? Or perhaps it is the case that the `Runtime.dll` and the `Tracer.dll` are agnostic of the QIR coming from Q#, whereas the `Core.dll` is specific to QIR that was generated from Q#? I could see the necessity for `Qsharp` in the name if that were the case. Otherwise I'm leaning towards dropping the `Qsharp` from core's name. Otherwise the naming you suggest makes perfect sense to me, and I like it. That's my 2 cents anyways.

swernli:
Yeah, I agree that `Microsoft.Quantum.Qir.QSharp.Core.dll` is both the wordiest and the hardest to reason about. A few points to respond to...

> Are there multiple dlls under the idea of `Qsharp`?

No, I don't expect there will be more under `QSharp` which makes it weird to have `Core` under there. I feel like our names end up trying to serve dual purposes, where the dots form a hierarchy like a folder structure, while also forming the logical separation in a descriptive name. The meaning I was aiming for is that this dll is the QIR support for the QSharp Core package, which is the default target package when compiling Q#. This is as opposed to `Type1.Core` or `Type2.Core` which are target packages with alternate gate sets and would require a different QIR dll to compile to an executable. We already have `Microsoft.Quantum.QSharp.Core.dll` which comes from this repo, under [src/Simulation/QSharpCore](https://github.com/microsoft/qsharp-runtime/tree/main/src/Simulation/QSharpCore). Since everything defined in here is under the `__quantum__qis` prefix because it corresponds to Q# callables defined as `body intrinsic`, maybe `Microsoft.Quantum.Qir.Intrinsic.QSharp`?

> It might me a bit redundant because Qir is already specific to Q#, right?

QIR itself is not Q# specific, which is an important point worth doubling down on. [Alan's blog post](https://devblogs.microsoft.com/qsharp/introducing-quantum-intermediate-representation-qir/) announcing QIR put it very directly:

> QIR is intended to serve as a common interface between many languages and target quantum computation platforms. Thus, while it supports Q#, QIR is not specific to Q#: any programming language for gate-based quantum computing can be represented in QIR. Similarly, QIR is hardware-agnostic: it does not specify a quantum instruction or gate set, leaving that to the target computing environment.

That idea was part of what motivated me to make this split into multiple dlls in the first place. Having a set of gates defined in the QIR runtime, even if it's intended as an optional interface developers can choose not to inherit from, implies tying the infrastructure to a specific quantum hardware. So the idea was to separate the quantum gates from the rest and make it easy for someone to swap in a different set of intrinsic gates. That said, this is all implementation for a specific QIR runtime; the bridge that lives in what we want to call `Microsoft.Quantum.Qir.Runtime.dll` is still quite Q# specific, given that most of the code is support for the types defined in the language (and intrinsics that come from `Microsoft.Quantum.QSharp.Foundation`). So there is QIR the specification, which is a completely language and hardware agnostic convention for LLVM IR, and then there is this particular runtime library that provides Q# specific implementations compatible with QIR but agnostic across gate sets, then separately there are implementations for intrinsics that match gate sets in specific Q# target packages. While that makes sense to me from a design standpoint, I struggle with how to convey the distinction between these three things in a short name :).

swernli:
Hmm... the more I think about it, the more I feel like conveying that this is a Q# compatible runtime for QIR is important, so maybe that means working that into the dll naming hierarchy. Something like:

- `Microsoft.Quantum.QSharp.Qir.Runtime` - the runtime support for Q# types and non-gate intrinsics.
- `Microsoft.Quantum.QSharp.Qir.Core` - the quantum gate set support for Q#'s default target package.
- `Microsoft.Quantum.QSharp.Qir.Tracer` - the infrastructure for tracing Q# programs compiled to QIR.

src/QirRuntime/build-qir-runtime.ps1
```diff
@@ -1,68 +1,86 @@
...
-        $env:CXX = "C:\Program Files\LLVM\bin\clang++.exe"
-        $env:RC = "C:\Program Files\LLVM\bin\clang++.exe"
-        $llvmExtras = (Join-Path $PSScriptRoot "externals/LLVM")
-        $env:PATH += ";$llvmExtras"
-    } else {
-        Write-Host "##vso[task.logissue type=error;]Failed to identify the OS. Will use default CXX compiler"
-    }
+if (($IsMacOS) -or ((Test-Path Env:AGENT_OS) -and ($Env:AGENT_OS.StartsWith("Darwin"))))
+{
+    Write-Host "On MacOS build QIR Runtim using the default C/C++ compiler (should be AppleClang)"
```
kuzminrobin, line 25: Typo: QIR Runtim**e** using

src/QirRuntime/build-qir-runtime.ps1
```diff
@@ -1,68 +1,86 @@
...
+    $env:CXX = "/usr/bin/clang++-11"
+    $env:RC = "/usr/bin/clang++-11"
+    $clangTidy = "-DCMAKE_CXX_CLANG_TIDY=clang-tidy-11"
+}
+elseif (($IsWindows) -or ((Test-Path Env:AGENT_OS) -and ($Env:AGENT_OS.StartsWith("Win"))))
+{
+    Write-Host "On Windows build QIR Runtime using Clang"
+    $env:CC = "C:\Program Files\LLVM\bin\clang.exe"
+    $env:CXX = "C:\Program Files\LLVM\bin\clang++.exe"
+    $env:RC = "C:\Program Files\LLVM\bin\clang++.exe"
```
kuzminrobin, line 40: (If you have time and desire )
:-\ Absolute paths... I understand that they were here prior to this PR... 
If you have time and desire then you may want to consider providing the path as an optional command line arg or env var. (in case the user has multiple versions installed and wants to run against a specific version)

cgranade: It's also worth noting that the intent of using PowerShell in these repos is for cross-platform compatibility, such that contributors compiling on non-Windows platforms won't have a Program Files folder.

kuzminrobin (approved):
Impressive number of changes.

src/QirRuntime/build-qir-runtime.ps1
kuzminrobin: @cgranade, I believe this particular fragment is Windows-specific, so the users will have the "Program Files", however your comment is still applicable. The users may have installed clang to the non-default location (outside of "Program Files").

swernli:
Just thinking out loud here, but I wonder if it makes sense to split this out even further, perhaps into another repo (yes, I know the last thing we want is another repo, but hear me out). Everything that is part of `__quantum__qis` is part of a specific target's quantum intrinsics. In our case, these come from `body intrinsic` Q# callables defined in either `Microsoft.Quantum.QSharp.Foundation` as part of the common language support or are part of a target package defining a quantum gate set like like `Microsoft.Quantum.QSharp.Core`. So maybe it makes sense to have all the truly generic elements in `__quantum__rt` part of their own repo along with the QIR spec that produces the library that all QIR consumers or developers would need, and then in this repo having only the Q# specific portions that are part of the qis.

@bettinaheim, @alan-geller Any thoughts on this?

src/QirRuntime/build-qir-runtime.ps1
swernli: Agreed, absolute paths are generally bad. In this case though, the script updates these variables based on the detected platform, and is already used in the build pipeline to run on OS X and Ubuntu in addition to Windows. Because of the way chocolatey installs LLVM on Windows and how we use the setup script to deploy dependencies, we need to use this full path or else the exe won't be found on the pipeline.

swernli: As an alternative, I could avoid needing to specify paths at all by updating the setup script to add the installed LLVM tools to the path. Let me try that approach; should hopefully make for a cleaner script.

src/QirRuntime/lib/QSharpCore/.clang-tidy
```diff
@@ -0,0 +1 @@
+Checks: '-*,bugprone-*'
```
cesarzc, line 1: I am wondering what were the specific reasons for not using the parent's clang-tidy configuration?

One pattern that might be useful if it is just a couple of checks that are different from the parent configuration would be using `InheritParentConfig: true` and adding/removing the necessary checks.

cesarzc (approved):
This change is great! I am a fan of the new naming convention and the ability to do dynamic linking :)

src/QirRuntime/build-qir-runtime.ps1
```diff
@@ -1,68 +1,94 @@
...
+}
 
-    if ($LastExitCode -ne 0) {
-        Write-Host "##vso[task.logissue type=error;]Failed to build QIR Runtime."
-    }
-} else {
-    Write-Host "To enable build of QIR Runtime set ENABLE_QIRRUNTIME environment variable to 'true'"
+$os = "win32"
+$pattern = "*.dll"
+if ($IsMacOS) {
```
kuzminrobin, line 77: Out of curiosity:
Why _32_ (in "win32")? Why not just "win"?

swernli: You know, to be honest, it's just a bad habit leftover from working on Windows... I type it without meaning to and don't even register it on the screen when I read it! There is no reason for the "32" here and it's probably more of a distraction than anything. I'll update my next PR to remove this!

swernli: Fix it in https://github.com/microsoft/qsharp-runtime/pull/553/commits/065589a3492580c0682ff40ef31eb873558fb581

PR #550, opened 2021-03-09
GetState variant for register

msoeken:
This adds an overload to `GetState` to support dumping states from a sub register in order to allow behavior as in `DumpRegister`.

* [x] Implementation
* [ ] Tests

Are there any comments on the interface? If not, I can add some tests to complete the PR.


bettinaheim:
@kuzminrobin This is actually related to the work to support dump register I believe. Take a look when you get to it.

msoeken:
@bettinaheim, @kuzminrobin, exactly, I was implementing this in order to have an operation such as `DumpRegister`.

kuzminrobin:
> @bettinaheim, @kuzminrobin, exactly, I was implementing this in order to have an operation such as `DumpRegister`.

Mathias, thanks for working on this. I'm also working on `DumpMachine()` enabling in QIR, and the `DumpRegister()` was the next item on my list, but I didn't yet get to it.
You can take a look at [the changes in my branch](https://github.com/microsoft/qsharp-runtime/compare/kuzminrobin/dumpMachine) to see where I'm moving to (the code does not compile yet, the branch has 2 commits).
Just in case I will also forward you the email discussion we had recently about `DumpMachine()`. 
And you may find the issue #592 slightly applicable to this PR.

PR #553, opened 2021-03-10
Refactor and rename QIR runtime headers and interfaces

### LINKED ISSUE #544: QIR Runtime should support alternate or custom target packages

swernli:
This is the second change that refactors the QIR runtime to split it into clearly divided chunks (see #549 for the first part). For this part, it renames `ISimulator` to `IRuntimeDriver` and moves gate set and diagnostic interfaces out into their own headers. While this requires users of the interface to use `dynamic_cast` to convert between interfaces on a object that implements `IRuntimeDriver` instead of using access methods, but that is a necessary evil to disentangle types and allow developers building their own intrinsic set (either for Q# or a different language) to take as a dependency just the headers and libs that correspond to the built-in QIR runtime and not any Q#-specific functionality.

This resolves #544.


cesarzc (approved):
Looks good to me! I like that now measurement is part of `IQuantumGateSet`.

PR #558, opened 2021-03-12
Copy generated import lib for QIR on Windows

swernli:
I forgot to include the generated import libs needed for compiling against the dlls on Windows in the build artifacts for the QIR runtime.


bettinaheim (approved):

swernli:
@cesarzc This is kind of annoying, and I wanted to give you the heads up. To use dlls at runtime on Windows we need to provide matching import libs which have the info needed for the linker. However, the only way to make sure those get "binplaced" by our install command in cmake is to include `ARCHIVE` which will also end up copying out any static libs that are generated. For now, I'm just putting all `*.lib` files into the drops folder for Windows, but once we get a formal packaging process we can be more discerning about which libs are included and avoid causing confusion by packing static libs that are redundant to the dynamic libraries produced.

cesarzc:
> 
> 
> @cesarzc This is kind of annoying, and I wanted to give you the heads up. To use dlls at runtime on Windows we need to provide matching import libs which have the info needed for the linker. However, the only way to make sure those get "binplaced" by our install command in cmake is to include `ARCHIVE` which will also end up copying out any static libs that are generated. For now, I'm just putting all `*.lib` files into the drops folder for Windows, but once we get a formal packaging process we can be more discerning about which libs are included and avoid causing confusion by packing static libs that are redundant to the dynamic libraries produced.

Thanks for the heads up @swernli!

PR #567, opened 2021-03-17
Update C# generation tests for changes from HM type inference

samarsha:
This PR updates C# generation tests to match changes in the syntax tree related to identifier type arguments caused by microsoft/qsharp-compiler#940.


src/Simulation/CSharpGeneration.Tests/SimulationCodeTests.fs
```diff
@@ -2183,32 +2183,48 @@ namespace N1
                     (__arg19__.Item1, __arg19__.Item2, (1L, Result.One))
                 ))
                 .Apply((0L, Result.Zero));"
-            "partialGeneric1__.Partial((0L, _, (1L, _))).Apply((Result.Zero, Result.One));"
-            "partialGeneric2__.Partial((0L, Result.Zero, (_, Result.One))).Apply(1L);"
-            "partialGeneric2__.Partial((_, _, (1L, Result.One))).Apply((0L, Result.Zero));"
-            "partialGeneric2__.Partial((0L, _, (1L, _))).Apply((Result.Zero, Result.One));"
+            "partialGeneric1__
```
bettinaheim, line 2186: How come the generated code for these changed? I don't think the type inference change should have an impact on the generated C# code, or change the representation in the syntax tree.

samarsha: The syntax tree did change, in that resolved identifiers now always have a type argument list attached to them, even if there was no `<T1, T2, ...>` list in the original source code.

swernli (approved):

src/Simulation/CSharpGeneration.Tests/SimulationCodeTests.fs
bettinaheim: Ok. Double check if there are any perf implications, but otherwise that's fine.

bettinaheim (approved):
Please quickly check the perf benchmarks before merging.

PR #576, opened 2021-03-21
Fix for dynamic type load when inheritance used

swernli:
If inheritance is used, the `this` in QCTraceSimulatorImpl wont resolve to the right assembly. This uses the QCTraceSimulatorImpl type itself so that the assembly matching the QDK can be resolved.


ScottCarda-MS (approved):

anpaz (approved):
lgtm

PR #580, opened 2021-03-23
Update reference C++ entry-point driver to use the latest QIR emission

cesarzc:
This change updates the reference C++ entry-point driver to be compatible with the latest QIR emission.

One of the important changes is that it is now assumed that interop QIR functions are always `void`.

### DIFF
--- a/src/QirRuntime/samples/StandaloneInputReference/qir-driver.cpp
+++ b/src/QirRuntime/samples/StandaloneInputReference/qir-driver.cpp
@@ -1,7 +1,6 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-#include <cstring> // for memcpy
 #include <fstream>
 #include <iostream>
 #include <map>
@@ -10,30 +9,47 @@
 
 #include "CLI11.hpp"
 
-#include "CoreTypes.hpp"
 #include "QirContext.hpp"
-#include "QirTypes.hpp"
-#include "QirRuntimeApi_I.hpp"
 #include "SimFactory.hpp"
-#include "QirRuntime.hpp"
 
 using namespace Microsoft::Quantum;
 using namespace std;
 
-// This is the function corresponding to the QIR entry-point.
-extern "C" int64_t Quantum__StandaloneSupportedInputs__ExerciseInputs__body( // NOLINT
-    int64_t intValue,
-    double doubleValue,
-    Result resultValue,
-    QirString* stringValue);
+struct InteropArray
+{
+    int64_t Size;
+    void* Data;
+};
 
-const char FalseAsChar = 0x0;
-const char TrueAsChar = 0x1;
+struct InteropRange
+{
+    int64_t Start;
+    int64_t Step;
+    int64_t End;
+};
+
+// This is the function corresponding to the QIR entry-point.
+extern "C" void Quantum__StandaloneSupportedInputs__ExerciseInputs( // NOLINT
+    int64_t intValue,
+    InteropArray* integerArray,
+    double doubleValue,
+    InteropArray* doubleArray,
+    char boolValue,
+    InteropArray* boolArray,
+    char pauliValue,
+    InteropArray* pauliArray,
+    InteropRange* rangeValue,
+    char resultValue,
+    InteropArray* resultArray,
+    const char* stringValue);
+
+const char InteropFalseAsChar = 0x0;
+const char InteropTrueAsChar = 0x1;
 map<string, bool> BoolAsCharMap{
-    {"0", FalseAsChar},
-    {"false", FalseAsChar},
-    {"1", TrueAsChar},
-    {"true", TrueAsChar}};
+    {"0", InteropFalseAsChar},
+    {"false", InteropFalseAsChar},
+    {"1", InteropTrueAsChar},
+    {"true", InteropTrueAsChar}};
 
 map<string, PauliId> PauliMap{
     {"PauliI", PauliId::PauliId_I},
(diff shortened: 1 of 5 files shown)


src/QirRuntime/samples/StandaloneInputReference/qsharp/qir-standalone-input-reference.qs
```diff
@@ -1,15 +1,28 @@
...
 
+    function TautologyPredicate<'T> (input : 'T) : Bool
+    {
+        return true;
+    }
+
     @EntryPoint()
-    operation ExerciseInputs (intValue : Int, doubleValue : Double, resultValue : Result, stringValue : String) : Int {
-        Message("Exercise Supported Inputs");
+    operation ExerciseInputs (intValue : Int, intArray : Int[], doubleValue : Double, doubleArray : Double[], boolValue : Bool, boolArray : Bool[], pauliValue : Pauli, pauliArray : Pauli[], rangeValue : Range, resultValue : Result, resultArray : Result[], stringValue : String) : Unit {
```
bettinaheim, line 13: Is there a reason that string arrays and range arrays are not included in this entry point?

bettinaheim (approved):
LGTM.

src/QirRuntime/samples/StandaloneInputReference/qir-driver.cpp
```diff
@@ -234,29 +236,22 @@ int main(int argc, char* argv[])
...
-
-    // Close opened file buffers;
-    if (operationOutputFileStream.is_open())
-    {
-        operationOutputFileStream.close();
-    }
-
     if (simulationOutputFileStream.is_open())
     {
         simulationOutputFileStream.close();
```
kuzminrobin, line 257: **Bug?**
Look at the following sequence:

```c++
// (1) The vector is created:
vector<int64_t> integerVector;
// (2) The vector is filled in:
app.add_option(.., integerVector, ..)..
```
(3) Then the call happens:
unique_ptr\<InteropArray> integerArray = CreateInteropArray(**integerVector**);

(4) The function called has the following signature:
unique_ptr<InteropArray> CreateInteropArray(**vector\<T> v**)

The parameter `v` is passed _by value_ (as opposed to "by reference"). The parameter `v` is created as a copy of the argument **integerVector**. The parameter `v` is similar to a local variable of the function `CreateInteropArray()`. It has a pointer to the heap-allocated block. That pointer can be acquired with the call `v.data()`.

Now here is what happens in the function:
```c++
unique_ptr<InteropArray> CreateInteropArray(vector<T> v)
{
    unique_ptr<InteropArray> array(new InteropArray());
    array->Size = v.size();
    array->Data = v.data();   // Fragment of interest.
    return array;
}   // Fragment of interest.
```
The fragment `array->Data = v.data()` memorizes the pointer to the heap block.
At the end of the function `}` the parameter `v` goes out of the scope and is destroyed, its destructor deallocates the heap-allocated block. The `array->Data` becomes a dangling pointer, and the local variable `array` (containing such a dangling pointer) is returned. 

After the call (3) the local variable `integerArray` contains a dangling pointer.

Possible solution: pass _by reference_ rather than _by value_. I.e. change the signature to 
unique_ptr\<InteropArray> CreateInteropArray(vector\<T>**&** v)
(the ampersand `&` is added in front of the parameter name `v`).

src/QirRuntime/samples/StandaloneInputReference/qir-driver.cpp
```diff
@@ -10,90 +9,103 @@
...
-#include "QirRuntime.hpp"
 
 using namespace Microsoft::Quantum;
 using namespace std;
 
+struct InteropArray
+{
+    int64_t Size;
+    void* Data;
+};
```
kuzminrobin, line 22: Consider adding 
either a constructor taking 2 parameters (and initializing both member variables with those parameters), see example below,
or a default constructor initializing both, e.g to 0.
Otherwise there is a risk of creating an instance of `InteropArray` with the memeber variables **un**initialized.
```c++
struct InteropArray
{
    int64_t Size;
    void* Data;
    InteropArray(int64_t size, void* data)
        : Size(size)
        , Data(data)
    {}
};
```
Then the instances can be created like this: 
```c++
unique_ptr<InteropArray> array(new InteropArray(v.size(), v.data()));
```

src/QirRuntime/samples/StandaloneInputReference/qir-driver.cpp
cesarzc: Great catch! I'll change the signature and make sure that we exercise a read to data to make sure the values are correct.

kuzminrobin (approved):
Approved with suggestion.

src/QirRuntime/samples/StandaloneInputReference/qir-driver.cpp
cesarzc: Added constructors to `InteropArray` and to `InteropRange`.

src/QirRuntime/samples/StandaloneInputReference/qir-driver.cpp
cesarzc: Updated the signature and added additional Q# code that prints the values of array to verify execution correctness.

src/QirRuntime/samples/StandaloneInputReference/qsharp/qir-standalone-input-reference.qs
cesarzc: When simulating an entry-point with a `String` or `Range` array, the simulation crashes. This seems to be related to the generated QIR. I'll investigate this as a separate issue and update this reference

src/QirRuntime/samples/StandaloneInputReference/qsharp/qir-standalone-input-reference.qs
```diff
@@ -4,6 +4,28 @@
...
+            if (first)
+            {
+                set itemsString = itemsString + $"{item}";
+            }
+            else
+            {
+                set itemsString = itemsString + $", {item}";
+            }
+
+            set first = false;
```
kuzminrobin, line 22: Consider moving `set first = false;` to inside of `if (first) { .. }`.

src/QirRuntime/samples/StandaloneInputReference/qsharp/qir-standalone-input-reference.qs
```diff
@@ -4,6 +4,28 @@
...
+        }
+
+        set itemsString = itemsString + $"]";
+        return itemsString;
```
kuzminrobin, line 26: Why do you mark (with `$` char) the string "]" as an interpolated string? What shall we lose if we remove `$`?

kuzminrobin:
Reapproving.

src/QirRuntime/samples/StandaloneInputReference/qsharp/qir-standalone-input-reference.qs
cesarzc: Moved to inside the if-statement.

src/QirRuntime/samples/StandaloneInputReference/qsharp/qir-standalone-input-reference.qs
cesarzc: There's nothing to lose, I was originally using another approach and did not remove the `$` symbol. Removed it in the latest version.

PR #581, opened 2021-03-23
Remove BitStates.hpp

### LINKED ISSUE #577: Remove BitStates.hpp in favor of vector<bool>
[BitStates.hpp](https://github.com/microsoft/qsharp-runtime/blob/main/src/QirRuntime/public/BitStates.hpp) is an odd element of the public headers for QIR, in that it is not needed for interacting with the runtime nor for building on top of it, and is instead a library used by our [ToffoliSimulator](https://github.com/microsoft/qsharp-runtime/blob/main/src/QirRuntime/lib/Simulators/ToffoliSimulator.cpp#L11) and the [QIR Runtime Unit Tests](https://github.com/microsoft/qsharp-runtime/blob/main/src/QirRuntime/test/unittests/QirRuntimeTests.cpp#L17). This seems like an internal implementation detail that shouldn't be part of the public API. Moreover, it seems the original intent was to provide an efficient, tight packed vector of Boolean values where each Boolean only used one bit. It turns out that `std::vector<bool>` is already as specialization of vector for exactly this purpose, and is intentionally optimized for space in much the same way the BitStates helper class is (see the [documentation for `vector<bool>`](https://www.cplusplus.com/reference/vector/vector-bool/) for more details).

We can remove this custom class and replace it's usage with `vector<bool>` and still meet the goal of space optimized bit tracking.

swernli:
BitStates.hpp defined a custom implementation for tracking vectors of bits. However, `vector<bool>` from the STL already is a specialization of `vector` designed to save space and use a single bit for each index. Since we don't want to support the custom type in our API nor do we really need one internally, it was removed and all instances replaced with `vector<bool>`.

Fixes #577


src/QirRuntime/lib/Simulators/ToffoliSimulator.cpp
```diff
@@ -67,21 +65,23 @@ namespace Quantum
...
         }
 
         void ReleaseQubit(Qubit qubit) override
         {
             const long id = GetQubitId(qubit);
             assert(id <= this->lastUsedId);
-            assert(!this->states.IsBitSetAt(id));
+            assert(!this->states.at(id));
+            this->lastUsedId--;
+            this->states.pop_back();
```
swernli, line 78: Note that this does have a change in behavior here, where previously an allocated qubit that was released would never be used again, here we allow for qubit "reuse" by decrementing the `lastUsedId` member such that future allocations can use that id again. This will prevent the bit vector from always growing, such that Q# like this:
```qsharp
for i in 1..10 {
    use q = Qubit();
    X(q);
    Reset(q);
}
```
Will now have a bit vector of at most length 1 instead of length 10.

src/QirRuntime/lib/Simulators/ToffoliSimulator.cpp
```diff
@@ -67,21 +65,23 @@ namespace Quantum
         Qubit AllocateQubit() override
         {
             this->lastUsedId++;
-            this->states.ExtendToInclude(this->lastUsedId);
+            this->states.emplace_back();
```
kuzminrobin, line 68: I have an impression that `this->states.emplace_back()` adds an *un*initialized bool to the vector.
If I'm mistaken then what is the value added to the vector?

src/QirRuntime/lib/Simulators/ToffoliSimulator.cpp
```diff
@@ -67,21 +65,23 @@ namespace Quantum
...
 
         void ReleaseQubit(Qubit qubit) override
         {
             const long id = GetQubitId(qubit);
             assert(id <= this->lastUsedId);
-            assert(!this->states.IsBitSetAt(id));
+            assert(!this->states.at(id));
+            this->lastUsedId--;
+            this->states.pop_back();
         }
```
kuzminrobin, line 79: I have an impression that this function works correctly for the _last-most qubit id only_. But if we release a qubit "in the middle" then still the last-most qubit will be released. E.g. we call `AllocateQubit()` 3 times. It returns 0, 1, 2. Then we call `ReleaseQubit(0)`, the fragment `this->states.pop_back()` removes the qubit with id 2 (rather than with id 0).

src/QirRuntime/test/unittests/QirRuntimeTests.cpp
```diff
@@ -14,34 +14,33 @@
...
         return static_cast<int>(reinterpret_cast<int64_t>(r));
     }
 
     ResultsReferenceCountingTestQAPI(int maxResults)
-        : maxResults(maxResults + 2)
+        : maxResults(maxResults)
     {
-        allocated.ExtendToInclude(maxResults);
+        allocated = std::vector<bool>(maxResults);
     }
```
kuzminrobin, line 37: I expect this fragment to default-construct the values in the vector. 
The type `bool` is a built-in type. For the built-in types the default-construction is implemented by means of "doing nothing", i.e. leaving the items *un*initialized. 
I expect the values of the vector to stay uninitialized after this constructor. Consider changing to something like this
```c++
    ResultsReferenceCountingTestQAPI(int maxResults)
        : maxResults(maxResults)
        , allocated(maxResults, false)    // Init all elems to false.
    {
    }
```

kuzminrobin (commented):
I would like to know your opinion to my notes.

src/QirRuntime/lib/Simulators/ToffoliSimulator.cpp
swernli: Actually, `emplace_back` will run the default constructor for the underlying type, which for `bool` means that the value will be initialized to false. I could make it explicit by instead using `emplace_back(false)` and that should avoid confusion for anyone working on this in the future.

src/QirRuntime/lib/Simulators/ToffoliSimulator.cpp
swernli: Yes, that's correct, but in this case I'm relying on Q#'s strict qubit scoping for a shortcut. In Q#, every qubit has a scope, either implied or explicit, such that any qubit allocations are always released in the reverse order they were allocated. This does mean that the ToffoliSimulator could exhibit the bug you describe if it is used for QIR that was generated from something other than Q# (or written manually). However, because the simulators are packages as part of `Microsoft.Quantum.Qir.QSharp.Core.dll` it seems reasonable to restrict their behavior to Q# generated QIR. If someone wanted to use these simulators for another language, they'd have to conform to Q# gateset AND Q#'s lifetime semantics, which seems reasonable to me.

That said, this is a very subtle requirement that might not be obvious, and I'm open to further discussion. @bettinaheim, any concerns or suggestions about how to handle this?

src/QirRuntime/lib/Simulators/ToffoliSimulator.cpp
swernli: Actually, given your other comment and double checking the documentation, I'm now less sure that the default constructor for `bool` initializes to `false`. At the very least, it seems like it could vary by platform. I'll update this logic as described (use `emplace_back(false)`) and then review other places to ensure that I'm not assuming default value of false.

src/QirRuntime/test/unittests/QirRuntimeTests.cpp
swernli: I could've sworn that I saw somewhere that the default constructor for `bool` always produced a value of `false` but now I can't find it, and the few explanations I can find say that the details are platform/implementation dependent. You are right, it would be much better to be explicit and use the syntax you suggest than leave it up to the individual behavior of the underlying language implementation, especially if that can vary.

swernli:
Great feedback, @kuzminrobin ! I've responded to the individual comments.

src/QirRuntime/lib/Simulators/ToffoliSimulator.cpp
kuzminrobin: If the qubits are always deallocated in the reverse order of allocation then we are fine here. Can mark this conversation as resolved.

kuzminrobin: In that case the parameter (`Qubit qubit`) becomes redundant and misleads:
```c++
void ReleaseQubit(Qubit qubit) override
```

kuzminrobin (approved):

PR #584, opened 2021-03-24
Fix broken documentation links

anjbur:


src/Simulation/Native/CMakeLists.txt
```diff
@@ -44,6 +44,9 @@ else (MSVC)
   set(CMAKE_CXX_FLAGS "-static-libgcc")
 endif (MSVC)
 
+set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} /Qspectre")
```
tcNickolas, line 47: This seems unrelated to the links modification? If it's a necessary change, I'd recommend doing it separately, figuring out why the build broke after only fixing some links would be a pain :-)

src/Simulation/Simulators/QuantumSimulator/Qubit.cs
```diff
@@ -20,10 +20,16 @@ public QSimQubit(int id, uint simulatorId) : base(id)
                 this.SimulatorId = simulatorId;
             }
 
+            public QSimQubit(int id, uint simulatorId, string label) : this(id, simulatorId) {
```
tcNickolas, line 23: This one also looks like it should be a separate change

tcNickolas (changes requested):
Two files have changes unrelated to fixing links - if they're needed, I recommend excluding them from this PR and doing them separately.

src/Simulation/Native/CMakeLists.txt
anjbur: Reverted, that was my mistake. Picked up some changes from other work that shouldn't have been included here.

src/Simulation/Simulators/QuantumSimulator/Qubit.cs
anjbur: Reverted as well, same mistake

tcNickolas (approved):

PR #586, opened 2021-03-25
Updating test to match the latest compiler version

bettinaheim:
This is the necessary test adjustment to match https://github.com/microsoft/qsharp-compiler/pull/914.


swernli (approved):
LGTM

PR #588, opened 2021-03-25
Enumerated File Names for Generated C#

ScottCarda-MS:
Changes the output for the C# Generation rewrite step to always put all of the generated files in the given output directory directly, not in subfolders and not respecting relative locations of their corresponding Q# source files. The naming convention for generated C# files is being updated to include a number when their is a name collision. For example, if `Main.g.cs` has already been written during a compilation and another `Main.g.cs` file is going to be written, it will instead write `Main.g1.cs`, and if that file also already exists, `Main.g2.cs`, and so on. There is a hardcoded limit at 100 enumerations of the same file name, so `Main.g99.cs` would be the last possible file name. The enumeration always precedes the final `.cs`.


ScottCarda-MS:
I will need to find a way to add tests for this. @bettinaheim if you have any suggestions or similar existing tests, that would be helpful to look at.

kuzminrobin (approved):
I don't understand the code for now.
To not delay the development I'm approving, but I recommend to get at least one more approval.

ScottCarda-MS:
> I will need to find a way to add tests for this. @bettinaheim if you have any suggestions or similar existing tests, that would be helpful to look at.

Tests have been added now.

src/Simulation/CSharpGeneration.Tests/SimulationCodeTests.fs
```diff
@@ -248,6 +248,53 @@ namespace N1
...
+
+        let mutable allFilesFound = true
+        
+        let deleteFile filePath =
+            if File.Exists filePath then
+                File.Delete filePath
+            else
+                allFilesFound <- false
+
+        intrinsic |> deleteFile
```
bettinaheim, line 282: In general I'd recommend to delete files from disk at the start of the test rather than at the end for tests that check for files on disk. Not deleting it at the beginning relies somewhat on the tests being run on a clean machine; if these files for some reason already exists on disk then the first time the test is run it will succeed.

bettinaheim (approved):
lgtm

src/Simulation/CSharpGeneration.Tests/SimulationCodeTests.fs
ScottCarda-MS: I've updated the test to do both the check/delete before the transformation, and the check/delete after. You are right that the test should delete the files if they exist beforehand, otherwise it will compromise the test, but I also don't want to leave these files on the system after the test either, so I updated the test to do both.

PR #591, opened 2021-03-26
Updated QirRuntime/README.md, build/test instructions

kuzminrobin:


src/QirRuntime/README.md
```diff
@@ -1,6 +1,7 @@
 # The Native QIR Runtime
 
-This folder contains QIR runtime project, which includes implementation of the
+This folder contains the Quantum Intermediate Representation (QIR) Runtime project. QIR is a subset of the [LLVM](https://llvm.org/) Intermediate Representation.
+QIR Runtime includes implementation of the
  [QIR specification](https://github.com/microsoft/qsharp-language/tree/main/Specifications/QIR) and the bridge to
  compile QIR to be run against the native full state simulator.
```
bettinaheim, line 6:
```suggestion
 run QIR against the native full state simulator.
```

src/QirRuntime/README.md
```diff
@@ -11,24 +12,17 @@ This folder contains QIR runtime project, which includes implementation of the
 
 ## Build
 
+### Prerequisites
 The QirRuntime project is using CMake (3.17) + Ninja(1.10.0) + Clang++(11.0.0). Other versions of the tools might work
- but haven't been tested. Only x64 architecture is supported.
+ but haven't been tested. Only x64 architecture is supported.  
+For runing the PowerShell scripts below use the powershell core (`pwsh`),
```
bettinaheim, line 18: Is this necessary? I believe with a minor tweak to the condition in prerequisites.ps1 the scripts run just fine on the out of the box installation (they did about three weeks ago in any case).
```suggestion
For running the PowerShell scripts below use the powershell core (`pwsh`),
```

src/QirRuntime/build.py
```diff
@@ -1,6 +1,8 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 
+# DEPRECATED !!!
```
bettinaheim, line 4: Is this script no longer meant to be used? If so, what is the reason to not just remove it?

src/QirRuntime/generateqir.py
```diff
@@ -1,6 +1,8 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 
+# DEPRECATED !!!
```
bettinaheim, line 4: Same question here.

src/QirRuntime/test.py
```diff
@@ -1,6 +1,8 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 
+# DEPRECATED !!!
```
bettinaheim, line 4: ... and here. :)

bettinaheim (commented):
I haven't tried to follow the instructions just yet, but will do so a bit later. For now just a few comments reading through it.

src/QirRuntime/README.md
```diff
@@ -1,6 +1,7 @@
 # The Native QIR Runtime
 
-This folder contains QIR runtime project, which includes implementation of the
+This folder contains the Quantum Intermediate Representation (QIR) Runtime project. QIR is a subset of the [LLVM](https://llvm.org/) Intermediate Representation.
```
cgranade, line 3: You've used the definite article "the" to refer to LLVM IR, but not to QIR; I'd suggest either adding a "the" before "QIR" or dropping the "the" before "LLVM."

src/QirRuntime/README.md
```diff
@@ -1,6 +1,7 @@
...
+QIR Runtime includes implementation of the
```
cgranade, line 4:
```suggestion
The QIR runtime includes an implementation of the
```

src/QirRuntime/README.md
```diff
@@ -11,24 +12,17 @@ This folder contains QIR runtime project, which includes implementation of the
 
 ## Build
 
+### Prerequisites
```
cgranade, line 15: Blank lines are recommended after Markdown headers.
```suggestion
### Prerequisites

```

src/QirRuntime/README.md
cgranade, line 18: Note that from version 7.0, "PowerShell Core" is just "PowerShell." (Similar to that .NET 5 is just ".NET" even though it's the next version of .NET Core.)

```suggestion
For runing the PowerShell scripts below use PowerShell Core or PowerShell (`pwsh`),
```

src/QirRuntime/README.md
```diff
@@ -11,24 +12,17 @@ This folder contains QIR runtime project, which includes implementation of the
...
 The QirRuntime project is using CMake (3.17) + Ninja(1.10.0) + Clang++(11.0.0). Other versions of the tools might work
- but haven't been tested. Only x64 architecture is supported.
+ but haven't been tested. Only x64 architecture is supported.  
+For runing the PowerShell scripts below use the powershell core (`pwsh`),
+e.g. [this one](https://github.com/PowerShell/PowerShell/releases/download/v7.1.2/PowerShell-7.1.2-win-x64.msi), not the inbox PowerShell.
```
cgranade, line 19: This links directly to a specific version of the Windows 64-bit installer; I'd suggest linking to the PowerShell installation page, both so that we don't need to update the link when new versions are released, and so that users on other operating systems aren't directed to Windows-only links. E.g.: https://docs.microsoft.com/powershell/scripting/install/installing-powershell.

src/QirRuntime/README.md
```diff
@@ -11,24 +12,17 @@ This folder contains QIR runtime project, which includes implementation of the
...
 
-You can use CMake directly. For example, to produce a release build:
+To install prerequisite tools for building the QIR runtime, you can set the `ENABLE_QIRRUNTIME` environment variable to the string `"true"` 
+and run [`prerequisites.ps1`](prerequisites.ps1), or manually install pre-reqs with the steps listed below
+(Windows steps rely on [Chocolatey](https://chocolatey.org/) for installation).
```
cgranade, line 23:
```suggestion
and run [`prerequisites.ps1`](prerequisites.ps1), or manually install pre-reqs with the steps listed below.
Note that on Windows, this script relies on the [Chocolatey package manager](https://chocolatey.org/), while on macOS, `prerequisites.ps1` relies on the `brew` package manager.
```

A link to homebrew may also be helpful here.

src/QirRuntime/README.md
```diff
@@ -55,35 +49,53 @@ Running cmake from the editors will likely default to MSVC or clang-cl and fail.
 
 See [https://code.visualstudio.com/docs/remote/wsl] on how to use VS Code with WSL.
 
-## Test
+#### Other Prerequisites
```
cgranade, line 52:
```suggestion
#### Other Prerequisites

```

src/QirRuntime/README.md
```diff
@@ -55,35 +49,53 @@ Running cmake from the editors will likely default to MSVC or clang-cl and fail.
...
 
-Some of the tests depend on Microsoft.Quantum.Simulator.Runtime library. To run them make sure to build Native simulator
- from this repository or provide your own version of the library in a folder the OS would search during dynamic library
- lookup.
 
-Some of the tests use generated QIR (*.ll) files as build input. Currently the files are checked-in as part of the project
- but in the future they will be replaced by automatic generation during build. To regenerate the files, run generateqir.py
- or build/test scripts without specifying `noqirgen`. To use the checked-in files without regenerating them, run build/test
- scripts with `noqirgen` argument.
+### Build Commands
```
cgranade, line 58:
```suggestion
### Build Commands

```

src/QirRuntime/README.md
```diff
@@ -55,35 +49,53 @@ Running cmake from the editors will likely default to MSVC or clang-cl and fail.
...
+```
+  
+The script will create the `build/{Debug|Release}` folder and place the build artifacts in it. The configuration `Debug|Release`
+is specified with the `BUILD_CONFIGURATION` environment variable.
+If the variable is not set then the default is specified in [`set-env.ps1`](../../build/set-env.ps1).  
+We strongly recommend doing local builds using this [`build-qir-runtime.ps1`](build-qir-runtime.ps1) build script 
+because it is used by the continuous integration (CI) infrastructure, and this script also runs clang-tidy if it is installed.
 
-### Running tests with test.py
+Or you can use CMake directly (not recommended). For example in Windows you can enter the following commands in the command line:
```
cgranade, line 70:
```suggestion
Though it is not recommended, you can also use CMake directly. For example, in Windows you can enter the following commands in the command line:
```

src/QirRuntime/README.md
```diff
@@ -55,35 +49,53 @@ Running cmake from the editors will likely default to MSVC or clang-cl and fail.
...
+```batch
+@rem Navigate to QirRuntime folder.
```
cgranade, line 72: This looks like it uses cmd notation, but cmd is no longer the default shell on Windows (PowerShell was made the default several years ago).

src/QirRuntime/README.md
```diff
@@ -55,35 +49,53 @@ Running cmake from the editors will likely default to MSVC or clang-cl and fail.
...
-All native tests, including QIR, use catch2 and are fully integrated with CTest. Navigate into
- `build/[Windows|Linux]/[Debug|Release]` folder and run `ctest`. No configuration options required. The results will be
- logged into the corresponding `build/[Windows|Linux]/[Debug|Release]/<target_path>/<test_binary_name>_results.xml` file.
- Tests from the "[skip]" category won't be run.
+@rem Return from the `build` directory:
+popd
+```
+## Tests
+
+### Running All Tests
```
cgranade, line 91:
```suggestion
### Running All Tests

```

src/QirRuntime/README.md
```diff
@@ -55,35 +49,53 @@ Running cmake from the editors will likely default to MSVC or clang-cl and fail.
...
 
-### Running test binaries individually
+```batch
```
cgranade, line 93: Similarly here, the `@rem` notation is specific to cmd.exe.

src/QirRuntime/build.py
```diff
@@ -1,6 +1,8 @@
...
+
```
cgranade, line 5: I'd suggest replacing this with a Python-style docstring:

```suggestion
"""
WARNING: Note that this script has been deprecated.
          Please use the `build-qir-runtime.ps1` script instead.
          See README.md for more details.
"""
```

src/QirRuntime/generateqir.py
cgranade, line 4: Similarly here.

src/QirRuntime/test.py
cgranade, line 4: Similarly here.

src/QirRuntime/README.md
kuzminrobin: After Stefan significantly refactored our build scripts 1 - 2 weeks ago, for me the .ps1 were failing without `pwsh`.

src/QirRuntime/README.md
kuzminrobin: Yes... I'm out-of-date ;-)
Don't you have on top of your head, what I should use instead of  ```batch
for PowerShell syntax highlighting? Locally I try `ps`, `ps1`, `PowerShell` but my MarkDown plugin for Chrome does not highlight the syntax if I use any of those.

kuzminrobin: Oh. looks like it is shown [here](https://github.com/MicrosoftDocs/PowerShell-Docs/issues/1511).

src/QirRuntime/README.md
kuzminrobin: > A link to homebrew may also be helpful here.

I assume it is https://brew.sh

src/QirRuntime/README.md
```diff
@@ -55,35 +49,53 @@ Running cmake from the editors will likely default to MSVC or clang-cl and fail.
...
+pushd build
 
-The script will trigger an incremental build unless `nobuild` options is specified. Tests from the "[skip]" category
- won't be run.
+@rem Prepare for the build:
+cmake -G Ninja -DCMAKE_BUILD_TYPE=Debug ..
 
-### Running tests with CTest
+@rem Build (you will likely need to issue this command several times, see issue #589):
+cmake --build .
```
swernli, line 89: I still don't understand this. Why does this need to be run multiple times? I haven't observed that behavior, though I'm trying now on a fresh repo to see if there is something I'm missing.

README.md
```diff
@@ -57,6 +57,37 @@ To build on other platforms:
 1. Install the pre-reqs:
     * Install [CMake](https://cmake.org/install/)
     * Install [.NET Core 3 SDK](https://dotnet.microsoft.com/download)
+    * On [WSL](https://docs.microsoft.com/en-us/windows/wsl/)/Linux:
+      * Install `g++` (e.g. in Ubuntu 20.04 `sudo apt-get install g++`).
+      * The build does not accept `dotnet-*-5.0`/`aspnetcore-*-5.0` packages, install `dotnet-*-3.1`/`aspnetcore-*-3.1`
+        (`sudo apt-get install aspnetcore-runtime-3.1 dotnet-sdk-3.1`). The possible result can be:
```
swernli, line 63: is the `aspnetcore-runtime-3.1` really necessary?

kuzminrobin: You're right. 
> sudo apt-get install **dotnet-sdk-3.1**

is enough.

src/QirRuntime/README.md
kuzminrobin: I saw similar behavior during multithreaded builds ( `make -j ..` ). Different targets are built by different threads. And sometimes the later targets (that depend on the results of the earlier target builds) start earlier than the earlier target build  fully completes. It felt like the earlier target step was completed (from the point of view of the build utility), but the file output of the earlier stage was cached for a few seconds (e.g. the object file was not written to the disk completely), but the linking stage was already kicked off, and the linker was failing to open the object file, and the linking was failing.
At that moment the cached file was flushed onto the disk and the repeated build attempt was succeeding.

swernli: I haven't observed this behavior with CMake, which does not perform parallel builds by default. In fact, CMake will build in dependency order based on detected dependencies across different targets, with the order of the folders in the top level CMakeLists.txt being the fallback. This ensures that CMake will only need to be run once. Have you gotten this error in your local builds?

I did just try it again, and after running bootstrap.ps1 and copying the relevant .ll files to the correct locations, I was able to manually create `src\QirRuntime\build\Debug` and then run these two commands to successfully build the whole QIR runtime and tests:
```
cmake -G Ninja ../..
cmake --build . --target install
```

src/QirRuntime/README.md
```diff
@@ -55,35 +51,57 @@ Running cmake from the editors will likely default to MSVC or clang-cl and fail.
...
-To execute all tests locally run `test.py` from the project's root folder:
+Though it is not recommended, you can also use CMake directly.
+For example, in Windows you can enter the following commands in the command line:
+```powershell
+# Navigate to QirRuntime folder.
 
-- (Windows) `python test.py [nobuild] [debug/release]`
-- (Linux) `python3 test.py [nobuild] [debug/release]`
+# Creatge the `build` directory:
+mkdir build
```
swernli, line 80: to match what the script does, the directory should be `build/Debug`, then navigate into that folder.

src/QirRuntime/README.md
```diff
@@ -55,35 +51,57 @@ Running cmake from the editors will likely default to MSVC or clang-cl and fail.
...
 
-The script will trigger an incremental build unless `nobuild` options is specified. Tests from the "[skip]" category
- won't be run.
+# Enter the `build` directory:
+pushd build
 
-### Running tests with CTest
+# Prepare for the build:
+cmake -G Ninja -DCMAKE_BUILD_TYPE=Debug ..
```
swernli, line 86: This needs to be `cmake -G Ninja -DCMAKE_BUILD_TYPE=Debug ../..` since the pwd should be `src\QirRuntime\build\Debug`

src/QirRuntime/README.md
swernli: Also worth noting, the `--target install` ensures that all the binary output is collected in the `bin` folder, which the test script expects.

src/QirRuntime/README.md
kuzminrobin: I removed those steps, they have some more issues.

src/QirRuntime/README.md
kuzminrobin: I removed those steps, they have some more issues.

src/QirRuntime/README.md
kuzminrobin: I just removed those steps.

kuzminrobin: I double checked out of curiosity. Looks like `--target install` was the missing part.

kuzminrobin:
All the feedback has been addressed. Please review.

README.md
```diff
@@ -57,6 +57,37 @@ To build on other platforms:
...
+ii  dotnet-apphost-pack-3.1   3.1.13-1     amd64        Microsoft.NETCore.App.Host 3.1.13
+ii  dotnet-host               5.0.4-1      amd64        Microsoft .NET Host - 5.0.4
+ii  dotnet-hostfxr-3.1        3.1.13-1     amd64        Microsoft .NET Core Host FX Resolver - 3.1.13 3.1.13
+un  dotnet-nightly            <none>       <none>       (no description available)
+ii  dotnet-runtime-3.1        3.1.13-1     amd64        Microsoft .NET Core Runtime - 3.1.13 Microsoft.NETCore.App 3.1.13
+ii  dotnet-runtime-deps-3.1   3.1.13-1     amd64        dotnet-runtime-deps-3.1 3.1.13
+ii  dotnet-sdk-3.1            3.1.407-1    amd64        Microsoft .NET Core SDK 3.1.407
+ii  dotnet-targeting-pack-3.1 3.1.0-1      amd64        Microsoft.NETCore.App.Ref 3.1.0
+
+qsharp-runtime$ dpkg -l *aspnet*
```
swernli, line 82: Now that you confirmed the aspnetcore package isn't required, this part can be removed, along with the mention of aspnetcore-*-5.0 above.

swernli (approved):
One last removal, and I think these updates are good to go. Thanks for taking the time to do this!

README.md
kuzminrobin: Done. Thanks.

src/QirRuntime/build.py
```diff
@@ -1,102 +0,0 @@
-# Copyright (c) Microsoft Corporation. All rights reserved.
```
DmitryVasilevsky, line 1: Why are these files removed?

kuzminrobin: The .py files are not applicable any more.
The .ps1 files are used for building and testing.

PR #594, opened 2021-03-31
Change to QIR folder in preparation for QIR controller project

cesarzc:
This change moves the contents of **_src/QirRuntime_** to **_src/Qir/Runtime_** in preparation to add the QIR controller project.


kuzminrobin:
LGTM! Approving.

kuzminrobin (approved):

swernli:
As I was working on top of these changes, I noticed a path that was off by a `../` and it just so happens to be related to the same failure shown in the logs here. So I took the liberty of trying the fix here to see if it resolves the e2e build.

cesarzc:
> 
> 
> As I was working on top of these changes, I noticed a path that was off by a `../` and it just so happens to be related to the same failure shown in the logs here. So I took the liberty of trying the fix here to see if it resolves the e2e build.

Thank you very much @swernli! Your fix worked :)

PR #596, opened 2021-03-31
Minor build script clean-up

swernli:
Cleans up a few minor bugs in the build scripts, adds the default docs output folder for Q# projects to the gitignore.


src/QirRuntime/build-qir-runtime.ps1
```diff
@@ -71,7 +71,7 @@ elseif (($IsWindows) -or ((Test-Path Env:AGENT_OS) -and ($Env:AGENT_OS.StartsWit
     Write-Host "##vso[task.logissue type=error;]Failed to identify the OS. Will use default CXX compiler"
 }
 
-$qirRuntimeBuildFolder = (Join-Path $PSScriptRoot "build\$Env:BUILD_CONFIGURATION")
+$qirRuntimeBuildFolder = (Join-Path $PSScriptRoot build $Env:BUILD_CONFIGURATION)
```
cgranade, line 74: I think this makes sense, but as a portability note, this is incompatible with PowerShell 5.1 (deprecated, but still the default on fresh Windows 10 installs). It's less elegant, but for compatibility with 5.1 as to @bettinaheim's recent point, can we modify this as follows?

```suggestion
$qirRuntimeBuildFolder = (Join-Path $PSScriptRoot (Join-Path build $Env:BUILD_CONFIGURATION))
```

swernli: (see for example the [prerequisites section](https://github.com/microsoft/qsharp-runtime/blob/main/src/QirRuntime/README.md#prerequisites) of the readme in this folder with @kuzminrobin recently updated to reflect the PowerShell 7 requirement)

cgranade: Ah, please disregard, then; I had gotten my wires crossed on 5.1 vs 7.0 requirements, then!

.gitignore
```diff
@@ -339,3 +339,6 @@ ASALocalRun/
 /src/Simulation/Simulators.Tests/TestProjects/QSharpExe/built
 /src/Simulation/Simulators.Tests/TestProjects/TargetedExe/built
 dbw_test
+
+# Generated Q# docs
+**/docs
```
samarsha, line 344: Is there a difference between `**/docs` and just `docs`?

src/Qir/Runtime/build-qir-runtime.ps1
```diff
@@ -105,7 +105,7 @@ $osQirDropFolder = Join-Path $Env:DROPS_DIR QIR $os
 if (!(Test-Path $osQirDropFolder)) {
     New-Item -Path $osQirDropFolder -ItemType "directory"
 }
-$pattern | Foreach-Object { Copy-Item (Join-Path . bin $_) $osQirDropFolder }
+$pattern | Foreach-Object { Copy-Item (Join-Path . bin $_) $osQirDropFolder -ErrorAction SilentlyContinue}
```
samarsha, line 108: Why do we want to silently continue on errors?

.gitignore
swernli: Now that you mention it, no there really isn't. I could change this to `[Dd]ocs` and but it up next to the `[Bb]in` and `[Oo]bj` exclusions for clarity.

src/Qir/Runtime/build-qir-runtime.ps1
swernli: It's a subtle but important reason. If the above generate step fails, then the source `bin` folder might not exist, in which case this line will throw. If it throws, then we won't execute the `Pop-Location` below, and the script execution will leave the present working directory as the folder in the subtree. This doesn't matter for the pipeline, as a failure will kill the powershell session and propagate up to a failure of the whole pipeline, but for local development it's extremely annoying to have a script fail and leave you in a random directory. There are two ways to handle this: wrap the copy in a try-catch or suppress the throw. Given that the copy already doesn't look for specific files and the failure only happens when the `bin` folder is missing completely, we know a line higher up in the script has already failed with the more relevant failure, so no need to distract with another throw here. Another alternative could be to put the whole script execution in a try-catch and have every failure throw, ensuring that the `Pop-Location` is in the finally block, but I erred on the side of keeping the overall structure of the script the same.

This goes along with my effort to make sure the build scripts work for local dev as well as they do for the pipeline, so I wanted to make sure to fix this to smooth out the experience.

samarsha, line 108:
```suggestion
$pattern | Foreach-Object { Copy-Item (Join-Path . bin $_) $osQirDropFolder -ErrorAction SilentlyContinue }
```

samarsha (approved):

cesarzc (approved):

PR #598, opened 2021-04-02
Separate QIR Runtime from Tests and Samples

swernli:
This change separates the QIR Runtime (production code) from the tests and samples. Since the Runtime has no Q# dependencies, we can now pull it into the native build step of the e2e pipeline to better produce cross-platform binaries from the builds.


src/Qir/.clang-format
```diff
@@ -1,37 +1,37 @@
...
+BreakConstructorInitializers: BeforeComma
+
+# misc
+Cpp11BracedListStyle: true
+FixNamespaceComments: true
+IncludeBlocks: Preserve
+SpaceBeforeInheritanceColon : true
+SpaceBeforeParens: ControlStatements
+DerivePointerAlignment: false
+PointerAlignment: Left
```
kuzminrobin, line 37: (Up to you)
[FYI A](https://teams.microsoft.com/l/message/19:bafaeb809d7342d68f4ad99cdc1675af@thread.skype/1617134257928?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=e6e22ac8-e11f-44d0-b599-46e9bea42c53&parentMessageId=1617058745524&teamName=Quantum%20Systems&channelName=QIR&createdTime=1617134257928)
[FYI B](https://teams.microsoft.com/l/message/19:bafaeb809d7342d68f4ad99cdc1675af@thread.skype/1617135209545?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=e6e22ac8-e11f-44d0-b599-46e9bea42c53&parentMessageId=1617058745524&teamName=Quantum%20Systems&channelName=QIR&createdTime=1617135209545)

src/Qir/Common/Include/qsharp__foundation_internal.hpp
```diff
@@ -6,18 +6,20 @@
...
 
-        void RandomizeSeed(bool randomize);
-        int64_t GetLastGeneratedRandomI64();
-        double GetLastGeneratedRandomDouble();
+        QIR_SHARED_API void RandomizeSeed(bool randomize);
+        QIR_SHARED_API int64_t GetLastGeneratedRandomI64();
+        QIR_SHARED_API double GetLastGeneratedRandomDouble();
     } // namespace Internal
 } // namespace Qis
 } // namespace Quantum
```
kuzminrobin, line 25: (Up to you)
FYI: Originally the `Qiantum::Qis::Internal`s APIs were supposed to be used by the tests only (and not exposed externally to the users).
With recent mass refactoring PRs I'm not sure if that is possible to retain. ;-)

src/Qir/Common/cmake/qir_cmake_include.cmake
```diff
@@ -65,4 +65,46 @@ macro(target_source_from_qir_obj target_name source_file)
...
+# compiling from IR
+macro(target_source_from_qir target_name source_file)
+    set(CLANG_ARGS "-c")
+    if (CMAKE_BUILD_TYPE STREQUAL "Debug")
+        set(CLANG_ARGS
+        "${CLANG_ARGS}"
+        "-O0"
+        "-D_DEBUG"
+        )
+    endif()
```
kuzminrobin, line 80: (Minor)
Among the debugging-related macros consider using the `NDEBUG` only. As I know, this is the only macro mentioned in the standard of C and C++. The other ones, like `_DEBUG`, `DEBUG`, are non-standard.

src/Qir/Common/cmake/unit_test_include.cmake
```diff
@@ -15,9 +13,10 @@ macro(add_unit_test target)
...
 
     set(TEST_DEPS2 "${CMAKE_BINARY_DIR}/bin")
+    set(TEST_DEPS3 "${PROJECT_SOURCE_DIR}/../Runtime/build/${CMAKE_BUILD_TYPE}/bin")
     set_property(TEST ${target} PROPERTY ENVIRONMENT
-        "LD_LIBRARY_PATH=${TEST_DEPS1}:${TEST_DEPS2}:${LD_LIBRARY_PATH}"
-        "PATH=${TEST_DEPS1}\;${TEST_DEPS2}\;${PATH}"
-        "DYLD_LIBRARY_PATH=${TEST_DEPS1}:${TEST_DEPS2}:${DYLD_LIBRARY_PATH}"
+        "LD_LIBRARY_PATH=${TEST_DEPS1}:${TEST_DEPS2}:${TEST_DEPS3}:${LD_LIBRARY_PATH}"
+        "PATH=${TEST_DEPS1}\;${TEST_DEPS2}\;${TEST_DEPS3}\;${PATH}"
+        "DYLD_LIBRARY_PATH=${TEST_DEPS1}:${TEST_DEPS2}:${TEST_DEPS3}:${DYLD_LIBRARY_PATH}"
```
kuzminrobin, line 20: (Probably something to figure out later (in a separate issue maybe?), it was existing before this change)
I'm rather surprised to see in a single block the env var `LD_LIBRARY_PATH` (typical for UNIX/Linux), and the `;` as a separator in the env var `PATH` (typical for Win but I have never seen `;` as a separator in UNIX/Linux, `:` only)

src/Qir/Samples/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+
+set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -DDEBUG")
+
+if(WIN32)
+    if(CMAKE_BUILD_TYPE STREQUAL "Debug")
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D_DLL -D_MT -Xclang --dependent-lib=msvcrtd")
+    else()
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D_DLL -D_MT -Xclang --dependent-lib=msvcrt")
+    endif()
+endif()
```
kuzminrobin, line 20: Minor. Consider
```suggestion
if(WIN32)
    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D_DLL -D_MT -Xclang --dependent-lib=msvcrt")
    if(CMAKE_BUILD_TYPE STREQUAL "Debug")
        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}d")
    endif()
endif()
```

src/Qir/Samples/build-qir-samples.ps1
```diff
@@ -0,0 +1,103 @@
...
+
+    if (Get-Command clang-tidy -ErrorAction SilentlyContinue) {
+        # Only run clang-tidy if it's installed. This is because the package used by chocolatey on
+        # the build pipeline doesn't include clang-tidy, so we allow skipping that there and let
+        # the Linux build catch tidy issues.
+        $clangTidy = "-DCMAKE_CXX_CLANG_TIDY=clang-tidy"
+    }
+} else {
+    Write-Host "##vso[task.logissue type=error;]Failed to identify the OS. Will use default CXX compiler"
+}
```
kuzminrobin, line 72: (Minor)
This looks more like a _warning_, rather than an _error_. But the fragment `type=error` makes me expect that it will be treated as an error.

src/Qir/Samples/test-qir-samples.ps1
```diff
@@ -0,0 +1,33 @@
...
+    Write-Host "##[info]Test QIR Samples ($Sample)"
+
+    Push-Location (Join-Path $PSScriptRoot build $Env:BUILD_CONFIGURATION $Sample)
+
+    ctest --verbose
+
+    if ($LastExitCode -ne 0) {
+        Write-Host "##vso[task.logissue type=error;]Failed to test QIR Samples ($Sample)"
+        $script:all_ok = $False
+    }
```
kuzminrobin, line 24: (Just curiosity)
I assume that `$script:all_ok` means the same as `$all_ok` (earlier and later in this script and in `build-qir-samples.ps1`)

src/Qir/Tests/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+
+set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -DDEBUG")
+
+if(WIN32)
+    if(CMAKE_BUILD_TYPE STREQUAL "Debug")
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D_DLL -D_MT -Xclang --dependent-lib=msvcrtd")
+    else()
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D_DLL -D_MT -Xclang --dependent-lib=msvcrt")
+    endif()
+endif()
```
kuzminrobin, line 20: (Minor)
I have an impression that this is done in more than one place.
Consider extracting the common part into a separate file invoked from other files.

src/Qir/Tests/build-qir-tests.ps1
```diff
@@ -0,0 +1,107 @@
...
+
+    if (Get-Command clang-tidy -ErrorAction SilentlyContinue) {
+        # Only run clang-tidy if it's installed. This is because the package used by chocolatey on
+        # the build pipeline doesn't include clang-tidy, so we allow skipping that there and let
+        # the Linux build catch tidy issues.
+        $clangTidy = "-DCMAKE_CXX_CLANG_TIDY=clang-tidy"
+    }
+} else {
+    Write-Host "##vso[task.logissue type=error;]Failed to identify the OS. Will use default CXX compiler"
+}
```
kuzminrobin, line 76: (Minor)
I see this also in more than one place. Consider extracting into a common file invoked from other files.

anpaz (approved):
lgtm. as we talked, we should add the qir dlls to the manifest so they get signed, but other than that i have no concerns.

build/build.ps1
```diff
@@ -16,6 +16,16 @@ if ($Env:ENABLE_NATIVE -ne "false") {
...
 
+if ($Env:ENABLE_QIRRUNTIME -ne "false") {
+    $qirRuntime = (Join-Path $PSScriptRoot "../src/Qir/Runtime")
+    & "$qirRuntime/build-qir-runtime.ps1"
+    if ($LastExitCode -ne 0) {
+        $script:all_ok = $False
+    }
+} else {
+    Write-Host "Skipping build of qir runtime because ENABLE_QIRRUNTIME variable is set to: $Env:ENABLE_QIRRUNTIME"
+}
```
kuzminrobin, line 27: (Minor)
I assume that if the env var `ENABLE_QIRRUNTIME`  
is not set
or is set to any value other than `false` (lowercase)
then the QIR Runtime will be built. Is that a correct assumption?

src/Qir/Samples/test-qir-samples.ps1
kuzminrobin: I believe I got the point of the `script:` fragment. It prevents creating another copy of `$all_ok` in a nested scope. If that's the case then please consider this as resolved.

kuzminrobin (approved):
Approving with suggestions.

src/Qir/Common/Include/qsharp__foundation_internal.hpp
swernli: Yeah, I couldn't have the test be separate from the runtime without making these `dllexport`. The good news is that by keeping the header out of the src/Qir/Runtime/public folder it's still not something we would expect any code external to our repo to utilize, but someone could copy the header and then access these functions if they were sufficiently motivated. I think we have two choices about how to proceed but I didn't want to wrap that into this change at the same time. Here are my ideas, and I think I'd defer to you on which would be preferable:

1.  Refactor the tests to be true unit tests that don't depend on any Q# compilation, such that they would only all into the random number generation code directly and either not use QIR or have simple hard-coded QIR wrappers. Then we could move it to the src/Qir/Runtime/unittests folder and put this header back into that part of the tree, removing the `dllexport` in the process.
2. Keep the tests where they are but refactor them such that they don't depend on any internal functionality. This would entail duplicating the error string into the test and then updating the test strategy to generate a spread of random numbers and confirm they are not the same, which would have a non-zero chance of failing if the two equal numbers are coincidentally generated. Given how much grief random tests have caused for us in the past, I think there is value in instead considering keeping `RandomizeSeed` as a part of the public API so that this and other tests can request a non-random seed and test with a well known sequence of results.
3. Some combination of both strategies, where the internal unit test that builds inside of the runtime tests the randomization and the test against the dll with Q#->QIR just verifies that the Q# callables work as expected without trying to validate the randomness of the results (ie: check that a number is returned but don't try to validate uniqueness). This one is the most work but gives us the most comprehensive test coverage.

I'll file an issue for this as pending clean-up work so it doesn't get lost.

src/Qir/Common/cmake/qir_cmake_include.cmake
swernli: Looking around at our codebase, it looks like we actually use `DEBUG` everywhere else (the rest of the QIR code use `set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -DDEBUG")`, see for example [here](https://github.com/microsoft/qsharp-runtime/blob/swernli/qir-build-native/src/Qir/Runtime/CMakeLists.txt#L12)). Given that, I'm going to update this location to use `DEBUG` instead of `_DEBUG` for consistency, and we can consider adopting to the alternate pattern later by updating all the locations together in a targeted PR.

swernli: I also realized looking back at this file that I forgot to remove the old QIR compilation macro which is no longer used!

src/Qir/Common/cmake/unit_test_include.cmake
swernli: yeah, this is a little odd but appears to just be a shortcut. It sets all three environment variables to avoid having an if-statement that sets only the variable for the current platform. We could clean this up later to avoid confusion, but it would require making the code here longer.

src/Qir/Samples/CMakeLists.txt
swernli: I think I'd prefer to keep these separate.

src/Qir/Samples/build-qir-samples.ps1
swernli: Good catch! This should be a warning, not an error.

src/Qir/Samples/test-qir-samples.ps1
swernli: PowerShell has some weird scoping rules that can cause problems if you aren't careful. Because this is in a function, using just `$all_ok` would create a local variable and assign the boolean value, not update the variable used elsewhere. The explicit `script` ensures PowerShell knows we intend to refer to the script-wide scope and will correctly update the variable. This is distinct from `global` which will set variables in the whole PowerShell session and could affect the execution of later scripts, and `env` which sets environment variables that will be inherited by launched processes. I've seen missing `script` lead to some very subtle bugs in the past!

See [about_Scoping](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_scopes?view=powershell-7.1) for more info.

src/Qir/Tests/CMakeLists.txt
swernli: That's true, and definitely clean-up we can consider for the future, especially as we grow the number of folders where we do native builds using cmake. I'll hold off on that for now though.

src/Qir/Tests/build-qir-tests.ps1
swernli: This one I agree should be done now, as we are likely to need to make changes to how cmake is invoked and don't want to have to do that in too many places. I'll work in this now.

src/Qir/Common/Include/SimulatorStub.hpp
```diff
@@ -1,139 +1,139 @@
...
-            throw std::logic_error("not_implemented");
-        }
-        Result UseOne() override
-        {
-            throw std::logic_error("not_implemented");
-        }
-    };
-
-} // namespace Quantum
+#pragma once
```
cesarzc, line 1: It seems like this file and a few others did not change the contents but have now different line endings. This might have happened because the file was moved "outside" of git and even though it was detected as a rename, line endings were not preserved.

Please consider using the `git mv` command which does preserve line endings.

build/build.ps1
swernli: Yes, that is correct. This an inversion of the previous behavior, where it only built if the variable was set to "true" and so would default to not building QIR Runtime. This was updated in #566 .

src/Qir/Samples/build-qir-samples.ps1
```diff
@@ -0,0 +1,103 @@
...
+
+$qirTestsBuildFolder = (Join-Path $PSScriptRoot build $Env:BUILD_CONFIGURATION)
```
cesarzc, line 74: NIT: Consider renaming `$qirTestsBuildFolder` to `$qirSamplesBuildFolder`.

cesarzc (approved):
Looks good! Approving with a couple of suggestions.

src/Qir/Samples/build-qir-samples.ps1
swernli: Per Robin's suggestion above, I'm factoring this out into common logic that will live in a util ps1, so this will just go away :)

src/Qir/Common/Include/SimulatorStub.hpp
swernli: I double checked, and unfortunately this is the right thing in this case. This repository was missing a `.gitattributes` file that would ensure that line endings get converted to Unix style when checked in to avoid this kind of issue. I added that file last month (see #545), but these files were created before then. This means that our repo has a mix of Unix line endings and Windows line endings in it, and someone cloning the repo on Linux would see every file with `CRLF` line endings as modified because git will convert them `LF`. Now that we have the gitattributes, future files will correctly be merged with Unix line endings but checked out locally with `CRLF` line endings if the local environment is Windows. Now that this is set, the first time we move a file git will see the line endings change, which is unfortunate but sets us up for better cross-platform development in the future.

See [Configuring Git to handle line endings](https://docs.github.com/en/github/getting-started-with-github/configuring-git-to-handle-line-endings) for more details on the guidance here. We should be making this change across all our repos to better enable cross-platform collaboration.

src/Qir/.clang-format
swernli: See my [comment below](https://github.com/microsoft/qsharp-runtime/pull/598/files#r606389734) to Cesar. For some of these files, the line ending update was intentional so the files are correctly set up to work well cross-platform.

src/Qir/Common/Include/qsharp__foundation_internal.hpp
swernli: Filed as #599

kuzminrobin: Some time later I will need to take a more detailed look at the new structure introduced by the earlier Cesar's PR and this PR, to get a broader picture.
For now I'm inclined to agree with you that those who will be using the random number generator, will also need to test their functionality, and they will also likely need a 100% deterministic pseudo-random number generators. So it makes sense to expose (`dllexport`) the discussed APIs.
Thank you for clarifying. Let's leave as is for now.

src/Qir/Common/cmake/qir_cmake_include.cmake
kuzminrobin: FYI: I have opened #600 (Consider using NDEBUG instead of DEBUG, _DEBUG) to not forget about this.

src/Qir/Common/cmake/unit_test_include.cmake
kuzminrobin: FYI: #601 Double-check that the PATH is set correctly

PR #605, opened 2021-04-06
Added more testcases for DrawRandomInt(), fixed documentation

### LINKED ISSUE #602: DrawRandomInt should work when min and max are equal
## Describe the bug
DrawRandomInt produces an ExecutionFailExecption when min and max parameters are equal.

## To Reproduce
Call Microsoft.Quantum.Random.DrawRandomInt with the same number for min and max. For example, `DrawRandomInt(0, 0)`.

## Expected behavior
Random integer functions typically allow a range of 1, for situations where the range may be variable. For example, I would expect to be able to select a random index of an array like this:

```
// assume "array" is of length > 0
let randomIndex = DrawRandomInt(0, Length(array) - 1);
```

## Actual behavior
The above code will fail if array is of length 1.

## System information
N/A

**Additional context**
To be clear, the current behavior is intended per the [documentation](https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.random.drawrandomint#remarks). Really the purpose of this issue is to ask if there is a reason why DrawRandomInt was implemented this way, and if the team would consider changing it.

Thanks :smile:

kuzminrobin:
Fixes #602

### DIFF
--- a/src/Simulation/QSharpFoundation/Random/Intrinsic.qs
+++ b/src/Simulation/QSharpFoundation/Random/Intrinsic.qs
@@ -17,7 +17,7 @@ namespace Microsoft.Quantum.Random {
     /// probability.
     ///
     /// # Remarks
-    /// Fails if `max < min`.
+    /// Fails if `max <= min`.
     ///
     /// # Example
     /// The following Q# snippet randomly rolls a six-sided die:
@@ -45,7 +45,7 @@ namespace Microsoft.Quantum.Random {
     /// uniform probability.
     ///
     /// # Remarks
-    /// Fails if `max < min`.
+    /// Fails if `max <= min`.
     ///
     /// # Example
     /// The following Q# snippet randomly draws an angle between $0$ and $2 \pi$:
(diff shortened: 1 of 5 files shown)


src/Simulation/QSharpFoundation/Random/Intrinsic.qs
```diff
@@ -17,7 +17,7 @@ namespace Microsoft.Quantum.Random {
     /// probability.
     ///
     /// # Remarks
-    /// Fails if `max <= min`.
+    /// Fails if `max < min`.
```
bettinaheim, line 20: We shouldn't edit this before the C# runtime support has been added for it.

kuzminrobin: Reverted this change.

ScottCarda-MS (approved):
Looks good to me. Best to double check with @cgranade that this is correct.

cgranade (approved):
Looks good to me, thanks for fixing this one!

PR #607, opened 2021-04-06
Minor script updates for QIR runtime

bettinaheim:
A couple of minor things I came across when I wanted to run a couple of things using the qir runtime.


swernli (changes requested):
It looks like this branch was missing the last refactoring to these folders and build scripts, which made major changes to the structure of these files and might have already addressed the script updates being made here (#598). I'd suggest rebasing the error reporting on main and reevaluating the script updates for the new folder structure.

src/Qir/Runtime/build-qir-runtime.ps1
```diff
@@ -106,6 +110,8 @@ if (!(Test-Path $osQirDropFolder)) {
     New-Item -Path $osQirDropFolder -ItemType "directory"
 }
 $pattern | Foreach-Object { Copy-Item (Join-Path . bin $_) $osQirDropFolder -ErrorAction SilentlyContinue }
+Copy-Item (Join-Path $env:NATIVE_SIMULATOR "Microsoft.Quantum.Simulator.Runtime.dll") $osQirDropFolder -ErrorAction SilentlyContinue
+$env:PATH += ";" + $osQirDropFolder
```
kuzminrobin, line 114: In Linux the `:` is used as the path separator. This line looks applicable for Win only.

PR #614, opened 2021-04-07
Remove duplication of Quantum SDK version number

### LINKED ISSUE #530: We need a better way to update the package versions across the repo
Several times when working on a change that needs to bring in new features from an updated QDK, the developer needs to update the package versions of the Microsoft.Quantum.Sdk (and any other explicitly referenced packages) across the whole repo. This can lead to missed update locations, mismatched packages, and subtle build failures. Rather than doing this by hand, we should consider either a helper script that updates the whole repo to a single provided QDK version or a single common location for defining the current QDK version to use across all projects.

For inspiration, we could draw from the script in the katas repo: https://github.com/microsoft/QuantumKatas/blob/main/scripts/Update-QDKVersion.ps1

samarsha:
Use global.json to specify the package version of Microsoft.Quantum.Sdk in one place.

The only other place this version number appears is in the dependency on Microsoft.Quantum.Compiler in the CSharpGeneration project. I'm not sure there is a good way to unify these two, but I suspect those package versions don't need to be in sync with each other in the same way that the SDK version numbers need to, so this could be good enough to close #530.


cgranade:
That's a really neat way to use global.json (and indeed, speaks well to the value of making Microsoft.Quantum.Sdk an SDK-style package)! I'd be interested in duplicating this on the libraries repo as well (samples less so, since users will copy and paste each sample such that it's good if samples work without needing repo-wide config files — as noted with @swernli there are exceptions to that, but those exceptions should be carefully considered).

cgranade (approved):
Approving as it looks good to me, but may want a second opinion from someone closer to runtime.

swernli (approved):

PR #619, opened 2021-04-08
Include QIR Runtime binaries and headers in drops

swernli:
This updates the build script to copy the QIR Runtime binaries and headers into the drops folder. This will ensure that during PR builds (non-e2e) the corresponding cross-platform binaries are still produced as build artifacts.

### DIFF
--- a/src/Qir/Runtime/build-qir-runtime.ps1
+++ b/src/Qir/Runtime/build-qir-runtime.ps1
@@ -6,3 +6,18 @@
 if (-not (Build-CMakeProject $PSScriptRoot "QIR Runtime")) {
     throw "At least one project failed to compile. Check the logs."
 }
+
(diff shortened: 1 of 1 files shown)


swernli:
The artifacts are showing up as expected, as you can see from this build: https://dev.azure.com/ms-quantum-public/Microsoft%20Quantum%20(public)/_build/results?buildId=25419&view=artifacts&pathAsName=false&type=publishedArtifacts

anpaz (approved):
lgtm.

the only comment is, should this go into qir-utils.ps1 to follow the same pattern?

swernli:
Good question. In this case, the only binaries that we want to copy out are the runtime, so it really is specific to that folder. When/if we have other folders whose binaries we want to produce as artifacts, it will probably make sense to refactor and pull this logic into the utils.

PR #620, opened 2021-04-08
QIR controller (part 1)

achocron:
This contains most of the QIR controller's fundamental logic; it's missing only the dependency implementations.


src/Qir/Controller/ErrorMessages.Designer.cs
```diff
@@ -0,0 +1,72 @@
+﻿//------------------------------------------------------------------------------
```
cesarzc, line 1: How was this code generated? What is its purpose?

achocron: This is a C# resources file. It's what we use for strings that might get localized in the future. (It's a best practice and also allows us to integrate with the localization tools if we set that up down the line. We have the same thing in the services code for error messages that get surfaced to users)

achocron: Or in other words, it's the autogenerated code behind the resx file containing error messages

src/Qir/Controller/Controller.cs
```diff
@@ -1,26 +1,100 @@
...
-        internal static void Execute(
-            FileInfo input,
-            FileInfo output,
-            FileInfo error)
+        public static async Task ExecuteAsync(
+            FileInfo inputFile,
+            FileInfo outputFile,
+            DirectoryInfo libraryDirectory,
+            FileInfo errorFile,
+            FileInfo bytecodeFile,
```
cesarzc, line 23: If for the driver file and the executable file we are using constants directly in this method, I think it would make sense to use the `BytecodeFilePath` here too rather than getting it as an argument to the method.

src/Qir/Controller/Executable/QuantumExecutableRunner.cs
```diff
@@ -0,0 +1,25 @@
...
+    public class QuantumExecutableRunner : IQuantumExecutableRunner
+    {
+        private readonly ILogger logger;
+
+        public QuantumExecutableRunner(ILogger logger)
+        {
+            this.logger = logger;
+        }
+
+        public Task RunExecutableAsync(FileInfo executableFile, EntryPointOperation entryPointOperation, FileInfo outputFile)
```
cesarzc, line 20: Since the type of linking we use when generating QIR executables is dynamic linking, I think this should receive another argument: `DirectoryInfo librariesDirectory`.

Before running the executable, the content of `librariesDirectory` should be copied into the same directory as the executable. The reason behind this is that it would work in all platforms. Otherwise we would have to edit the `PATH` environment variable or its equivalent depending on the platform.

src/Qir/Controller/Constant.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace Microsoft.Quantum.Qir
+{
+    public static class Constant
```
cesarzc, line 6: I would advocate for a `src` and a `bin` directory. So I think it would make sense to have a `DirectoryPath` static class that elements in the `FilePath` static class use.

Related to this [comment](https://github.com/microsoft/qsharp-runtime/pull/620/files#r610220933), the `DirectoryPath` static class would also be leveraged in that case.

src/Qir/Controller/Executable/ClangClient.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        public ClangClient(ILogger logger)
+        {
+            this.logger = logger;
+        }
+
+        public async Task CreateExecutableAsync(string[] inputFiles, string[] libraries, string libraryPath, string outputPath)
+        {
+            var inputsArg = string.Join(' ', inputFiles);
+            var librariesArg = $"{LinkFlag} {string.Join(LinkFlag, libraries)}";
```
cesarzc, line 23: Is the first `{LinkFlag}` intended?

src/Qir/Controller/Program.cs
```diff
@@ -31,6 +40,15 @@ static void Main(string[] args)
...
             rootCommand.AddOption(outputOption);
+
+            var libraryDirectoryOption = new Option<DirectoryInfo>(
+            aliases: new string[] { "--libraryDirectory" })
+            {
+                Description = "Path to the directory containing the libraries that must be linked to the driver executable.",
+                IsRequired = true
+            };
+
+            rootCommand.AddOption(libraryDirectoryOption);
```
cesarzc, line 51: We will also need to add a `--includeDirectory` option which will be used by the executable generator in a similar way that the `--libraryDirectory` is used.

src/Qir/Controller/Executable/ClangClient.cs
```diff
@@ -0,0 +1,33 @@
...
+    {
+        private const string LinkFlag = " -l ";
+        private readonly ILogger logger;
+
+        public ClangClient(ILogger logger)
+        {
+            this.logger = logger;
+        }
+
+        public async Task CreateExecutableAsync(string[] inputFiles, string[] libraries, string libraryPath, string outputPath)
```
cesarzc, line 20: Related to this [comment](https://github.com/microsoft/qsharp-runtime/pull/620/files#r610263153), this method would need a new parameter (`string includePath`) which should be added to the arguments passed to `clang` using the `-I` option.

src/Qir/Controller/Executable/ClangClient.cs
```diff
@@ -0,0 +1,33 @@
...
+
+using System.CommandLine.Invocation;
+using System.Threading.Tasks;
+using Microsoft.Quantum.Qir.Utility;
+
+namespace Microsoft.Quantum.Qir.Executable
+{
+    public class ClangClient : IClangClient
+    {
+        private const string LinkFlag = " -l ";
```
cesarzc, line 12: I believe the `-l` flag is meant to be used with no space between "l" and the name of the library.

src/Qir/Controller/QirController.csproj
```diff
@@ -1,12 +1,44 @@
...
+
+  <ItemGroup>
+    <BondCodegen Remove="test-artifacts\**" />
+    <BondCodegen Remove="Tests.QirController\**" />
+    <Compile Remove="test-artifacts\**" />
+    <Compile Remove="Tests.QirController\**" />
+    <EmbeddedResource Remove="test-artifacts\**" />
+    <EmbeddedResource Remove="Tests.QirController\**" />
+    <None Remove="test-artifacts\**" />
+    <None Remove="Tests.QirController\**" />
```
cesarzc, line 21: Wondering if all these remove actions are needed. The one that most stands out to me is the BondCodeGen since there are no Bond schemas defined for this project.

src/Qir/Controller/QirController.csproj
```diff
@@ -1,12 +1,44 @@
...
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>netcoreapp3.1</TargetFramework>
+    <RootNamespace>Microsoft.Quantum.Qir</RootNamespace>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <WarningLevel>4</WarningLevel>
```
cesarzc, line 10: I am curious about why setting this warning level when the build configuration is Debug?

src/Qir/Controller/Tests.QirController/ControllerTests.cs
```diff
@@ -0,0 +1,246 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
cesarzc, line 1:
```suggestion
// Copyright (c) Microsoft Corporation.
```

cesarzc: Nit: the header we are using now removed the "All rights reserved." part. Consider changing it in the other new files.

src/Qir/Controller/Executable/ClangClient.cs
achocron: It works either way. It seemed to cause errors on Windows when I did it as you suggest, but adding the space (which is actually documented as well) worked.

src/Qir/Controller/QirController.csproj
achocron: I didn't add any of this manually (i selected the folder and said "exclude from project" because it seemed to get added when I would run tests. Since this was automatically generated, I prefer not to change it.

src/Qir/Controller/QirController.csproj
achocron: I am not sure why but this was somehow added with the warning level as '5' which caused issues, so I downgraded it to '4'. If you don't think it is needed, I will remove it entirely. I'm not sure how it got added, though.

src/Qir/Controller/Executable/ClangClient.cs
achocron: It is - string.Join does not add one before the first element.

src/Qir/Controller/Controller.cs
achocron: The issue is that since the controller code directly creates the file, tests do not work because the path cannot be changed or mocked. I agree that the way you suggest is cleaner, but unfortunately, it is less testable.

src/Qir/Controller/Constant.cs
achocron: Makes sense. I'll address this in a future PR and we can discuss a little more offline.

src/Qir/Controller/Executable/ClangClient.cs
cesarzc: I see.

Maybe there is a better alternative to using `string.Join` because adding the `{LinkFlag}` at the beginning looks like an edge case that doesn't really exists. Using Linq `ForEach` and a `StringBuilder` might be a good alternative.

src/Qir/Controller/Controller.cs
cesarzc: I think tests don't need to be able to control the file where the bytecode is written from outside the controller. The bytecode is already present in the input file.

Additionally, the bytecode file is an implementation detail, so I think we shouldn't expose as an argument to this method.

achocron: It affects unit tests actually, not integration tests, because the path doesn't exist and an error gets thrown.

cesarzc (approved):

src/Qir/Controller/test-qir-controller.ps1
```diff
@@ -19,23 +20,38 @@ Foreach-Object {
...
+        $actualOutput = Get-Content -Path $outputFile -Raw
+        if (-not ($expectedOutput -ceq $actualOutput)) {
+            Write-Host "##vso[task.logissue type=error;]Failed QIR Controller test case: $($_.BaseName)"
+            Write-Host "##[info]Expected output:"
+            Write-Host $expectedOutput
+            Write-Host "##[info]Actual output:"
+            Write-Host $actualOutput
+            $script:all_ok = $False
+        }
+        elseif ((Test-Path $errorFile)) {
```
cesarzc, line 40: Just noticed this. The existence of an output files does not mean an error did not occur. A simulation can fail at any point but it could still output something before that, which is important to debug where it failed.

We could have two independent `if` statements (one for the error file and one for the output file) and the logic inside should only check for the contents of that particular file.

achocron: Got it, I think I misunderstood something in the spec.

PR #624, opened 2021-04-13
Create IQirMachine interface

cesarzc:
This change implements the interface to submit a Q# program for execution in Azure Quantum using a QIR executable.

### DIFF
--- a/src/Simulation/Core/IAzureMachine.cs
+++ b/src/Simulation/Core/IAzureMachine.cs
@@ -1,0 +1,28 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Threading.Tasks;
+using Microsoft.Quantum.Simulation.Core;
+
+namespace Microsoft.Quantum.Runtime
+{
+    public interface IAzureMachine
+    {
+        /// <summary>
+        /// Function that configures a job object before submission.
+        /// </summary>
+        public delegate void ConfigureJob(object job);
+
+        /// <summary>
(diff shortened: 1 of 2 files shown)


cesarzc:
@samarsha this is the idea for the interface that the QIR machine class would implement. It is very similar to `IQuantumMachine`.

The assumption that I am making is that we would extend the `EntryPointInfo` class to contain data about the names of the arguments.

One idea to extend the `EntryPointInfo` class would be to add a new property called `InData` which is of type `Tuple<Type, ITuple<string>?>`. The `Type`  portion of the tuple is the same as the `InType` property while the `ITuple<string>?` portion has the same structure as the `Type` portion but the values represent the names of each argument. Note that it is nullable in case it is a `QVoid` operation. Does it make sense?

samarsha:
To clarify, the client needs three pieces of information for each entry point argument, right?

1. Name
2. Type
3. Value

Would an encoding like this be sufficient? (Using F# for brevity, but we can convert it to C#.)

```fsharp
type ArgumentValue =
    | Bool of bool
    | Integer of int64
    | Double of float
    | Pauli of Pauli
    | Range of Range
    | Result of Result
    | String of string
    | Array of ArgumentValue[]

type Argument = { Name: string; Value: ArgumentValue }

val SubmitAsync:
    args: Argument list
    * qir: byte[]
    * configureJob: ConfigureJob
    -> IQuantumMachineJob Task
```

cesarzc:
Yes, an encoding like that would be sufficient.

PR #626, opened 2021-04-14
Implement dependencies and tests

achocron:
This implements all the dependencies and tests. The test is marked "skip" for now because the driver generation is not yet merged, but when I manually used the driver.cpp that should get generated, the test passed.


src/Qir/Controller/Driver/QirDriverGenerator.cs
```diff
@@ -5,22 +5,48 @@
...
         {
             this.logger = logger;
         }
 
-        public Task GenerateQirDriverCppAsync(DirectoryInfo sourceDirectory, EntryPointOperation entryPointOperation, ArraySegment<byte> bytecode)
+        public async Task GenerateQirDriverCppAsync(DirectoryInfo sourceDirectory, EntryPointOperation entryPointOperation, ArraySegment<byte> bytecode)
         {
-            throw new NotImplementedException();
+            // Wrapped in a task because DirectoryInfo operations are not asynchronous.
+            await Task.Run(async () =>
```
cesarzc, line 27: I think that it would be ok to not run this through a task. The only advantage of using a task is that creating the directory and opening the files to write would run async, but those are not intensive enough operations to warrant a task in my opinion.

src/Qir/Controller/Driver/QirDriverGenerator.cs
```diff
@@ -5,22 +5,48 @@
...
+                // Create bytecode file.
+                var bytecodeFile = new FileInfo(Path.Combine(sourceDirectory.FullName, BytecodeFileName));
+                using var bytecodeFileStream = bytecodeFile.OpenWrite();
+                await bytecodeFileStream.WriteAsync(bytecode.Array, bytecode.Offset, bytecode.Count);
+                logger.LogInfo($"Created bytecode file at {bytecodeFile.FullName}.");
+            });
+        }
+
+        // Virtual method wrapper is to enable mocking in unit tests.
+        public virtual void GenerateQirDriverCppHelper(EntryPointOperation entryPointOperation, Stream stream)
```
cesarzc, line 47: Since the main purpose of this wrapper is to enable mocking in tests, this could be `internal`.

src/Qir/Controller/Driver/QirDriverGenerator.cs
```diff
@@ -5,22 +5,48 @@
...
+        private const string DriverFileName = "driver.cpp";
         private readonly ILogger logger;
 
         public QirDriverGenerator(ILogger logger)
         {
             this.logger = logger;
         }
 
-        public Task GenerateQirDriverCppAsync(DirectoryInfo sourceDirectory, EntryPointOperation entryPointOperation, ArraySegment<byte> bytecode)
+        public async Task GenerateQirDriverCppAsync(DirectoryInfo sourceDirectory, EntryPointOperation entryPointOperation, ArraySegment<byte> bytecode)
```
cesarzc, line 24: nit: since this method creates both the C++ driver and the QIR bytecode files, consider renaming it to something like `GenerateQirExecutableSourceFiles`.

src/Qir/Controller/Executable/QirExecutableGenerator.cs
```diff
@@ -18,10 +24,44 @@ public QirExecutableGenerator(IClangClient clangClient, ILogger logger)
...
+                {
+                    CopyFileIfNotExists(file, binDirectory);
+                }
+
+                // Copy all include contents to bin.
+                logger.LogInfo("Copying include directory contents into the executable's folder.");
+                var includeFiles = includeDirectory.GetFiles();
+                foreach (var file in includeFiles)
+                {
+                    CopyFileIfNotExists(file, binDirectory);
```
cesarzc, line 49: Include files should be copied to the `sourceDirectory` instead of the `binDirectory`.

cesarzc: To not make clang compile headers, this should be done after determining the files that clang actually has to compile.

src/Qir/Controller/Tests.QirController/QirDriverGeneratorTests.cs
```diff
@@ -0,0 +1,63 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
cesarzc, line 1:
```suggestion
// Copyright (c) Microsoft Corporation.
```

cesarzc: nit: files in this repo no longer have the "All rights reserved" portion of the header.

src/Qir/Controller/Tests.QirController/QirExecutableGeneratorTests.cs
```diff
@@ -0,0 +1,127 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
cesarzc, line 1:
```suggestion
// Copyright (c) Microsoft Corporation.
```

cesarzc: nit: files in this repo no longer have the "All rights reserved" portion of the header.

src/Qir/Controller/Driver/QirDriverGenerator.cs
achocron: I think it's generally best to do File IO asynchronously. But since this is not a multithreaded program, I suppose it's probably fine to do it synchronously.

src/Qir/Controller/Driver/QirDriverGenerator.cs
achocron: Yeah if this were a library I'd mark it internal, but since it isn't, there is no appreciable difference with that approach (except for the requirement then to add an InternalsVisibleTo)

src/Qir/Controller/Executable/QirExecutableGenerator.cs
achocron: Fixed, but I had to filter by extension. (Windows doesn't always let the program delete files if they are already present, and thus we have to assume that the files might have already been copied. I don't love it, but I suggest we don't spend time investigating that for now.)

cesarzc: Reviewing the code again, include files don't need to be copied at all because the path to the folder that contains them is passed to clang through the `-I` option.

achocron: Ah okay that's what I thought at first but I thought I was maybe missing something. I'll remove that!

achocron: Removed it and it's still working!

cesarzc (approved):
Looks good!

PR #627, opened 2021-04-14
Add `IQirSubmitter` interface

samarsha:
This PR is based on #624 but adds data types for representing the arguments to an entry point.


src/Simulation/Core/IQirMachine.cs
```diff
@@ -0,0 +1,197 @@
...
+            /// <summary>
+            /// The values of the argument.
+            /// </summary>
+            public IQArray<ArgumentValue> Values { get; }
+
+            /// <summary>
+            /// Creates an array argument value.
+            /// </summary>
+            /// <param name="values">The values of the argument.</param>
+            public Array(IQArray<ArgumentValue> values) => this.Values = values;
```
samarsha, line 194: Should the condition that every value in the array is the same type be enforced here, or later?

src/Simulation/Core/IQirMachine.cs
```diff
@@ -0,0 +1,197 @@
...
+        /// <param name="entryPoint">The fully-qualified name of the entry point to execute.</param>
+        /// <param name="arguments">The arguments to the entry point in the order in which they are declared.</param>
+        /// <returns>The submitted job.</returns>
+        Task<IQuantumMachineJob> SubmitAsync(Stream qir, string entryPoint, IReadOnlyList<Argument> arguments);
+    }
+
+    /// <summary>
+    /// An argument to a QIR callable.
+    /// </summary>
+    public class Argument
```
samarsha, line 32: The `Argument` and `ArgumentValue` classes could go in separate files, but there are already a lot of files in this folder... maybe if all of the quantum machine files were in a separate folder it would be more organized?

cesarzc: Yes, having the classes in separate files would be good.

I created this other [PR](https://github.com/microsoft/qsharp-runtime/pull/629) moving the rest of the related source files to the _Machines_ folder.

While the PR is merged into the feature branch, you can places these new source files into a _Machines_ and after both PRs are merged we will have things more organized.

src/Simulation/Core/IQirMachine.cs
cesarzc: I think it would be good to do it here.

src/Simulation/Core/IQirMachine.cs
swernli: nit: I know it's become standard terminology for us to refer to these classes as "Machines" because of the naming convention there, but I don't think that makes a very good folder name. It's so generic a term that it's not clear what one would expect to find inside of that folder unless already familiar with our class hierarchy and names. Maybe we can find something more descriptive? At the very least, I think we are getting to the point where the folder hierarchy and how that influences the division of feature implementation is something we should address in on of the readmes (similar to the suggestion from @kuzminrobin regarding more clarity regarding the split between QSharp.Core and QSharp.Foundation).

swernli: I'm realizing this comment and any further discussion about naming/organization probably better belongs on Cesar's other PR. Please ignore it here, and I'll move over there!

src/Simulation/Core/IQirMachine.cs
samarsha: I added a check.

src/Simulation/Core/IQirMachine.cs
```diff
@@ -0,0 +1,218 @@
...
+                .All(pair => pair switch 
+                {
+                    (Bool _, Bool _) => true,
+                    (Int _, Int _) => true,
+                    (Double _, Double _) => true,
+                    (Pauli _, Pauli _) => true,
+                    (Range _, Range _) => true,
+                    (Result _, Result _) => true,
+                    (String _, String _) => true,
+                    (Array _, Array _) => true,
```
samarsha, line 213: I think this won't prevent arrays of arrays from having different item types - I'll replace this with a better check.

src/Simulation/Core/IQirMachine.cs
samarsha: > While the PR is merged into the feature branch, you can places these new source files into a Machines and after both PRs are merged we will have things more organized.

@cesarzc, is it right that the argument classes apply not just to submitting a QIR program to Azure, but also to running a QIR program locally if/when the entry point driver supports that? So would it make sense not to put them in the Submitters folder?

cesarzc: @samarsha good point. Yes, I think it makes sense to not put them in the _Submitters_ folder.

cesarzc:
It seems like the only thing left to agree on is whether the new interfaces should have the `*Submitter` suffix.

samarsha:
I renamed the new interfaces to `IQirSubmitter` and `IAzureSubmitter`. Also, as a note for the future, `IQuantumMachine` should probably be `IQSharpSubmitter`, since I think it specifically submits Q# programs.

swernli:
I like the rename! Thanks for landing at something more descriptive.

cesarzc:
I like the rename too!

src/Simulation/Core/IAzureSubmitter.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) Microsoft Corporation.
```
cesarzc, line 1: Now that this [PR](https://github.com/microsoft/qsharp-runtime/pull/629) has been merged, should _IAzureSubmitter.cs_ and _IQirSubmitter.cs_ be placed in the _Submitter_ folder?

samarsha:
I'm wondering if the Windows build errors would be fixed by #627 if main is merged into feature/azure-quantum-simulator-client?

cesarzc:
I think merging main will likely fix the problem. Thanks for creating the merge PR.

cesarzc (approved):

Issue #631, opened 2021-04-15
Fix the Native Simulator to Expose the Fixed-Size Types.


kuzminrobin:
The [Native Simulator](https://github.com/microsoft/qsharp-runtime/tree/main/src/Simulation/Native) is compiled (to a dynamic library) with `g++` on WSL, and presumably with MSVC++ on Win.
Our QIR wrappers around it, e.g. [`CFullstateSimulator`](https://github.com/microsoft/qsharp-runtime/blob/main/src/Qir/Runtime/lib/Simulators/FullstateSimulator.cpp) in QIR Runtime, are compiled with a different compiler - `clang`.

The Native Simulator dynamic library [exposes its API](https://github.com/microsoft/qsharp-runtime/blob/main/src/Simulation/Native/src/simulator/capi.hpp) using the _built-in_ types like `unsigned`, `bool`, `double`, etc. The size of such types can be different in different compilers, which can cause problems when crossing the dynamic library boundary (see more details [here](https://stackoverflow.com/questions/22797418/how-do-i-safely-pass-objects-especially-stl-objects-to-and-from-a-dll/22797419#22797419), search for "Datatype size".)

Consider replacing all the built-in types with the fixed-size types, like `uint32_t`, `uint8_t`, <to do: fixed-size type for `double`>.

Issue #603, opened 2021-04-05
Minor Engineering Dept Items


kuzminrobin:
https://github.com/microsoft/qsharp-runtime/search?q=MICROSOFT_QUANTUM_DECL_IMPORT
`MICROSOFT_QUANTUM_DECL_IMPORT` is likely never used. Consider removing from [config.hpp.in](https://github.com/microsoft/qsharp-runtime/blob/main/src/Simulation/Native/src/config.hpp.in), [config.hpp](https://github.com/microsoft/qsharp-runtime/blob/main/src/Simulation/Native/src/config.hpp)

To be continued.

PR #634, opened 2021-04-16
Added DumpMachine() and DumpRegister(), some refactoring

### LINKED ISSUE #578: Move the Message() implementation from QIS to RT
Currently the [Message()](https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.intrinsic.message) function is implemented in [QIS](https://github.com/microsoft/qsharp-runtime/search?q=quantum__qis__message__body) and is visible as [`__quantum__qis__message__body`](https://github.com/microsoft/qsharp-runtime/search?q=__quantum__qis__message__body).

We have a second implementation - [`quantum__rt__message`](https://github.com/microsoft/qsharp-runtime/search?q=quantum__rt__message), that is not visible outside.

**qsharp-runtime**
Make `__quantum__rt__message` visible.
Replace the implementation of `void quantum__rt__message(QirString* msg)` with the implementation of `void quantum__qis__message__body(QirString* qstr)`.

**qsharp_compiler**
Change the compiler to call `__quantum__rt__message` instead of `__quantum__qis__message__body`.

**qsharp-runtime**
Remove `__quantum__qis__message__body` (and everything that is called from it only).

kuzminrobin:
(Feel free to let me know if you'd like me to break this up into multiple PRs)
Added DumpMachine() and DumpRegister() functions implementation.
Added some README.md files (that I've written for myself, to get a more clear understanding for the future of what to add where, and decided to share those).
Did some refactoring of the code.
Resolves #578.


src/Azure/Azure.Quantum.Client/Microsoft.Azure.Quantum.Client.csproj
```diff
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
```
kuzminrobin, line 1: (feel free to let me know if you'd prefer me to revert this change)
Looks like this is a remainder of the rebasing/merging epopee. This changes the first few bytes (not visible in the diff) that determine the file encoding.

src/Qir/Runtime/lib/QIR/README.md
```diff
@@ -0,0 +1,61 @@
...
+## Level 4
+**callables.cpp**           Defines the {internals of `QirTupleHeader`, `QirCallable`}, `quantum__rt__tuple_*()`, `quantum__rt__callable_*()`  
+                            Depends on `QirArray`, `Microsoft::Quantum::GlobalContext()`, `quantum__rt__fail()`, `quantum__rt__string_create()`, `TupleWithControls`,  
+                            Consider breaking up into **Tuples.cpp** and **Callables.cpp**.
+
+## Level 5
+**bridge-rt.ll**            Defines the `@__quantum__rt__*` entry points (to be called by the `.ll` files generated from users' `.qs` files).  
+                            The C++ Standard reserves the identifiers starting with double underscores `__`, that is why the definitions of `@__quantum__rt__*`
+                            have been put to `.ll` file rather than `.cpp` file.  
+                            Depends on `quantum__rt__*` implementations (in **utils.cpp**, **strings.cpp**, **delegated.cpp**, **arrays.cpp**, **callables.cpp**).
```
kuzminrobin, line 61: I added the `README.md` files like this for myself for the future (to see better the dependencies in the code) and decided to share with others.

src/Qir/build_all.ps1
```diff
@@ -0,0 +1,3 @@
+& Runtime\build-qir-runtime.ps1
+& Tests\build-qir-tests.ps1
+& Samples\build-qir-samples.ps1
```
kuzminrobin, line 3: This and `test_all.ps1` are also just for myself, rather raw files (they don't analyze the result of the command), they list the commands to run manually.
I suspect the `qir-utils.ps1` is for the same or similar purpose, feel free to refer me to how to use it (some quick start/examples).

src/Simulation/Common/Microsoft.Quantum.Simulation.Common.csproj
```diff
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
```
kuzminrobin, line 1: Another result of rebase/merge epopee.

src/Simulation/Native/src/simulator/simulator.hpp
```diff
@@ -302,6 +302,18 @@ class Simulator : public Microsoft::Quantum::Simulator::SimulatorInterface
         }
     }
 
+    void dump(TDumpToLocationCallback callback, TDumpLocation location) override
+    {
+        recursive_lock_type l(mutex());
```
kuzminrobin, line 307: (For myself: If making changes to this PR, remove this line because of the second part of #603)

kuzminrobin:
Please review while I'm resolving the build breaks

src/Azure/Azure.Quantum.Client/Microsoft.Azure.Quantum.Client.csproj
cgranade: In general, the first two bytes should not be 0xFEFF, as that is the UTF-16 and UTF-32 byte order mark; its presence in a UTF-8 encoded file is a nonstandard Windows extension and reduces portability.

src/Qir/Runtime/lib/QIR/strings.cpp
```diff
@@ -11,7 +11,7 @@
 #include "QirTypes.hpp"
 #include "QirRuntime.hpp"
 
-std::unordered_map<std::string, QirString*>& AllocatedStrings()
+std::unordered_map<std::string, QirString*>& AllocatedStrings()     // Cannot be static, is called by tests.
```
cgranade, line 14: I'm not sure I follow why making this static blocks calling it from tests?

src/Qir/Runtime/lib/QSharpCore/README.md
```diff
@@ -0,0 +1,30 @@
...
+Same-level entities are independent of each other (unless specified otherwise). Entities depend on the levels listed earlier only.  
+
+
+## Level 0. External To This Directory
+**public\CoreTypes.hpp**    (QUBIT, PauliId, RESULT)[, `public\QirTypes.hpp` (QirArray)].  
+**public\QirContext.hpp**   Declares `GlobalContext()`.
+**public\QSharpSimApi_I.hpp**
+                            Defines `IQuantumGateSet`.
+
+## Level
```
cgranade, line 18: Looks like something is missing here?

src/Qir/Runtime/lib/QSharpCore/qsharp__core__qis.hpp
```diff
@@ -1,24 +1,11 @@
...
-#include <cstdint>
-
-#include "CoreTypes.hpp"
+#include "CoreTypes.hpp"    // QUBIT, PauliId, RESULT
 
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
 struct QirArray;
-struct QirCallable;
```
cgranade, line 11: Why were these declarations removed?

src/Qir/Runtime/lib/QSharpFoundation/qsharp-foundation-qis.ll
```diff
@@ -43,10 +44,11 @@ define dllexport void @__quantum__qis__message__body(%String* %.str) {
 }
 
 ;===============================================================================
-; quantum.qis math functions
+; quantum.qis math functions declarations
 ;
 
 ; LLVM intrinsics (https://llvm.org/docs/LangRef.html):
+; To do: consider calling these directly from the compiler-generated .ll code, rather than throug the QIR.
```
cgranade, line 51:
```suggestion
; TODO: consider calling these directly from the compiler-generated .ll code, rather than through the QIR.
```

src/Qir/Runtime/lib/QSharpFoundation/qsharp-foundation-qis.ll
```diff
@@ -202,12 +215,8 @@ define dllexport double @__quantum__qis__drawrandomdouble__body(double %min, dou
 
 
 ;===============================================================================
-; quantum.qis conditional functions
+; quantum.qis conditional functions implementation
```
cgranade, line 218: In Q#, these are operations rather than functions; does "function" here refer to the Q# signature, or to the projection of that into QIR?

src/Qir/Runtime/lib/Simulators/FullstateSimulator.cpp
```diff
@@ -122,6 +128,7 @@ namespace Quantum
             return ids;
         }
 
+        // To do: remove after the `function DumpMachine<'T> (location : 'T) : Unit` is implemented.
```
cgranade, line 131:
```suggestion
        // TODO: remove after the `function DumpMachine<'T> (location : 'T) : Unit` is implemented.

```

src/Qir/Runtime/lib/Simulators/FullstateSimulator.cpp
```diff
@@ -171,6 +178,7 @@ namespace Quantum
             }
         }
 
+        // To do: remove after the `function DumpMachine<'T> (location : 'T) : Unit` is implemented.
```
cgranade, line 181: This might not need to be removed; as discussed earlier, host programs can make use of structured diagnostics (e.g.: the state vector represented as a vector of complex numbers, rather than as a human-readable string).

src/Qir/Runtime/lib/Simulators/FullstateSimulator.cpp
```diff
@@ -408,7 +419,109 @@ namespace Quantum
...
             return (std::abs(actualProbability - probabilityOfZero) < precision);
         }
-    };
+
+      private:
+        std::ostream& GetOutStream(const void* location, std::ofstream& outFileStream);
+        void DumpMachineImpl(std::ostream& outStream);
+        void DumpRegisterImpl(std::ostream& outStream, const QirArray* qubits);
+        void GetStateTo(TDumpLocation location, TDumpToLocationCallback callback);
+        bool GetRegisterTo(TDumpLocation location, TDumpToLocationCallback callback, const QirArray* qubits);
```
cgranade, line 428: Critically, this private method does not get a register, but the quantum state of that register (this is a has-a versus is-a distinction, in that a register has a state, but a register is not a state). We should make absolutely sure to keep that distinction clear in our documentation and our naming conventions.

src/Qir/Runtime/lib/Simulators/FullstateSimulator.cpp
```diff
@@ -408,7 +419,109 @@ namespace Quantum
...
+        outStream << "*********************" << std::endl;
+        outStream.flush();
+    }
+
+    void CFullstateSimulator::DumpRegisterImpl(std::ostream& outStream, const QirArray* qubits)
+    {
+        outStream << "*********************" << std::endl;
+        if(!this->GetRegisterTo((TDumpLocation)&outStream, dumpToLocationCallback, qubits))
+        {
+            outStream << "## Qubits were entangled with an external qubit. Cannot dump corresponding wave function. ##" << std::endl;
```
cgranade, line 456: Separate from this PR, but this has been a longstanding limitation, such that we should consider relaxing it at some point in the future.

src/Qir/Runtime/lib/Simulators/FullstateSimulator.cpp
```diff
@@ -408,7 +419,109 @@ namespace Quantum
...
+                {
+                    outStream << "|" << std::bitset<8>(idx) << ">: " << re << "+" << im << "i" << std::endl;
+                }
+                return true;
+            };
+    }; // class CFullstateSimulator
+
+    void CFullstateSimulator::DumpMachineImpl(std::ostream& outStream)
+    {
+        outStream << "*********************" << std::endl;
```
cgranade, line 445: Are these rows of asterisks needed? I don't believe we have similar in any of our other diagnostic output functionality, but would need to double-check.

src/Qir/Runtime/lib/Simulators/FullstateSimulator.cpp
```diff
@@ -408,7 +419,109 @@ namespace Quantum
...
+
+    bool CFullstateSimulator::GetRegisterTo(TDumpLocation location, TDumpToLocationCallback callback, const QirArray* qubits)
+    {
+        // `qubits->buffer` points to the sequence of qubit ids each of type `void*` (logically `qubits->buffer` is of type `void**`).
+        // We need to pass to the native simulator the pointer to the sequence of `unsigned`.
+        std::vector<unsigned> unsQbIds(qubits->count);
+        size_t i = 0;
+        for (unsigned& qbId : unsQbIds)
+        {
+            qbId = (unsigned)(uintptr_t)(((void**)(qubits->buffer))[i]);
```
cgranade, line 470: It feels a little odd to see as many C-style casts here all of a sudden, is there a reason for that?

src/Simulation/Native/src/simulator/capi.cpp
```diff
@@ -216,6 +221,18 @@ extern "C"
         return Microsoft::Quantum::Simulator::get(id)->dumpQubits(qs, callback);
     }
 
+    MICROSOFT_QUANTUM_DECL bool DumpQubitsToLocation(
+        _In_ unsigned id,
+        _In_ unsigned n,
+        _In_reads_(n) unsigned* q,
+        _In_ TDumpToLocationCallback callback, 
+        _In_ TDumpLocation location)
```
cgranade, line 229: I'm a little confused about the API for this export; in particular, why does this function need both the location and a callback that takes that location? Given that the callbacks provided for this argument are all lambdas, I'd expect the  location to be captured in a closure cell of that lambda, rather than explicitly built into the API here. If you went with that approach, I don't think you would need this and the other new API declaration to be added to the native simulator, as you would then again have a callback of type `bool (*)(size_t, double, double)`.

src/Qir/Runtime/lib/Simulators/FullstateSimulator.cpp
```diff
@@ -408,7 +419,109 @@ namespace Quantum
...
+    std::ostream& CFullstateSimulator::GetOutStream(const void* location, std::ofstream& outFileStream)
+    {
+        // If the location is not nullptr and not empty string then dump to a file:
+        if((location != nullptr) &&
+            (((static_cast<const QirString *>(location))->str) != ""))
+        {
+            // Open the file for appending:
+            const std::string& filePath = (static_cast<const QirString *>(location))->str;
+
+            outFileStream.open(filePath, std::ofstream::out | std::ofstream::app);
```
cgranade, line 496: It looks like there's no exception handling for any I/O issues that could arise here; in general, diagnostic routines such as DumpMachine and DumpRegister should not cause a quantum program to fail, such that I would suggest catching I/O exceptions here and providing actionable error messages to the user at the console instead.

src/Qir/Runtime/lib/Simulators/FullstateSimulator.cpp
```diff
@@ -408,7 +419,109 @@ namespace Quantum
...
+            if((outFileStream.rdstate() & std::ofstream::failbit) != 0)
+            {
+                QirString qstr {std::string("Failed to open dump file \"") +
+                                            filePath +
+                                            "\""};
+                quantum__rt__fail(&qstr);
```
cgranade, line 502: Similarly here, hitting a `fail` call isn't ideal for a diagnostic routine.

src/Qir/Runtime/lib/Simulators/ToffoliSimulator.cpp
```diff
@@ -109,11 +109,22 @@ namespace Quantum
             return std::abs(actualZeroProbability - probabilityOfZero) < precision;
         }
 
+        // To do: remove after the `function DumpMachine<'T> (location : 'T) : Unit` is implemented.
```
cgranade, line 112: As before, I'm not sure that I agree with removing this; it's still useful to host programs.
```suggestion
        // TODO: remove after the `function DumpMachine<'T> (location : 'T) : Unit` is implemented.

```

src/Qir/Runtime/lib/Simulators/ToffoliSimulator.cpp
```diff
@@ -109,11 +109,22 @@ namespace Quantum
...
 
+        // To do: remove after the `function DumpMachine<'T> (location : 'T) : Unit` is implemented.
         void GetState(TGetStateCallback callback) override
         {
             throw std::logic_error("operation_not_supported");
         }
 
+        void DumpMachine(const void* location) override
+        {
+            throw std::logic_error("operation_not_supported");
```
cgranade, line 120: Is there a reason this isn't implemented? See https://github.com/microsoft/qsharp-runtime/blob/21f5fab97682d9bc0ae8e3b58f679fdb952f1c24/src/Simulation/Simulators/ToffoliSimulator/ToffoliSimulator.cs#L336 for an example of what diagnostic output for Toffoli simulation should look like.

In any case, this shouldn't throw an exception, but rather should provide a clear message at the console informing the user that this diagnostic is not yet supported.

cgranade (commented):
This is a large PR, so I haven't had the chance to finish reviewing it yet; wanted to give initial feedback on what parts I had looked at so far, though.

src/Qir/Runtime/lib/QIR/OutputStream.cpp
```diff
@@ -0,0 +1,39 @@
...
+    std::ostream& OutputStream::Get()
+    {
+        return *currentOutputStream;
+    }
+
+    std::ostream& OutputStream::Set(std::ostream& newOStream)
+    {
+        std::ostream& oldOStream = *currentOutputStream;
+        currentOutputStream = &newOStream;
+        return oldOStream;
```
ScottCarda-MS, line 26: Is it common practice for a setter to return the old value?

src/Qir/Runtime/lib/QIR/OutputStream.cpp
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+// https://stackoverflow.com/a/5419388/6362941
```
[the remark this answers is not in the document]
kuzminrobin: Added.

src/Qir/Runtime/lib/QIR/README.md
```diff
@@ -0,0 +1,61 @@
+# API Dependency
+
+(See the raw file. Please keep the raw file readable rather than the browser-rendered one)
```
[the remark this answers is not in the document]
kuzminrobin: I experimented with different content, including tables. I ended up in some balance between the [raw file](https://raw.githubusercontent.com/microsoft/qsharp-runtime/34040a7abab04ad041ec897bf2399397a93dc674/src/Qir/Runtime/lib/QIR/README.md) readability and the [web view](https://github.com/microsoft/qsharp-runtime/blob/34040a7abab04ad041ec897bf2399397a93dc674/src/Qir/Runtime/lib/QIR/README.md) readability. 
Since the file is intended to be mostly used _during coding_, I wanted the raw file readability to be higher priority.

Rewording this fragment to:
> (Try to keep the readability balance between the web view and raw file, give the preference to the raw file)

src/Qir/Runtime/lib/QIR/README.md
```diff
@@ -0,0 +1,61 @@
...
+**public\CoreTypes.hpp**    Defines `QIR_SHARED_API`, `QUBIT`, `Qubit`, `RESULT`, `Result`, `ResultValue`, `PauliId`.  
+                            Does not depend on anything of our code.
+
+**public\QirTypes.hpp**     Defines `QirArray`, `QirString`, `PTuple`, `QirTupleHeader`, `TupleWithControls`, `QirCallable`, `QirRange`.  
+                            Depends on the listed earlier ones.
+
+**public\QirRuntime.hpp**   Declares `quantum__rt__*()`. Depends on the listed earlier ones.  
+
+
+## Level 1
```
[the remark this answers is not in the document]
kuzminrobin: Added.

src/Qir/Runtime/lib/QIR/bridge-rt.ll
```diff
@@ -233,7 +233,7 @@ define dllexport %Array* @__quantum__rt__array_copy(%Array* %.ar, i1 %force) {
 
 ; TODO: This bridge isn't cross-platform!
 ; it works on Windows but on Linux %args ends up not being a valid pointer.
-declare void @DebugLogPtr(i8*)
+;declare void @DebugLogPtr(i8*)
```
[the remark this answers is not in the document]
kuzminrobin: Removed.

src/Qir/Runtime/lib/QIR/context.cpp
```diff
@@ -9,58 +9,90 @@
...
-        if (this->trackAllocatedObjects)
+        if(this->trackAllocatedObjects)
         {
-            this->allocationsTracker = std::make_unique<AllocationsTracker>();
+            this->allocationsTracker->OnAllocate(object);
         }
     }
-    QirExecutionContext::~QirExecutionContext() = default;
+
+    IRuntimeDriver* QirExecutionContext::GetDriver() const
```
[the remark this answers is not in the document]
kuzminrobin: @swernli, any comments about why the simulator is called Driver?
I believe it has been renamed to Driver during the recent mass-refactoring couple of weeks ago.

src/Simulation/Native/src/simulator/capi.cpp
kuzminrobin: I agree with your idea, and that was the first thing I considered (and as I now understand, I probably should have tried that, at least out of curiosity). However,

There are a few topics in this decision.
1. We are crossing the dynamic lib boundary (ABI boundary) during these calls. Passing the C++ instances over this boundary is [unreliable](https://stackoverflow.com/questions/22797418/how-do-i-safely-pass-objects-especially-stl-objects-to-and-from-a-dll/22797419#22797419). That is why the callback is passed as a _function pointer_ (typical/reliable in C ABI), rather than [`std::function()`](http://www.cplusplus.com/reference/functional/function/?kw=function).
2. The lambda that _has no captures_, is (and/or can be implicitly cast to) the function pointer. That's what happens here. But if the lambda _does have captures_, then it cannot be represented/passed just as a function pointer. It becomes a dual entity - the function pointer + the reference to the captured data, i.e. it becomes a _function object_, sometimes called _functor_ (speaking in C++). Such an object cannot be just passed as an argument corresponding to the parameter of function pointer type.

That's what I'm doing here - passing the function object - "lambda with captures" - as 2 separate pointers explicitly.

---
I looked at how C# handles this. And for now I failed to figure out how they _magically_ pass a lambda with captures to this function pointer in
```c++
MICROSOFT_QUANTUM_DECL void Dump(_In_ unsigned id, _In_ bool (*callback)(size_t, double, double))
```
I realize that my understanding described above can be mistaken. Any explanation is welcome. Without the reliable explanation, I'm afraid passing a lambda with captures as a function pointer can be a hack that can break upon the next compiler upgrade. And [this](https://stackoverflow.com/questions/7852101/c-lambda-with-captures-as-a-function-pointer) makes me think that it will hardly compile at all.

_Requesting for an explanation/example how to implement better_.

src/Qir/Runtime/lib/Simulators/FullstateSimulator.cpp
kuzminrobin: Thanks. Good point.

The [`open()`](http://www.cplusplus.com/reference/fstream/ofstream/open/) throws exceptions if "member `exceptions` was set to throw". My code does not deal with "setting to throw" for this `std::ofstream` instance, that's why exception handling can be redundant. However, I see that this function gets the instance as a parameter, so "setting to throw" could have been done before this function. To make the function reliable I'm adding the exception handling.

Also, in case of `open()` failure, I'm adding the actionable complaint to `std::cerr` and dumping to `std::cout` (instead of calling `quantum__rt__fail()`).

Separate thanks for this feedback!

src/Qir/Runtime/lib/Simulators/FullstateSimulator.cpp
kuzminrobin: Fixed.

src/Qir/Runtime/lib/QIR/strings.cpp
kuzminrobin: The `static` function is only visible in the current translation unit (current .cpp file after preprocessing). If I make it `static`, then the linker will not be able to resolve the call to this function from the test's translation unit.

(I feel that you already know that and you mean something different. If that's the case then please elaborate)

src/Qir/Runtime/lib/QSharpCore/README.md
kuzminrobin: Yep, the Markdown format is missing the automatic enumeration of the headers here :-D... 
Adding the explicit enumeration (but for myself personally would prefer to leave as is :-).

src/Qir/Runtime/lib/QSharpCore/qsharp__core__qis.hpp
kuzminrobin: Because they are either not used below or are declared in the included header.

src/Qir/Runtime/lib/QIR/context.cpp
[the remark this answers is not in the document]
swernli: This terminology is something I switched to because this interface is not really exclusive for the simulator or for host program, but rather the object that allows for invoking functions on the backend target. That might be a simulator, a direct link to hardware, or some other intermediate component that forwards the invocations to an entirely different infrastructure. In this configuration, what was the host program is really expected to be in the QIR itself (or perhaps some thin logic in the `main` function the QIR is compiled with).

src/Qir/Runtime/lib/QSharpFoundation/qsharp-foundation-qis.ll
kuzminrobin: Fixed.

src/Qir/Runtime/lib/QSharpFoundation/qsharp-foundation-qis.ll
kuzminrobin: Changed to 
> quantum.qis conditional functions/**operations** implementation

src/Qir/Runtime/lib/Simulators/FullstateSimulator.cpp
kuzminrobin: Fixed.

src/Qir/Runtime/lib/Simulators/FullstateSimulator.cpp
kuzminrobin: The `GetState()` is only called by `CFullstateSimulator::DumpState()`, which is never called (not to be confused with the C#'s `DumpState()`).
The `DumpMachine()` and `DumpRegister()` added in this PR, provide the same functionality. 
After this PR I plan to remove `GetState()` and `DumpState()`.

kuzminrobin: #636

src/Qir/Runtime/lib/Simulators/FullstateSimulator.cpp
kuzminrobin: I'm not sure what you mean. The parameter `qubits` is a pointer to `QirArray` containing the qubit _ids_ (not qubit _values_). 
Do you mean that I need to rename `qubits` to `qubitIds` or something like that?
Please elaborate, feel free to give me a Teams call.

src/Qir/Runtime/lib/Simulators/FullstateSimulator.cpp
kuzminrobin: #636. Feel free to provide more details in that issue.

src/Qir/Runtime/public/QirContext.hpp
```diff
@@ -14,30 +14,33 @@ namespace Quantum
...
+        static void Deinit();
 
-        QirExecutionContext(IRuntimeDriver* sim, bool trackAllocatedObjects);
+        QirExecutionContext(IRuntimeDriver* simulator, bool trackAllocatedObjects);
         ~QirExecutionContext();
-    };
-    extern thread_local std::unique_ptr<QirExecutionContext> g_context;
-    extern QIR_SHARED_API std::unique_ptr<QirExecutionContext>& GlobalContext();
 
-    struct QIR_SHARED_API QirContextScope
```
swernli, line 31: Removing this type is technically a breaking change, such that I think we should reconsider. Both our generated main.cpp and any clients that have started building against the runtime will need to change to adapt to this. I think we may need a more general discussion to identify which parts of our API are up for refactor vs which ones would be part for breaking change.

swernli (changes requested):
I'll have a more full review later, but some of these changes (most notably the ones to the headers in the public folder) are breaking changes, such that I think we should have a more general discussion about which parts of our API boundary have dependencies vs not so we can avoid breaking any consumers.

src/Qir/Runtime/lib/Simulators/FullstateSimulator.cpp
kuzminrobin: (This has been copied from `DumpState()`)

Changed to be similar to what C# outputs.
C#: `channel($"# wave function for qubits with ids (least to most significant): {string.Join(";", ids)}");`

src/Qir/Runtime/lib/Simulators/FullstateSimulator.cpp
kuzminrobin: Yep, otherwise the compiler complains. :-D

1. (regarding the C-style casts as opposed to C++-style casts)
The C-style casts are shorter and more readable than the C++-style casts (when we are comparing with `reinterpret_cast<>()`).

2. (regarding the number of casts) 

By C++ definition, the type of `qubits->buffer` is `char *` (pointer to the sequence of `char`s).

In reality the `qubits->buffer` (for an array of _qubits_) points to the sequence of qubit IDs, each qubit ID having a size of a pointer (i.e. the size of qubit ID is currently implementation-dependent, and each qubit ID gets cast to, and stored as, a `void*`). To summarize, the `qubits->buffer` points to the sequence of pointers to `void` i.e. the real type of the `qubits->buffer` is `void**` (hence the cast to `void**`).
Then we index that pointer - `[i]` - i.e. we get the corresponding `void*` out of the sequence. On a 64-bit platform that `void*` is 64-bit. Eventually we need to cast it to `unsigned` (which in current (compiler) implementation is 32-bit). A direct _narrowing_ cast from a 64-bit _pointer_ to a 32-bit _integer_ causes a Clang compiler error or warning (regardless of whether the cast is implicit or explicit, and whether it is a C-style cast or C++-style cast). That is why I have to use an intermediate cast to an integer type -  [`uintptr_t`](https://man7.org/linux/man-pages/man0/stdint.h.0p.html) - whose size is the same as the pointer type (regardless of whether the architecture is 64-bit or 32-bit). And then I apply the _narrowing_ cast from 64-bit _integer_ to 32-bit _integer_ (`unsigned`), and the compiler is happy.

May be the last cast to `unsigned` can be _implicit_, I didn't check or don't remember, but the static analyzers typically complain for the _implicit narrowing_ casts. So I have cast explicitly.

src/Qir/Runtime/lib/Simulators/ToffoliSimulator.cpp
kuzminrobin: Fixed.

src/Qir/Runtime/lib/Simulators/ToffoliSimulator.cpp
kuzminrobin: Thank you for the direct link. But I don't know C# and it is very hard for me to translate the C# code to C++. May be you can provide a couple of examples of the eventual output? Or can refer me to how to redirect the Q# project to use a Toffoli simulator (with C# runtime)? (such that I can generate the example output by myself)

I have written these fragments (exception throw) by example of `ToffoliSimulator::GetState()` (one function above), and based on observation that currently in QIR the `ToffoliSimulator` is only used in tests. I thought the `ToffoliSimulator` is something dying (or already dead), but looks like it is not.

src/Qir/Runtime/lib/QIR/OutputStream.cpp
kuzminrobin: I believe in typical cases - _not_. But some standard functions do.

My code was inspired by 
* [`std::set_new_handler()`](http://www.cplusplus.com/reference/new/set_new_handler/), described in more details in chapter 8 [here](https://ptgmedia.pearsoncmg.com/images/9780321334879/samplepages/0321334876.pdf) or [here](http://wavelino.coffeecup.com/pdf/Effective%20Cshapshap.pdf)
* and (2) of [`std::ios::rdbuf()`](http://www.cplusplus.com/reference/ios/ios/rdbuf/), but this one is more a getter with an extra job of a setter ;-).

src/Qir/Runtime/lib/Simulators/ToffoliSimulator.cpp
cgranade: The Toffoli simulator is rather important to a number of different domain areas, especially reversible oracle synthesis. As such, I would suggest that it's important to make sure that even if a feature isn't implemented, the Toffoli simulator at least fails gracefully — in this case, replacing the `throw` with an insertion to `std::cout` may be enough to do that.

As for implementing the feature, the Toffoli simulator takes advantage of the fact that if you start in a computational basis state and only apply `X` and `Controlled X` operations, the state of the system always has the form |𝑏₀𝑏₁…𝑏ₙ₋₁⟩ for some classical bitstring 𝑏₀𝑏₁…𝑏ₙ₋₁. Thus, a reasonable diagnostic for Toffoli simulation is to display the state 𝑏₀𝑏₁…𝑏ₙ₋₁. When 𝑛 is large enough, then hex output can be particularly helpful; most of the C# code above is intended to switch between outputting binary and hex formats based on the size of the register.

src/Qir/Runtime/public/QirContext.hpp
kuzminrobin: I'm not removing this type, I'm encapsulating/moving it into `QirExecutionContext` and renaming to `Scoped`, i.e. now it is called `QirExecutionContext::Scoped` (I'm converting a separate class to a nested class).

The reason for encapsulating is that this class is built on top of `QirExecutionContext` class. If `QirExecutionContext` is moved or deleted, then it is reasonable for this class to be automatically moved or deleted together with the `QirExecutionContext`.

kuzminrobin:
Looks like I missed an important information earlier. I didn't know that our QIR code has customers.
I need more info about that. If yo have time, then I would really appreciate if you (or anybody else) could tell me more about how the customers use our QIR code.

src/Azure/Azure.Quantum.Client/Microsoft.Azure.Quantum.Client.csproj
kuzminrobin: I see that this particular file is ASCII-encoded (the first 2 chars are 3C 50, which are the first 2 visible characters `<` and `P`).
I'll double-check the other similar changes.

src/Simulation/Common/Microsoft.Quantum.Simulation.Common.csproj
kuzminrobin: (no invisible chars in front of `<Project`)

src/Xunit/Microsoft.Quantum.Xunit.csproj
```diff
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
```
kuzminrobin, line 1: (No invisible chars in front of `<Project`)

src/Azure/Azure.Quantum.Client/Microsoft.Azure.Quantum.Client.csproj
kuzminrobin: All such changes have no invisible chars in the beginning.

src/Qir/Runtime/lib/Simulators/ToffoliSimulator.cpp
kuzminrobin: For now changing to
```c++
        void DumpMachine(const void* location) override
        {
            std::cerr << __func__ << " is not yet implemented" << std::endl;
            // TODO: How to implement: https://github.com/microsoft/qsharp-runtime/pull/634#discussion_r615183791
        }

        void DumpRegister(const void* location, const QirArray* qubits) override
        {
            std::cerr << __func__ << " is not yet implemented" << std::endl;
            // TODO: How to implement: https://github.com/microsoft/qsharp-runtime/pull/634#discussion_r615183791
        }
```

src/Qir/Runtime/public/QirContext.hpp
kuzminrobin: I didn't know that our QIR code has customers.
Let's discuss that separately. Thank you for letting me know.

swernli:
Yeah, it has definitely been the case up until now that we have been more free to do refactoring based on no external dependencies on this API, but that seems to be changing quickly. The changes you are making here are well thought out and reasonable from the point of view of a C++ API, and give us important things to consider. At the same time, we also should start considering ease of interop with other languages, in which case it might be worth leaning toward a more C-style API that relies on extern functions and basic types instead of objects. All these API design concerns are things that only recently have started coming up, so I think it will be important to have your experience and knowledge to take into account with that before we settle on a firm approach, so between that and the few partners who are taking a dependency on the API it seems we are due for a larger team discussion about how we want to handle updates to the design of the API.

src/Qir/Runtime/public/QirRuntime.hpp
```diff
@@ -162,7 +149,6 @@ extern "C"
 
     // Initializes the callable with the provided function table and capture tuple. The capture tuple pointer
     // should be null if there is no capture.
-    typedef void (*t_CallableEntry)(PTuple, PTuple, PTuple);                                                // NOLINT
```
kuzminrobin, line 165: This was a redefinition of the same in `QirTypes.hpp` included above.

kuzminrobin:
I have reverted the breaking changes. Please give another round of review.

src/Qir/Runtime/lib/QIR/bridge-rt.ll
```diff
@@ -233,7 +233,6 @@ define dllexport %Array* @__quantum__rt__array_copy(%Array* %.ar, i1 %force) {
 
 ; TODO: This bridge isn't cross-platform!
 ; it works on Windows but on Linux %args ends up not being a valid pointer.
-declare void @DebugLogPtr(i8*)
```
swernli, line 236: nit: now that `DebugLogPtr` has been removed, we can remove the commented out call sites below.

src/Qir/Runtime/lib/QIR/context.cpp
```diff
@@ -46,21 +31,92 @@ namespace Quantum
...
         }
 
-        ResultOne = nullptr;
-        ResultZero = nullptr;
         g_context.reset(nullptr);
     }
 
-    QirExecutionContext::QirExecutionContext(IRuntimeDriver* sim, bool trackAllocatedObjects)
-        : driver(sim)
+    QirExecutionContext::QirExecutionContext(IRuntimeDriver* simulator, bool trackAllocatedObjects)
```
swernli, line 37: nit: This shouldn't really be named simulator, as it doesn't have to be one. It could be a driver that sends instructions to hardware or collections gates to output to something else. Not something I think you need to change here, but it makes it seem like this runtime infrastructure is specific to simulation, which it is not.

src/Qir/Runtime/lib/QIR/context.cpp
```diff
@@ -46,21 +31,92 @@ namespace Quantum
...
+    {
+        QirExecutionContext::Init(sim, trackAllocatedObjects);
+    }
+
+    QirExecutionContext::Scoped::~Scoped()
+    {
+        QirExecutionContext::Deinit();
+    }
+
+    QirContextScope::QirContextScope(IRuntimeDriver* sim, bool trackAllocatedObjects /*= false*/)
```
swernli, line 111: Thanks for putting these back!

src/Qir/Runtime/lib/QSharpFoundation/qsharp-foundation-qis.ll
```diff
@@ -31,22 +31,24 @@
 %struct.QirString = type opaque
 %PauliId = type i32
 
-declare void @quantum__rt__message(%"struct.QirString"* %str)
+; TODO: remove this declaration after the https://github.com/microsoft/qsharp-runtime/issues/578 is resolved.
+declare dllimport void @quantum__rt__message(%"struct.QirString"* %str)
```
swernli, line 35: Do we have tests that execute this on Linux? I remember running into an issue with `dllimport` decorations causing a crash on Linux, which was part of the motivation for switching from global variables for `ResultOne` and `ResultZero` to functions (Windows required `dllimport` for it to link, but adding it somehow caused Linux tests to fail at runtime despite `dllimport` supposedly being ignored on non-Windows platforms). I agree with the comment that the real fix here is to resolve #578, but I wouldn't want it to leave us in a state where only Windows works in the meantime.

src/Qir/Runtime/lib/Simulators/FullstateSimulator.cpp
swernli: I wonder, instead of copying the qubit id vector, couldn't we just pass the buffer directly to the simulator? something like:
`return dumpQubitsToLocation(this->simulatorId, (unsigned)(qubits->count), (unsigned*)(qubits->buffer), callback, location);`
Is there really anything to be gained by converting the ids one at a time instead of all at once?

src/Qir/Runtime/lib/Simulators/ToffoliSimulator.cpp
```diff
@@ -109,11 +110,24 @@ namespace Quantum
...
+        void DumpMachine(const void* location) override
+        {
+            std::cerr << __func__ << " is not yet implemented" << std::endl;
+            // TODO: How to implement: https://github.com/microsoft/qsharp-runtime/pull/634#discussion_r615183791
+        }
+
+        void DumpRegister(const void* location, const QirArray* qubits) override
+        {
+            std::cerr << __func__ << " is not yet implemented" << std::endl;
+            // TODO: How to implement: https://github.com/microsoft/qsharp-runtime/pull/634#discussion_r615183791
```
swernli, line 128: Probably worth separating out into it's own issue specific to the Toffoli simulator. Then you can link the issue here, and link to the PR comment in the issue.

src/Qir/Runtime/public/QirRuntime.hpp
```diff
@@ -66,7 +53,7 @@ extern "C"
     QIR_SHARED_API char* quantum__rt__memory_allocate(uint64_t size); // NOLINT
 
     // Fail the computation with the given error message.
-    QIR_SHARED_API void quantum__rt__fail(QirString* msg); // NOLINT
+    [[noreturn]] QIR_SHARED_API void quantum__rt__fail(QirString* msg); // NOLINT
```
swernli, line 56: I'm not familiar with the `[[noreturn]]` decoration. Does that indicate that the function unconditionally throws exception or asserts or otherwise terminates? Seems handy!

src/Qir/Tests/QIR-static/qir-test-noqsharp.ll
```diff
@@ -9,7 +9,7 @@ declare i8* @__quantum__rt__array_get_element_ptr(%Array*, ...)
 declare i32 @__quantum__rt__array_get_dim(%Array*)
 declare %Array* @__quantum__rt__array_project(%Array*, i32, i64)
 declare void @__quantum__rt__array_update_reference_count(%Array*, i32)
-declare void @DebugLog(i64)
+;declare void @DebugLog(i64)
```
swernli, line 12: nit: can be removed now

src/Qir/Runtime/lib/QIR/bridge-rt.ll
kuzminrobin: I'm not sure I understand.
The `DebugLog[Ptr]()` have relocated to `TestUtils.{h,c}pp`. And I don't see any calls to them (they exist in case anybody needs to use them, but nobody calls/links them at the moment).

swernli: There are commented out calls later in this file, or at least [one here](https://github.com/microsoft/qsharp-runtime/pull/634/files/ecae5566d3bdecd699e88cbe999a4904fef154a6#diff-c9b697f7183ad4664bf2423e5205f04b07fda1ef2654d756872430292a246118R241) that I can't comment on directly because it is too far from a changed line.

swernli: I suppose because the function hasn't been fully removed we could leave the  commented out lines behind, but it looks to me like those were debug lines accidentally left in by earlier investigation, and so there is no reason to leave them in.

swernli (approved):
Some minor comments for you to consider, but overall changes look good to me (though it does look like there is some ongoing discussion with Chris in a few places).

src/Qir/Runtime/lib/Simulators/FullstateSimulator.cpp
kuzminrobin: @swernli, thanks for admitting.
What you are suggesting, was my original attempt. But `dumpQubitsToLocation()` expects as the 3rd argument a pointer to the sequence of `unsigned` (32-bit each), but the `qubits->buffer` is a pointer to the sequence of `void*` (64-bt each). We have to convert.
Some more info on the topic - #631

swernli: ah, I understand now. Yeah, that is annoying, but this appears to be the best way to accomplish it for now. Thanks for explaining!

src/Qir/Runtime/lib/QIR/context.cpp
kuzminrobin: Renamed.

src/Qir/Runtime/lib/QSharpFoundation/qsharp-foundation-qis.ll
kuzminrobin: Nice catch!
I double-checked and it looks like the **qsharp_compiler** part of #578 is already fixed (compiler generates `__quantum__rt__message` instead of `__quantum__qis__message__body`).
I just need to remove `__quantum__qis__message__body` and the declaration you pointed at.

src/Qir/Runtime/lib/Simulators/ToffoliSimulator.cpp
kuzminrobin: Thanks for the hint!  #645. Changing the code to
```c++
        void DumpMachine(const void* location) override
        {
            std::cerr << __func__ << " is not yet implemented" << std::endl;    // #645
        }

        void DumpRegister(const void* location, const QirArray* qubits) override
        {
            std::cerr << __func__ << " is not yet implemented" << std::endl;    // #645
        }
```

src/Qir/Runtime/public/QirRuntime.hpp
kuzminrobin: See [here](https://en.cppreference.com/w/cpp/language/attributes/noreturn).
I believe it affects optimization and likely something else.

src/Qir/Tests/QIR-static/qir-test-noqsharp.ll
kuzminrobin: Hm... I thought I removed that... Removed again ;-)

PR #643, opened 2021-04-20
QEP 2: Enhanced Array Literals

samarsha:
Depends on microsoft/qsharp-compiler#983.


src/Simulation/QSharpFoundation/Microsoft.Quantum.QSharp.Foundation.csproj
```diff
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.Quantum.Sdk">
+<Project Sdk="Microsoft.Quantum.Sdk">
```
swernli, line 1: nit: Something is different with the encoding settings lately, such that this is showing up in other PRs too (see https://github.com/microsoft/qsharp-runtime/pull/634#discussion_r614818183). Hopefully we don't have multiple people with different settings such that these files keep showing up as changed anytime anyone checks in...

swernli (approved):

src/Simulation/QSharpFoundation/Microsoft.Quantum.QSharp.Foundation.csproj
samarsha: Yeah, this PR is removing the BOM. I think BOMs in UTF-8 are discouraged, but there might be some editor that's adding it back in...

cgranade: If I recall, the issue was that one of the bootstrap.ps1 scripts was getting called with Windows PowerShell (5.1 and earlier), which only supports the nonstandard BOM-in-UTF-8 encoding. 6.0 and later added `Uft8NoBom` as an encoding option, but we don't use it everywhere for compatibility with 5.1 clients.

samarsha: Hmm, are older PowerShells not able to interpret BOM-less files as ASCII? The project files are generally limited to ASCII characters (other than a © in AssemblyCommon.props).

cgranade: Good point; it may work to use `Out-File -Encoding Ascii`. I checked, and that is supported on 5.1. If we go that route in bootstrap scripts, it may be worth a quick comment as to why, since otherwise using `Ascii` would be a bit of a red flag.

samarsha: I'm not aware of any PowerShell scripts that write to .csproj files in this repository. Do they exist?

samarsha: Keeping as is since I don't think any script writes to these files; if the BOM comes back from some script or editor, we can consider leaving it there...

PR #647, opened 2021-04-21
Enabling Control Flow Guard for Qir Runtime

swernli:
This enables the Windows security feature Control Flow Guard for our clang compiled native binaries in the Qir Runtime. See here for details: https://msrc-blog.microsoft.com/2020/08/17/control-flow-guard-for-clang-llvm-and-rust/


kuzminrobin (approved):

src/Simulation/Native/CMakeLists.txt
```diff
@@ -36,10 +36,12 @@ option(USE_GATE_FUSION "Use gate fusion" ON)
...
   # always create debug info
   add_definitions("/Zi")
+  add_definitions("/Z7")
 
   # build with no VC runtime depedencies:
-  set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} /MT /Qspectre /guard:cf")
+  set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} /MT /Qspectre /guard:cf /Zi /Z7")
   set(CMAKE_CXX_FLAGS_DEBUG   "${CMAKE_CXX_FLAGS_DEBUG}   /MTd /Qspectre /guard:cf")
+  set(CMAKE_SHARED_LINKER_FLAGS_RELEASE "${CMAKE_SHARED_LINKER_FLAGS_RELEASE} /DEBUG")
 else (MSVC)
```
kuzminrobin, line 45: (Not a part of this PR changes. I don't know CMake and I can be mistaken)

It looks suspicious that `if` and `else` part both have `(MSVC)` in them:

if **(MSVC)**
...
else **(MSVC)**

swernli: yeah, it's an odd convention to me too. Apparently it is optional, but quite common. It's similar to the way you often see comments on namespace close to clarify which scope is ending, as in
```c++
namespace Foo
{
namespace Bar
{
    ...
} // namespace Bar
} // namespace Foo
```

src/Qir/Tests/CMakeLists.txt
```diff
@@ -6,7 +6,7 @@ message(INFO "*** build config: ${CMAKE_BUILD_TYPE}")
 project(qir-tests)
 
 # specify the C++ standard, compiler and other tools
-set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_STANDARD 17)
```
kuzminrobin, line 9: (nit) It is suspicious that C++ version (17) is chosen in more than one place (suspicion is like "Are these all the places where the change is needed?").

swernli: I did a search through the code base for all instances of `set(CMAKE_CXX_STANDARD` and saw that the Native simulator was using C++17 while the Qir Runtime was using 14, so I updated it here in the hopes that it would resolve the issue with control flow guard. So far no luck, such that the BinSkim tool we rely on for binary validation still flags the dlls as invalid. I've filed an issue with the BinSkim team in the hopes that we can find a resolution: https://github.com/microsoft/binskim/issues/365

swernli:
Converting to draft while I continue to work out why this flag isn't working. Both the BinSkim tool and dotPeek report that CFG protections are not enabled, which means either the LLVM support is broken or I am still not properly enabling it somehow.

PR #651, opened 2021-04-22
Explicitly disable docs generation in all test projects.

cgranade:
This PR explicitly disables Q# API documentation generation for unit test projects that do not depend on `Simulators.Test.props` (e.g.: QIR and trace simulation tests), so that test functions and operations are not incorrectly included in documentation.

(+@KittyYeungQ for broken links impact; the documentation that is incorrectly generated from unit tests contains broken links, since there's no corresponding package to document.)


samarsha:
Hmm, I feel like `QSharpDocsOutputPath` being set should not imply `QSharpDocsGeneration=true`, which would prevent this kind of mistake?

ricardo-espinoza (approved):

PR #656, opened 2021-04-26
Create QIR Runtime nuget package

swernli:
Adds the infrastructure for creating the QIR Runtime package with native dependencies (libraries and headers) needed for compiling and executing QIR with the runtime. Note that this package is not ready for publish yet, so upload to the nuget feed is suppressed.

### DIFF
--- a/src/Qir/Runtime/build-qir-runtime.ps1
+++ b/src/Qir/Runtime/build-qir-runtime.ps1
@@ -7,17 +7,22 @@ if (-not (Build-CMakeProject $PSScriptRoot "QIR Runtime")) {
     throw "At least one project failed to compile. Check the logs."
 }
 
-# Copy the results of runtime compilation and the corresponding headers to the drops folder so
+# Copy the results of runtime compilation and the corresponding headers to the QIR drops folder so
 # they can be included in pipeline artifacts.
-$qirDropsFolder = (Join-Path $Env:DROPS_DIR QIR)
-$qirDropsBin = (Join-Path $qirDropsFolder bin)
-$qirDropsInclude = (Join-Path $qirDropsFolder include)
-if (-not (Test-Path $qirDropsFolder)) {
-    New-Item -Path $qirDropsFolder -ItemType "directory"
+$osDir = "win-x64"
+if ($IsLinux) {
+    $osDir = "linux-x64"
+} elseif ($IsMacOS) {
+    $osDir = "osx-x64"
+}
+$qirDropsBin = (Join-Path $Env:QIR_DROPS bin $osDir native)
+$qirDropsInclude = (Join-Path $Env:QIR_DROPS include)
+if (-not (Test-Path $Env:QIR_DROPS)) {
+    New-Item -Path $Env:QIR_DROPS -ItemType "directory"
     New-Item -Path $qirDropsBin -ItemType "directory"
     New-Item -Path $qirDropsInclude -ItemType "directory"
 }
 $qirBinaries = (Join-Path $PSScriptRoot build $Env:BUILD_CONFIGURATION bin *)
 $qirIncludes = (Join-Path $PSScriptRoot public *)
-Copy-Item $qirBinaries $qirDropsBin
-Copy-Item $qirIncludes $qirDropsInclude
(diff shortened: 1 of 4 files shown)


src/Qir/Runtime/build-qir-runtime.ps1
```diff
@@ -7,17 +7,22 @@ if (-not (Build-CMakeProject $PSScriptRoot "QIR Runtime")) {
...
-$qirDropsBin = (Join-Path $qirDropsFolder bin)
-$qirDropsInclude = (Join-Path $qirDropsFolder include)
-if (-not (Test-Path $qirDropsFolder)) {
-    New-Item -Path $qirDropsFolder -ItemType "directory"
+$osDir = "win-x64"
+if ($IsLinux) {
+    $osDir = "linux-x64"
+} elseif ($IsMacOS) {
+    $osDir = "osx-x64"
+}
```
kuzminrobin, line 17: (Minor)
We don't build on the same machine for different platforms, do we? For example, on the same machine we don't create the dirs "win-x64" **and** "linux-x64". All these dirs ("win-x64", "linux-x64", "osx-x64") are mutually exclusive.
What if we just skip them? (skip everything related to `$osDir`)

kuzminrobin (approved):

src/Qir/Runtime/build-qir-runtime.ps1
swernli: There's actually an interesting and valuable subtlety here that makes this folder structure critical. When our e2e pipeline runs, it builds all native components (simulator and QIR Runtime) across all three platforms. Each platform creates a folder matching the names in this section of code, and all three get uploaded to the same artifacts, such that the pipeline artifacts after that step completes have this folder structure:
```
xplat/src/Qir/drops/
    bin/
        linux-x64/
        osx-x64/
        win-x64/
    include/
```
You can see this in one of the manual e2e builds I triggered to test these changes combined with some pipeline yml changes: https://ms-quantum.visualstudio.com/Quantum%20Program/_build/results?buildId=138114&view=artifacts&pathAsName=false&type=publishedArtifacts. Then a later pipeline step will copy down the contents of `xplat` and put them back into the source tree, such that when the rest of the pipeline runs on Windows, it will have the binaries from all three platforms in the `src/Qir/drops/bin` folder. Then the [Microsoft.Quantum.Qir.Runtime.nuspec](https://github.com/microsoft/qsharp-runtime/blob/e9f55f9ea7dfd2f9f2b7ea82cafe67ef955a282a/src/Qir/Runtime/Microsoft.Quantum.Qir.Runtime.nuspec#L19) has a line that copies the contents of the `bin` folder into the package `runtimes` folder:
```xml
<file src="..\drops\bin\**\*" target="runtimes" />
```
This means the package will also have the platform specific folders inside of it. Later, if this package is included as a dependency in a dotnet project, loading the library by name will automatically include the folder under runtimes that matches the Runtime Identifier of the current platform, as described in [.NET RID Catalog](https://docs.microsoft.com/en-us/dotnet/core/rid-catalog#:~:text=RID%20is%20short%20for%20Runtime%20Identifier.%20RID%20values,packages%20to%20represent%20platform-specific%20assets%20in%20NuGet%20packages.). So a project can use [NativeLibrary.Load](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.nativelibrary.load?view=net-5.0#System_Runtime_InteropServices_NativeLibrary_Load_System_String_System_Reflection_Assembly_System_Nullable_System_Runtime_InteropServices_DllImportSearchPath__) without needing to specify the full path to the dll, like this:
```csharp
var qsharpFoundationLibrary = NativeLibrary.Load("Microsoft.Quantum.Qir.QSharp.Foundation", Assembly.GetExecutingAssembly(), null);
```
And that will automatically know to load `runtimes/win-x64/native/Microsoft.Quantum.Qir.QSharp.Foundation.dll` on Windows and instead load `runtimes/linux-x64/native/libMicrosoft.Quantum.Qir.QSharp.Foundation.so` on Linux.

kuzminrobin: Thank you for explaining. Very helpful.

PR #658, opened 2021-04-27
QIR Build From Q# DLL

ScottCarda-MS:
Fills out the `BuildFromQSharDll` function on the Qir.Tools API.

Depends on compiler PR: https://github.com/microsoft/qsharp-compiler/pull/996


src/Qir/Execution/Tools/QirTools.cs
```diff
@@ -19,9 +24,22 @@ public static class QirTools
         /// <param name="qsharpDll">.NET DLL generated by the Q# compiler.</param>
         /// <param name="libraryDirectory">Directory where the libraries to link to are located.</param>
         /// <param name="includeDirectory">Directory where the headers needed for compilation are located.</param>
-        public static Task BuildFromQSharpDll(FileInfo qsharpDll, DirectoryInfo libraryDirectory, DirectoryInfo includeDirectory)
+        public static async Task BuildFromQSharpDll(FileInfo qsharpDll, DirectoryInfo libraryDirectory, DirectoryInfo includeDirectory)
```
cesarzc, line 27: Thinking about the signature of this method more closely, there's no way for the caller to know where the executables were placed. Maybe the following signature would be better:
```csharp
public static async Task<IList<FileInfo>> BuildFromQSharpDll(FileInfo qsharpDll, DirectoryInfo libraryDirectory, DirectoryInfo includeDirectory, DirectoryInfo executablesDirectory);
```
Where: 
- The method returns a list of the executables created.
- It receives a directory to place the executables as argument.

src/Qir/Execution/Tools/QirTools.cs
```diff
@@ -12,16 +16,41 @@ namespace Microsoft.Quantum.Qir.Tools
...
+
+            if (!executablesDirectory.Exists)
+            {
+                executablesDirectory.Create();
+            }
+
+            var dirStr = executablesDirectory.FullName + Path.DirectorySeparatorChar;
+            var tasks = EntryPointOperationLoader.LoadEntryPointOperations(qsharpDll).Select(entryPointOp =>
+            {
+                var exeFileInfo = new FileInfo(dirStr + entryPointOp.Name);
```
cesarzc, line 46:
```suggestion
                var exeFileInfo = new FileInfo(Path.Combine(executablesDirectory.FullName, entryPointOp.Name);
```

src/Qir/Execution/Tools/QirTools.cs
```diff
@@ -12,16 +16,41 @@ namespace Microsoft.Quantum.Qir.Tools
...
+            {
+                throw new ArgumentException("The given DLL does not contain QIR byte code.");
+            }
+
+            if (!executablesDirectory.Exists)
+            {
+                executablesDirectory.Create();
+            }
+
+            var dirStr = executablesDirectory.FullName + Path.DirectorySeparatorChar;
```
cesarzc, line 43: With the suggestion using `Path.Combine`, this variable is no longer needed.

ScottCarda-MS: removed

src/Qir/Execution/Tools/QirTools.cs
ScottCarda-MS: done

src/Qir/Execution/Tools/QirTools.cs
ScottCarda-MS: I've implemented the executable directory parameter. From a conversation with you offline, it sounds like we want to wait on returning a list of FileInfos for the created executables.

cesarzc: The part to return the list of executables will be done once the `BuildAsync` method returns the `FileInfo` of the executable.

cesarzc (approved):

PR #661, opened 2021-04-28
Publish QIR Runtime package as experimental

swernli:
This change updates the name of the QIR Runtime package to `Microsoft.Quantum.Qir.Runtime.Experimental` in order to further underscore that the API is in-progress and likely to change. It also removes the workaround that prevented the nuget package from being published since #647 fixed the CFG issues. This will allow our early workloads to begin taking a dependency on this package from the qdk-alpha feed.

### DIFF
--- a/src/Qir/Runtime/lib/Tracer/CMakeLists.txt
+++ b/src/Qir/Runtime/lib/Tracer/CMakeLists.txt
@@ -27,7 +27,7 @@ target_link_libraries(Microsoft.Quantum.Qir.Tracer
   "-L${CMAKE_BINARY_DIR}/lib/QIR"
   -lMicrosoft.Quantum.Qir.Runtime
 )
-add_dependencies(Microsoft.Quantum.Qir.QSharp.Foundation Microsoft.Quantum.Qir.Runtime)
+add_dependencies(Microsoft.Quantum.Qir.Tracer Microsoft.Quantum.Qir.Runtime)
 
 target_include_directories(Microsoft.Quantum.Qir.Tracer PUBLIC ${includes})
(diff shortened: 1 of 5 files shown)


src/Qir/Runtime/public/QirContext.hpp
```diff
@@ -7,6 +7,8 @@
 
 #include "CoreTypes.hpp"
 
+    extern "C" QIR_SHARED_API void InitializeQirContextEx(IRuntimeDriver* driver, bool trackAllocatedObjects = false);
```
swernli, line 10: Oops.., this change was not supposed to be included. Removing...

src/Qir/Runtime/lib/Tracer/CMakeLists.txt
```diff
@@ -27,7 +27,7 @@ target_link_libraries(Microsoft.Quantum.Qir.Tracer
   "-L${CMAKE_BINARY_DIR}/lib/QIR"
   -lMicrosoft.Quantum.Qir.Runtime
 )
-add_dependencies(Microsoft.Quantum.Qir.QSharp.Foundation Microsoft.Quantum.Qir.Runtime)
+add_dependencies(Microsoft.Quantum.Qir.Tracer Microsoft.Quantum.Qir.Runtime)
```
swernli, line 30: This is just a minor bug fix. The first argument to `add_dependencies` is supposed to be the name of the target that you want to declare dependencies for. In this case, copy-paste mistake had the wrong name here, so it was possible for the tracer to sometimes build before the runtime, resulting in a failure to find the dependent dll. This was causing intermittent failures in the QIR Runtime build in the pipeline.

kuzminrobin (approved):

cesarzc (approved):

swernli:
This PR is superseded by #663

PR #662, opened 2021-04-28
Updating devcontainer to include Mono+Nuget

swernli:
This change updates the devcontainer setup to include Mono+Nuget, as well as common extensions and tools. This matches the set up already used in qsharp-compiler.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -10,11 +10,14 @@ RUN apt-get -y update && \
     apt-get clean && rm -rf /var/lib/apt/lists/
 
 # Install CMake, Ninja, LLVM/Clang tools
-RUN apt update \
-    && apt-get install -y cmake \
-    && apt-get install -y ninja-build \
-    && apt-get install -y clang-11 \
-    && apt-get install -y clang-tidy-11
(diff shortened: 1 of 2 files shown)


.devcontainer/Dockerfile
```diff
@@ -1,8 +1,20 @@
...
+    apt-get -y update && \
+    apt-get -y install --no-install-recommends mono-complete nuget && \
+    apt-get clean && rm -rf /var/lib/apt/lists/
+
+# Install CMake, Ninja, LLVM/Clang tools
 RUN apt update \
     && apt-get install -y cmake \
     && apt-get install -y ninja-build \
     && apt-get install -y clang-11 \
     && apt-get install -y clang-tidy-11
```
cgranade, line 17: For reducing layer sizes, I'd recommend clearing the apt cache in the same layer as `apt update`. That does slow down building the devcontainer, but saves a lot of disk space.

cgranade (approved):

.devcontainer/Dockerfile
swernli: Good suggestion! I'll update this to incorporate clearing the cache.

PR #663, opened 2021-04-29
Force QIR package to have a pre-release version

anpaz:
Currently, the QIR package is not ready to be officially released, so it will always build with a `-alpha` version to mark it as a pre-release package.

### DIFF
--- a/build/pack.ps1
+++ b/build/pack.ps1
@@ -101,7 +101,7 @@ Pack-Dotnet '../src/Simulation/Type3Core/Microsoft.Quantum.Type3.Core.csproj'
 Pack-One '../src/Simulation/Simulators/Microsoft.Quantum.Simulators.nuspec'
 Pack-One '../src/Quantum.Development.Kit/Microsoft.Quantum.Development.Kit.nuspec'
 Pack-One '../src/Xunit/Microsoft.Quantum.Xunit.csproj'
-Pack-One '../src/Qir/Runtime/Microsoft.Quantum.Qir.Runtime.nuspec' -ForcePrelease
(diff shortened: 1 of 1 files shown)


cgranade (approved):

swernli (approved):
This is great, thanks for doing it so quickly!

swernli:
Hmm... I'm seeing the same samples failure on a different PR, and ran an e2e build with not changes (just normal main) and got them there too. I think this may be fallout from https://github.com/microsoft/Quantum/pull/478, which enabled QIR Generation for a subset of the samples. I didn't notice it before, but that PR to samples only ran local build and not e2e, so it's possible something about the changes is not compatible with the e2e build. We should investigate, but to get other PRs unblocked in the meantime it might be worth reverting that PR. 
@ScottCarda-MS @cesarzc FYI

swernli:
@anpaz One detail I just noticed: the change is working as expected, but that means that for PR builds like this, the `-pull` gets overridden as `-alpha` as well. Would it be worth making an exception for that, or does it matter less because the PR pipeline artifacts don't get pushed to any nuget feeds? The only folks this would affect are those who pull down artifacts from PR builds to use in a local nuget source, in which case they would need to remember that this one package is always `-alpha` even when from a PR, which doesn't seem unreasonable to me. Any thoughts?

anpaz:
Yes, also, the update version script assumes all packages follow the same convention, so this will be problematic for project that has the Qir.Runtime package listed as a dependency. Let me think about it.

swernli:
Good point, I hadn't thought about the version update script. For now, nothing uses this package as a dependency, so it won't be a problem immediately, but I expect that to change soon as we start finding more ways to utilize the package including from within dotnet languages.

PR #665, opened 2021-04-30
Add test case and fix test script

achocron:


src/Qir/Controller/test-qir-controller.ps1
```diff
@@ -80,22 +82,21 @@ Foreach-Object {
...
-        Write-Host "##[info]Expected error:"
-        Write-Host $expectedError
-        Write-Host "##[info]Actual error:"
-        Write-Host $actualError
-        $script:all_ok = $False
-        continue
     }
     else {
-        Write-Host "##[info]Test case '$($_.BaseName)' passed"
+        $expectedError = Get-Content -Path $expectedErrorFile -Raw
```
cesarzc, line 87: An error file can exist even when an output file exists. I think the right thing to do would be to treat the error file in a similar way than the output file. Check if a expected error file is present, if so, check if it is equal to the actual error file. If not, no error file should have been produced.

src/Qir/Execution/Tools/Executable/ClangClient.cs
```diff
@@ -23,13 +24,15 @@ public async Task CreateExecutableAsync(string[] inputFiles, string[] libraries,
 
             // string.Join does not automatically prepend the delimiter, so it is included again in the string here.
             var librariesArg = $"{LinkFlag} {string.Join(LinkFlag, libraries)}";
-            var arguments = $"{inputsArg} -I {includePath} -L {libraryPath} {librariesArg} -o {outputPath}";
+            var arguments = $"{inputsArg} -I {includePath} -L {libraryPath} {librariesArg} -o {outputPath} -std=c++17 -v";
             logger.LogInfo($"Invoking clang with the following arguments: {arguments}");
             var taskCompletionSource = new TaskCompletionSource<bool>();
             using var process = new Process();
+            Environment.SetEnvironmentVariable("DYLD_LIBRARY_PATH", libraryPath);
```
cesarzc, line 31: I believe this is only relevant for execution, so we should remove this from here and maybe add it into the execution section. I am hesitant to set environment variables but we might not have a way around it.

achocron: I think the argument can be made for having it in the clang client as it is rather specific to c++ execution and the very libraries that must be set in the environment variable are known at link-time.

cesarzc (approved):

PR #667, opened 2021-05-03
Update QIR tests for latest calling conventions

swernli:
This change updates the calling conventions used in the QIR Runtime tests to use `__Interop` wrappers for calling functions. This also required updating some older tests to the multiple entry point strategy for testing so that each test function is considered an externally callable operation/function. In addition, this updates the CMakeLists.txt to consume the bitcode files instead of the human readable .ll files wherever possible.

### DIFF
--- a/src/Qir/Tests/QIR-static/qir-test-ouput.cpp
+++ b/src/Qir/Tests/QIR-static/qir-test-ouput.cpp
@@ -4,11 +4,10 @@
 
 #include "catch.hpp"
 
-#include "QirTypes.hpp"
 #include "QirRuntime.hpp"
 #include "OutputStream.hpp"
 
-extern "C" void Microsoft__Quantum__Testing__QIR__Out__MessageTest__body(void*); // NOLINT
+extern "C" void Microsoft__Quantum__Testing__QIR__Out__MessageTest__Interop(const char[]); // NOLINT
 
 TEST_CASE("QIR: Out.Message", "[qir.Out][qir.Out.Message]")
 {
@@ -22,10 +21,8 @@ TEST_CASE("QIR: Out.Message", "[qir.Out][qir.Out.Message]")
         Microsoft::Quantum::OutputStream::ScopedRedirector qOStreamRedirector(outStrStream);
 
         // Log something (to the redirected output):
-        QirString qstr{std::string(testStr1)};
-        Microsoft__Quantum__Testing__QIR__Out__MessageTest__body(&qstr);
-        qstr.str = testStr2;
-        Microsoft__Quantum__Testing__QIR__Out__MessageTest__body(&qstr);
+        Microsoft__Quantum__Testing__QIR__Out__MessageTest__Interop(testStr1.c_str());
+        Microsoft__Quantum__Testing__QIR__Out__MessageTest__Interop(testStr2.c_str());
 
     } // Recover the output stream.
 
--- a/src/Qir/Tests/QIR-static/qsharp/qir-test-arrays.qs
+++ b/src/Qir/Tests/QIR-static/qsharp/qir-test-arrays.qs
@@ -8,7 +8,7 @@ namespace Microsoft.Quantum.Testing.QIR {
     open Microsoft.Quantum.Testing.QIR.Out;
 
     @EntryPoint()
-    operation Test_Arrays(array : Int[], index : Int, val : Int, compilerDecoy : Bool) : Int {
+    operation Test_Arrays(array : Int[], index : Int, val : Int) : Int {
         // exercise __quantum__rt__array_copy
         mutable local = array;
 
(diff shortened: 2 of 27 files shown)


kuzminrobin:
(nit) This is the first time I hear about renaming to `__Interop`. Would be nice to see a link to a task that discusses or provides more details about that renaming.

swernli:
@kuzminrobin This wasn't a rename so much as a switch in pattern where the expectation is that callers should use the interop functions we generate instead of trying to rely on QIR specific types. The `__Interop` wrappers were introduced as part of the initial QIR generation work (see https://github.com/microsoft/qsharp-compiler/pull/889) but we didn't enforce their usage because we were lacking multiple entry point support. Now that this is done, we can call into these instead and move forward with marking all non-entrypoint functions as internal.

src/Qir/Tests/QIR-static/qir-test-ouput.cpp
```diff
@@ -22,10 +21,8 @@ TEST_CASE("QIR: Out.Message", "[qir.Out][qir.Out.Message]")
         Microsoft::Quantum::OutputStream::ScopedRedirector qOStreamRedirector(outStrStream);
 
         // Log something (to the redirected output):
-        QirString qstr{std::string(testStr1)};
-        Microsoft__Quantum__Testing__QIR__Out__MessageTest__body(&qstr);
-        qstr.str = testStr2;
-        Microsoft__Quantum__Testing__QIR__Out__MessageTest__body(&qstr);
+        Microsoft__Quantum__Testing__QIR__Out__MessageTest__Interop(testStr1.c_str());
+        Microsoft__Quantum__Testing__QIR__Out__MessageTest__Interop(testStr2.c_str());
```
kuzminrobin, line 25: Wow! Rather dramatic change.

Before this change the `QirString*` was passed to the `Microsoft__Quantum__Testing__QIR__Out__MessageTest__body()`, 
the `Microsoft__Quantum__Testing__QIR__Out__MessageTest__body()` was passing that pointer to the `__quantum__rt__message()` (in the generated .ll), and `__quantum__rt__message()` was handling that `QirString*`.

Now after this change I see that the `const char*` is passed instead. But in this PR I don't see any changes in `__quantum__rt__message()`. This makes me think that in the generated .ll the function `Microsoft__Quantum__Testing__QIR__Out__MessageTest__Interop()` creates the instance of `QirString` and passes the pointer to it to the `__quantum__rt__message()`. Which means that the Q# compiler has also changed.
But I don't see that this PR changes the compiler package version.

So for now I have an impression that this test after this change started behaving differently. If the test still _succeeds_ then I would investigate why.

src/Qir/Tests/QIR-static/qsharp/qir-test-arrays.qs
```diff
@@ -31,40 +31,6 @@ namespace Microsoft.Quantum.Testing.QIR {
...
-            let res20 = PauliArrayAsIntTest();
-            let res21 = PauliArrayAsIntFailTest();
-            let res22 = TestDrawRandomDouble(0.0, 1.0);
-            MessageTest("Test");
-
-            // Conditionals:
-            TestApplyIf();
-            TestApplyIfWithFunctors();
-            TestApplyConditionally();
-        }
```
kuzminrobin, line 67: Nice change! Thanks!

src/Qir/Tests/QIR-static/qir-test-ouput.cpp
swernli: The interop functions that do the conversion from C types to QIR typles have been produced since the first version of the QIR was merged to main (see https://github.com/microsoft/qsharp-compiler/pull/889 for the original PR that added this to feature branch). We really should have been using these interop functions all along, but we were also waiting for multiple entrypoint support (added in https://github.com/microsoft/qsharp-compiler/pull/881). I am finally circling back to these call points so that we can follow our own guidance, which is to use the `__Interop` functions everywhere and not call into any `__body` functions.

kuzminrobin: Can you please post here the body of the generated `Microsoft__Quantum__Testing__QIR__Out__MessageTest__Interop()`?

I assume that the following functions (called from `Microsoft__Quantum__Testing__QIR__Out__MessageTest__Interop()`) are still the same:
```ll
define dllexport void @__quantum__rt__message(%String* %.str) {
  %str = bitcast %String* %.str to %"struct.QirString"*
  call void @quantum__rt__message(%"struct.QirString"* %str)
  ret void
}
```
```c++
    void quantum__rt__message(QirString* qstr)   // NOLINT
    {
        Microsoft::Quantum::OutputStream::Get() << qstr->str << std::endl;
    }
```

swernli: Yup, this is all the same IR that has been generated since March, we just haven't been using it to it's full extent. Here is what the interop version of MessageTest looks like:
```ll
define void @Microsoft__Quantum__Testing__QIR__Out__MessageTest__Interop(i8* %msg) #0 {
entry:
  %0 = call %String* @__quantum__rt__string_create(i8* %msg)
  call void @Microsoft__Quantum__Testing__QIR__Out__MessageTest__body(%String* %0)
  call void @__quantum__rt__string_update_reference_count(%String* %0, i32 -1)
  ret void
}
```

kuzminrobin (approved):

PR #668, opened 2021-05-03
Azure QIR submission in entry point driver

samarsha:
Missing functionality:
* `IQirSubmitter` doesn't have support for dry runs (validation only) or a submission context (friendly name, number of shots).
* No non-mock implementations of `IQirSubmitter` exist yet, so there is no factory.


src/Simulation/CSharpGeneration/EntryPoint.fs
```diff
@@ -24,7 +23,7 @@ type private Parameter =
 let entryPointClassName = "__QsEntryPoint__"
 
 /// The namespace containing the non-generated parts of the entry point driver.
-let private driverNamespace = "Microsoft.Quantum.EntryPointDriver"
+let private driverNamespace = "global::Microsoft.Quantum.EntryPointDriver"
 
 /// A sequence of all of the named parameters in the argument tuple and their respective C# and Q# types.
 let rec private parameters context doc = function
```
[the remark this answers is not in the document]
samarsha: Yes, `entryPointClass` calls the `parameters` function. The parameters are flattened by `Seq.concat` in the recursive case. The resulting sequence goes from left-to-right (A, B, C, D).

```fsharp
| QsTuple items -> items |> Seq.map (parameters context doc) |> Seq.concat
```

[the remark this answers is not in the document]
cesarzc: Thanks for the explanation Sarah!

src/Simulation/EntryPointDriver/Azure/Azure.cs
```diff
@@ -0,0 +1,292 @@
...
+{
+    using Environment = System.Environment;
+
+    /// <summary>
+    /// Provides entry point submission to Azure Quantum.
+    /// </summary>
+    public static class Azure
+    {
+        /// <summary>
+        /// Submits an entry point to Azure Quantum. If <paramref name="qirSubmission"/> is non-null and a QIR submitter
```
cesarzc, line 24: How are `settings` populated? How is it determined if a QIR submitter is available for the target?

samarsha: `AzureSettings` is created by System.CommandLine using some reflection magic that sets the properties based on comparing the property name to the command line option names.

https://github.com/microsoft/qsharp-runtime/blob/7cff67b6bbcc6ebc1509b634b1dbceb21a62c6db/src/Simulation/EntryPointDriver/Driver.cs#L409-L412

The availability of a QIR submitter is determined by `Azure.QirSubmitter` returning a non-null value, which will be determined by the return value of the QIR submitter factory once that is implemented. That's what the chain of `if`s here do - they try each submitter/machine type in order until one returns non-null.

https://github.com/microsoft/qsharp-runtime/blob/7cff67b6bbcc6ebc1509b634b1dbceb21a62c6db/src/Simulation/EntryPointDriver/Azure/Azure.cs#L39-L55

samarsha:
@cesarzc, I made the changes we talked about to QIR detection and improved the error message when QIR is missing.

src/Simulation/EntryPointDriver/Azure/Azure.cs
cesarzc: Thanks for the explanation Sarah!

cesarzc (approved):
Looks good, thanks Sarah!

PR #670, opened 2021-05-04
Microsoft.Quantum.Qir.Tools package

cesarzc:


swernli:
There is one other file to update that does some tracking of published dlls and packages: https://github.com/microsoft/qsharp-runtime/blob/feature/azure-quantum-simulator/build/manifest.ps1. It doesn't affect whether or not the package will get published, as you saw from your e2e build, but I think it gets used for some of the signing and validation that happens as part of the release pipeline, so it's a good idea to keep it in sync.

cesarzc:
> 
> 
> There is one other file to update that does some tracking of published dlls and packages: https://github.com/microsoft/qsharp-runtime/blob/feature/azure-quantum-simulator/build/manifest.ps1. It doesn't affect whether or not the package will get published, as you saw from your e2e build, but I think it gets used for some of the signing and validation that happens as part of the release pipeline, so it's a good idea to keep it in sync.

@swernli Added the new package and DLL to the appropriate places in the file.

swernli (approved):

PR #672, opened 2021-05-04
Fix bug in joint measurement decomposition

swernli:
This fixes a bug in the target package decomposition for joint measurement.

To Do:

- [x] Add joint measurement tests for QSharp.Core and decompositions

### DIFF
--- a/src/Simulation/TargetDefinitions/Decompositions/Measure.qs
+++ b/src/Simulation/TargetDefinitions/Decompositions/Measure.qs
@@ -54,9 +54,9 @@ namespace Microsoft.Quantum.Intrinsic {
             use q = Qubit() {
                 H(q);
                 for k in 0 .. Length(bases) - 1 {
-                    if( bases[k] == PauliX ) { Controlled X ([qubits[k]], q); }
-                    if( bases[k] == PauliZ ) { Controlled Z ([qubits[k]], q); }
-                    if( bases[k] == PauliY ) { Controlled Y ([qubits[k]], q); }
+                    if( bases[k] == PauliX ) { Controlled X ([q], qubits[k]); }
+                    if( bases[k] == PauliZ ) { Controlled Z ([q], qubits[k]); }
+                    if( bases[k] == PauliY ) { Controlled Y ([q], qubits[k]); }
                 }
                 H(q);
                 set res =  M(q);
--- a/src/Simulation/Simulators.Tests/Circuits/JointMeasureTest.qs
+++ b/src/Simulation/Simulators.Tests/Circuits/JointMeasureTest.qs
@@ -4,38 +4,144 @@
 namespace Microsoft.Quantum.Simulation.Simulators.Tests.Circuits {
     
     open Microsoft.Quantum.Intrinsic;
-    
-    
-    operation JointMeasureTest () : Unit {
-        
-        
-        using (qubits = Qubit[3]) {
-            let q1 = qubits[0];
-            let q2 = qubits[1];
-            let q3 = qubits[2];
-            let basis = [PauliZ, PauliZ, PauliZ];
-            let r1 = Measure(basis, qubits);
-            AssertEqual(Zero, r1);
-            X(q1);
-            let r2 = Measure(basis, qubits);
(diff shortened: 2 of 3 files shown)


cgranade (approved):
I think this looks correct, yes. Is it possible to add regression check to make sure that the results agree with `Measure` on simulators that support joint measurements directly without decomposition? If not, we may still want to add tests that (e.g.:) measuring parity of Bell states, and GHZ states work correctly without collapsing the state of their input registers (using AssertProbability to check tomographically complete measurements after a call to `Measure`). Let me know how I can help get those up and running.

I think @robinkuzmin was looking at writing some similar unit tests as well, so we may be able to combine some effort there, too.

swernli:
@kuzminrobin, as this overlaps with your ongoing effort to get some joint measurement tests for QIR.

src/Simulation/TargetDefinitions/Decompositions/Measure.qs
```diff
@@ -54,9 +54,9 @@ namespace Microsoft.Quantum.Intrinsic {
...
                 for k in 0 .. Length(bases) - 1 {
-                    if( bases[k] == PauliX ) { Controlled X ([qubits[k]], q); }
-                    if( bases[k] == PauliZ ) { Controlled Z ([qubits[k]], q); }
-                    if( bases[k] == PauliY ) { Controlled Y ([qubits[k]], q); }
+                    if( bases[k] == PauliX ) { Controlled X ([q], qubits[k]); }
+                    if( bases[k] == PauliZ ) { Controlled Z ([q], qubits[k]); }
+                    if( bases[k] == PauliY ) { Controlled Y ([q], qubits[k]); }
                 }
                 H(q);
                 set res =  M(q);
```
kuzminrobin, line 62: There is still a lot for me to study and practice in, regarding the measurement, it is a tricky topic (at least the multi-qubit one). But this particular change seemed to me suspicious at the first sight. ;-) I'm sure I'm trying to raise a false-alarm but I will still share.

We allocate the qubit `q` (line 54). It is in |0> state.
We take it to the equally-probable state/superposition of |0> and |1> (line 55).
Then we do the _read-only_ operations on `q` (green lines 57 - 59). We _do not update_ `q`.
Then we take `q` back to |0> (line 61).
And finally we measure |0> and return that.

I have a naïve impression that the same can be done with the following lines :-D
```q#
use q = Qubit();
return M(q);
```

kuzminrobin (approved):

src/Simulation/TargetDefinitions/Decompositions/Measure.qs
cgranade: @kuzminrobin: In general, controlled operations can affect both the control and target qubits. For example, the following two Q# programs are equivalent (up to hardware noise):

```qsharp
within {
    H(ctrl);
    H(target);
} apply {
    CNOT(ctrl, target);
}
// is equivalent to
CNOT(target, ctrl);
```

In the above lines 57–59, `H(q)` is used to reverse the direction of each controlled operation, as the decomposition above is much easier to derive in the Hadamard basis than in the computational basis. To see that, we need one more trick from the stabilizer formalism; namely, that `X` operations propagate "down" CNOTs while `Z` calls propagate "up":

```qsharp
// Xs propagate "down"
X(ctrl);
CNOT(ctrl, target);
// is equivalent to:
CNOT(ctrl, target);
X(ctrl);
X(target);

// Zs propagate "up"
Z(target);
CNOT(ctrl, target);
// is equivalent to
CNOT(ctrl, target);
Z(ctrl);
Z(target);
```

Using that trick, the above decomposition is used to ensure that operations that anticommute with the measurement basis propagate "up" each controlled operation as an `X`, so that they can flip the measurement at line 62.

Based on that analysis, I believe this PR is correct as written (though additional unit test coverage would still be helpful!).

kuzminrobin: Thank you very much for explaining, Chris!
(but for now I'm still not mature enough to fully understand that ;-) Nonetheless I will likely be getting back to this from time to time)

PR #673, opened 2021-05-05
Add IQSharpSubmitter and mark IQuantumMachine and related types as obsolete

samarsha:
* No factory for IQSharpSubmitter (or IQirSubmitter) yet since there are no concrete implementations.
* Naming oddity: IQuantumMachineJob is used by the new submitters. Not sure if it's worth renaming this, or what to rename it to. AzureJob? AzureQuantumJob?

### DIFF
--- a/src/Simulation/Core/Submitters/IQSharpSubmitter.cs
+++ b/src/Simulation/Core/Submitters/IQSharpSubmitter.cs
@@ -1,0 +1,38 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+#nullable enable
+
+using Microsoft.Quantum.Simulation.Core;
+using System.Threading.Tasks;
+
+namespace Microsoft.Quantum.Runtime.Submitters
+{
+    /// <summary>
+    /// An interface for submitting Q# programs to Azure.
+    /// </summary>
+    public interface IQSharpSubmitter : IAzureSubmitter
+    {
+        /// <summary>
+        /// Submits a job to execute a Q# program without waiting for execution to complete.
+        /// </summary>
+        /// <typeparam name="TIn">The entry point argument type.</typeparam>
+        /// <typeparam name="TOut">The entry point return type.</typeparam>
+        /// <param name="entryPoint">The entry point information for the submitted program.</param>
+        /// <param name="argument">The argument to the entry point.</param>
+        /// <param name="options">Additional options for the submission.</param>
+        /// <returns>The submitted job.</returns>
+        Task<IQuantumMachineJob> SubmitAsync<TIn, TOut>(
+            EntryPointInfo<TIn, TOut> entryPoint, TIn argument, SubmissionOptions options);
+
+        /// <summary>
+        /// Validates a Q# program for execution on Azure Quantum.
+        /// </summary>
+        /// <typeparam name="TIn">The entry point argument type.</typeparam>
+        /// <typeparam name="TOut">The entry point return type.</typeparam>
+        /// <param name="entryPoint">The entry point information for the submitted program.</param>
+        /// <param name="argument">The argument to the entry point.</param>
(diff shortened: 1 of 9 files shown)


src/Simulation/Core/Submitters/IQSharpSubmitter.cs
```diff
@@ -0,0 +1,38 @@
...
+        /// <summary>
+        /// Submits a job to execute a Q# program without waiting for execution to complete.
+        /// </summary>
+        /// <typeparam name="TIn">The entry point argument type.</typeparam>
+        /// <typeparam name="TOut">The entry point return type.</typeparam>
+        /// <param name="entryPoint">The entry point information for the submitted program.</param>
+        /// <param name="argument">The argument to the entry point.</param>
+        /// <param name="options">Additional options for the submission.</param>
+        /// <returns>The submitted job.</returns>
+        Task<IQuantumMachineJob> SubmitAsync<TIn, TOut>(
```
cesarzc, line 25: I think this interface should also provide an option without the submission options.

src/Simulation/Core/Submitters/IQirSubmitter.cs
```diff
@@ -20,7 +20,9 @@ public interface IQirSubmitter : IAzureSubmitter
         /// <param name="qir">The QIR program as a byte stream.</param>
         /// <param name="entryPoint">The fully-qualified name of the entry point to execute.</param>
         /// <param name="arguments">The arguments to the entry point in the order in which they are declared.</param>
+        /// <param name="options">Additional options for the submission.</param>
         /// <returns>The submitted job.</returns>
-        Task<IQuantumMachineJob> SubmitAsync(Stream qir, string entryPoint, IReadOnlyList<Argument> arguments);
+        Task<IQuantumMachineJob> SubmitAsync(
```
cesarzc, line 25: Similarly to [this](https://github.com/microsoft/qsharp-runtime/pull/673/files#r626707394) comment, I think this interface should also provide an option without the submission options.

src/Simulation/Core/Submitters/IQSharpSubmitter.cs
samarsha: What is the behavior when the options are not provided? Is it specific to each submitter instance, or are there defaults that can be used across all submitters?

I think it would also be reasonable to allow e.g. the default friendly name, but a custom number of shots, or vice versa. So we could have a default instance of SubmissionOptions as a static property there, which would make the submitter APIs something like:

```csharp
submitter.SubmitAsync(entryPoint, argument, SubmissionOptions.Default);
submitter.SubmitAsync(entryPoint, argument, SubmissionOptions.Default.With(shots: 1000));
submitter.SubmitAsync(entryPoint, argument, SubmissionOptions.Default.With(friendlyName: "foo", shots: 1000));
```

(This would work better with a record, but we don't have C# 9 yet.)

samarsha: If necessary, we could provide a `submitter.SubmitAsync(entryPoint, argument)` that automatically used `SubmissionOptions.Default`. But I am wary of overloads and default parameters, so I would prefer to keep `SubmissionOptions.Default` explicit in the calling code.

samarsha: I added a default instance for SubmissionOptions, where the friendly name is empty string, and the number of shots is 500.

cesarzc: I really liked your idea of a default instance for `SubmissionOptions`. It makes the interface simpler too because we don't have a bunch of overloads.

src/Simulation/Core/Submitters/IQirSubmitter.cs
cesarzc: Closing given that now you provide a default instance for `SubmissionOptions`.

cesarzc (approved):

PR #677, opened 2021-05-06
Better reuse of qubits within block

DmitryVasilevsky:
Better reuse of qubits within one block.
When multiple qubits are released at the same time, they are sorted on the start of their busy time. This heuristic should result in a better reuse of qubits within one block.

### DIFF
--- a/src/Simulation/QCTraceSimulator/DepthCounter.cs
+++ b/src/Simulation/QCTraceSimulator/DepthCounter.cs
@@ -244,19 +244,28 @@ namespace Microsoft.Quantum.Simulation.QCTraceSimulatorRuntime
 
             if (optimizeDepth)
             {
-                // Doing width reuse heuristics and width computation.
+                // Performing width computation and applying reuse heuristics.
+                // First we fix busy time periods for all qubits involved.
+                foreach (QubitTimeMetrics q in qubitsTraceData.Cast<QubitTimeMetrics>())
+                {
+                    // If qubit is not fixed in time we fix it at zero.
+                    if (q.StartTime.IsEqualTo(ComplexTime.MinValue)) {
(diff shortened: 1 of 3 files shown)


src/Simulation/QCTraceSimulator/DepthCounter.cs
```diff
@@ -244,20 +244,29 @@ public void OnRelease(object[] qubitsTraceData)
...
+                // (1) We were not allowed to move gates in time.
+                // (2) We were sorting all qubits of the cirquit, not just this block.
+                // In reality these statements aren't true, so this is only a heuristic rather than the optimal reuse of qubits.
+                Array.Sort(qubitsTraceData, CompareAsQubitTimeMetricsByStartTime);
+
+                // Reuse qubits if possible or increase count of qubits used.
+                foreach (QubitTimeMetrics q in qubitsTraceData.Cast<QubitTimeMetrics>())
+                {
+                    // If qubit wasn't used in any gate. We don't allocate it.
+                    if (q.EndTime.IsEqualTo(ComplexTime.Zero)) {
```
bettinaheim, line 267: Is this correct even when there are gates right at the beginning that all have duration set to zero, and the qubit is only ever involved in these?

bettinaheim (approved):

src/Simulation/QCTraceSimulator/DepthCounter.cs
DmitryVasilevsky: Yes, this is correct. ComplexTime.Zero means not only zero depth, but also zero gates.

bettinaheim: Got it, thanks!

PR #681, opened 2021-05-10
Command-Line Compiler

ScottCarda-MS:
Adds a project for a command-line tool for creating executables from .NET DLLs with QIR embedded resources.


src/Qir/Execution/Tools/QirTools.cs
```diff
@@ -29,27 +29,49 @@ public static class QirTools
             DirectoryInfo includeDirectory,
             DirectoryInfo executablesDirectory)
         {
-            using var qirContentStream = new MemoryStream();
-            if (!AssemblyLoader.LoadQirByteCode(qsharpDll, qirContentStream))
-            {
-                throw new ArgumentException("The given DLL does not contain QIR byte code.");
-            }
+            await Foo(qsharpDll, libraryDirectory, includeDirectory, executablesDirectory);
```
ScottCarda-MS, line 32: This is just temporary for testing. It will be removed and this function restored before merging.

src/Qir/Execution/CommandLineCompiler/CommandLineCompiler.csproj
```diff
@@ -0,0 +1,13 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
```
cesarzc, line 5: The last thing to do to make the project self-contained, would be to add the `<CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>` property. This copies the dependencies to the output folder, which is what we want for the `Microsoft.Quantum.Simulators` and `Microsoft.Quantum.Qir.Runtime` packages.

src/Qir/Execution/CommandLineCompiler/CommandLineCompiler.csproj
```diff
@@ -0,0 +1,13 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
```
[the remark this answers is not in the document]
ScottCarda-MS: added

src/Qir/Execution/CommandLineCompiler/CommandLineCompiler.csproj
```diff
@@ -0,0 +1,13 @@
...
+    <PackageReference Include="System.CommandLine" Version="2.0.0-beta1.20213.1" />
+    <ProjectReference Include="..\Tools\Microsoft.Quantum.Qir.Tools.csproj" />
```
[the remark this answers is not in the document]
cesarzc: Microsoft.Quantum.Qir.Tools

src/Qir/Execution/CommandLineCompiler/Program.cs
```diff
@@ -0,0 +1,146 @@
...
+
+            Option<FileInfo> qsharpDllOption = new Option<FileInfo>(
+                aliases: new string[] { "--qsharpDll", "--dll" },
+                description: "The path to the QIR DLL file that is to be compiled.")
+            {
+                Required = true
+            };
+            buildCommand.AddOption(qsharpDllOption);
+
+            Option<DirectoryInfo> libraryDirectory = new Option<DirectoryInfo>(
```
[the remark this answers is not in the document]
ScottCarda-MS: What would the default value be for the library directories option if it is not provided? Should the parameter to the `BuildFromQSharpDll` be nullable, as well as the corresponding parameters to subsequent API functions, or should it be default to some other path. Maybe the includeDirectory path could be used as the default value? Since I'm not really familiar with how this is used in practice it is difficult for me to tell what the right behavior is on my own.

src/Qir/Execution/CommandLineCompiler/CommandLineCompiler.csproj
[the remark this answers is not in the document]
ScottCarda-MS: Thanks! I've updated the reference now.

src/Qir/Execution/CommandLineCompiler/Program.cs
```diff
@@ -0,0 +1,152 @@
...
+        {
+            var buildCommand = new Command("build", "(default) Build the executables from a QIR DLL.")
+            {
+                Handler = CommandHandler.Create((BuildOptions settings) =>
+                {
+                    if (settings.LibraryDirectories.Length < 1)
+                    {
+                        throw new ArgumentException("The '--libraryDirectories' option requires at least one argument.");
+                    }
+                    return QirTools.BuildFromQSharpDll(settings.QsharpDll, settings.LibraryDirectories[0], settings.IncludeDirectory, settings.ExecutablesDirectory);
```
bettinaheim, line 51: It is very odd to me that the libraries directories takes an array of libraries and then ignores all but the first. 
Why not allow a list of directories for both the libraries and the includes? Alternatively, if it should be one, then the command line should reflect that.

src/Qir/Execution/CommandLineCompiler/Program.cs
```diff
@@ -0,0 +1,152 @@
...
+                })
+            };
+            buildCommand.TreatUnmatchedTokensAsErrors = true;
+
+            Option<FileInfo> qsharpDllOption = new Option<FileInfo>(
```
bettinaheim, line 56: It would be nice if it was an either-or option of either giving the path to the qsharp dll or the path to the bitcode directly (or maybe even the human readable format would be nice as a third option). I haven't worked extensively with System.CommandLine, but usually it is possible to define alternative set of arguments. This is e.g. done for the Q# command line compiler by distinguishing code mode, snippet mode or responsefile mode (see [here](https://github.com/microsoft/qsharp-compiler/blob/5a90bed8554044d99d89953b922ddb8b804c4628/src/QsCompiler/CommandLineTool/Options.cs#L155)).

src/Qir/Execution/CommandLineCompiler/CommandLineCompiler.csproj
```diff
@@ -0,0 +1,16 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
bettinaheim, line 1: To avoid confusion with the Q# command line compiler, I would rename this to CommandLineBuildTools.

src/Qir/Execution/CommandLineCompiler/CommandLineCompiler.csproj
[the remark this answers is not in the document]
bettinaheim: Where would this move to? I think it makes sense to have this where the qir tools are, too, and then it would be better if it was built against the source code version of that package. Similarly for all the package references to the other packages that are built on this repo; I don't think having them as package references will work well with our e2e build, and I don't see a reason why they would need to be project references. Upon packaging this, we can configure that the project references are replaced by package references. What is your thinking here, @cesarzc?

src/Qir/Execution/CommandLineCompiler/CommandLineCompiler.csproj
bettinaheim: I would also probably solve this as part of packaging rather than as using CopyLocalLockFileAssemblies here; I believe it will work better with our build (but am happy to be convinced otherwise).

src/Qir/Execution/CommandLineCompiler/Program.cs
bettinaheim: I realize right now only having the path to the bitcode wouldn't work since you are creating the entry point object using the Q# AST. So I guess this is more my thinking for future evolution.

ScottCarda-MS: I am unsure what is the distinction being drawn between the qsharp dll and the bitcode. The QIR bitcode is a resource found inside the dll, right? Or have I been doing this wrong so far?

src/Qir/Execution/CommandLineCompiler/Program.cs
ScottCarda-MS: I believe the API calls will take multiple library directory paths in the near future, perhaps in this PR. @cesarzc?

src/Qir/Execution/CommandLineCompiler/CommandLineCompiler.csproj
ScottCarda-MS: Renamed to QirCommandLineTool.

src/Qir/Execution/CommandLineCompiler/CommandLineCompiler.csproj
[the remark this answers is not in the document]
cesarzc: Yes, it makes sense to have it here. I agree then that package references should be changed to project references and depending on how we package this tool, we might not even have to change them to package references at any point.

src/Qir/Execution/CommandLineCompiler/CommandLineCompiler.csproj
cesarzc: Yes, this was before deciding to use `PackAsTool`. `CopyLocalLockFileAssemblies` is not meant for packaging but to make the project self-contained in case something like `dotnet publish` was used. So, I think it's ok to remove `CopyLocalLockFileAssemblies` from the project file.

src/Qir/Execution/CommandLineCompiler/Program.cs
cesarzc: Yes, Scott is right. In the near future `BuildFromQSharpDll` will take a `IEnumerable` of `DirectoryInfo` for both libraries and include directories.

src/Qir/Execution/CommandLineCompiler/Program.cs
cesarzc: QIR bitcode is a resource found inside the DLL. I think @bettinaheim's comment is an idea on how this tool can evolve in the future but in the meantime I think it's ok to just have a `--qsharpDll` option.

src/Qir/Execution/CommandLineCompiler/Program.cs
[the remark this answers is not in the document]
cesarzc: I am when creating the option setting `Required = false`. In the near future, the `libraryDirectories` parameter for `BuildFromQSharpDll` will also be optional.

src/Qir/Execution/CommandLineCompiler/CommandLineCompiler.csproj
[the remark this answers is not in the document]
ScottCarda-MS: Which project does the `Microsoft.Quantum.Qir.Runtime` package refer to?

src/Qir/Execution/CommandLineCompiler/Program.cs
ScottCarda-MS: @cesarzc Should I change the include directories parameter to take multiple arguments as well?

ScottCarda-MS: I've updated the include directories to take multiple arguments as well.

src/Qir/Execution/QirCommandLineTool/QirCommandLineTool.csproj
```diff
@@ -0,0 +1,15 @@
...
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="System.CommandLine" Version="2.0.0-beta1.20213.1" />
+    <ProjectReference Include="..\..\..\Simulation\Simulators\Microsoft.Quantum.Simulators.csproj" />
+    <PackageReference Include="Microsoft.Quantum.Qir.Runtime" Version="0.16.210526824-alpha" />
```
cgranade, line 11: Should this be a project reference, since it's intra-repo?

src/Qir/Execution/QirCommandLineTool/Program.cs
```diff
@@ -0,0 +1,156 @@
...
+
+        /// <summary>
+        /// A class for encapsulating the different options for the build command.
+        /// </summary>
+        public sealed class BuildOptions
+        {
+            /// <summary>
+            /// The path to the .NET DLL file generated by the Q# compiler.
+            /// </summary>
+            public FileInfo QsharpDll { get; set; }
```
cgranade, line 138:
```suggestion
            public FileInfo QSharpDll { get; set; }
```

src/Qir/Execution/Tools/QirTools.cs
```diff
@@ -29,27 +29,49 @@ public static class QirTools
...
+        private static async Task<int> Foo(
+            FileInfo qsharpDll,
+            DirectoryInfo libraryDirectory,
+            DirectoryInfo includeDirectory,
+            DirectoryInfo executablesDirectory)
+        {
+            Console.WriteLine("From BuildFromQSharpDll!");
 
-            // ToDo: Return list of created file names
+            Console.WriteLine("QsharpDll:");
```
cgranade, line 65:
```suggestion
            Console.WriteLine("QSharpDll:");
```

src/Qir/Execution/QirCommandLineTool/Program.cs
```diff
@@ -0,0 +1,156 @@
...
+using System.CommandLine.Builder;
+using System.CommandLine.Help;
+using System.CommandLine.Invocation;
+using System.CommandLine.Parsing;
+using System.IO;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Quantum.Qir.Tools;
+
+namespace CommandLineCompiler
```
cgranade, line 15: Should this namespace be prefixed with `Microsoft.Quantum.`?

src/Qir/Execution/QirCommandLineTool/Program.cs
```diff
@@ -0,0 +1,156 @@
...
+        /// <returns>The Build command.</returns>
+        private static Command CreateBuildCommand()
+        {
+            var buildCommand = new Command("build", "(default) Build the executables from a QIR DLL.")
+            {
+                Handler = CommandHandler.Create((BuildOptions settings) =>
+                {
+                    if (settings.LibraryDirectories.Length < 1)
+                    {
+                        throw new ArgumentException("The '--libraryDirectories' option requires at least one argument.");
```
cgranade, line 49: In general, `--` style options tend to be typeset as `kebab-case`. In particular, our Q# entry point generator translates `nQubits` to `--n-qubits` as per that expectation, such that I'd suggest following the same pattern here.

src/Qir/Execution/QirCommandLineTool/Program.cs
```diff
@@ -0,0 +1,156 @@
...
+                    }
+                    if (settings.IncludeDirectories.Length < 1)
+                    {
+                        throw new ArgumentException("The '--includeDirectories' option requires at least one argument.");
+                    }
+                    return QirTools.BuildFromQSharpDll(settings.QsharpDll, settings.LibraryDirectories[0], settings.IncludeDirectories[0], settings.ExecutablesDirectory);
```
cgranade, line 55: Why does this take the first element? If only one `--includeDirectories` option is allowed, then I would suggest enforcing that with `Option.Arity` (https://github.com/dotnet/command-line-api/blob/main/src/System.CommandLine/Option.cs#L23) or at least using `.Single()` to raise an exception if multiple options are provided.

src/Qir/Execution/QirCommandLineTool/Program.cs
```diff
@@ -0,0 +1,156 @@
...
+                })
+            };
+            buildCommand.TreatUnmatchedTokensAsErrors = true;
+
+            Option<FileInfo> qsharpDllOption = new Option<FileInfo>(
```
cgranade, line 60: Given that `qsharpDllOption` is not referred to later, you may be able to simply the code by passing it directly to `AddOption` instead of saving it to a variable.

src/Qir/Execution/QirCommandLineTool/Program.cs
```diff
@@ -0,0 +1,156 @@
...
+                // Hide long argument descriptors.
+                var descriptor = base.ArgumentDescriptor(argument);
+                return descriptor.Length > 30 ? argument.Name : descriptor;
+            }
+        }
+
+        /// <summary>
+        /// A class for encapsulating the different options for the build command.
+        /// </summary>
+        public sealed class BuildOptions
```
cgranade, line 133: This looks like an ideal case for a record class (I think .NET 5 is used to build?). To use record classes in projects compatible with .NET Core SDK 3.1, you just need to make sure that the following declaration occurs somewhere in your project:

```
namespace System.Runtime.CompilerServices
{
    public class IsExternalInit { }
}
```

src/Qir/Execution/QirCommandLineTool/Program.cs
ScottCarda-MS: done

src/Qir/Execution/QirCommandLineTool/Program.cs
ScottCarda-MS: done

src/Qir/Execution/QirCommandLineTool/Program.cs
ScottCarda-MS: done

src/Qir/Execution/QirCommandLineTool/Program.cs
ScottCarda-MS: done

src/Qir/Execution/Tools/QirTools.cs
ScottCarda-MS: done

src/Qir/Execution/QirCommandLineTool/QirCommandLineTool.csproj
ScottCarda-MS: See this comment: https://github.com/microsoft/qsharp-runtime/pull/681#discussion_r631972088

cgranade: Ah, sorry for bifurcating the discussion, then!

src/Qir/Execution/QirCommandLineTool/Program.cs
ScottCarda-MS: This is related to an earlier conversation here: https://github.com/microsoft/qsharp-runtime/pull/681#discussion_r631538735
I don't seem to have access to the constructor you linked to. I don't think we have access to Option arity at all in this code.

src/Qir/Execution/QirCommandLineTool/Program.cs
ScottCarda-MS: I'm not very excited to use the new C# record types, and I don't really see the advantage here. Using a public class like this is consistent with examples and tutorials of the System.CommandLine Commands, and the compatibility with the new record feature is questionable. We have something that works well here.

src/Qir/Execution/CommandLineCompiler/CommandLineCompiler.csproj
ScottCarda-MS: removed

src/Qir/Execution/CommandLineCompiler/Program.cs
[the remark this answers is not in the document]
cesarzc: Talked about this offline. Since the underlying API currently does not support a list of library directories and a list of include directories, it's ok to leave these options as required now.

src/Qir/Execution/CommandLineCompiler/CommandLineCompiler.csproj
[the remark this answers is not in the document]
cesarzc: `Microsoft.Quantum.Qir.Runtime` package does not have a corresponding project file since it's all native binaries and headers. Since this in mostly relevant for packaging, for this PR the package reference can be removed.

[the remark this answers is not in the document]
ScottCarda-MS: removed

src/Qir/Execution/QirCommandLineTool/Program.cs
ScottCarda-MS: This has been fixed now. The API correctly accepts an array for the libraries and headers paths. This has allow for the arguments to the command line tool to also be made optional. I also figured out how to use the Arity property, although it seems to work a little differently than the online documentation shows.

src/Qir/Execution/CommandLineCompiler/Program.cs
ScottCarda-MS: This is addressed now that the API also accepts arrays for the libraries and headers paths.

cesarzc (approved):
Looks good. Thanks for making this change Scott!

PR #685, opened 2021-05-12
Fix Linux Clang build by including cstddef.h header explicitly

ricardo-espinoza:
Include cstddef.h header explicitly to ensure a deterministic behavior across platforms.


ricardo-espinoza:
Abandoning since #686 is doing this too.

PR #688, opened 2021-05-13
Add Azure Quantum submitter factory

samarsha:


src/Simulation/EntryPointDriver/Azure/Azure.cs
```diff
@@ -271,26 +270,24 @@ private static void Log(AzureSettings settings, string message)
...
         /// <returns>A QIR submitter.</returns>
-        /// <exception cref="ArgumentNullException"><paramref name="settings"/>.Target is null.</exception>
         private static IQirSubmitter? QirSubmitter(AzureSettings settings) => settings.Target switch
         {
-            null => throw new ArgumentNullException(nameof(settings), "Target is null"),
+            null => null,
             NoOpQirSubmitter.Target => new NoOpQirSubmitter(),
             NoOpSubmitter.Target => new NoOpSubmitter(),
-            _ => null // TODO: Add a factory to Microsoft.Azure.Quantum.Client.
+            _ => SubmitterFactory.QirSubmitter("TODO", settings.Target, settings.CreateWorkspace(), settings.Storage)
```
samarsha, line 290: @cesarzc, `AzureQirSubmitter` in qdk-private has a `providerId` parameter in its constructor, but I'm not sure what to put there. There isn't any provider option in the existing command-line options for Azure submission.

samarsha: This was based on our conversation where you said the providers would have the same constructor shape as `AzureQirSubmitter` - but it seems like the provider classes won't need their own provider ID, so will their constructors be:

```csharp
string target
IWorkspace workspace
string? storageConnectionString
```

instead?

cesarzc: Sorry I missed that. Yes, there should be another class that uses `AzureQirSubmitter` but does not expose the `providerId` in its constructor. I'll create a `MicrosoftSubmitter` class following the same pattern that we have for IonQ and Honeywell. I'll open a PR shortly.

src/Azure/Azure.Quantum.Client/SubmitterFactory.cs
```diff
@@ -0,0 +1,122 @@
...
+    /// Creates Azure Quantum submitters for specific targets.
+    /// </summary>
+    public static class SubmitterFactory
+    {
+        /// <summary>
+        /// Information about each supported QIR submitter.
+        /// </summary>
+        private static readonly ImmutableList<SubmitterInfo> QirSubmitters = ImmutableList.Create(
+            new SubmitterInfo(
+                "microsoft",
```
samarsha, line 25: @cesarzc, do you know what the target prefix for MicrosoftSimulatorSubmitter should be? I guessed `microsoft.*`, but maybe it's `microsoft.simulator.*` or something else.

cesarzc: Currently it is `microsoft.*`, and for the full state simulator the target name is currently `microsoft.simulator`.

cesarzc: This is just the current name that the service team is currently working with. We are coordinating with them to come up with a naming convention that makes sense.

src/Azure/Azure.Quantum.Client/SubmitterFactory.cs
```diff
@@ -0,0 +1,123 @@
...
+    /// </summary>
+    public static class SubmitterFactory
+    {
+        /// <summary>
+        /// Information about each supported QIR submitter.
+        /// </summary>
+        private static readonly ImmutableList<SubmitterInfo> QirSubmitters = ImmutableList.Create(
+            new SubmitterInfo(
+                new Regex(@"^microsoft\.simulator$"),
+                "Microsoft.Quantum.Providers.Targets.MicrosoftSimulatorSubmitter, Microsoft.Quantum.Providers.Core",
```
cesarzc, line 27: Is the first part (before the `,`) the name of the type and the second part the assembly name? Is this how fully-qualified names are defined?

cesarzc (approved):

src/Azure/Azure.Quantum.Client/SubmitterFactory.cs
samarsha: Yes, it's an [assembly-qualified name](https://docs.microsoft.com/en-us/dotnet/api/system.type.assemblyqualifiedname?view=netcore-3.1).

PR #692, opened 2021-05-17
Introduce ClassicalControl namespace, depreicate old namespace

swernli:
This moves the implementation of the callables used by the Classical Control rewrite step into a new namespace, Microsoft.Quantum.ClassicalControl. The old callables in the Microsoft.Quantum.Simulation.QuantumProcessor.Extensions are marked as depricated and will be removed in a later PR. Follow up will need to be done across the remaining repos where explicit references to this namespace exist (such as in the compiler rewrite step itself) to move to the new namespace.


ScottCarda-MS (approved):
Looks good! There should be a related PR made on the compiler repo that updates to use the new namespace as well. One place that needs updating that I can think of is here: https://github.com/microsoft/qsharp-compiler/blob/main/src/QsCompiler/Core/Dependencies.fs#L27

In the mean time, the compiler would just end up using the depreciated definitions defined in this PR.

PR #695, opened 2021-05-19
Added the assert message printing

kuzminrobin:


ScottCarda-MS (approved):
lgtm. I am fairly unfamiliar with this contented however, so it would be good to get someone who is more knowledgeable to review as well.

Issue #557, opened 2021-03-12
QIR FullStateSimulator does not reuse qubit mappings


swernli:
The QIR FullStateSimulator wrapper does not include any logic to reuse qubit logical identifiers, instead always incrementing the logical qubit id with each call to allocate, as seen in [FullStateSimulator.cpp](https://github.com/microsoft/qsharp-runtime/blob/main/src/QirRuntime/lib/Simulators/FullstateSimulator.cpp#L186):
```c++
        Qubit AllocateQubit() override
        {
            typedef void (*TAllocateQubit)(unsigned, unsigned);
            static TAllocateQubit allocateQubit = reinterpret_cast<TAllocateQubit>(this->GetProc("allocateQubit"));

            const unsigned id = this->nextQubitId;
            allocateQubit(this->simulatorId, id);
            this->nextQubitId++; // <-- UNCONDITIONALLY INCREASES ID
            return reinterpret_cast<Qubit>(id);
        }
```
Looking at how the native simulator implements the `allocateQubit` API, the assumption is that the caller will reuse a previously released qubit id if possible. Since the QIR wrapper does not, it means each allocation will grow the size of the qubit map even if there are unused qubits already present in the map. This is effectively a memory leak, though admittedly a very small one.

While we could implement the logic in FullStateSimulator.cpp to track in-use qubit ids and reuse the next available (similar to what the C# mapping implemented in [QubitManager.cs](https://github.com/microsoft/qsharp-runtime/blob/8b4a430956f7f0cfa731eb93458e4df93ee03d56/src/Simulation/Common/QubitManager.cs#L21) does), we should instead consider exposing the qubit management logic already present in the native simulator through the C API. The simulator already supports a [variant of allocate](https://github.com/microsoft/qsharp-runtime/blob/main/src/Simulation/Native/src/simulator/simulator.hpp#L75) that takes no arguments and instead returns the qubit id:
```c++
    logical_qubit_id allocate()
    {
        recursive_lock_type l(mutex());
        return psi.allocate_qubit();
    }
```
The [corresponding function on the wave function implementation](https://github.com/microsoft/qsharp-runtime/blob/main/src/Simulation/Native/src/simulator/wavefunction.hpp#L478) then uses its internal map to find the first available qubit, only adding a new qubit if no existing entries are available:
```c++
    /// Allocate a qubit with implicitly assigned logical qubit id.
    logical_qubit_id allocate_qubit()
    {
#ifndef NDEBUG
[...]

PR #696, opened 2021-05-19
Added first version of Qubit manager to QIR

DmitryVasilevsky:
This is the first version of the native Qubit Manager for QIR. Initial unit tests are included.

Qubit Manager supports
* Allocation and release of qubits including arrays of qubits.
* Reuse encouraged/discouraged. Extending capacity enabled/disabled.
* Restricted reuse areas and segments to support low-depth circuits.

Qubit manager builds and tests are run as part of the qir-runtime-unittests. It is used in the bridge to the full state simulator.


src/Qir/Runtime/lib/QIR/QubitManagerRestrictedReuse.hpp
```diff
@@ -0,0 +1,210 @@
+// Copyright (c) Microsoft Corporation.
```
kuzminrobin, line 1: Up to you: Would be nice to have a description of what you told me, what the qubit manager is for, etc.
Either as a section in the existing `README.md` or in a separate `QubitManager.md`.

kuzminrobin, line 1: I feel a strong need of high-level description of how different parts are supposed to work (with examples using particular numbers).
Especially `QubitListInSharedArray`. There is a number of places, understanding which takes more time than it could, and still no guarantee that the understanding is correct. At least the places where the indices and values are manipulated, like this:
```c++
        sharedQubitStatusArray[lastElement] = id;
        sharedQubitStatusArray[id] = NoneMarker;
```

src/Qir/Runtime/lib/QIR/QubitManagerRestrictedReuse.hpp
```diff
@@ -0,0 +1,210 @@
...
+    struct IQubitManager
+    {
+        virtual ~IQubitManager() = default;
+        virtual Qubit Allocate() = 0;
+        virtual void Release(Qubit qubit) = 0;
+    };
+
+    // The end of free lists are marked with NoneMarker value. It is used like null for pointers.
+    // This value is non-negative just like other values in the free lists.
+    constexpr int NoneMarker = INT_MAX;
```
kuzminrobin, line 34: Nit: In a number of places the _signed_ integer types are used (for an _unsigned_ concept like array index or number of elements) and then checked against `< 0`. Consider the _unsigned_ integer types (and remove the checks `< 0`).

FYI: For my own code I consider _unsigned_ integer types by default.

src/Qir/Runtime/lib/QIR/QubitManagerRestrictedReuse.hpp
```diff
@@ -0,0 +1,210 @@
...
+
+    // Explicitly allocated qubits are marked with Allocated value.
+    // Qubits implicitly allocated for borrowing are "refcounted" with values [Allocated+1 .. -2].
+    // When refcount needs to be decreased to Allocated value, the qubit is automatically released.
+    constexpr int AllocatedMarker = INT_MAX;
```
kuzminrobin, line 39: FYI: The [`std::numeric_limits<int>::max()`](https://en.cppreference.com/w/cpp/types/numeric_limits) is considered the C++ counterpart of C's `INT_MAX`.

src/Qir/Runtime/lib/QIR/QubitManagerRestrictedReuse.hpp
```diff
@@ -0,0 +1,210 @@
...
+        // Returns a newly instantiated qubit.
+        public: virtual Qubit CreateQubitObject(int id);
+
+        // May be overriden to get a qubit id from a custom qubit object.
+        // Must be overriden if CreateQubitObject is overriden.
+        // When not overriden, it just reinterprets pointer to qubit as a qubit id.
+        // qubit: pointer to QUBIT
+        // Returns id of a qubit pointed to by qubit.
+        public: virtual int QubitToId(Qubit qubit);
+    };
```
kuzminrobin, line 206: Consider grouping the members by access (`public` first, then `protected`, then `private`):
```c++
class C ..
{
public:
    ..
protected:
    ..
private:
    ..
};
```
I understand that sometimes it creates code that cannot be compiled in a single parse (e.g. the member variables are mentioned in the `public` constructor first, and then they are defined in the `private` section). But today's compilers are multi-parse. And that's the approach many people use (list the `public` first).

src/Qir/Runtime/lib/QIR/QubitManagerRestrictedReuse.hpp
```diff
@@ -0,0 +1,210 @@
...
+        CRestrictedReuseAreaStack& operator = (const CRestrictedReuseAreaStack&) = delete;
+        ~CRestrictedReuseAreaStack() = default;
+
+        void PushToBack(RestrictedReuseArea area);
+        RestrictedReuseArea PopFromBack();
+        RestrictedReuseArea& PeekBack();
+        // TODO: Remove and use size directly?
+        int Count();
+    };
+
```
kuzminrobin, line 110: > `class CQubitManagerRestrictedReuse`

FYI: Class names are typically _nouns_. So I intuitively I feel that your code may benefit if you rename the class (and files)
CQubitManagerRestrictedReuse to
C**RestrictedReuse**QubitManager

src/Qir/Runtime/lib/QIR/QubitManagerRestrictedReuse.hpp
```diff
@@ -0,0 +1,210 @@
...
+        // No complex scenarios for now. Don't need to support copying/moving.
+        public: CQubitManagerRestrictedReuse(const CQubitManagerRestrictedReuse&) = delete;
+        public: CQubitManagerRestrictedReuse& operator = (const CQubitManagerRestrictedReuse&) = delete;
+        public: ~CQubitManagerRestrictedReuse();
+        private: void EnsureCapacity(int requestedCapacity);
+
+        // Disable, disabled qubit count and checks
+
+        public: int DisabledQubitsCount = 0;
```
kuzminrobin, line 146: FYI: In the C++ code I mostly saw, the member variables were typically grouped in a single `private` section in the end of the class. Such grouping helps to quickly estimate the total size of the class instance, and helps to check if all the member variables are initialized in the constructor.

src/Qir/Runtime/lib/QIR/QubitManagerRestrictedReuse.hpp
```diff
@@ -0,0 +1,210 @@
...
+
+    class CQubitManagerRestrictedReuse : public IQubitManager
+    {
+        // Restricted reuse area control, allocation and release
+
+        public: void StartRestrictedReuseArea();
+        public: void NextRestrictedReuseSegment();
+        public: void EndRestrictedReuseArea();
+        private: long AllocateQubit(); // Computation complexity is O(number of nested restricted reuse areas).
+        private: void ReleaseQubit(long id);
```
kuzminrobin, line 119: > **long** AllocateQubit()

FYI for the future: I'm not sure if the qubit manager will be used _in addition_ to the [`FullStateSimulator`](https://github.com/microsoft/qsharp-runtime/blob/3f7ea4b7f0174b58f665737452da64710eb227c3/src/Qir/Runtime/lib/Simulators/FullstateSimulator.cpp#L193) wrapper class and [actual simulator](https://github.com/microsoft/qsharp-runtime/blob/3f7ea4b7f0174b58f665737452da64710eb227c3/src/Simulation/Native/src/simulator/capi.hpp#L72), or _instead_ of them, but eventually I believe we will likely use the same type everywhere to designate the qubit ID.

You may want to reuse one of the types they currently use for the qubit ID.

src/Qir/Runtime/lib/QIR/QubitManagerRestrictedReuse.hpp
```diff
@@ -0,0 +1,210 @@
...
+
+        // Borrow (We treat borrowing as allocation currently)
+
+        public: Qubit Borrow();
+        public: Qubit* Borrow(long qubitCountToBorrow);
+
+        // Return (We treat returning as release currently)
+
+        public: void Return(Qubit qubit);
+        public: void Return(Qubit* qubitsToReturn, int qubitCountToReturn);
```
kuzminrobin, line 188: Nit: If the _borrowing_ feature is optional then it might make sense to extract it into a separate interface (e.g. `IQubitLender`), either implemented or not implemented by the qubit manager classes.

src/Qir/Runtime/lib/QIR/QubitManagerRestrictedReuse.hpp
```diff
@@ -0,0 +1,210 @@
...
+        public: void Disable(Qubit* qubitsToDisable, int qubitCount);
+
+        // Allocate, allocated qubit count and checks
+
+        public: int AllocatedQubitsCount = 0;
+        public: bool IsExplicitlyAllocated(long id);
+        public: void ChangeStatusToAllocated(long id);
+        // Allocates a qubit. Extend capacity if necessary and possible.
+        // Fails if the qubit cannot be allocated. 
+        public: Qubit Allocate();
```
kuzminrobin, line 163: `Qubit Allocate()` and `void Release(Qubit qubit)` are `CQubitManagerRestrictedReuse`'s overrides of the virtual functions inherited from `IQubitManager`:
```c++
virtual Qubit Allocate() = 0;        
virtual void Release(Qubit qubit) = 0;
```
Consider using the keyword `override` for the overrides. See [Item 12: Declare overriding functions `override`](https://moodle.ufsc.br/pluginfile.php/2377667/mod_resource/content/0/Effective_Modern_C__.pdf)

src/Qir/Runtime/lib/QIR/QubitManagerRestrictedReuse.hpp
```diff
@@ -0,0 +1,210 @@
...
+    private:
+        int firstElement = NoneMarker;
+        int lastElement = NoneMarker;
+
+    public:
+        // Initialize empty list
+        QubitListInSharedArray() = default;
+
+        // Initialize as a list with sequential elements from minElement to maxElement inclusve.
+        QubitListInSharedArray(int minElement, int maxElement, int* sharedQubitStatusArray);
```
kuzminrobin, line 68: (For the future)
From the description it is not evident how large should the array be, pointed to by the `sharedQubitStatusArray` (how many elements it should have - e.g. `(maxElement - minElement)` elements or other?). It is also not evident (from the description) that the elements `sharedQubitStatusArray[minElement]` through the `sharedQubitStatusArray[maxElement]` are _accessed_ in this function.

Also for me personally it is unexpected that the constructor updates anything _outside_ of the instance it constructs.

src/Qir/Runtime/lib/QIR/QubitManagerRestrictedReuse.cpp
```diff
@@ -0,0 +1,468 @@
...
+    }
+    sharedQubitStatusArray[maxElement] = NoneMarker;
+    firstElement = minElement;
+    lastElement = maxElement;
+}
+
+bool QubitListInSharedArray::IsEmpty()
+{
+    return firstElement == NoneMarker;
+}
```
kuzminrobin, line 37: Nit: Consider using `const` member functions.
> bool QubitListInSharedArray::IsEmpty() **const**

See "const Member Functions" in "Item 3: Use `const` whenever possible" either [here](http://wavelino.coffeecup.com/pdf/Effective%20Cshapshap.pdf) or [here](https://www.amazon.com/gp/product/0321334876?ie=UTF8&tag=aristeia.com-20&linkCode=as2&camp=1789&creative=9325&creativeASIN=0321334876).

src/Qir/Runtime/lib/QIR/QubitManagerRestrictedReuse.cpp
```diff
@@ -0,0 +1,468 @@
...
+    return CreateQubitObject(newQubitId);
+}
+
+Qubit* CQubitManagerRestrictedReuse::Allocate(long numToAllocate)
+{
+    FailIf(numToAllocate < 0, "Attempt to allocate negative number of qubits.");
+    if (numToAllocate == 0)
+    {
+        return new Qubit[0];
+    }
```
kuzminrobin, line 332: It looks like a dusty/arguable corner of the C++ language when we allocate an array with `0` number of elements.
Consider documenting that case (there is a risk of undefined behavior when accessing the returned pointer, see "undefined behavior" [here](https://www.cplusplus.com/reference/new/operator%20new[]/)).

kuzminrobin:
I believe I'm done with the review. I assume you don't need approval (requesting for comments in this PR).

swernli:
An interesting note: the clang-tidy on Linux tends to be stricter than the one on Windows, and has caught a few issues that were reported as errors on that platform causing the native Linux build step to fail. You can see those starting in the log [here](https://dev.azure.com/ms-quantum-public/Microsoft%20Quantum%20(public)/_build/results?buildId=27083&view=logs&j=b4736319-4c84-58ca-d81e-81ed9fcd8955&t=b7224d1d-00c8-5354-2b25-576812c6c32f&l=46).

src/Qir/Runtime/lib/QIR/QubitManagerRestrictedReuse.cpp
```diff
@@ -0,0 +1,468 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+// !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+// This code compiles, but was never executed, be careful!
```
swernli, line 5: Now that you have a unit test for this, perhaps the comment can be updated to say this is prototype code that the runtime does not yet use?

src/Qir/Runtime/lib/QIR/QubitManagerRestrictedReuse.hpp
DmitryVasilevsky: Thanks! Updated.

src/Qir/Runtime/lib/QIR/QubitManagerRestrictedReuse.hpp
DmitryVasilevsky: I renamed parameters. This class may be moved to be a nested class of QubitManager. That way it would be easier to understand what's going on. However, as this is called "list in shared array" and this shared array is passed in, I think it's reasonable to assume that this will adjust the array.

src/Qir/Runtime/lib/QIR/QubitManager.hpp
```diff
@@ -0,0 +1,218 @@
...
+namespace Quantum
+{
+
+    // TODO: Remove this! Use standard ones.
+    static void FailNow(const char* message) { std::cout << message; throw message; }
+    static void FailIf(bool condition, const char* message) {
+        if (condition) {
+            FailNow(message);
+        }
+    }
```
kuzminrobin, line 32: FYI:
* If a global function (as opposed to member function) is fully _defined_ in a header, then every .c/.cpp file that includes that header, will have a separate copy of such a function in the .obj file. Later on the linker is not guaranteed to merge the multiple copies into a single one (especially if the function is `static`).  
So, for global functions, you may want to leave in the header the _declaration_ only: `void FailNow(const char* message)` (optionally prepended with `extern`), and move the _definition_ to the .cpp file.
* I see that these particular functions are only used in one .cpp file. So it makes sense to completely move them from the header to that .cpp file and leave them `static`.
* Instead of `FailNow()` you may want to use [`quantum__rt__fail_cstr()`](https://github.com/microsoft/qsharp-runtime/blob/b10447eff672b7ba4f2e8564302888a8eb6bce8a/src/Qir/Runtime/lib/QIR/utils.cpp#L60).
* If you are interested about the future of `quantum__rt__fail_cstr()`, you can search for "quantum__rt__fail_cstr()" [here](https://ms-quantum.visualstudio.com/Quantum%20Program/_git/Quantum-OKRs.wiki?path=%2FMicrosoft-Quantum-Program%2FTechnical%2FEngineering-Specifications%2FCompiler-%26-Runtime%2FQIR-RT-API-Design-Guidelines.md&version=GBrokuzmin%2FqirRtApiGuidelines&_a=preview) (the doc is under development).

src/Qir/Runtime/lib/QIR/QubitManager.hpp
```diff
@@ -0,0 +1,218 @@
...
+    using QubitIdType = ::int32_t;
+
+    // The end of free lists are marked with NoneMarker value. It is used like null for pointers.
+    // This value is non-negative just like other values in the free lists.
+    constexpr QubitIdType NoneMarker = std::numeric_limits<QubitIdType>::max();
+
+    // Explicitly allocated qubits are marked with Allocated value.
+    // Qubits implicitly allocated for borrowing are "refcounted" with values [Allocated+1 .. -2].
+    // When refcount needs to be decreased to Allocated value, the qubit is automatically released.
+    constexpr QubitIdType AllocatedMarker = std::numeric_limits<QubitIdType>::max();
```
kuzminrobin, line 43: FYI: I'm a bit confused.
* What is "Allocated value"? Did you mean `AllocatedMarker`? An example would help...
* The fragment `[Allocated+1 .. -2]` (the value `-2`) makes an impression, that you meant: 
`[Allocated+1, Allocated+0, Allocated-1, Allocated-2, ... , 2, 1, 0, -1, -2]`
But intuitively I feel that you meant something different.

src/Qir/Runtime/lib/QIR/QubitManager.hpp
```diff
@@ -0,0 +1,218 @@
...
+
+    // TODO: Remove this! Use standard ones.
+    static void FailNow(const char* message) { std::cout << message; throw message; }
+    static void FailIf(bool condition, const char* message) {
+        if (condition) {
+            FailNow(message);
+        }
+    }
+
+    using QubitIdType = ::int32_t;
```
kuzminrobin, line 34: Up to you: I would suggest to consider the _unsigned type_ for the unsigned concept - e.g. **u**int32_t.

src/Qir/Runtime/lib/QIR/QubitManager.hpp
```diff
@@ -0,0 +1,218 @@
...
+    // Disabled qubits are marked with this value.
+    constexpr QubitIdType DisabledMarker = -1;
+
+    // We want status array to be reasonably large if initialization for it is wrong (i.e. requested size < 1).
+    constexpr QubitIdType FallbackQubitCapacity = 8;
+
+    // Indexes in the status array can potentially be in range 0 .. int.MaxValue-1.
+    // This gives maximum capacity as int.MaxValue.
+    // Index equal to int.MaxValue doesn't exist and is reserved for 'NoneMarker' - list terminator.
+    constexpr QubitIdType MaximumQubitCapacity = std::numeric_limits<QubitIdType>::max();
```
kuzminrobin: FYI:
(I pretend for myself that I'm reading this from scratch, familiarizing myself with the Qubit Manager)

Would be nice to see some introduction about the data structures used to store the instances of `QubitIdType`. At a glance I have an impression that the instances of `QubitIdType` are to be stored in the _arrays_. If that's the case then would be nice to see the answer to the following questions
* Is it a single array or multiple arrays? If multiple, then used for the same purpose or for different purposes? (e.g. one array for available qubits, another one for allocated, the third one for blocked, borrowed, etc.);
* Can `NoneMarker` and `AllocatedMarker` occur in the same data structure? If yes then how do the algorithms distinguish between them if they have the same value?
* What is "status array"?

And these questions:
> // The end of free lists are marked with NoneMarker value

What is free list? Is it one or many? If many then why (what is each one for)? The fragment says how the end is marked, but what are the values _in_ the free list, _before the end_?

> // Explicitly allocated qubits are marked with ..

Marked where? (in which data structure?)

> // Disabled qubits are marked with ..

Marked where? (in which data structure?)

---
I have an impression that the same data type - `QubitIdType` - will be used 
* to store the values in the array elements
* and to index the array, and also to specify the capacity (or number of elements) of the array.

If that's the case then some more comments would help about why such a decision is better than the other options. 

Because in general those are two separate/independent concepts, and they can change (get refactored) regardless of each other. E.g. the capacity can be fixed to 10 (indexes fixed to the range 0..9), but the values written to the array can grow from 0 to 0xFFFF, and then reset back to 0 and grow again.

---
To summarize, the intro comment answering all those questions would help.

src/Qir/Runtime/lib/QIR/QubitManager.cpp
```diff
@@ -0,0 +1,468 @@
...
+        lastElement = id; // The new element will be the last element in the chain.
+    }
+}
+
+// TODO: Set status to the taken qubit here. Then counting is reasonable here, but not possible?
+// TODO: Rename 'RemoveQubitFromFront'?
+int QubitListInSharedArray::TakeQubitFromFront(QubitIdType* sharedQubitStatusArray)
+{
+    // First element will be returned. It is 'NoneMarker' if the list is empty.
+    int result = firstElement;
```
kuzminrobin, line 66: Forgot to replace `int` with `QubitIdType`? (2 occurrences)

src/Qir/Runtime/public/QubitManager.hpp
```diff
@@ -0,0 +1,220 @@
...
+    public:
+        CQubitManager(
+            QubitIdType initialQubitCapacity = DefaultQubitCapacity,
+            bool mayExtendCapacity = true,
+            bool encourageReuse = true);
+
+        // No complex scenarios for now. Don't need to support copying/moving.
+        CQubitManager(const CQubitManager&) = delete;
+        CQubitManager& operator = (const CQubitManager&) = delete;
+        ~CQubitManager();
```
kuzminrobin, line 130: Class has at least one virtual function - `CreateQubitObject()`. Consider making the destructor also _virtual_. See [Item 7: Declare destructors virtual in polymorphic base classes](http://wavelino.coffeecup.com/pdf/Effective%20Cshapshap.pdf).

src/Qir/Runtime/lib/QIR/QubitManager.cpp
```diff
@@ -0,0 +1,465 @@
...
+{
+    this->mayExtendCapacity = mayExtendCapacity;
+    this->encourageReuse = encourageReuse;
+
+    qubitCapacity = initialQubitCapacity;
+    if (qubitCapacity <= 0) {
+        qubitCapacity = DefaultQubitCapacity;
+    }
+    sharedQubitStatusArray = new int[qubitCapacity];
+
```
kuzminrobin, line 164: Did you mean the following?
> sharedQubitStatusArray = new **QubitIdType**[qubitCapacity];

src/Qir/Runtime/lib/QIR/QubitManager.cpp
```diff
@@ -0,0 +1,465 @@
...
+{
+    // No need to adjust existing values (NonMarker or indexes in the array).
+    // TODO: Borrowing/returning are not yet supported.
+    if (requestedCapacity <= qubitCapacity)
+    {
+        return;
+    }
+
+    // Prepare new shared status array
+    QubitIdType* newStatusArray = new int[requestedCapacity];
```
kuzminrobin, line 373: Did you mean the following?
> QubitIdType* newStatusArray = new **QubitIdType**[requestedCapacity];

src/Qir/Runtime/lib/QIR/QubitManager.cpp
```diff
@@ -0,0 +1,465 @@
...
+
+//
+// RestrictedReuseArea
+//
+
+RestrictedReuseArea::RestrictedReuseArea(QubitListInSharedArray freeQubits)
+{
+    //FreeQubitsReuseProhibited = QubitListInSharedArray(); // This is initialized by default.
+    FreeQubitsReuseAllowed = freeQubits; // Default shallow copying.
+}
```
kuzminrobin, line 116: (Up to you)
Consider
* Making the parameter a reference to `const`
* And using the member initialization list in all the constructors.

```c++
RestrictedReuseArea::RestrictedReuseArea(const QubitListInSharedArray& freeQubits) // Param is ref to const
    : FreeQubitsReuseAllowed(freeQubits)  // Member initialization list.
{    
}
```
See [Item 4: Make sure that objects are initialized before they're used](http://wavelino.coffeecup.com/pdf/Effective%20Cshapshap.pdf).

src/Qir/Runtime/public/QubitManager.hpp
```diff
@@ -0,0 +1,220 @@
...
+        CRestrictedReuseAreaStack(const CRestrictedReuseAreaStack&) = delete;
+        CRestrictedReuseAreaStack& operator = (const CRestrictedReuseAreaStack&) = delete;
+        ~CRestrictedReuseAreaStack() = default;
+
+        void PushToBack(RestrictedReuseArea area);
+        RestrictedReuseArea PopFromBack();
+        RestrictedReuseArea& PeekBack();
+        // TODO: Remove and use size directly?
+        int Count() const;
+    };
```
kuzminrobin, line 117: (Up to you)
Consider naming the member functions the same way as in [`std::stack`](https://www.cplusplus.com/reference/stack/stack/?kw=stack): e.g. `Push()`, `Pop()`, `Top()`, `Size()` (and referring to the documentation of the `std::stack`).

src/Qir/Runtime/lib/QIR/QubitManager.cpp
```diff
@@ -0,0 +1,465 @@
...
+    this->insert(this->end(), area);
+}
+
+RestrictedReuseArea CRestrictedReuseAreaStack::PopFromBack()
+{
+    FailIf(this->empty(), "Cannot remove element from empty set.");
+    RestrictedReuseArea result = this->back();
+    this->pop_back();
+    return result;
+}
```
kuzminrobin, line 135: (FYI)
From the exception-safety and thread-safety point of view it is important that `Pop()` does not return a value.
I don't remember the details well enough to explain, but, if you are interested, they can be found in
* [3.2.3 Spotting race conditions inherent in interfaces](https://www.bogotobogo.com/cplusplus/files/CplusplusConcurrencyInAction_PracticalMultithreading.pdf)
* [[exc_h:fsos]](https://github.com/kuzminrobin/code_review_notes/blob/master/article_list.md)

src/Qir/Runtime/public/QubitManager.hpp
```diff
@@ -0,0 +1,220 @@
...
+
+        // State
+        QubitIdType* sharedQubitStatusArray; // Tracks allocation state of all qubits. Stores lists of free qubits.
+        QubitIdType qubitCapacity; // qubitCapacity is always equal to the array size.
+        CRestrictedReuseAreaStack freeQubitsInAreas; // Fresh Free Qubits are located in freeQubitsInAreas[0].FreeQubitsReuseAllowed
+
+        // Counts
+        int disabledQubitCount;
+        int allocatedQubitCount;
+        int freeQubitCount;
```
kuzminrobin, line 215: (FYI)
I didn't see any discussion of [this](https://github.com/microsoft/qsharp-runtime/pull/696#discussion_r635652520). To make sure that this particular part is not forgotten, I'll remind, consider using the unsigned types for the unsigned concepts (or I'd love if you could explain me the benefits of the signed types).

Looking forward I also recommend the _fixed-size types_. See details in [Do Not Expose the Non-fixed-size Data Types](https://ms-quantum.visualstudio.com/Quantum%20Program/_git/Quantum-OKRs.wiki?path=%2FMicrosoft-Quantum-Program%2FTechnical%2FEngineering-Specifications%2FCompiler-%26-Runtime%2FQIR-RT-API-Design-Guidelines.md&version=GBrokuzmin%2FqirRtApiGuidelines&_a=preview&anchor=do-not-expose-the-non-fixed-size-data-types) section (the doc is under development).

kuzminrobin:
I feel tired/exhausted today. I still need to figure out the logic behind the calls of `CQubitManager` (`Allocate()`, `Release()`, `Disable()`, ...). I will postpone that until some time later.

src/Qir/Runtime/lib/QIR/QubitManager.hpp
DmitryVasilevsky: This was obviously an integration point to be fixed with a TODO in front of it. I think the best way for now it to define these functions with quantum__rt__fail_cstr() as you suggested and move them to the .cpp file. That's what I did.

src/Qir/Runtime/lib/QIR/QubitManager.hpp
DmitryVasilevsky: It is AllocatedMarker and it is supposed to be int min value, not max. I also removed the wording about borrowing as it is not supported. We lack some unit tests (which I will add) and some comments about how status array is structured (which I will also add).

src/Qir/Runtime/lib/QIR/QubitManager.hpp
```diff
@@ -0,0 +1,218 @@
...
+    // Explicitly allocated qubits are marked with Allocated value.
+    // Qubits implicitly allocated for borrowing are "refcounted" with values [Allocated+1 .. -2].
+    // When refcount needs to be decreased to Allocated value, the qubit is automatically released.
+    constexpr QubitIdType AllocatedMarker = std::numeric_limits<QubitIdType>::max();
+
+    // Disabled qubits are marked with this value.
+    constexpr QubitIdType DisabledMarker = -1;
+
+    // We want status array to be reasonably large if initialization for it is wrong (i.e. requested size < 1).
+    constexpr QubitIdType FallbackQubitCapacity = 8;
```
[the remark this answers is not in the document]
DmitryVasilevsky: I renamed this DefaultQubitCapacity as I changed defaults for the qubit manager constructor. As for types - I will post a separate discussion.

src/Qir/Runtime/lib/QIR/QubitManager.cpp
DmitryVasilevsky: Thanks for catching! Updated.

src/Qir/Runtime/public/QubitManager.hpp
DmitryVasilevsky: Of course! We allow derived classes here. Thanks for catching. Updated.

src/Qir/Runtime/lib/QIR/QubitManager.cpp
DmitryVasilevsky: Yes! Updated.

src/Qir/Runtime/lib/QIR/QubitManager.cpp
DmitryVasilevsky: Yes! Updated.

src/Qir/Runtime/lib/QIR/QubitManager.cpp
DmitryVasilevsky: No need to make it reference to const. This is a small class, think of it as an int or long long. No one would do this for ints, as they are intended to be copied.
As for member initialization - I will think about it. I know the reasons, but I don't like the style.

src/Qir/Runtime/public/QubitManager.hpp
DmitryVasilevsky: You didn't see any discussion because this github interface ate my replies. I will post them again.

src/Qir/Runtime/lib/Simulators/FullstateSimulator.cpp
```diff
@@ -156,6 +159,7 @@ namespace Quantum
             typedef unsigned (*TInit)();
             static TInit initSimulatorInstance = reinterpret_cast<TInit>(this->GetProc("init"));
 
+            qubitManager = new CQubitManager();
```
swernli, line 162: If you make this member variable an `std::unique_ptr<CQubitManager>` then this line can be `qubitManager = std::make_unique<CQubitManager>();` and you can remove the explicit delete in the destructor below.

swernli (approved):
Just one minor comment from me, but otherwise looks good from my side. Also, I believe merging this will resolve #557.

src/Qir/Runtime/lib/QIR/QubitManager.cpp
kuzminrobin: FYI: I see the update in .cpp but not in .hpp

src/Qir/Runtime/lib/QIR/QubitManager.cpp
```diff
@@ -0,0 +1,487 @@
...
+
+//
+// QubitListInSharedArray
+//
+
+QubitListInSharedArray::QubitListInSharedArray(QubitIdType startId, QubitIdType endId, QubitIdType* sharedQubitStatusArray)
+{
+    FailIf(startId > endId || startId < 0 || endId == std::numeric_limits<QubitIdType>::max(),
+        "Incorrect boundaries in the linked list initialization.");
+
```
kuzminrobin, line 38: (I'm trying to dive deeper into the logic this time)

Does this occurrence of `std::numeric_limits<QubitIdType>::max()` mean `NoneMarker` or `MaximumQubitCapacity` or other? Consider replacing with the corresponding one, to be more informative.

src/Qir/Runtime/lib/QIR/QubitManager.cpp
```diff
@@ -0,0 +1,487 @@
...
+        QubitIdType newQubitId = AllocateQubitId();
+        if (newQubitId == NoneMarker)
+        {
+            for (int k = 0; k < i; k++)
+            {
+                Release(qubitsToAllocate[k]);
+            }
+            FailNow("Not enough qubits.");
+            return;
+        }
```
kuzminrobin, line 263: Just curious, can this be optimized out by comparing `qubitCountToAllocate > freeQubitCount`? (before the `for` loop)

kuzminrobin (approved):
Other than [this](https://github.com/microsoft/qsharp-runtime/pull/696#discussion_r643455697) looks good to me. 
Approving.

src/Qir/Runtime/lib/QIR/QubitManagerRestrictedReuse.hpp
DmitryVasilevsky: I added concise description as a comment in the .hpp file.

DmitryVasilevsky: I added comments to the places in QubitListInSharedArray where liked lists are manipulated. I believe it should be easy to understand as this is just a standard singly linked list, managed by a struct named QubitListInSharedArray. Method names are also self-explanatory.
I will add description of the content of the shared array later.

src/Qir/Runtime/lib/QIR/QubitManagerRestrictedReuse.hpp
DmitryVasilevsky: Long time ago I was also advocating unsigned types for unsigned concepts. I changed my mind after seeing how convenient signed types are for these concepts in C#. I will make a separate comment about this later.

src/Qir/Runtime/lib/QIR/QubitManagerRestrictedReuse.hpp
DmitryVasilevsky: This is not about a single pass compilation. This is about C# heritage. In C# each member carries access rights. So the members were grouped according to similar functionality. I.e. everything that concerns qubit release were in one region, etc. Yes it's more common in C++ to group members by access rights and I changed the code. This way it became a bit more convenient for _users_ of this class rather than _implementors_.

src/Qir/Runtime/lib/QIR/QubitManagerRestrictedReuse.hpp
DmitryVasilevsky: Yes. Actually we don't need any clarifications. It's the only qubit manager we provide. So it's going to be just CQubitManager.

src/Qir/Runtime/lib/QIR/QubitManagerRestrictedReuse.hpp
DmitryVasilevsky: As I commented earlier this is the C# heritage. I rearranged the code now.

src/Qir/Runtime/lib/QIR/QubitManagerRestrictedReuse.hpp
DmitryVasilevsky: The qubit manager will be in addition to the wrapper. I will replace the simple ++ logic of qubit allocation. Yes, it would be beneficial to designate a separate type for qubit id. I added QubitIdType. I made it signed. See separate discussion about signed/unsigned.

src/Qir/Runtime/lib/QIR/QubitManagerRestrictedReuse.hpp
DmitryVasilevsky: I don't think a separate interface is necessary. Borrowing functions are fully supported now, they just do not do borrowing. When user needs a qubit to borrow, we can't guarantee that such qubit exists; and if it does not, a new fresh qubit is allocated. So current implementation of borrowing is just takes this fallback approach. So it may be better to say that _true_ borrowing isn't supported. I removed remarks about borrowing in a few places.

src/Qir/Runtime/lib/QIR/QubitManagerRestrictedReuse.hpp
DmitryVasilevsky: IQubitManager was an potential integration point. I don't think we need it at the moment as it just gets in the way. When we figure out our extensibility story we will extract the interface. I removed IQubitManager for now.

src/Qir/Runtime/lib/QIR/QubitManagerRestrictedReuse.hpp
DmitryVasilevsky: I will actually unresolve this as I'm still contemplating making auxiliary classes nested in CQubitManager.

src/Qir/Runtime/lib/QIR/QubitManagerRestrictedReuse.cpp
DmitryVasilevsky: I made a few members const that seem to have no reason to modify containing classes. Although that required QubitToId to become const as well. And that's a virtual function. It doesn't modify containing class in my implementation, but I can imagine some overrides may actually need it. OK, let's be more restrictive for now.

src/Qir/Runtime/lib/QIR/QubitManagerRestrictedReuse.cpp
DmitryVasilevsky: This was a direct conversion from C#. And it wasn't a good idea to allocate array and return it anyway. I changed the code to accept the array as an input parameter instead. That way user may decide to use statically allocated (or stack allocated) array and avoid dynamic allocation altogether.

src/Qir/Runtime/lib/QIR/QubitManager.cpp
DmitryVasilevsky: Thanks. Probably the last place that I overlooked.

src/Qir/Runtime/lib/Simulators/FullstateSimulator.cpp
DmitryVasilevsky: Makes sense. Automatic memory management is a good idea when it is applicable.

src/Qir/Runtime/lib/QIR/QubitManager.cpp
DmitryVasilevsky: Agreed, MaximumQubitCapacity is a better id here. Potential qubit id range is from 0 to MaximumQubitCapacity - 1.

src/Qir/Runtime/lib/QIR/QubitManager.cpp
DmitryVasilevsky: Not really. First, we would have to have this check for robustness anyway. So we cannot just remove the check. Also even if qubitCountToAllocate > freeQubitCount, we may still be able to allocate if we are allowed to extend capacity.

src/Qir/Runtime/lib/QIR/QubitManager.hpp
DmitryVasilevsky: A note on the use of signed vs. unsigned integers for qubit ids.

First, I would like to establish that indexes and capacities should be of the same type. This is simply because they are compared. Comparing signed/unsigned will result in a mismatch.

Second, capacities and counts should be the same as well as they are also compared.

Third, the indexes and the values stored in the status array should be of the same type because I do store indexes directly.

So, indexes, stored values, counts and capacities should be of the same type. Can they be all unsigned? Yes. Can they all be signed? Yes. Both variants are possible. However, I would argue that signed type is more convenient.

One problem with unsigned types is that one of the typical iteration patterns doesn’t work. Specifically a loop that iterates indexes backwards may not always work:
`for (int i = count-1; i >= 0; i-- ) {}`
This will work perfectly for a signed type but will loop forever with an unsigned type.

Working with non-negative integers may occasionally result in a small negative integer (For example, x-1 < x is always true for non-negative signed integers). For signed integers, an absolute difference between any two positive integers will always fit into a signed integer (abs(index1-index2) is always what you want for non-negative signed integers). Counting items may occasionally get negative and a check count >= 0 is more intuitive that what it would have to be otherwise. All of this is because with signed integers you typically work in the middle of the range rather than on the boundary.

Another remark specific to the qubit manager is that the range of values stored in the array can be naturally split between non-negative and negative integers. Non-negative integers are used for free qubits and negative integers are used for other statuses. A simple comparison value >= 0 establishes if a qubit is free or not.

In the end, all of this is possible with unsigned integers and there’re different problems with signed integers that can be overlooked. But having worked with STL (which uses unsigned integers) and C# (which uses signed integers) I would prefer signed integers.

A few pointers to discussions can be found here: https://stackoverflow.com/questions/3658560/why-is-count-not-an-unsigned-integer

src/Qir/Runtime/lib/QIR/QubitManagerRestrictedReuse.hpp
DmitryVasilevsky: I moved all auxiliary classes, types and constants to be nested in CQubitManager. This made it more clear that these classes are bound to the Qubit Manager.

src/Qir/Runtime/lib/QIR/QubitManager.cpp
DmitryVasilevsky: Added non-static data member initialization and initialization lists where appropriate.

src/Qir/Runtime/public/QubitManager.hpp
DmitryVasilevsky: This is not a pure stack. While Push and Pop are only done only LIFO style, we also iterate it back and forth. I'll keep the names. I did switch to std iterator for iteration.

src/Qir/Runtime/lib/QIR/QubitManager.cpp
DmitryVasilevsky: This is not part of a public interface and is used on an object not intended for multithreading. So this is acceptable.
Yes, in general, this code would definitely be unacceptable for multithreading. Also, if part of the public interface, it would not be a good idea to have such a function as it would be impossible to tell if the element was actually removed in case of exception thrown.

src/Qir/Runtime/public/QubitManager.hpp
DmitryVasilevsky: Signed vs unsigned is now explained (see QubitIdType comment). This class is not part of the interface, but I think it is beneficial to use int32_t instead of int as it will simplify any wrapper that will expose this functionality. I kept the bool though - it's convenient and can be easily handled in the wrapper if needed.

src/Qir/Runtime/lib/QIR/QubitManager.hpp
DmitryVasilevsky: Updated comments on the classes, markers and added comment to the sharedQubitStatusArray. I believe I provided enough answers now.

DmitryVasilevsky: Values sored are indexes. So these are not separate concepts. See QubitIdType comment.

src/Qir/Runtime/lib/QIR/QubitManager.hpp
kuzminrobin: Thank you very much, this is a very good topic to know.

src/Qir/Runtime/lib/QIR/QubitManager.cpp
```diff
@@ -0,0 +1,508 @@
...
+CQubitManager::QubitIdType CQubitManager::GetQubitId(Qubit qubit) const
+{
+    FailIf(!IsValidQubit(qubit), "Not a valid qubit.");
+    return QubitToId(qubit);
+}
+
+
+Qubit CQubitManager::CreateQubitObject(QubitIdType id)
+{
+    FailIf(id < 0 || id > INTPTR_MAX, "Qubit id is out of range.");
```
kuzminrobin, line 384: `INTPTR_MAX` is a 64-bit value on a 64-bit platform, but `id` is 32-bit.
```suggestion
    FailIf(id < 0 || id >= MaximumQubitCapacity, "Qubit id is out of range.");
```

kuzminrobin: Or maybe even `qubitCapacity` instead of `MaximumQubitCapacity`.

src/Qir/Runtime/lib/QIR/QubitManager.cpp
```diff
@@ -0,0 +1,508 @@
...
+
+void CQubitManager::DeleteQubitObject(Qubit qubit)
+{
+    // Do nothing. By default we store qubit Id in place of a pointer to a qubit.
+}
+
+CQubitManager::QubitIdType CQubitManager::QubitToId(Qubit qubit) const
+{
+    intptr_t pointerSizedId = reinterpret_cast<intptr_t>(qubit);
+    FailIf(pointerSizedId < 0 || pointerSizedId > std::numeric_limits<QubitIdType>::max(), "Qubit id is out of range.");
```
kuzminrobin, line 397: Did you mean
```suggestion
    FailIf(pointerSizedId < 0 || pointerSizedId >= MaximumQubitCapacity, "Qubit id is out of range.");
```
or
```suggestion
    FailIf(pointerSizedId < 0 || pointerSizedId >= qubitCapacity, "Qubit id is out of range.");
```
?

src/Qir/Runtime/lib/QIR/QubitManager.cpp
```diff
@@ -0,0 +1,508 @@
...
+    } else
+    {
+        // Released qubits are added to reuse area/segment in which they were released
+        // (rather than area/segment where they are allocated).
+        // Although counterintuitive, this makes code simple.
+        // This is reasonable because qubits should be allocated and released in the same segment. (This is not enforced)
+        freeQubitsInAreas.PeekBack().FreeQubitsReuseAllowed.AddQubit(id, encourageReuse, sharedQubitStatusArray);
+    }
+
+    freeQubitCount++;
```
kuzminrobin, line 486: I have a suspicion that `freeQubitCount++` (line 486) should only happen in the `else` part (lines 478 - 484), i.e. only if the qubit will be _reused_.

kuzminrobin:
Reapproving with minor comments.

src/Qir/Runtime/lib/QIR/QubitManager.cpp
DmitryVasilevsky: The intention of the check is to make sure the qubit id fits into intptr_t to which it is casted on the next line. This is a good candidate for a debug only check or even static assert.

DmitryVasilevsky: On a second though, this should stay as is, at least for now. Regardless of how wide QubitIdType and intptr_t types are, this ensures that static cast that follows works just fine as it checks that the value will fit.

src/Qir/Runtime/lib/QIR/QubitManager.cpp
DmitryVasilevsky: Again, the intention of this check is to verify that the value we are casting in the next line will fit into the variable of destination type.

src/Qir/Runtime/lib/QIR/QubitManager.cpp
DmitryVasilevsky: This is a great observation and a great question. It comes down to the definition of what freeQubitCount is, which is, unfortunately, not provided anywhere! freeQubitCount is the count of free qubits tracked by qubit manager. Technically these qubits are tracked so increase is always needed. Similar question may be asked - why freeQubitCount is let's say 4, when qubit manager may extend capacity? I will add comments.

src/Qir/Runtime/lib/QIR/QubitManager.cpp
kuzminrobin: FYI:
```
pwsh bootstrap.ps1

. . .
[14/35] Building CXX object lib/QIR/CMakeFiles/qir-rt-support-obj.dir/QubitManager.cpp.obj
../../lib/QIR/QubitManager.cpp:395:25: warning: result of comparison of constant 9223372036854775807 with expression of type 'Microsoft::Quantum::CQubitManager::QubitIdType' (aka 'int') is always false [-Wtautological-constant-out-of-range-compare]
    FailIf(id < 0 || id > std::numeric_limits<intptr_t>::max(), "Qubit id is out of range.");
                     ~~ ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 warning generated.
```
If you get a chance soon please fix.
(we are using clang-11, but locally I have clang 12 and it reports. So you might not be able to repro with clang-11)

kuzminrobin: FYI: Has been changed in [my PR](https://github.com/microsoft/qsharp-runtime/pull/718#discussion_r647046753). Feel free to change/correct in the subsequent PRs.

PR #697, opened 2021-05-19
Replace generated files in Quantum.Client with reference to the Azure.QuantumJobs package

anpaz:


src/Quantum.Development.Kit/Microsoft.Quantum.Development.Kit.nuspec
```diff
@@ -20,7 +20,7 @@
     <tags>Quantum Q# QSharp</tags>
 
     <dependencies>
-      <dependency id="Microsoft.Quantum.Simulators" version="$version$" />
+      <dependency id="Microsoft.Quantum.Simulators" version="0.16.2105140472" />
```
vxfield, line 23: Is this intentional? Or we should leave `$version$` instead?

src/Xunit/Microsoft.Quantum.Xunit.nuspec
```diff
@@ -22,7 +22,7 @@
     <dependencies>
       <dependency id="xunit" version="2.3.1" />
       <dependency id="Microsoft.NET.Test.Sdk" version="15.3.0" />
-      <dependency id="Microsoft.Quantum.Runtime.Core" version="0.16.2104138281-beta" />
+      <dependency id="Microsoft.Quantum.Runtime.Core" version="0.16.2105140472" />
```
vxfield, line 25: Is this intentional? Or we should leave `$version$` instead?

src/Simulation/CSharpGeneration/Microsoft.Quantum.CSharpGeneration.fsproj
```diff
@@ -22,7 +22,7 @@
 
   <ItemGroup>
     <PackageReference Update="FSharp.Core" Version="4.7.0" />
-    <PackageReference Include="Microsoft.Quantum.Compiler" Version="0.16.2104138281-beta" />
+    <PackageReference Include="Microsoft.Quantum.Compiler" Version="0.16.2105140472" />
```
vxfield, line 25: Is this intentional? Or we should leave `$version$` instead?

vxfield (commented):
Looks good to me.
Just a question whether the `.nuspec` files should have `$version$` instead of fixed version numbers.

src/Azure/Azure.Quantum.Client.Test/Helpers/TestConstants.cs
```diff
@@ -5,10 +5,11 @@ namespace Microsoft.Azure.Quantum.Test
 {
     public static class TestConstants
     {
-        public const string SubscriptionId = "sub1";
-        public const string ResourceGroupName = "rg1";
-        public const string WorkspaceName = "ws1";
-        public const string ProviderId = "provider1";
-        public const string Endpoint = "https://test";
+        // Used when connecting live
+        public const string LiveSubscriptionId = "916dfd6d-030c-4bd9-b579-7bb6d1926e97";
```
vxfield, line 9: Shouldn't this values live in Environment Variables instead of the code?

src/Azure/Azure.Quantum.Client.Test/WorkspaceTest.cs
```diff
@@ -3,246 +3,239 @@
...
-            string newJobId = Guid.NewGuid().ToString();
-            SetJobResponseMessage(newJobId);
+            try
+            {
+                var result = workspace.CancelJob(job.Id);
+                AssertJob(result);
+            }
+            catch (WorkspaceClientException e)
+            {
+                Assert.AreEqual((int)HttpStatusCode.Conflict, e.Status);
```
vxfield, line 80: Is this exception expected to happen? If yes, this would be better `Assert.Throws<Exception>(() => SomethingThatThrowsAnException());`

src/Azure/Azure.Quantum.Client/JobManagement/Workspace.cs
```diff
@@ -21,159 +24,59 @@ namespace Microsoft.Azure.Quantum
...
             string workspaceName,
             string location,
-            TokenCredential tokenCredential = null)
-            : this(
-                subscriptionId,
-                resourceGroupName,
-                workspaceName,
-                tokenCredential,
-                new Uri($"https://{location}.{Constants.DefaultLocationlessEndpoint}/"))
+            TokenCredential credential = null,
```
vxfield, line 41:
```suggestion
            TokenCredential credential = new DefaultAzureCredential(true),
```

Shouldn't we set `DefaultAzureCredential` as default here?

vxfield: Never mind, I see you are dealing with it in the method implementation.

src/Quantum.Development.Kit/Microsoft.Quantum.Development.Kit.nuspec
```diff
@@ -20,7 +20,7 @@
     <tags>Quantum Q# QSharp</tags>
 
     <dependencies>
-      <dependency id="Microsoft.Quantum.Simulators" version="$version$" />
+      <dependency id="Microsoft.Quantum.Simulators" version="0.16.2105143425-alpha" />
```
vxfield, line 23: Just checking if the removal of `$version$` is intentional.

vxfield (approved):
Approved with comments

src/Quantum.Development.Kit/Microsoft.Quantum.Development.Kit.nuspec
anpaz: In general, it is better if it uses $version$ so they are all in-sync. In practice, this get updated with updateQDKVersion (and this package is deprecated), so I'm going keep the specific version in it.

ScottCarda-MS (approved):
LGTM

src/Azure/Azure.Quantum.Client.Test/Helpers/TestConstants.cs
anpaz: I have changed the tests to give preference to the same env variables we use for e2e.

PR #700, opened 2021-05-20
Ignore Empty Library Files During C# Generation

ScottCarda-MS:
Puts the null check back in to prevent empty library files from getting generated during C# generation.


swernli (approved):
LGTM! Thanks for fixing this so quickly!

PR #703, opened 2021-05-24
QIR runtime heap alloc and heap free functions removed

### LINKED ISSUE #702: QIR runtime heap alloc and heap free functions should be removed
The QIR runtime currently includes functions for allocating and freeing memory from the heap:
https://github.com/microsoft/qsharp-runtime/blob/380003f44cea08ae0b91132ad168ea34ba56297d/src/Qir/Runtime/lib/QIR/bridge-rt.ll#L124
https://github.com/microsoft/qsharp-runtime/blob/380003f44cea08ae0b91132ad168ea34ba56297d/src/Qir/Runtime/lib/QIR/bridge-rt.ll#L129

These two methods are neither part of the QIR specification, nor are they used anywhere in the runtime implementation or in QIR generation, making them dead code. If we expect them to be needed in the future then we should update the QIR spec to describe their intended use, otherwise we should remove them and the underlying support to avoid the cost of maintaining them.

@bettinaheim @alan-geller @kuzminrobin FYI.

Patil2099:
Fixes #702

### DIFF
--- a/src/Qir/Runtime/lib/QIR/bridge-rt.ll
+++ b/src/Qir/Runtime/lib/QIR/bridge-rt.ll
@@ -121,15 +121,6 @@ declare void @quantum__rt__message(%"struct.QirString"* %str)
 ;------------------------------------------------------------------------------
 ; classical bridge
 ;
-define dllexport i8* @__quantum__rt__heap_alloc(i64 %size) {
-  %mem = call i8* @quantum__rt__heap_alloc(i64 %size)
-  ret i8* %mem
-}
-
-define dllexport void @__quantum__rt__heap_free(i8* %mem) {
-  call void @quantum__rt__heap_free(i8* %mem)
-  ret void
-}
 
 ; Returns a pointer to the malloc-allocated block.
 define dllexport i8* @__quantum__rt__memory_allocate(i64 %size) {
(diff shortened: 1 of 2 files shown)


swernli:
This is the right start, but there are a few other spots with code that can be removed as well. Now that `__quantum__rt__heap_alloc` and `__quantum__rt__heap_free` are removed, the declaration and implementation of the internal functions backing them can also be removed. Those are located in these two places:
https://github.com/microsoft/qsharp-runtime/blob/380003f44cea08ae0b91132ad168ea34ba56297d/src/Qir/Runtime/lib/QIR/bridge-rt.ll#L40-L41
https://github.com/microsoft/qsharp-runtime/blob/380003f44cea08ae0b91132ad168ea34ba56297d/src/Qir/Runtime/lib/QIR/utils.cpp#L25-L47

swernli (commented):
Thanks for the quick updates! The changes look good to me, I'll formally sign off to unblock merge once the issue is triaged.

Patil2099:
Can you review it again? @swernli

swernli (approved):

swernli:
Apologies, @Patil2099, we seem to be having some problems in the build pipeline unrelated to this change. Hopefully I can get a clean build soon.

PR #704, opened 2021-06-01
QIR Command Line Tool - Automatically Use Required Resources

ScottCarda-MS:
Adds packaging functionality, and have the tool automatically include the required headers and libraries in addition to any resources the user explicitly pass to the tool.


src/Qir/Execution/QirCommandLineTool/Program.cs
```diff
@@ -44,7 +46,36 @@ private static Command CreateBuildCommand()
...
+                            osID = "win-x64";
+                            break;
+                        case PlatformID.Unix:
+                            osID = "linux-x64";
+                            break;
+                        case PlatformID.MacOSX:
+                            osID = "osx-x64";
+                            break;
+                        default:
+                            throw new ArgumentException("TODO");
```
ScottCarda-MS, line 72: TODO: Write error message

src/Qir/Execution/QirCommandLineTool/Program.cs
```diff
@@ -44,7 +46,36 @@ private static Command CreateBuildCommand()
             {
                 Handler = CommandHandler.Create((BuildOptions settings) =>
                 {
-                    return QirTools.BuildFromQSharpDll(settings.QSharpDll, settings.LibraryDirectories, settings.IncludeDirectories, settings.ExecutablesDirectory);
+                    var thisModulePath = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
+                    if (string.IsNullOrWhiteSpace(thisModulePath))
+                    {
+                        throw new InvalidOperationException("TODO");
```
ScottCarda-MS, line 52: TODO: write error message

src/Qir/Execution/QirCommandLineTool/Program.cs
```diff
@@ -44,7 +46,36 @@ private static Command CreateBuildCommand()
...
+                    }
+
+                    var headersPath = new DirectoryInfo(Path.Combine(thisModulePath, "runtimes", "any", "native", "include"));
+                    var externalsPath = new DirectoryInfo(Path.Combine(thisModulePath, "Externals", "CLI11"));
+                    
+                    var osArch = Environment.OSVersion.Platform;
+                    var osID = "";
```
ScottCarda-MS, line 59: I think there is some nicer switch syntax that could be used here.

src/Qir/Execution/QirCommandLineTool/QirCommandLineTool.csproj
```diff
@@ -1,15 +1,43 @@
...
     <PackAsTool>true</PackAsTool>
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="System.CommandLine" Version="2.0.0-beta1.20213.1" />
-    <ProjectReference Include="..\..\..\Simulation\Simulators\Microsoft.Quantum.Simulators.csproj" />
-    <ProjectReference Include="..\Tools\Microsoft.Quantum.Qir.Tools.csproj" />
+    <PackageReference Include="System.CommandLine" Version="2.0.0-beta1.20213.1"/>
+    <ProjectReference Include="..\Tools\Microsoft.Quantum.Qir.Tools.csproj"/>
+    <PackageReference Include="Microsoft.Quantum.Qir.Runtime" Version="0.16.210526824-alpha"/>
```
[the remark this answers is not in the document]
swernli: Yes, though it requires that the QIR Runtime has built already. If you take a look at the build script:
https://github.com/microsoft/qsharp-runtime/blob/85bbb3eaa4755510d224ae53cc5acaa65e956fa1/src/Qir/Runtime/build-qir-runtime.ps1#L18-L28
it includes logic to copy into the drops folder, as used by the e2e build to collect build artifacts. So you can use the path `src/Qir/drops/bin/**` to grab all the binaries available. Locally, this will just get the ones matching your local build, whereas the e2e will get all three platforms.

src/Qir/Execution/QirCommandLineTool/Program.cs
ScottCarda-MS: done

src/Qir/Execution/QirCommandLineTool/Program.cs
ScottCarda-MS: done

src/Qir/Execution/QirCommandLineTool/Program.cs
ScottCarda-MS: found better syntax

src/Qir/Execution/QirCommandLineTool/QirCommandLineTool.csproj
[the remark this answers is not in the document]
ScottCarda-MS: Is there some way to add a dependency on the QIR Runtime build without having a package reference, to ensure that we have these libraries available?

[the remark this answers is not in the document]
swernli: No, there really isn't. But this same problem exists for the simulator which you already take a dependency on below. Both the simulator and the QIR Runtime are built during an early step of the build scripts (which in my opinion is always how you should build locally to match the CI pipeline), so this pattern is something we already use and depend on.

That said, it looks like the QIR Runtime is NOT built in the bootstrap.ps1 script, which it should be, since that is how the simulator mitigates this dependency. I'll work on a PR to fix that.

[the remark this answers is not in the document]
swernli: #706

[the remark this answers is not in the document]
cesarzc: I was thinking that it might be better to include QIR runtime libraries and headers as part of the `Microsoft.Quantum.Qir.Tools` project rather than the command line project. Since the tools code already assumes that the QIR runtime libraries with specific names exist, I believe it makes sense to distribute the package with the libraries too. It is also more convenient for users of QIR tools because for cases in which the full state simulator is used, they wouldn't have to specify where libraries are located.

Additionally, the CLI11 header is already part of the `Microsoft.Quantum.Qir.Tools` package (see [here](https://github.com/microsoft/qsharp-runtime/blob/1b514cc7e180ffb46b4c6505a90083b86f3db9b4/src/Qir/Execution/Tools/Microsoft.Quantum.Qir.Tools.csproj#L18)).

What do you think @swernli, @ScottCarda-MS?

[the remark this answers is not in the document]
ScottCarda-MS: I've removed the package reference and added logic similar to that used to get the simulator dll. Thanks @swernli!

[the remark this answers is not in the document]
swernli: Huh, that's an interesting question, @cesarzc. I don't have a strong opinion which package should or shouldn't have it, and since it's so small there isn't much cost to duplicating the dlls into multiple projects. If QIR.Tools does pull the runtime as a dependency then it shouldn't also need to be pulled here, as the dlls and headers will get copied as a transitive dependency, so that would allow some simplification, so I guess it would make sense to have one or the other include them but not both.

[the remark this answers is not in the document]
ScottCarda-MS: I've moved the itemgroups for including these resources to the Qir.Tools project, as suggested.

ScottCarda-MS:
I believe @cesarzc has suggested that the logic for using these included resources be moved to the Qir.Tools project.

src/Qir/Execution/QirCommandLineTool/Program.cs
```diff
@@ -44,7 +46,27 @@ private static Command CreateBuildCommand()
             {
                 Handler = CommandHandler.Create((BuildOptions settings) =>
                 {
-                    return QirTools.BuildFromQSharpDll(settings.QSharpDll, settings.LibraryDirectories, settings.IncludeDirectories, settings.ExecutablesDirectory);
+                    var thisModulePath = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
```
ScottCarda-MS, line 49: I believe @cesarzc has suggested that the logic for using these included resources be moved to the Qir.Tools project.

ScottCarda-MS: The logic has been moved.

src/Qir/Execution/Tools/Executable/QirFullStateExecutable.cs
```diff
@@ -13,20 +16,40 @@ namespace Microsoft.Quantum.Qir.Tools.Executable
...
-            };
+            var osID = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? "win-x64"
+                : RuntimeInformation.IsOSPlatform(OSPlatform.Linux) ? "linux-x64"
+                : RuntimeInformation.IsOSPlatform(OSPlatform.OSX) ? "osx-x64"
+                : throw new ArgumentException("Unsupported operating system architecture.");
 
-        public override string DriverFileExtension => "cpp";
+            LibraryDirectories.Add(new DirectoryInfo(Path.Combine(thisModulePath, "runtimes", osID, "native")));
+            LibraryDirectories.Add(new DirectoryInfo(Path.Combine(thisModulePath, "Libraries", osID)));
+            LibraryDirectories.Add(new DirectoryInfo(Path.Combine(thisModulePath, "Libraries")));
```
swernli, line 52: Since library directories are copied into the bin folder, won't this have the effect of copying the contents of `Libraries/<osId>` twice, once directly into the bin folder and once as subdir of the bin folder?

ScottCarda-MS: These are the directories that get passed to the clang compiler. I don't believe it checks for files recursively, and I don't think we do on our end either, so the reference to the Libraries directory is not sufficient to get the files in the subdirectory by itself. The simulator library file is directly under the Libraries directory, while other platform-specific libraries are further nested. We tried adding the simulator library to the platform-specific subdirectories, but that had other side-effects related to how MSBuild expects the files to be organized. Listing both directories here should not cause duplicate copies to be copied to the output folder of a Q# program compiled with the tool.

swernli: Oh, I see! The subdirectory is not copied by `CopyDirectoryContents` because that method doesn't include folders. That makes sense, thanks for explaining!

swernli (approved):
LGTM!

PR #706, opened 2021-06-02
Add QIR Runtime to bootstrap script

swernli:
This ensures that the release version of the QIR Runtime is present in the drops folder for other build steps that might depend on it, and matches the pattern we use for the native simulator dlls.


ScottCarda-MS (approved):
lgtm

PR #708, opened 2021-06-02
Increasing the CMAKE required version for QIR

swernli:
This updates the required version for CMAKE in the CMakeLists.txt files for QIR, since we are now using features not correctly supported on older versions.


cgranade (approved):
Looks good to me! This should help avoid some of the confusing errors that come up with 3.18.1; thanks for doing this!

Issue #504, opened 2021-02-06
Noisy simulation in Q#


crazy4pi314:
**Please describe what you would like the feature to accomplish.**  
I would like a way to simulate real/noisy hardware in Q#. 

**Describe the solution you'd like**  
I am working now full time developing [Mitiq](https://github.com/unitaryfund/mitiq), a Python library for doing error mitigation techniques for quantum programs. The reason we have gone with Python and other tools like Circ and Qiskit to date is because they have built-in ways to simulate the noise we are trying to mitigate. I am running into other situations where I would rather try and implement these techniques in Q# (one such being compiling to QIR) but without a noisy simulator it would be pretty much impossible.

**Describe alternatives you've considered**  
I could try and write my own noisy simulator backend for Q#, or try and hook it up to other existing noisy simulators via the simulator interface. This is a well explored area and basically I don't want to duplicate effort on.

**Additional context**  
The reason doing this in Q# is really interesting to me is that I need to design experiments where sending an entire program that can generate random characterization gate sequences _on the device_ is **essential**. I am constantly fighting other AIPs where I can only send individual quantum circuits and are thus limited in how much data I can get per request.

cgranade:
Thanks for the suggestion, @crazy4pi314! If you're interested, I've been working on an **experimental** and **untested** approach at https://github.com/microsoft/qsharp-runtime/tree/cgranade/experimental/opensim. This experiment may not go anywhere in particular, and there's currently not a working build for it, but I'd be very interested in your feedback on whether something along those _rough_ lines would be useful to you in your work on Mitiq.

I'll keep you posted here if and when there's an **experimental** alpha build you can try, but until then, thanks again for the feature request!

cgranade:
OK, here's an **experimental** build for Q# standalone and Q# + C# (no IQ# or Python support yet) that should be able to run some simple Q# applications (e.g.: only CNOT as far as two-qubit gates go, no single-qubit rotations, so forth):

```xml
<!-- example.csproj -->
<Project Sdk="Microsoft.Quantum.Sdk/0.15.210223293-alpha">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>netcoreapp3.1</TargetFramework>
    <DefaultSimulator>Microsoft.Quantum.Experimental.OpenSystemsSimulator</DefaultSimulator>
  </PropertyGroup>
</Project>
```

```xml
<!-- NuGet.Config -->
[...]

PR #709, opened 2021-06-03
Start experimental simulators feature

cgranade:
This PR starts a new feature branch for **experimental** open systems and stabilizer / CHP simulator work. As this PR represents the start of that work, there's a fair bit that still needs done, but I wanted to open the PR for **early initial feedback** on the design of these experimental features.

The core functionality for the simulator is written as a Rust library with Python and native C APIs, along with a C# class to expose this library to Quantum Development Kit users. The Rust portions of the library are tested through unit and integration tests, and are checked against rustfmt and clippy for style and correctness. Performance of the Rust library is measured using criterion.rs,  both at the C API boundary, and in microbenchmarks that exercise internal details; unit test and performance benchmarking results are integrated into CI.

TODO:
- [ ] Add decompositions for `Controlled H` and other similar operations.
- [ ] Finish exposing stabilizer noise models with new `ChpDecomposition` variant for `qdk_sim::Process`
- [ ] Finish C# and Python representations of noise models, allowing calling into stabilizer simulation from Python or C# hosts..
- [ ] Set up deployment pipeline for crate and wheel artifacts.
- [ ] Finish including new simulators in e2e build.
- [ ] Enable control flow guard at C / Rust boundaries.
- [ ] Improve performance of applying quantum processes.
- [ ] Improve test coverage across the board.

See https://github.com/microsoft/qsharp-runtime/issues/504 for further discussion, and matching PR at https://github.com/microsoft/iqsharp/pull/449.


swernli:
This is really exciting to see, great work! To understand the process here, is the checklist above the set of things you want to accomplish before merging this PR into the feature branch or the list of things you'd like to accomplish before merging the feature branch into main? I think either approach would work, but there are trade-offs in terms of tracking, review, and PR overhead since feature branches trigger builds (but do not require them to succeed to merge).

.pre-commit-config.yaml
```diff
@@ -0,0 +1,11 @@
+repos:
+  - repo: https://github.com/doublify/pre-commit-rust
+    rev: v1.0
+    hooks:
+      - id: fmt
+        args: ['--manifest-path', 'src/Simulation/qdk_sim_rs/Cargo.toml', '--']
+      - id: cargo-check
+        args: ['--manifest-path', 'src/Simulation/qdk_sim_rs/Cargo.toml', '--']
+      - id: clippy
```
swernli, line 9: I'm actually not familiar with clippy, any links you'd recommend to get more info on it?

bootstrap.ps1
```diff
@@ -7,13 +7,28 @@ Push-Location (Join-Path $PSScriptRoot "build")
...
+    # Since that Cargo.toml is referenced by CMake lists in the QIR
+    # runtime, this injection has to be the first thing we do.
+    dotnet tool restore
+    dotnet script inject-version.csx -- `
+        --template Cargo.toml.template `
+        --out-path Cargo.toml `
+        --version $Env:NUGET_VERSION;
+Pop-Location
+
 if (-not (Test-Path Env:AGENT_OS)) {
```
swernli, line 22: I just recently added a pre-emptive build of the QIR Runtime to this block (see #706). Just calling that out to make sure that when that change gets merged to the working branch or feature branch it is done in the right order relative to the qdk_sim_rs added below.

src/Qir/Runtime/.gitignore
```diff
@@ -0,0 +1 @@
+build
```
swernli, line 1: Is this file still needed? As of https://github.com/microsoft/qsharp-runtime/pull/647 I changed the CMAKE output folder to `bin` instead of `build` to ensure that the SDL pipeline picked up the generated binaries for validation, so the build folder should no longer be used for anything.

src/Simulation/qdk_sim_rs/.gitignore
```diff
@@ -0,0 +1,12 @@
+linux
+osx
+win10
+target
+
+# We inject version numbers into Cargo.toml, so don't want them stored in repo.
```
swernli, line 6: I'm curious why you chose this strategy. Is it possible to leave the Cargo.toml checked in and instead have it load version numbers from environment variables or something similar? I only ask because I worry that having Cargo.toml not checked in might set us up for cases where dev tools helpfully automatically update the local file and the change gets missed in commit and causes a failing PR build.

src/Simulation/qdk_sim_rs/Cargo.lock
```diff
@@ -0,0 +1,1212 @@
+# This file is automatically @generated by Cargo.
```
swernli, line 1: I think this file is usually added to the gitignore, since it is generated by the build.

.pre-commit-config.yaml
cgranade: Ah, yes, let me add that link... thanks for calling that out! In the meantime, https://github.com/rust-lang/rust-clippy and the list of lints checked by clippy at https://rust-lang.github.io/rust-clippy/master/index.html should be helpful.

src/Simulation/qdk_sim_rs/.gitignore
cgranade: Agreed entirely. I was unable to find a way to inject version numbers from the environment, but if you have any suggestions I would strongly prefer to do it that way for all the reasons you lay out in your comment.

src/Simulation/qdk_sim_rs/benches/c_api_benchmark.rs
```diff
@@ -0,0 +1,70 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! This set of benchmarks exercises the open systems simulator exclusively via
```
swernli, line 4: This benchmarking is really cool! Not for this PR, but I wonder if there is a way we could generalize some of this (or at least put a guide somewhere in the repo as a readme) so that future Rust components can easily include their own benchmarks?

src/Simulation/qdk_sim_rs/Cargo.lock
cgranade: Depends on if we build with `--locked`. That would be great to do for reproducible builds, but that's unfortunately not really consistent with generating Cargo.toml on the fly.

build/set-env.ps1
```diff
@@ -20,10 +20,27 @@ if ($Env:ENABLE_QIRRUNTIME -ne "false" -and $Env:QIR_DROPS -eq $null) {
 }
 
 If ($Env:DROPS_DIR -eq $null) { $Env:DROPS_DIR =  [IO.Path]::GetFullPath((Join-Path $PSScriptRoot "..\drops")) }
+if ($Env:DROP_NATIVE -eq $null) {
+    $Env:DROP_NATIVE = (Join-Path $PSScriptRoot ".." "xplat")
```
swernli, line 24: Hmm... I think this may have been a bug previously that accidentally worked because of how the e2e builds consume the native binaries, and in fixing the value to be correctly populated in this repo it might not coexist well with the e2e logic. Either way, I think the behavior of copying the simulators to this folder are no longer needed for the existing simulator, but might mean that the OpenSim won't show up in the right places in the e2e build. Let's sync up on this one to make sure we understand which pattern the OpenSim should use, and perhaps have a separate PR to remove redundant logic from within this repo.

src/Simulation/qdk_sim_rs/.gitignore
swernli: Ah, I wasn't sure if you had already tried something like that. It's certainly possible the template-style approach is our best bet then. I'll circle back if I happen to find another approach worth trying!

build/set-env.ps1
cgranade: Sounds good, would much appreciate your help. At the moment, if this line is removed, then the builds pass but make a NuGet package that's missing the required DLL; would love to figure out the right way to do things instead.

src/Simulation/qdk_sim_rs/.gitignore
cgranade: Resolving for now, but happy to revisit this if you find anything. Would definitely love to have something more robust than templating!

src/Qir/Runtime/.gitignore
cgranade: Good point!

bootstrap.ps1
cgranade: Looks like catching up to main went cleanly, so will spin up a new build with that change. Thanks!

cgranade:
As a quick note, end-to-end builds are currently expected to fail, as the logic required to coordinate cross-platform builds of Rust components has not yet been added to the end-to-end pipeline. That logic will hopefully be added soon, allowing end-to-end builds to proceed correctly.

src/Qir/Tests/OpenSystemsSimulator/OpenSystemsSimulatorTests.cpp
```diff
@@ -0,0 +1,36 @@
...
+#include "catch.hpp"
+
+#include "QuantumApi_I.hpp"
+#include "SimFactory.hpp"
+#include "context.hpp"
+
+using namespace Microsoft::Quantum;
+using namespace std;
+
+TEST_CASE("Open systems simulator: X and measure", "[open_simulator]")
```
efratshabtai, line 18: Please make sure you cover all of the implemented functionality.

cgranade: Yes, absolutely. These tests are for the QIR runtime integration, which won't be supported in the initial version, given focus on usage from Python host programs for now.

src/Simulation/qdk_sim_rs/README.md
```diff
@@ -0,0 +1,41 @@
+# Quantum Development Kit Experimental Simulators
```
efratshabtai, line 1: Can you please add some usage examples - how will someone call the simulator and get results from Python, C# and C++

cgranade: Definitely agreed, but this may not be the right place for C# examples in particular — this file is the README specifically for the native library portion of the experimental simulators, and the C and Python APIs built from that library (notably the API for calling directly into the simulator, and not the Python + Q# API covered by https://github.com/microsoft/iqsharp/pull/449). Following that, this README.md is formatted for use with rustdoc / https://docs.rs. Will look for appropriate places to add such examples.

cgranade:
As a quick update, the e2e builds are currently blocked on https://github.com/microsoft/QuantumKatas/pull/627.

src/Simulation/qdk_sim_rs/README.md
cgranade: Added significantly to the Rust documentation generated by `rustdoc` or https://docs.rs, and added new .md files explaining the C# bindings. Will add docs on Python +Q# usage (as opposed to the Python-only usage documented here) to the IQ# repo.

src/Qir/Runtime/lib/Simulators/OpenSystemsSimulator.cpp
```diff
@@ -0,0 +1,292 @@
...
+    i64 s_adj(size_t sim_id, size_t idx); // NOLINT(readability-identifier-naming)
+    i64 t(size_t sim_id, size_t idx); // NOLINT(readability-identifier-naming)
+    i64 t_adj(size_t sim_id, size_t idx); // NOLINT(readability-identifier-naming)
+    i64 cnot(size_t sim_id, size_t idx_control, size_t idx_target); // NOLINT(readability-identifier-naming)
+    i64 m(size_t sim_id, size_t idx, size_t* result_out); // NOLINT(readability-identifier-naming)
+    const char* lasterr(); // NOLINT(readability-identifier-naming)
+    const char* get_noise_model(size_t sim_id); // NOLINT(readability-identifier-naming)
+    i64 set_noise_model(size_t sim_id, const char* new_model); // NOLINT(readability-identifier-naming)
+    const char* get_current_state(size_t sim_id); // NOLINT(readability-identifier-naming)
+}
```
kuzminrobin, line 40: I feel that there can be multiple entities (files) that may need this `extern "C"` declarations, but not the rest of this file.
If that's the case or possible scenario then you might want to consider extracting this `extern "C"` block into a separate header.

src/Qir/Runtime/lib/Simulators/OpenSystemsSimulator.cpp
```diff
@@ -0,0 +1,292 @@
...
+
+namespace Microsoft
+{
+namespace Quantum
+{
+    // FIXME: support methods from public IDiagnostics; they currently
+    //        just throw.
+    class COpenSystemSimulator : public IRuntimeDriver, public IQuantumGateSet, public IDiagnostics
```
kuzminrobin, line 48: Consider breaking up the class into a separate definition of the class (in the .hpp header), e.g. like in [QubitManager.hpp](https://github.com/microsoft/qsharp-runtime/blob/main/src/Qir/Runtime/public/QubitManager.hpp),
and the definition of the member functions (in the .cpp source), e.g. like in [QubitManager.cpp](https://github.com/microsoft/qsharp-runtime/blob/main/src/Qir/Runtime/lib/QIR/QubitManager.cpp).

In the definition of the class (.hpp) consider 
* listing the `public`s first, then `protected`s, then `private`s (such that the users of the class see the most used API first);
* grouping all the member variables together in the end of the class in the separate `private` section (such that the users can quickly estimate the size of the class instance and whether all the member variables are initialized in the constructors).

src/Qir/Runtime/lib/Simulators/OpenSystemsSimulator.cpp
```diff
@@ -0,0 +1,292 @@
...
+#include <iostream>
+#include <memory>
+#include <vector>
+
+#include "QirRuntimeApi_I.hpp"
+#include "QSharpSimApi_I.hpp"
+#include "SimFactory.hpp"
+
+using namespace std;
+typedef long long int i64;
```
kuzminrobin, line 16: What is the advantage of `i64` over the [`uint64_t`](https://man7.org/linux/man-pages/man0/stdint.h.0p.html)?

src/Simulation/Simulators/OpenSystemsSimulator/NativeInterface.cs
```diff
@@ -0,0 +1,265 @@
...
+            CheckCall(_Init(initialCapacity, representation, out var simulatorId));
+            return simulatorId;
+        }
+
+        [DllImport(DLL_NAME, ExactSpelling=true, CallingConvention=CallingConvention.Cdecl, EntryPoint="destroy")]
+        private static extern Int64 _Destroy(ulong simId);
+
+        public static void Destroy(ulong simId)
+        {
+            LogCall("init");
```
kuzminrobin, line 94:
```suggestion
            LogCall("destroy");
```

src/Qir/Runtime/lib/Simulators/OpenSystemsSimulator.cpp
cgranade: Agreed. As noted at https://github.com/microsoft/qsharp-runtime/blob/72977df6a9ddbc17b9dcecc3dfb7fd355f348a81/src/Simulation/qdk_sim_rs/docs/c-api.md, it may make sense to even remove this entirely and replace it by a header generated automatically by `cbindgen`.

src/Qir/Runtime/lib/Simulators/OpenSystemsSimulator.cpp
cgranade: This was directly copied from the full-state simulator, then modified for the new API. I think the full-state simulator has diverged fairly substantially since then, including the cleanup that you mention above.

src/Qir/Runtime/lib/Simulators/OpenSystemsSimulator.cpp
cgranade: None in particular; this type is used only for error return codes from C API calls, at which point it's fairly arbitrary whether we assign errors to positive or negative numbers. [In `c_api.rs`](https://github.com/microsoft/qsharp-runtime/blob/72977df6a9ddbc17b9dcecc3dfb7fd355f348a81/src/Simulation/qdk_sim_rs/src/c_api.rs#L41), I used `-1` as the error code, thinking of the convention used by Unix processes. That said, if you'd strongly suggest changing the return code convention to `u64` (Rust) / `uint64_t` (C / C++), I'm happy to do so.

bootstrap.ps1
```diff
@@ -7,13 +7,28 @@ Push-Location (Join-Path $PSScriptRoot "build")
...
     if ($Env:ENABLE_NATIVE -ne "false") {
         Write-Host "Build release flavor of the native simulator"
         $Env:BUILD_CONFIGURATION = "Release"
         Push-Location (Join-Path $PSScriptRoot "src/Simulation/Native")
             .\build-native-simulator.ps1
         Pop-Location
+        Push-Location (Join-Path $PSScriptRoot "src/Simulation/qdk_sim_rs")
+            .\build-qdk-sim-rs.ps1
```
swernli, line 30: Since we don't have logic in build/prerequisites.ps1 that will install Rust automatically, we should either include logic in build-qdk-sim-rs.ps1 to no-op if cargo is not in the path or update the top-level readme.md to indicate that manually installing Rust is a prerequisite. As it is, it looks like this will unconditionally run and then fail during bootstrap for users without Rust installed.

src/Qir/Runtime/prerequisites.ps1
```diff
@@ -8,6 +8,9 @@ if ($Env:ENABLE_QIRRUNTIME -ne "false") {
         }
         if (!(Get-Command ninja -ErrorAction SilentlyContinue)) {
             choco install ninja
+            # NB: The chocolatey package for rustup is currently nonfunctional,
```
swernli, line 11: Ah, this is a shame. I actually think it would be worth trying to add a prerequistes.ps1 to src/Simulation/qdk_sim_rs (to match the ones here and in src/Simulation/Native) that tries to invoke rustup via another method. As mentioned in the comment on bootstrap.ps1, the opensim feature can cause our build scripts to start failing for users without Rust installed, so providing some automated help for getting it installed would be good.

build/build.ps1
```diff
@@ -26,6 +26,16 @@ if ($Env:ENABLE_QIRRUNTIME -ne "false") {
     Write-Host "Skipping build of qir runtime because ENABLE_QIRRUNTIME variable is set to: $Env:ENABLE_QIRRUNTIME"
 }
 
+if ($Env:ENABLE_EXPERIMENTALSIM -ne "false") {
```
swernli, line 29: While the feature is still experimental (or at least until we have a good automatic install story for rustup), should this part of the build also get skipped if cargo is not found in the path? That would bridge the gap for a bit for local build while some folks' environments don't yet have Rust. Something like `-and (Get-Command cargo -ErrorAction SilentlyContinue)` would do the trick.

src/Qir/Tests/OpenSystemsSimulator/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
+
+compile_from_qir(qir-test-opensimulator qir_test_opensimulator_target)
```
swernli, line 2: This macro has been updated since this was worked on, so this folder won't build (and is indeed skipped intentionally for now, I see). Since it's not part of the build and not yet ready for integration, maybe it makes sense to tease out the QIR-opensim infrastructure for now and leave that for a later PR? Not a big deal while this lives in a feature branch, but not something I think we'd want in main, and merging to the feature branch as-is puts the burden on us to remember to re-review this part before merging the feature branch to main, when it might be eaiser to miss in a large diff.

src/Simulation/Common/Simulators.Dev.props
```diff
@@ -5,8 +5,11 @@
     <DocumentationFile>bin\$(BuildConfiguration)\$(TargetFramework)\$(AssemblyName).xml</DocumentationFile>
     <EnlistmentRoot>$([MSBuild]::NormalizeDirectory($(MSBuildThisFileDirectory)..\..\..\))</EnlistmentRoot>
     <NativeBuildPath>$([MSBuild]::NormalizePath($(EnlistmentRoot)src/Simulation/Native/build/drop))</NativeBuildPath>
+    <ExperimentalSimBuildPath>$([MSBuild]::NormalizePath($(EnlistmentRoot)src/Simulation/qdk_sim_rs/drop))</ExperimentalSimBuildPath>
   </PropertyGroup>
 
+  <!-- Copy Microsoft.Quantum.Simulator.Runtime.dll from src/Simulation/Native build to where we can see and use it. -->
+
```
swernli, line 12: nitpick: I think removing this newline will make it clearer this comment goes with the block below, and matches what is done for the experimental sim too.

src/Simulation/Simulators/OpenSystemsSimulator/Decompositions/README.md
```diff
@@ -0,0 +1,4 @@
+# TODO
+
+The decompositions in this folder should eventually be converted into a new "type 4" targeting package.
+We use the legacy approach to making intrinsics available to simulators, however, on a temporary basis for compatibility with Python and IQ#.
```
swernli, line 4: Makes sense, and hopefully we can finish up the support soon such that this workaround won't be needed anymore!

src/Simulation/qdk_sim_rs/Cargo.toml.template
```diff
@@ -0,0 +1,88 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+[package]
+name = "qdk_sim"
```
swernli, line 5: This is perhaps a bit of a nitpick, but would it make sense to use a name other than `qdk_sim` for the crate? Knowing the project it makes sense, but imagining coming at it fresh I might find it surprising that this crate does not contain all the simulators from the QDK. Are there long term plans to include the fullstate simulator as well?

In some ways, this is handled by the `-alpha` or `-beta` in the semantic versioning, so perhaps we can make a similar exception like what was done for the QIR Runtime package and ensure that this crate keeps the `-alpha` in its version even when published as part of a beta or release build?

On a similar note, I'm less familiar with the naming convention for crates, but maybe it would make sense to name this something like `qdk_sim_experimental` and then when we are ready to publish the non-experimental version we use the `qdk_sim` name along with the updated description and documentation?

src/Simulation/qdk_sim_rs/NuGet.Config
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <clear />
```
swernli, line 4: I realize this is the latest advice for how to avoid nuget injection attacks, but it also means that this folder will not have access to the qdk-alpha feed configured at the root which may be surprising and lead to confusion later. Was there specific motivation behind including this here?

swernli (commented):
I'm slowly working my way through the changes, but figured I'd post what I have so far to start engaging in discussion. Really excited to see this work!

src/Simulation/qdk_sim_rs/Cargo.toml.template
cgranade: Don't think it's a nitpick at all, thanks for raising this! Perhaps for now, `qdk_sim_experimental` would be the cleanest crate name (can keep the actual DLL the same name since it gets renamed by the pipeline already). That seems to match crates like `serde_json_experimental`, or all the `wasm` experimental crates, such that I think that'd clearly communicate to users the maturity and support levels of the crate. At that point, forcing `-alpha` even on RC builds may be overkill? Anyway, let me go on and make that change, then. Thanks again for raising this!

src/Simulation/qdk_sim_rs/NuGet.Config
cgranade: This is used only for restoring the `dotnet-script` tool; ideally, that shouldn't be looking at the qdk-alpha feed, so that it always gets the right version off of the public nuget.org feed.

build/set-env.ps1
```diff
@@ -20,10 +20,27 @@ if ($Env:ENABLE_QIRRUNTIME -ne "false" -and $Env:QIR_DROPS -eq $null) {
...
 If (-not (Test-Path -Path $Env:DOCS_OUTDIR)) { [IO.Directory]::CreateDirectory($Env:DOCS_OUTDIR) }
 
+Get-ChildItem @(
+    "Env:\DROPS_DIR",
+    "Env:\DROP_NATIVE",
+    "Env:\NUGET_OUTDIR",
+    "Env:\CRATE_OUTDIR",
+    "Env:\WHEEL_OUTDIR",
+    "Env:\DOCS_OUTDIR"
+ ) | Format-Table
```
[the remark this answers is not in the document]
cgranade: Thanks! Added it to debug where I had the wrong path at some point, and figured it was worth leaving in to help make the logs a bit more useful.

src/Qir/Runtime/lib/QSharpCore/CMakeLists.txt
```diff
@@ -26,6 +26,7 @@ target_link_libraries(Microsoft.Quantum.Qir.QSharp.Core
   ${CMAKE_DL_LIBS}
   qsharp-core-qis-support-obj
   simulators-obj
+  # opensim
```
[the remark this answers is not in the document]
cgranade: Sounds good, let me go add that then. Thanks!

build/build.ps1
cgranade: Agreed. Broke it up a little bit more than your suggestion to make it obvious to the user what they need to do to proceed, and to make sure that we don't have surprising CI failures if prereq installation breaks, but I think this should bridge the gap that you point out. Thanks for flagging that!

src/Qir/Runtime/lib/Simulators/OpenSystemsSimulator.cpp
```diff
@@ -0,0 +1,292 @@
...
+        Qubit AllocateQubit() override
+        {
+            const unsigned id = this->nextQubitId;
+            this->nextQubitId++;
+            return reinterpret_cast<Qubit>(id);
+        }
+
+        void ReleaseQubit(Qubit q) override
+        {
+            // TODO
```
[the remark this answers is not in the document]
cgranade: Sounds good, let me add that and @kuzminrobin's feedback above to https://github.com/microsoft/qsharp-runtime/issues/713 for tracking. If it's alright with both of you, I'll defer that work until work on the QIR integration piece of this feature is resumed so as to keep that work as up to date with the QIR runtime at that point as possible. Thanks!

src/Qir/Runtime/lib/Simulators/OpenSystemsSimulator.cpp
cgranade: Added to https://github.com/microsoft/qsharp-runtime/issues/713; if it's alright with you, I'll defer to when the QIR work is resumed to address separating the C API and relying on `cbindgen`.

src/Qir/Runtime/lib/Simulators/OpenSystemsSimulator.cpp
cgranade: Added to the issue for QIR integration at https://github.com/microsoft/qsharp-runtime/issues/713; if it's alright with you, I'd like to defer that along with other QIR integration work. Thanks for raising these code quality issues, appreciate your help with the C++ side!

src/Qir/Runtime/lib/Simulators/OpenSystemsSimulator.cpp
```diff
@@ -0,0 +1,292 @@
...
+        }
+
+        void X(Qubit q) override
+        {
+            x(this->simulatorId, GetQubitId(q));
+        }
+
+        void ControlledX(long numControls, Qubit controls[], Qubit target) override
+        {
+            throw std::logic_error("not yet implemented");
```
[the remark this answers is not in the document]
cgranade: I think that could make a lot of sense, agreed. Honestly, my dream would be eventually to not just have feature parity with the full state simulator, but complete integration with — that is, using a full-state representation when allowed by the current noise model, and promoting to open systems as needed.

src/Qir/Runtime/prerequisites.ps1
cgranade: Makes sense. Added the guard that you suggested to bootstrap.ps1, so that should help CI and local usage to have a nicer fallback, but I agree with adding a prerequisites.ps1 here as well. Let me see if the rustup choco package has been fixed since that comment back in February; otherwise can always do the curl-style download and run command (after all, what is chocolatey other than a really awesome curl-and-eval as a service?).

src/Qir/Tests/OpenSystemsSimulator/CMakeLists.txt
cgranade: Would it perhaps make sense, then, to revert all changes to QIR runtime as part of this feature work? As per blocker on target packages in Python, QIR integration is likely deferred for now, and that keeps outdated code from landing in main by accident.

src/Simulation/Common/Simulators.Dev.props
cgranade: Good point, fixed. Thanks!

src/Simulation/Simulators/OpenSystemsSimulator/Decompositions/README.md
cgranade: Sounds good! Do we have an issue for Python/IQ# ↔ QIR integration that I can link to here?

src/Simulation/qdk_sim_rs/benches/c_api_benchmark.rs
cgranade: Thanks! Maybe Rust benchmarking best practices would be something to add to the repo wiki? Would love to figure out the best way to pick up benchmarks from the CI artifacts as part of writing that guidance too (+@cesarzc for discussion here as well).

bootstrap.ps1
cgranade: Sounds good, let me add that, then. If it's alright, I'll similarly gate that on non-CI usage, so that CI builds fail correctly if prerequisites are missing?

cgranade:
> I'm slowly working my way through the changes, but figured I'd post what I have so far to start engaging in discussion. Really excited to see this work!

Thanks for the review, @swernli! I really appreciate the help! I've responded to each comment and addressed what I can at this point — I think the big question that I still have is whether it makes sense to just revert all changes to the QIR folder until we can address https://github.com/microsoft/qsharp-runtime/issues/713 more holistically. Thoughts?

cgranade:
Thanks for the review, @kuzminrobin, and for the help in improving the C++ code in this PR. As per above, would you be alright with deferring your feedback until we return to https://github.com/microsoft/qsharp-runtime/issues/713? Thanks!

src/Qir/Tests/OpenSystemsSimulator/CMakeLists.txt
swernli: Yeah, I do think that makes sense, especially now that you also filed the issue that helps track the QIR integration feedback you've already received.

src/Simulation/Simulators/OpenSystemsSimulator/Decompositions/README.md
swernli: We don't, actually... let me connect with @bettinaheim and get the right thing filed in the right place for tracking.

src/Simulation/qdk_sim_rs/Cargo.toml.template
swernli: That approach sounds good!

swernli:
> I think the big question that I still have is whether it makes sense to just revert all changes to the QIR folder until we can address #713 more holistically. Thoughts?
I think that makes a lot of sense. I know it's annoying to have to separate out some of the work you already did, but it will make this PR just about the C# integration (which is the current focus anyway) and easier to review. Then a targeted QIR integration PR later where you can return to the feedback captured in #713 will likewise be easier to bring up to the latest QIR features and infrastructure.

src/Simulation/qdk_sim_rs/Cargo.toml.template
cgranade: Awesome, let me go on and do that, then. I'll apply the same to the Python interface as well for uniformity. Thanks again for raising this!

src/Qir/Tests/OpenSystemsSimulator/CMakeLists.txt
cgranade: Sounds good, went on and reverted in f3c038258b54eec549169d1717f3dd8ec2411da6. If we need to, can always look at the parents for that commit. Thanks!

src/Qir/Runtime/lib/Simulators/OpenSystemsSimulator.cpp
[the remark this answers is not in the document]
cgranade: Deferred by revert in f3c038258b54eec549169d1717f3dd8ec2411da6.

src/Simulation/qdk_sim_rs/Cargo.toml.template
cgranade: Addressed in 42c03efb180786feb2d61589a1c46a61206209d3. Thank you!

src/Qir/Runtime/prerequisites.ps1
cgranade: Looks like the rustup package for chocolatey is still both unmoderated and failing validation; I'll fall back to the PowerShell snippet used by CI to configure rustup. Thanks!

cgranade: Added in a80fb6fe52a4ffdd74ba33231cc77ad23ddc189f. Thanks!

src/Simulation/qdk_sim_rs/src/instrument.rs
```diff
@@ -0,0 +1,153 @@
...
+            }
+            assert!(
+                (output_state.trace() - 1.0).norm() <= 1e-10,
+                "Expected output of instrument to be trace 1."
+            );
+            return (idx, output_state);
+        }
+    }
+    let (idx, output_state, _) = possible_outcomes.pop().unwrap();
+    drop(possible_outcomes);
```
swernli, line 151: Out of curiosity, why the explicit drop of `possible_outcomes` here? Wouldn't it automatically be dropped when it goes out of scope during the return below? of note, it does not get explicitly dropped before the early return in the if-statement above (line 147).

src/Simulation/qdk_sim_rs/src/lib.rs
```diff
@@ -0,0 +1,72 @@
...
+// for discussion.
+#![cfg_attr(doc, feature(extended_key_value_attributes))]
+#![cfg_attr(doc, cfg_attr(doc, doc = include_str!("../README.md")))]
+// Set linting rules for documentation. We will stop the build on missing docs,
+// or docs with known broken links. We only enable this when all relevant
+// features are enabled, otherwise the docs build will fail on links to
+// features that are disabled for the current build.
+#![cfg_attr(all(doc, feature = "python"), deny(rustdoc::broken_intra_doc_links))]
+#![cfg_attr(doc, deny(missing_docs))]
+#![cfg_attr(doc, warn(missing_doc_code_examples))]
```
swernli, line 16: Of note, this causes quite a few warnings to be logged for missing code examples, which is good but distracting for now. I don't think this needs to be resolved for this PR, but worth calling out as folks will see those warnings pretty prominently during local development (you already explained this one to me offline, just noting here for tracking/follow-up).

src/Simulation/qdk_sim_rs/src/python.rs
```diff
@@ -0,0 +1,243 @@
...
+
+#[pymethods]
+impl PyInstrument {
+    pub fn sample(&self, idx_qubits: Vec<usize>, state: &PyState) -> (usize, PyState) {
+        let (result, data) = self.data.sample(idx_qubits.as_slice(), &state.data);
+        (result, PyState { data })
+    }
+
+    #[staticmethod]
+    #[args(pr_readout_error = "0.0")]
```
swernli, line 155: Is this line setting a default value for the argument? If so, does that work for Rust callers or is that purely a Python interop feature/attribute?

src/Simulation/qdk_sim_rs/src/python.rs
```diff
@@ -0,0 +1,243 @@
...
+        py_enum.call1(args).unwrap().into_py(py)
+    }
+}
+
+impl FromPyObject<'_> for Pauli {
+    fn extract(ob: &'_ PyAny) -> PyResult<Self> {
+        // We want to support either a primitive type that can extract to u8,
+        // or a value of type qdk_sim.Pauli from the root module.
+        Python::with_gil(|py| {
+            let root = PyModule::import(py, "qdk_sim").unwrap();
```
swernli, line 231: Now that the crate's package name has changed to `qdk_sim_experimental` will this logic break? Or does this Python module name come from the library name instead of the package name?

src/Simulation/qdk_sim_rs/src/tableau.rs
```diff
@@ -0,0 +1,298 @@
...
+impl Tableau {
+    /// Returns a serialization of this stabilizer tableau as a JSON object.
+    #[cfg(feature = "python")]
+    pub fn as_json(&self) -> PyResult<String> {
+        serde_json::to_string(self)
+            .map_err(|e| PyErr::from(Error::new(ErrorKind::Other, e.to_string())))
+    }
+
+    /// Applies a Hadamard operation in-place to the given qubit.
+    pub fn apply_h_mut(&mut self, idx_target: usize) {
```
swernli, line 120: nitpick: is there any reason besides establishing a convention that these methods can't just avoid the `_mut` in their names? I was a little surprised that there isn't an `apply_h` to match `apply_h_mut`. Do you expect immutable variants of these to exist in the future (possibly by returning the updated table instead of modifying in-place)?

swernli (approved):
The changes look good! I have a few more questions/feedback below and one typo, but otherwise I feel good about signging off. Thanks for separating out the QIR part for now, that made this much easier to review.

src/Simulation/Simulators/OpenSystemsSimulator/DataModel/NoiseModel.cs
```diff
@@ -0,0 +1,73 @@
...
+
+namespace Microsoft.Quantum.Experimental
+{
+    // NB: To make this compatible with Newtonsoft.Json as well as
+    //     System.Text.Json, we use a Newtonsoft converter that delegates to
+    //     S.T.Json.
+    [Newtonsoft.Json.JsonConverter(typeof(DelegatedConverter<NoiseModel>))]
+    public class NoiseModel
+    {
+        [JsonPropertyName("initial_state")]
```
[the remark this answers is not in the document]
cgranade: Not that nitpicky, it's a good point! I think in this case, there will necessarily be some friction in that Rust and C# use different naming conventions such that either side will be "wrong" in some sense, requiring us to explicitly specify property names on at least one side of the interoperability boundary.

We ran into something similar with Microsoft.Jupyter.Core, where the JSON property names are most definitely not under our control, and all follow `snake_case`.

I'm not sure of the best solution here other than making sure to follow the pattern explicitly in core review, or to use a custom naming policy in System.Text.Json (https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-customize-properties#use-a-custom-json-property-naming-policy).

src/Simulation/Simulators/OpenSystemsSimulator/DataModel/Process.cs
```diff
@@ -0,0 +1,204 @@
...
+                }
+            );
+        }
+
+        public override void Write(Utf8JsonWriter writer, Process value, JsonSerializerOptions options)
+        {
+            var arrayConverter = new ComplexArrayConverter();
+
+            writer.WriteStartObject();
+                writer.WriteNumber("n_qubits", value.NQubits);
```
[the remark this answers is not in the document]
cgranade: Agreed, and I'd like to improve serialization more generally throughout this feature — it turns out to be painful to represent Rust's discriminated unions in C#, but I have a few ideas as to how to improve that going forward. If it's alright with you, I'd like to defer and open an issue for improving serialization by reducing the amount of custom converter logic required to handle Rust-style DUs.

src/Simulation/Simulators/OpenSystemsSimulator/DataModel/State.cs
```diff
@@ -0,0 +1,108 @@
...
+    public class StabilizerState : State
+    {
+        // Design note:
+        // We could use the same array converter as used of complex-like arrays
+        // of floats, but it's a bit easier in this case to directly
+        // deserialize into a type that represents the underlying data that
+        // we need.
+        public class TableArray
+        {
+            [JsonPropertyName("v")]
```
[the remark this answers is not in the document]
cgranade: This comes from how the `ndarray` crate exposes its data structures to `serde` for serialization. In particular, `ndarray` hard codes the `"v"` property at https://github.com/rust-ndarray/ndarray/blob/master/src/array_serde.rs#L96, such that trait coherence rules make it difficult to override that name. I'll add a comment explaining that here, if that makes sense?

src/Simulation/Simulators/OpenSystemsSimulator/OpenSystemsSimulator.cs
```diff
@@ -0,0 +1,286 @@
...
+        {
+            NativeInterface.Z(this.Id, target);
+        }
+
+        void IIntrinsicZ.ControlledBody(IQArray<Qubit> controls, Qubit target)
+        {
+            // TODO: pass off to decompositions for more than one control.
+            if (controls is { Count: 0 })
+            {
+                (this as IIntrinsicY).Body(target);
```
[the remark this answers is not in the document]
cgranade: Indeed at that, thanks for catching that!

src/Simulation/qdk_sim_rs/test-qdk-sim-rs.ps1
```diff
@@ -0,0 +1,60 @@
...
+        $script:allOk = $script:allOk -and $LASTEXITCODE -eq 0;
+
+        $testJson `
+            | cargo2junit `
+            <# We use this name to match the *_results.xml pattern that is used
+               to find test results in steps-wrap-up.yml. #> `
+            | Out-File -FilePath opensim_results.xml -Encoding utf8NoBOM
+    } else {
+        # Outside of CI, show human-readable output.
+        cargo +nightly test
```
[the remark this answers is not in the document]
cgranade: I've gated the `cargo clean` call so that it doesn't run in local builds, as per your earlier feedback. That said, I agree that in CI builds this will cause the crate to get built twice. I don't know how to avoid that in our current e2e builds given the hard disk space limitations and the order in which we call build and test commands. Would love to get @anpaz's input as to what may be a good way of cutting down on extraneous builds may be.

src/Simulation/qdk_sim_rs/src/instrument.rs
cgranade: Good catch...! I think this was needed back when I wrote the innovation week project back in December, as I was handling the return value a bit differently such that possible_outcomes holding owned values of other states was a problem here. I think refactoring since has obviated the need here, such that I can now drop this statement. Thanks for pointing this out!

src/Simulation/qdk_sim_rs/src/python.rs
cgranade: This is purely a Python interop feature; Python functions allow declaring keyword arguments with default values, such that this is a nice convenience to allow Python users to call `Instrument.new_z_measurement()` without specifying a readout error probability.

src/Simulation/qdk_sim_rs/src/python.rs
cgranade: Good catch, this will need to change. Thanks!

src/Simulation/qdk_sim_rs/src/tableau.rs
cgranade: I chose the names here to parallel usage in `ndarray`, where `_mut` is often used to indicate in-place even when no immutable variant exists. That said, I think it would be worthwhile at some point to expand to include `apply_h` as well as `apply_h_mut`.

src/Simulation/qdk_sim_rs/src/lib.rs
cgranade: I'll add a comment, as you suggest.

src/Simulation/qdk_sim_rs/test-qdk-sim-rs.ps1
```diff
@@ -0,0 +1,60 @@
+& (Join-Path $PSScriptRoot ".." ".." ".." "build" "set-env.ps1");
+
+$script:allOk = $true;
+
+Push-Location $PSScriptRoot
+    # Start with the quick check first and make sure that Rust sources
+    # meet formatting and style guide rules.
+    cargo fmt -- --check
```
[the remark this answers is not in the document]
cgranade: I'm pretty ambivalent, and will admit lumping into build instead of test was pretty arbitrary. Happy to move if you think that would be better.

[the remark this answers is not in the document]
swernli: I don't feel that strongly, but it does seem like a better fit to me. There's also the benefit that it means the developer will get the format and clippy feedback a little earlier (I know I personally tend to run the build script pretty frequently and then run the test script only when I think I'm done making changes, but that's me).

src/Simulation/Simulators/OpenSystemsSimulator/DataModel/Process.cs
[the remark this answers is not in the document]
cgranade: Tracked at https://github.com/microsoft/qsharp-runtime/issues/717.

src/Simulation/qdk_sim_rs/test-qdk-sim-rs.ps1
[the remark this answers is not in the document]
cgranade: Sounds good, let me go on and move it, then!

src/Qir/Runtime/lib/Simulators/OpenSystemsSimulator.cpp
kuzminrobin: Sorry, the `u` was redundant in my question. I was trying to suggest `int64_t` instead of `i64`.
The way the `i64` is currently defined does not guarantee that `i64` is 64-bit type (can be larger than that). Whereas `int64_t` is strictly 64-bit type.

anpaz:
Looks great, adding a couple of general comments for review/discussion:

anpaz (approved):
Looks great! Have a couple of general comments:

* [ ] As I mentioned somewhere else, I'm not sure this should be called "experimental". I would name this something less scary/more catchy: `powertools`, `toybox`, `sandbox`.
* [ ] I'm not sure this should be under an `Experimental` namespace. It does make it explicit, but switching to non-experimental will make it harder.
* [ ] I **highly** encourage you to write some simple tests for the qdk e2e tests, to make sure the new simulators run successfully on all environments (OS/conda/docker/etc)
* [ ] Is it possible to configure the noisy simulator to run without noise? I think it would be important to have a test for each intrinsic operation that makes sure every operation applies the corresponding correctly; we have something like that for the FullState simulator, so let me know if you need pointers on that.

.devcontainer/Dockerfile
```diff
@@ -9,15 +9,14 @@ RUN apt-get -y update && \
...
     apt-get clean && rm -rf /var/lib/apt/lists/
 
-# Install CMake, Ninja, LLVM/Clang tools
-RUN apt-get -y update && \
-    apt-get install -y cmake && \
-    apt-get install -y ninja-build && \
-    apt-get install -y clang-11 && \
-    apt-get install -y clang-tidy-11 && \
-    apt-get clean && rm -rf /var/lib/apt/lists/
+RUN apt update \
```
anpaz, line 12: fyi: left a bunch of general items on my review comment for you to consider. thanks!

cgranade:
> Looks great!

Thanks for the review, @anpaz, and for the kind words!

> Have a couple of general comments:
> 
> * [ ]  As I mentioned somewhere else, I'm not sure this should be called "experimental". I would name this something less scary/more catchy: `powertools`, `toybox`, `sandbox`.
> * [ ]  I'm not sure this should be under an `Experimental` namespace. It does make it explicit, but switching to non-experimental will make it harder.

For both of these, I'd like to get @bromeg's and @efratshabtai's input as well, to make sure that aligns with experimental features more broadly.

> * [ ]  I **highly** encourage you to write some simple tests for the qdk e2e tests, to make sure the new simulators run successfully on all environments (OS/conda/docker/etc)

Absolutely agreed. If it's alright, I'd like to do that after this PR completes, so that will allow for making independent PRs on each repo more easily.

> * [ ]  Is it possible to configure the noisy simulator to run without noise? I think it would be important to have a test for each intrinsic operation that makes sure every operation applies the corresponding correctly; we have something like that for the FullState simulator, so let me know if you need pointers on that.

Similarly here, I agree entirely with adding more test coverage. Ensuring that the noisy simulator agrees with the full-state simulator when configured to use the `ideal` noise model would be a great check on that. Is this the `Microsoft.Quantum.Simulation.Simulators.Tests.QuantumSimulatorTests.QSimVerify*` set of test cases? In any case, if it's alright, I'd similarly like to add that coverage following this PR, so as to unblock independent progress on the iqsharp, qsharp-runtime, and qdk repos.

swernli:
> Is this the `Microsoft.Quantum.Simulation.Simulators.Tests.QuantumSimulatorTests.QSimVerify*` set of test cases? 

Yes, that would be the right set of test cases. In fact, if you take a look at any of the target package simulator tests, such as those built by [Tests.Microsoft.Quantum.Simulators.Type1.csproj](https://github.com/microsoft/qsharp-runtime/blob/main/src/Simulation/Simulators.Type1.Tests/Tests.Microsoft.Quantum.Simulators.Type1.csproj), you can see how they build just the subset of the tests that do gate validation. Something similar might work well for the opensim, though we may have to do some small work to generalize the constructor the test invokes. I agree that this can come in during a later PR to the feature branch before the merge to main.

cgranade:
Sounds good, going to go on and merge this to the feature branch, then, with the following items in PRs to follow once https://github.com/microsoft/iqsharp/pull/449 is merged:

- [ ] Finish decompositions in qsharp-runtime
- [ ] Expand test coverage in qsharp-runtime
- [ ] Expand test coverage in iqsharp.
- [ ] Add e2e tests to QDK build

cgranade:
The e2e build failed due to a known issue in the full-state simulator:

image

Given that e2e builds have been passing for a while, that local builds now work, and that this is to a feature branch, I'd like to go on and merge to unblock work on decompositions and test cases. Any objections, @bettinaheim, @anpaz, @swernli?

swernli:
Sounds good to me!

PR #710, opened 2021-06-04
Update Simulator Release to check for entanglement

### LINKED ISSUE #552: QIR simulator should check result of qubit Release
Our existing C# wrapper for the fullstate simulator checks the result of `release` on a qubit to ensure that the qubit was in a well-known state at release time (see #206 for the detailed discussion of the release semantics). Releasing a qubit that could be entangled can result in problems for the rest of the program, and the simulator gives developers a way to use Q# to validate their algorithms for this. The C++ wrapper around the simulator used in QIR does not check the return of `release` and therefore doesn't provide this feedback, allowing for algorithms to effectively leak state.

To test this, consider the following Q#:
```qsharp
operation Test() : Unit {
    use q = Qubit();
    X(q);
}
```
Because the qubit is not in the ground state and hasn't been measured, running this code on the C# simulator will result in a `ReleasedQubitsAreNotInZeroState` exception being thrown from [QubitManager.cs](https://github.com/microsoft/qsharp-runtime/blob/main/src/Simulation/Simulators/QuantumSimulator/QubitManager.cs#L59). We need an equivalent check in the QIR fullstate simulator wrapper.

@bettinaheim @cgranade FYI

swernli:
This change adds a new utility to the full state simulator, `is_classical_or_entangled` that returns a pair of bools indicating when the given qubit is classical with regard to the computational basis and whether the qubit is entangled with any other qubits. The former is needed to ensure the state of the qubit is collapsed before release, while the latter is the correct feedback to return from release to indicate whether releasing that qubit had any side effects on other qubits in the wave function.

This updates the logic in the C# infrastructure to use the returned bool as the indication of whether or not the release was valid (as opposed to the old check that used a cache of whether or not the last operation on the qubit was a measurement). THIS IS A CHANGE IN BEHAVIOR. Notably, a qubit is now considered valid for release as long as it is not entangled with any other qubit, regardless of superposition. As a result, the exception name and text for `ReleasedQubitsAreNotInZeroState` is changed to `ReleaseQubitsAreEntangled`.

This also fixes #552 by having the FullstateSimulator wrapper in the QIR Runtime check the return value from release and call `quantum__rt__fail_cstr` if the qubit released was entangled.


swernli:
Looks like `ReleasedQubitsAreNotInZeroState` is used directly in samples, so this will require a cross-repo change.

swernli:
On second thought, looking at how samples uses the exception, I can simply leave the existing type with a deprecation. I'll go that route instead.

swernli:
@dbwz8 It turns out the entanglement check can be further simplified! Because of the way we are iterating over the wave function, we are always checking the two states where the only difference is the qubit in question. If both of those have non-zero probability, that means there is a chance of measuring the identical values for every other qubit regardless of the state of the qubit we care about, which by definition means it cannot be entangled with any of the other qubits. If we never see that case, then it must mean that every measurement of the qubit we care about corresponds to a specific measurement of at least one other qubit. The check I was doing before (comparing the two indices for differences other than the target qubit) was redundant, because in the case where the probabilities were non-zero (the case we care about) that check would always be true.

src/Simulation/Native/src/simulator/kernels.hpp
```diff
@@ -144,6 +144,46 @@ bool isclassical(
...
+#pragma omp parallel for schedule(static) reduction(|| : have0, have1, notentangled)
+    for (std::intptr_t i = 0; i < static_cast<std::intptr_t>(wfn.size()); i += 2 * offset)
+        for (std::intptr_t j = 0; j < static_cast<std::intptr_t>(offset); ++j)
+        {
+            bool has0 = std::norm(wfn[i + j]) >= eps;
+            bool has1 = std::norm(wfn[i + j + offset]) >= eps;
+            have0 = have0 || has0;
+            have1 = have1 || has1;
+            notentangled = notentangled || (has0 && has1);
+        }
```
kuzminrobin, line 169: FYI (may be a false alarm):
I have a suspicion that during the last-most iteration of the outer loop (for `i`) nothing prevents `i` to be exactly 1 less than `wfn.size()`, see the condition `i < static_cast<std::intptr_t>(wfn.size())`.

In that case the fragments
* `wfn[i + j]`
* `wfn[i + j + offset]`

are likely to read _outside_ of `wfn`.

To provide more guarantees of the code correctness consider changing the condition to something like
> i <= (static_cast<std::intptr_t>(wfn.size()) **- 2 * offset**)

src/Simulation/Native/src/simulator/kernels.hpp
```diff
@@ -144,6 +144,46 @@ bool isclassical(
...
+        bool has0 = std::norm(wfn[i + j]) >= eps;
+        bool has1 = std::norm(wfn[i + j + offset]) >= eps;
+        have0 = have0 || has0;
+        have1 = have1 || has1;
+        notentangled = notentangled || (has0 && has1);
+    }
+#endif
+
+    return std::make_pair<bool, bool>(have0 ^ have1, have0 && have1 && !notentangled);
+}
```
kuzminrobin, line 185: Would be nice to see the comments explaining the logic of this function.

src/Simulation/Native/src/simulator/simulator.hpp
```diff
@@ -71,6 +71,12 @@ class Simulator : public Microsoft::Quantum::Simulator::SimulatorInterface
         return psi.isclassical(q);
     }
 
+    std::pair<bool, bool> is_classical_or_entangled(logical_qubit_id q)
+    {
+        recursive_lock_type l(mutex());
+        return psi.is_classical_or_entangled(q);
+    }
```
kuzminrobin, line 78: FYI:
> recursive_lock_type l(mutex());

I saw this fragment a number of times in the existing code and I don't understand what this fragment does. It locks not the _existing mutex_, but a mutex created on the fly (new mutex created every time the function is entered). That is why the fragment does not protect from re-entrancy and does not synchronize the access during concurrency.

Can you please explain to me what this fragment does?

kuzminrobin (approved):

src/Simulation/Native/src/simulator/simulator.hpp
swernli: Interesting... I hadn't noticed this until you pointed it out, so I'm glad you highlighted it! To call the constructor for a mutex this would actually have to be `std::mutex()` where in this case it is instead intended to invoke the (rather poorly named) member function `mutex()` from simulationinterface.hpp: https://github.com/microsoft/qsharp-runtime/blob/b908e8d62b2c67df6999546dd45425f6e6a45db1/src/Simulation/Native/src/simulator/simulatorinterface.hpp#L127-L130
That will get the pointer to the member mutex for this class. That said, its possible that depending on how this is compiled and if any header files have a `using namespace std;` in them, it could change how this call works and cause it to behave exactly as you describe. That could even be the result of the intermitten failures that we've seen in functions in this class when running multithreaded tests. To make the use clearer and explicitly avoid this possible source of error, I'll update all these call points to use `this->mutex()` instead. Maybe that will even fix our intermittent issue!

src/Simulation/Native/src/simulator/kernels.hpp
swernli: Good point... I'll add more comments throughout.

src/Simulation/Native/src/simulator/kernels.hpp
swernli: I added a comment that hopefully helps clarify the intent of the iteration, but it's worth answering your specific question about bounds. Note that `i` is always incremented by `2 * offset` where `offset` is at minimum 1 when `q` is 0. Combining this with the condition `i < static_cast<std::intptr_t>(wfn.size());` means that `i` will always be at most equal to `wfn.size() - 2` if the wave function has an even number of entries. Because the wave function represents the possible states of the N qubits, which is 2^N, we know the length of the wave function will always be even. In this same `q` equals 0 case, `j` is limited by `j < static_cast<std::intptr_t>(offset)` which means `j` will be at most 0, giving us the two final indices of `i + j` or `wfn.size() - 2 + 0` and `i + j + offset` or `wfn.size() - 2 + 0 + 1`. For larger values of `q` the maximum possible value of `i` is lower and the range for `j` is higher, but the largest index checked is still always be `wfn.size() - 1`, so we know we are safely within the bounds of the array for any target qubit.

swernli:
@kuzminrobin thanks for the review and feedback! Especially the observation about the use of `mutex()` which could possibly be the resolution to our long standing intermediate access violation in CI. Since this PR includes some changes in behavior for the simulator that affect the observed behavior for users of Q#, I'll hold off on merging until a few more stakeholders weigh in.
@bettinaheim, @cgranade, @tcNickolas, any opinions on that part?

cgranade:
> @kuzminrobin thanks for the review and feedback! Especially the observation about the use of `mutex()` which could possibly be the resolution to our long standing intermediate access violation in CI. Since this PR includes some changes in behavior for the simulator that affect the observed behavior for users of Q#, I'll hold off on merging until a few more stakeholders weigh in.
> @bettinaheim, @cgranade, @tcNickolas, any opinions on that part?

Sorry for the delay, been a bit head down trying to fix the e2e build for https://github.com/microsoft/qsharp-runtime/pull/709, and got behind on this one. Anyway, I'm in the awkward position of thinking that both the old and the new behavior makes sense... effectively, my thinking comes back to that when a user allocates a qubit with `use`, we currently document that they're promising that they will either measure before deallocating or will coherently return it to |0⟩. That promise can't be checked on real hardware, so running on simulators with checks like these is helpful to ensure that condition is met. From that perspective, checking that a released qubit is in |0⟩ is a much stronger guarantee than that it is separable, such that allowing users to pick a weaker condition can be helpful when running against targets that implicitly reset on deallocation (that is, where coherent return to |0⟩ isn't allowed). In other cases, though, the old behavior of checking for |0⟩ makes sense, such that if possible, I feel like making it configurable would be really helpful?

Apologies for the incoherent (pun fully intended) wall of text, just my 2¢ from an initial pass.

swernli:
No worries at all, @cgranade, I definitely wasn't expecting a response over the weekend!

> From that perspective, checking that a released qubit is in |0⟩ is a much stronger guarantee than that it is separable, such that allowing users to pick a weaker condition can be helpful when running against targets that implicitly reset on deallocation (that is, where coherent return to |0⟩ isn't allowed). In other cases, though, the old behavior of checking for |0⟩ makes sense, such that if possible, I feel like making it configurable would be really helpful?

Making it configurable is a really interesting idea, and stands to reason that would allow the simulation to meet the needs of a broader audience.  In fact, the C# simulator already has a setting that can disable this check entirely with QIR simulation infrastructure does not replicate. Just thinking out loud here, the measurement tracking felt like it was a stand in for entanglement (we know a measured qubit couldn't be entangled) but was already weaker than the zero check. If we have `release` in the full state simulator return a pair of bools, one indicating if the qubit was in the ground state and one indicating if it was entangled, then we could pretty easily have three configurable settings: allow release of any qubit, fail on release of entangled qubits, fail on release of non-zero qubits. Then both C# and QIR could offer the same functionality with minimal additional overhead (no need for measurement frame tracking). We'd have to decide which of those three is the default and how we'd want to expose the setting for QIR (environment variable perhaps?), but it would not be hard to implement given the information the new `is_classical_or_entangled` function returns.

src/Simulation/Native/src/simulator/simulator.hpp
kuzminrobin: :-DDDDDDDDDDDDDDDDDDDDDDD
Incredible!
Thank you very much!

DmitryVasilevsky: Can we rename this function so that there's no confusion now AND in future code?

swernli: Might as well... the naming conventions in this simulator code are a bit odd and seem to prioritize terseness over readability or maintainability, so I have no problem making that change. In that case, I should be able to remove the explicit use of the `this` pointer.

swernli:
I've converted this to a draft while we work out the questions @cgranade brought up about configurability. That said, #552 should be considered a critical issue for QIR-based simulation, so we should aim to have this included in the next release if at all possible.

swernli:
@cgranade Regarding configuration for simulation, there is still an outstanding general question about how we want to support configuring QDK, ranging from settings in the qubit manager to this qubit release behavior and even noise models for the experimental open simulator. I believe both @DmitryVasilevsky and @kuzminrobin  have started looking into this and might want to weigh in for this part.

bettinaheim (approved):
I think this change makes sense. I did look at the test cases and think they cover everything.

bettinaheim:
@cgranade
"... my thinking comes back to that when a user allocates a qubit with use, we currently document that they're promising that they will either measure before deallocating or will coherently return it to |0⟩. [...], checking that a released qubit is in |0⟩ is a much stronger guarantee than that it is separable, such that allowing users to pick a weaker condition can be helpful when running against targets that implicitly reset on deallocation (that is, where coherent return to |0⟩ isn't allowed). In other cases, though, the old behavior of checking for |0⟩ makes sense, such that if possible, I feel like making it configurable would be really helpful?"

What is not clear to me is the value of requiring that a qubit is measured (i.e. in the 0,1 basis) rather than in any unentangled state. The whole idea of these checks was that it is completely fine to measure a qubit after it is release (so it can be reused) without that changing the algorithm. What am I missing - why would the stricter condition be beneficial?

bettinaheim:
After some offline discussion, the concern is captured below. 
I don't really see a good way to determine when a qubit is coherently reset to zero if qubits can be released in an arbitrary state as long as they are unentangled. What is outlined below is actually a pretty important factor that I didn't think of in the rush of the moment yesterday. @swernli Thoughts?

cgranade:
When qubits are released, the target resets them to |0⟩ automatically before they are allocated again? In that case, how does the target differentiate between automatic reset behaviour and the coherent unpreparation case where a developer has already reset to |0⟩ before releasing?
At the moment, the demand that a qubit is either in |0⟩ or is measured individually allows for both cases to be covered without introducing measurements in the coherent unpreparation case; weakening to "reset if unentangled" means that since a target cannot in general prove they're unentangled, the target needs to always reset.
Could imagine a weakening where the target can prove exactly what state the qubits being released are in from whatever measurements have been performed (e.g.: let r1 = Measure([PauliZ, PauliZ], qs); let r2 = Measure([PauliX, PauliX], qs); is enough to nail down exactly which state qs is in and to thus reset w/o additional measurements), but weakening all the way to unentangled seems to me like it'd necessarily introduce extra measurements in the unpreparation case."

bettinaheim:
"Hm, I can see where you are coming from. I'd like to avoid opt in or opt out for a more strict check here, since otherwise we basically need to propagate with what setting the program has been compiled with to the code generation stage to avoid exactly that. That seem rather ad-hoc. On the other hand, a good way might be that releases are annotated by the compiler with a flag indicating that there is no need to reset the qubit when the more strict checking is turned on. That could make sense.

cgranade:
Right. It's at the least a change in behavior that would change how libraries are optimized (for example, wouldn't want to coherently unprepare unless the compiler knows either automatically or via a flag that those qubits don't then need to be reset).

swernli:
This PR is superseded by #796

PR #712, opened 2021-06-05
Qubit manager optimizations

DmitryVasilevsky:
Optimized computation complexity of allocate() to be O(1) amortized (worst case of a single call is still O(number of nested areas)).
Removed support for EncourageReuse. This is always assumed to be true.
Added more unit tests.


DmitryVasilevsky:
EncourageReuse=false is now removed. It is not an interesting case and wasn't used in the previous qubit manager. The only interesting case for EncourageReuse=true is when qubit ids aren't reused at all and capacity grows indefinitely. I.e., qubit manager functionality is reduced to id++. It makes sense to provide a different qubit manager for this case.

src/Qir/Runtime/public/QubitManager.hpp
```diff
@@ -185,6 +183,7 @@ namespace Quantum
         public:
             QubitListInSharedArray FreeQubitsReuseProhibited;
             QubitListInSharedArray FreeQubitsReuseAllowed;
+            int32_t prevAreaWithFreeQubits = 0;
```
avasch01, line 186: I would add a comment explaining this logic. At lease mention that this is a hint. This previous area with free qubits might not actually have free qubits. It is more of a "skip to" area. it is guaranteed that there are no free qubits between current area and this skip-to area.

avasch01 (approved):

kuzminrobin (approved):

src/Qir/Runtime/public/QubitManager.hpp
DmitryVasilevsky: Correct. The guarantee is that there're no free qubits between current area and the area pointed to by this variable (bounds non-inclusive). However it would be non-practical to name the variable to reflect exact definition. So I think I'll keep the variable name. I'll definitely add the comments.

src/Qir/Runtime/lib/QIR/QubitManager.cpp
```diff
@@ -180,10 +172,8 @@ int32_t CQubitManager::CRestrictedReuseAreaStack::Count() const
...
 CQubitManager::CQubitManager(
     QubitIdType initialQubitCapacity,
-    bool mayExtendCapacity,
-    bool encourageReuse):
+    bool mayExtendCapacity):
         mayExtendCapacity(mayExtendCapacity),
-        encourageReuse(encourageReuse),
         qubitCapacity(initialQubitCapacity)
 {
     FailIf(qubitCapacity <= 0, "Qubit capacity must be positive.");
```
kuzminrobin, line 179: (For the future)
What do you think about using `assert()` instead of `FailIf()` for the cases when the programming error is detected (as in this particular case when `qubitCapacity` or `initialQubitCapacity` is negative)?

DmitryVasilevsky: Excellent question! In general, we should have two failure mechanisms. One for internal errors that should never happen (data structure corruption, data mismatches, etc.). And another one for incorrect parameters provided from outside. Ideally, internal errors should crash the program (fail fast), and incorrect parameters should cause exceptions. This matches general code guidelines.
We'll definitely need to go over the codebase and make sure this cleaned up, including Qubit Manager. I'm deferring this work to a later time.

PR #715, opened 2021-06-08
Add required props files for target packages

swernli:
This adds the props files needed so that the target packages produced can be loaded with `IsTargetPackage="true"` in a consumers csproj file.


src/Simulation/Type1Core/Microsoft.Quantum.Type1.Core.props
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <PropertyGroup>
```
ScottCarda-MS, line 4: Is this property group formatted correctly? It looks a bit off. Does it need the extra line in it?

ScottCarda-MS (approved):

src/Simulation/Type1Core/Microsoft.Quantum.Type1.Core.props
swernli: It doesn't look odd to me; don't we usually have an empty line between the opening of the project and the first property group? Either way, it doesn't have any functionality besides separating the property group out for legibility, but if it actually makes it more confusing then I can take it out. This is where I copied the pattern from:
https://github.com/microsoft/qsharp-runtime/blob/5a379fd1c72ea19cb50d6acc501c6896894e96ac/src/Simulation/CSharpGeneration/RewriteStep.props#L1-L8

ScottCarda-MS: The line I was referring to was the one after the start of the property group, before the <QscRef_Microsoft_Quantum_CSharpGeneration> node. You are right in that this is a minor style comment, which we agree long ago not to fuss over too much, so if it looks fine to you, feel free to close out this comment and merge it. Everything looks good to me otherwise.

PR #718, opened 2021-06-08
Added treating warnings as errors

kuzminrobin:
Added treating warnings as errors

### DIFF
--- a/src/Qir/Runtime/lib/QIR/QubitManager.cpp
+++ b/src/Qir/Runtime/lib/QIR/QubitManager.cpp
@@ -392,7 +392,7 @@ CQubitManager::QubitIdType CQubitManager::GetQubitId(Qubit qubit) const
 Qubit CQubitManager::CreateQubitObject(QubitIdType id)
 {
     // Make sure the static_cast won't overflow:
-    FailIf(id < 0 || id > std::numeric_limits<intptr_t>::max(), "Qubit id is out of range.");
+    FailIf(id < 0 || id >= MaximumQubitCapacity, "Qubit id is out of range.");
     intptr_t pointerSizedId = static_cast<intptr_t>(id);
     return reinterpret_cast<Qubit>(pointerSizedId);
 }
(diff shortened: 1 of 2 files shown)


src/Qir/Runtime/lib/QIR/QubitManager.cpp
```diff
@@ -392,7 +392,7 @@ CQubitManager::QubitIdType CQubitManager::GetQubitId(Qubit qubit) const
 Qubit CQubitManager::CreateQubitObject(QubitIdType id)
 {
     // Make sure the static_cast won't overflow:
-    FailIf(id < 0 || id > std::numeric_limits<intptr_t>::max(), "Qubit id is out of range.");
+    FailIf(id < 0 || id >= MaximumQubitCapacity, "Qubit id is out of range.");
```
kuzminrobin, line 395: This was giving a warning:
```
pwsh bootstrap.ps1

. . .
[14/35] Building CXX object lib/QIR/CMakeFiles/qir-rt-support-obj.dir/QubitManager.cpp.obj
../../lib/QIR/QubitManager.cpp:395:25: warning: result of comparison of constant 9223372036854775807 with expression of type 'Microsoft::Quantum::CQubitManager::QubitIdType' (aka 'int') is always false [-Wtautological-constant-out-of-range-compare]
    FailIf(id < 0 || id > std::numeric_limits<intptr_t>::max(), "Qubit id is out of range.");
                     ~~ ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 warning generated.
```

src/Qir/qir-utils.ps1
```diff
@@ -35,9 +35,15 @@ function Build-CMakeProject {
     $oldCC = $env:CC
     $oldCXX = $env:CXX
     $oldRC = $env:RC
+    $oldCCFLAGS  = $env:CCFLAGS
+    $oldCXXFLAGS = $env:CXXFLAGS
 
     $clangTidy = ""
 
+    $warningFlags = "-Werror"
```
kuzminrobin, line 43: This variable is to be extended with a number of warning flags.

swernli (approved):
Change looks good, thanks for doing this! The build failure looks like an unlucky non-deterministic test, so it should pass on rerun:
``` Microsoft.Quantum.Simulation.Core.ExecutionFailException: Variance of normal(-2.0, 5.0) distribution should be 5, was 4.978016489850558.```

PR #719, opened 2021-06-08
Use of exceptions recommendation

DmitryVasilevsky:
Updated section on exceptions

### DIFF
--- a/src/Qir/QIR-RT-API-Design-Guidelines.md
+++ b/src/Qir/QIR-RT-API-Design-Guidelines.md
@@ -363,7 +363,7 @@ exist on how to use exceptions properly and minimize performance penalties.
 * As part of [C++ core guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines).
 * MSVC Documentation: [Modern C++ best practices for exceptions and error handling](https://docs.microsoft.com/en-us/cpp/cpp/errors-and-exception-handling-modern-cpp?view=msvc-160) and [How to: Design for exception safety](https://docs.microsoft.com/en-us/cpp/cpp/how-to-design-for-exception-safety?view=msvc-160).
 * [Exceptions and Error Handling](https://isocpp.org/wiki/faq/exceptions) As part of Modern C++ FAQ.
-* Google guidelines recommend against exceptions, but [allow them in windows code, especially STL](https://google.github.io/styleguide/cppguide.html#Windows_Code).
+* Google guidelines recommend against exceptions, but [allow them in windows code, especially while using STL](https://google.github.io/styleguide/cppguide.html#Windows_Code).
 * [Vishal Chovatiya Guidelines](http://www.vishalchovatiya.com/7-best-practices-for-exception-handling-in-cpp-with-example/), which may be easier to read.
 This section will describe only exceptions with regards to the runtime API.
 
@@ -371,9 +371,8 @@ This section will describe only exceptions with regards to the runtime API.
 
 It is not recommended to catch asynchronous harware exceptions, this is sometimes called
 [Structured Exception Handling](https://docs.microsoft.com/en-us/cpp/cpp/structured-exception-handling-c-cpp?view=msvc-160).
-An example of a hardware exception is a segmentation fault. If such exceptional situation arises, application should
-fail fast and quit. Typically hardware exceptions are handled in a system level code, and our runtime is an application with
-regards to the host OS. NOTE: MSVC compiler accepts the [`/EHs`](https://docs.microsoft.com/en-us/cpp/build/reference/eh-exception-handling-model?view=msvc-160) switch to disable hardware exception handing in catch() blocks.
+Examples of hardware exceptions are the exception thrown when accessing a forbidden memory location as a result of a null pointer access, stack overflow, corrupt pointer dereferencing, unaligned memory access, etc.. If such exceptional situation arises, application should fail fast and quit. Typically hardware exceptions are handled in a system level code, and our runtime is an application with
+regards to the host OS. NOTE: The [MSVC's default exception handling behavior](https://docs.microsoft.com/en-us/cpp/build/reference/eh-exception-handling-model?view=msvc-160#default-exception-handling-behavior) does not conform to the C++ standard. Use the `/EHsc` flag to [conform to the standard](https://docs.microsoft.com/en-us/cpp/build/reference/eh-exception-handling-model?view=msvc-160#standard-c-exception-handling).
 
 It is also not recommended to handle out-of-memory exception as there's a very little chance of handling it right.
 For such handling to work correctly there shouldn't be any memory allocation from the point of exception, in the stack
@@ -384,17 +383,36 @@ is resolved. Just let the application fail. For the same reason, using non-throw
 
 If you implement an interface function mark it with [noexcept specifier](https://en.cppreference.com/w/cpp/language/noexcept_spec):
 
+Inside of your binary if you implement a function that should not throw exceptions, then mark it with the [noexcept specifier](https://en.cppreference.com/w/cpp/language/noexcept_spec).
+
 > void interface_func() **noexcept**
+
+extern "C" functions aren't considered noexcept by default. Consider marking them noexcept.
+Keep in mind that `noexcept` specifier is a C++ feature. If you need to add `noexcept` specifier
+to a public header or any other header that should be compatible with a pure C compiler
+make sure to use `__cplusplus` guards. For example:
+
+```c++
+#ifdef __cplusplus
+#define NOEXCEPT noexcept     // `noexcept` in C++ only.
+extern "C"
+{
+#else
+#define NOEXCEPT      // In C this macro is empty.
+#endif
+
+.. // `#include`s
+void f() NOEXCEPT;    // Exposed function. It is `noexcept` if the header is included to the C++ source file.
+
+#ifdef __cplusplus
+} // extern "C"
+#endif
+```
 
 If exception is thrown (and not caught) within the noexcept function, the application will terminate. Termination will
 ensure that exceptions will never cross the API boundaries. An overhead is incurred if a noexcept function
 calls functions that can throw exceptions, so consider marking all functions that shouldn't throw exceptions as noexcept.
 If you are calling functions that can throw exceptions, make sure to handle them properly.
-
-extern "C" functions aren't considered noexcept by default. Consider marking them noexcept, but only when they are
-compiled by the C++ compiler (and not by a pure C compiler). NOTE: MSVC compiler accepts
-[`/EHsc`](https://docs.microsoft.com/en-us/cpp/build/reference/eh-exception-handling-model?view=msvc-160)
-flag to treat extern "C" functions as not throwing.
 
 * If using the [`dynamic_cast<>()`](https://en.cppreference.com/w/cpp/language/dynamic_cast), 
 consider casting the _pointers_ rather than _references_, e.g. 
--- a/src/Qir/QIR-RT-API-Design-Guidelines.md
+++ b/src/Qir/QIR-RT-API-Design-Guidelines.md
@@ -369,9 +369,9 @@ This section will describe only exceptions with regards to the runtime API.
 
 * Do not handle hardware exceptions and out-of-memory exception
 
-It is not recommended to catch asynchronous harware exceptions, this is sometimes called
-[Structured Exception Handling](https://docs.microsoft.com/en-us/cpp/cpp/structured-exception-handling-c-cpp?view=msvc-160).
-Examples of hardware exceptions are the exception thrown when accessing a forbidden memory location as a result of a null pointer access, stack overflow, corrupt pointer dereferencing, unaligned memory access, etc.. If such exceptional situation arises, application should fail fast and quit. Typically hardware exceptions are handled in a system level code, and our runtime is an application with
+It is not recommended to catch asynchronous harware exceptions. (MSVC compiler calls this
+[Structured Exception Handling](https://docs.microsoft.com/en-us/cpp/cpp/structured-exception-handling-c-cpp?view=msvc-160).)
+Examples of hardware exceptions are the exceptions thrown when accessing a forbidden memory location as a result of a null pointer access, stack overflow, corrupt pointer dereferencing, unaligned memory access, etc.. If such exceptional situation arises, application should fail fast and quit. Typically hardware exceptions are handled in a system level code, and our runtime is an application with
 regards to the host OS. NOTE: The [MSVC's default exception handling behavior](https://docs.microsoft.com/en-us/cpp/build/reference/eh-exception-handling-model?view=msvc-160#default-exception-handling-behavior) does not conform to the C++ standard. Use the `/EHsc` flag to [conform to the standard](https://docs.microsoft.com/en-us/cpp/build/reference/eh-exception-handling-model?view=msvc-160#standard-c-exception-handling).
 
 It is also not recommended to handle out-of-memory exception as there's a very little chance of handling it right.
--- a/src/Qir/QIR-RT-API-Design-Guidelines.md
+++ b/src/Qir/QIR-RT-API-Design-Guidelines.md
@@ -359,8 +359,13 @@ reported by using exceptions (such as [`out_of_range`](https://en.cppreference.c
 that the C++ code we are developing will throw exceptions even if they aren't thrown right in the code we write.
 
 Recommendation is to use std namespace and exceptions in C++ code **only where appropriate**. Many recommendations
-exist on how to use exceptions properly and minimize performance penalties. This section will describe only exceptions
-with regards to the runtime API.
+exist on how to use exceptions properly and minimize performance penalties.
+* As part of [C++ core guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines).
+* [MSVC Documentation](https://docs.microsoft.com/en-us/cpp/cpp/errors-and-exception-handling-modern-cpp?view=msvc-160).
+* [Exceptions and Error Handling](https://isocpp.org/wiki/faq/exceptions) As part of Modern C++ FAQ.
+* Google guidelines recommend against exceptions, but [allow them in windows code, especially STL](https://google.github.io/styleguide/cppguide.html#Windows_Code).
+* [Vishal Chovatiya Guidelines](http://www.vishalchovatiya.com/7-best-practices-for-exception-handling-in-cpp-with-example/), which may be easier to read.
+This section will describe only exceptions with regards to the runtime API.
 
 * Do not handle hardware exceptions and out-of-memory exception
 
--- a/src/Qir/QIR-RT-API-Design-Guidelines.md
+++ b/src/Qir/QIR-RT-API-Design-Guidelines.md
@@ -361,7 +361,7 @@ that the C++ code we are developing will throw exceptions even if they aren't th
 Recommendation is to use std namespace and exceptions in C++ code **only where appropriate**. Many recommendations
 exist on how to use exceptions properly and minimize performance penalties.
 * As part of [C++ core guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines).
-* [MSVC Documentation](https://docs.microsoft.com/en-us/cpp/cpp/errors-and-exception-handling-modern-cpp?view=msvc-160).
+* MSVC Documentation: [Modern C++ best practices for exceptions and error handling](https://docs.microsoft.com/en-us/cpp/cpp/errors-and-exception-handling-modern-cpp?view=msvc-160) and [How to: Design for exception safety](https://docs.microsoft.com/en-us/cpp/cpp/how-to-design-for-exception-safety?view=msvc-160).
 * [Exceptions and Error Handling](https://isocpp.org/wiki/faq/exceptions) As part of Modern C++ FAQ.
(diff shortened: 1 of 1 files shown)


src/Qir/QIR-RT-API-Design-Guidelines.md
```diff
@@ -353,103 +353,52 @@ The plan was to replace the exception throw with the [`exit()`](https://www.cplu
...
-* Consider the _nothrow_ version of the [`operator new()`](https://www.cplusplus.com/reference/new/operator%20new/) 
-(for single elements) and 
-[`operator new[]()`](https://www.cplusplus.com/reference/new/operator%20new[]/) (for arrays), e.g.
-```c++
-int * myDynamicIntArray = new                int[10]; // In case of allocation failure THROWS! Avoid using this.
-int * myDynamicIntArray = new (std::nothrow) int[10]; // In case of allocation failure returns the nullptr. Consider this first.
-```
+On the other hand, exceptions are a natural part of the C++ standard. It is virtually impossible to use
+the standard library (STL, std namespace) and not to use exceptions. All _exceptional_ situations are
+reported by using exceptions (such as out_of_range or invalid_argument to name a few). So it is safe to assume
```
kuzminrobin, line 358:
```suggestion
reported by using exceptions (such as [`out_of_range`](https://en.cppreference.com/w/cpp/error/out_of_range) or [`invalid_argument`](https://en.cppreference.com/w/cpp/error/invalid_argument) to name a few). So it is safe to assume
```

src/Qir/QIR-RT-API-Design-Guidelines.md
```diff
@@ -353,103 +353,52 @@ The plan was to replace the exception throw with the [`exit()`](https://www.cplu
...
+that the C++ code we are developing will throw exceptions even if they aren't thrown right in the code we write.
+
+Recommendation is to use std namespace and exceptions in C++ code **only where appropriate**. Many recommendations
+exist on how to use exceptions properly and minimize performace penalties. This section will describe only exceptions
+with regards to the runtime API.
```
kuzminrobin, line 363: > Many recommendations exist on how to use exceptions properly and minimize performace penalties

Any references? I would love to know a good explanation on the topic.

src/Qir/QIR-RT-API-Design-Guidelines.md
```diff
@@ -353,103 +353,52 @@ The plan was to replace the exception throw with the [`exit()`](https://www.cplu
...
-int * myDynamicIntArray = new                int[10]; // In case of allocation failure THROWS! Avoid using this.
-int * myDynamicIntArray = new (std::nothrow) int[10]; // In case of allocation failure returns the nullptr. Consider this first.
-```
+On the other hand, exceptions are a natural part of the C++ standard. It is virtually impossible to use
+the standard library (STL, std namespace) and not to use exceptions. All _exceptional_ situations are
+reported by using exceptions (such as out_of_range or invalid_argument to name a few). So it is safe to assume
+that the C++ code we are developing will throw exceptions even if they aren't thrown right in the code we write.
+
+Recommendation is to use std namespace and exceptions in C++ code **only where appropriate**. Many recommendations
+exist on how to use exceptions properly and minimize performace penalties. This section will describe only exceptions
```
kuzminrobin, line 362:
```suggestion
exist on how to use exceptions properly and minimize performance penalties. This section will describe only exceptions
```

src/Qir/QIR-RT-API-Design-Guidelines.md
DmitryVasilevsky: Added.

src/Qir/QIR-RT-API-Design-Guidelines.md
```diff
@@ -353,103 +353,57 @@ The plan was to replace the exception throw with the [`exit()`](https://www.cplu
...
+that the C++ code we are developing will throw exceptions even if they aren't thrown right in the code we write.
+
+Recommendation is to use std namespace and exceptions in C++ code **only where appropriate**. Many recommendations
+exist on how to use exceptions properly and minimize performance penalties.
+* As part of [C++ core guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines).
+* MSVC Documentation: [Modern C++ best practices for exceptions and error handling](https://docs.microsoft.com/en-us/cpp/cpp/errors-and-exception-handling-modern-cpp?view=msvc-160) and [How to: Design for exception safety](https://docs.microsoft.com/en-us/cpp/cpp/how-to-design-for-exception-safety?view=msvc-160).
+* [Exceptions and Error Handling](https://isocpp.org/wiki/faq/exceptions) As part of Modern C++ FAQ.
+* Google guidelines recommend against exceptions, but [allow them in windows code, especially STL](https://google.github.io/styleguide/cppguide.html#Windows_Code).
```
msoeken, line 366:
```suggestion
* Google guidelines recommend against exceptions, but [allow them in Windows code, especially STL](https://google.github.io/styleguide/cppguide.html#Windows_Code).
```

src/Qir/QIR-RT-API-Design-Guidelines.md
```diff
@@ -353,103 +353,57 @@ The plan was to replace the exception throw with the [`exit()`](https://www.cplu
...
+* [Vishal Chovatiya Guidelines](http://www.vishalchovatiya.com/7-best-practices-for-exception-handling-in-cpp-with-example/), which may be easier to read.
+This section will describe only exceptions with regards to the runtime API.
+
+* Do not handle hardware exceptions and out-of-memory exception
+
+It is not recommended to catch asynchronous harware exceptions, this is sometimes called
+[Structured Exception Handling](https://docs.microsoft.com/en-us/cpp/cpp/structured-exception-handling-c-cpp?view=msvc-160).
+An example of a hardware exception is a segmentation fault. If such exceptional situation arises, application should
```
msoeken, line 374:
```suggestion
An example of a hardware exception is a segmentation fault. If such exceptional situation arises, the application should
```

src/Qir/QIR-RT-API-Design-Guidelines.md
```diff
@@ -353,103 +353,57 @@ The plan was to replace the exception throw with the [`exit()`](https://www.cplu
...
+fail fast and quit. Typically hardware exceptions are handled in a system level code, and our runtime is an application with
+regards to the host OS. NOTE: MSVC compiler accepts the [`/EHs`](https://docs.microsoft.com/en-us/cpp/build/reference/eh-exception-handling-model?view=msvc-160) switch to disable hardware exception handing in catch() blocks.
```
msoeken, line 376:
```suggestion
regards to the host OS. NOTE: MSVC compiler accepts the [`/EHs`](https://docs.microsoft.com/en-us/cpp/build/reference/eh-exception-handling-model?view=msvc-160) switch to disable hardware exception handing in `catch` blocks.
```

src/Qir/QIR-RT-API-Design-Guidelines.md
```diff
@@ -353,103 +353,57 @@ The plan was to replace the exception throw with the [`exit()`](https://www.cplu
...
+
+It is also not recommended to handle out-of-memory exception as there's a very little chance of handling it right.
+For such handling to work correctly there shouldn't be any memory allocation from the point of exception, in the stack
+unwinding process, till the handling of exception and in the handling of exception until the out-of-memory situation
+is resolved. Just let the application fail. For the same reason, using non-throwing version of operator new is not recommended.
+
+* Mark interface functions with noexcept
+
+If you implement an interface function mark it with [noexcept specifier](https://en.cppreference.com/w/cpp/language/noexcept_spec):
```
msoeken, line 385:
```suggestion
If you implement an interface function mark it with the [noexcept specifier](https://en.cppreference.com/w/cpp/language/noexcept_spec):
```

src/Qir/QIR-RT-API-Design-Guidelines.md
```diff
@@ -353,103 +353,57 @@ The plan was to replace the exception throw with the [`exit()`](https://www.cplu
...
+
+> void interface_func() **noexcept**
+
+If exception is thrown (and not caught) within the noexcept function, the application will terminate. Termination will
```
msoeken, line 389:
```suggestion
If an exception is thrown (and not caught) within the noexcept function, the application will terminate. Termination will
```

msoeken (approved):
Looks good to me. Thanks @DmitryVasilevsky

src/Qir/QIR-RT-API-Design-Guidelines.md
```diff
@@ -353,103 +353,57 @@ The plan was to replace the exception throw with the [`exit()`](https://www.cplu
...
+* MSVC Documentation: [Modern C++ best practices for exceptions and error handling](https://docs.microsoft.com/en-us/cpp/cpp/errors-and-exception-handling-modern-cpp?view=msvc-160) and [How to: Design for exception safety](https://docs.microsoft.com/en-us/cpp/cpp/how-to-design-for-exception-safety?view=msvc-160).
+* [Exceptions and Error Handling](https://isocpp.org/wiki/faq/exceptions) As part of Modern C++ FAQ.
+* Google guidelines recommend against exceptions, but [allow them in Windows code, especially STL](https://google.github.io/styleguide/cppguide.html#Windows_Code).
+* [Vishal Chovatiya Guidelines](http://www.vishalchovatiya.com/7-best-practices-for-exception-handling-in-cpp-with-example/), which may be easier to read.
+This section will describe only exceptions with regards to the runtime API.
+
+* Do not handle hardware exceptions and out-of-memory exception
+
+It is not recommended to catch asynchronous harware exceptions, this is sometimes called
+[Structured Exception Handling](https://docs.microsoft.com/en-us/cpp/cpp/structured-exception-handling-c-cpp?view=msvc-160).
```
kuzminrobin, line 373: By following the link I see that Structured Exception Handling is "a Microsoft extension to C". 

I believe the readers don't expect us to mention anything Microsoft-specific in our cross-platform code, especially compiled with clang.

src/Qir/QIR-RT-API-Design-Guidelines.md
```diff
@@ -353,103 +353,57 @@ The plan was to replace the exception throw with the [`exit()`](https://www.cplu
...
+that the C++ code we are developing will throw exceptions even if they aren't thrown right in the code we write.
+
+Recommendation is to use std namespace and exceptions in C++ code **only where appropriate**. Many recommendations
+exist on how to use exceptions properly and minimize performance penalties.
+* As part of [C++ core guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines).
+* MSVC Documentation: [Modern C++ best practices for exceptions and error handling](https://docs.microsoft.com/en-us/cpp/cpp/errors-and-exception-handling-modern-cpp?view=msvc-160) and [How to: Design for exception safety](https://docs.microsoft.com/en-us/cpp/cpp/how-to-design-for-exception-safety?view=msvc-160).
+* [Exceptions and Error Handling](https://isocpp.org/wiki/faq/exceptions) As part of Modern C++ FAQ.
+* Google guidelines recommend against exceptions, but [allow them in Windows code, especially STL](https://google.github.io/styleguide/cppguide.html#Windows_Code).
+* [Vishal Chovatiya Guidelines](http://www.vishalchovatiya.com/7-best-practices-for-exception-handling-in-cpp-with-example/), which may be easier to read.
```
kuzminrobin, line 367: Thank you for this set of links. Very useful.

src/Qir/QIR-RT-API-Design-Guidelines.md
```diff
@@ -353,103 +353,57 @@ The plan was to replace the exception throw with the [`exit()`](https://www.cplu
...
+that the C++ code we are developing will throw exceptions even if they aren't thrown right in the code we write.
+
+Recommendation is to use std namespace and exceptions in C++ code **only where appropriate**. Many recommendations
+exist on how to use exceptions properly and minimize performance penalties.
+* As part of [C++ core guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines).
+* MSVC Documentation: [Modern C++ best practices for exceptions and error handling](https://docs.microsoft.com/en-us/cpp/cpp/errors-and-exception-handling-modern-cpp?view=msvc-160) and [How to: Design for exception safety](https://docs.microsoft.com/en-us/cpp/cpp/how-to-design-for-exception-safety?view=msvc-160).
+* [Exceptions and Error Handling](https://isocpp.org/wiki/faq/exceptions) As part of Modern C++ FAQ.
+* Google guidelines recommend against exceptions, but [allow them in Windows code, especially STL](https://google.github.io/styleguide/cppguide.html#Windows_Code).
```
kuzminrobin, line 366: The fragment "allow them in Windows code, especially STL" makes an impression that STL is a particular case of Windows code.
I believe you intended to say something slightly different.

src/Qir/QIR-RT-API-Design-Guidelines.md
```diff
@@ -353,103 +353,57 @@ The plan was to replace the exception throw with the [`exit()`](https://www.cplu
...
+An example of a hardware exception is a segmentation fault. If such exceptional situation arises, the application should
+fail fast and quit. Typically hardware exceptions are handled in a system level code, and our runtime is an application with
+regards to the host OS. NOTE: MSVC compiler accepts the [`/EHs`](https://docs.microsoft.com/en-us/cpp/build/reference/eh-exception-handling-model?view=msvc-160) switch to disable hardware exception handing in `catch` blocks.
+
+It is also not recommended to handle out-of-memory exception as there's a very little chance of handling it right.
+For such handling to work correctly there shouldn't be any memory allocation from the point of exception, in the stack
+unwinding process, till the handling of exception and in the handling of exception until the out-of-memory situation
+is resolved. Just let the application fail. For the same reason, using non-throwing version of operator new is not recommended.
+
+* Mark interface functions with noexcept
```
kuzminrobin, line 383:
```suggestion
* Mark interface functions with `noexcept`
```

src/Qir/QIR-RT-API-Design-Guidelines.md
```diff
@@ -353,103 +353,57 @@ The plan was to replace the exception throw with the [`exit()`](https://www.cplu
...
+
+If you implement an interface function mark it with the [noexcept specifier](https://en.cppreference.com/w/cpp/language/noexcept_spec):
```
kuzminrobin, line 385: I would suggest something like this:
> **Inside of your binary** if you implement an ~~interface~~ function **that should not throw exceptions**, then mark it with the [noexcept specifier](https://en.cppreference.com/w/cpp/language/noexcept_spec). **However do not use the `noexcept` specifier in the headers exposed externally, since those headers can be included to C source files, and the `noexcept` specifier is a C++ feature:**

I crossed out the word "interface" since it brings some ambiguity, a confusion about whether "interface function" means the _exposed API function_ or _the function implementing an interface_ - the base abstract class.

src/Qir/QIR-RT-API-Design-Guidelines.md
```diff
@@ -353,103 +353,57 @@ The plan was to replace the exception throw with the [`exit()`](https://www.cplu
...
+
+If an exception is thrown (and not caught) within the noexcept function, the application will terminate. Termination will
+ensure that exceptions will never cross the API boundaries. An overhead is incurred if a noexcept function
+calls functions that can throw exceptions, so consider marking all functions that shouldn't throw exceptions as noexcept.
+If you are calling functions that can throw exceptions, make sure to handle them properly.
+
+extern "C" functions aren't considered noexcept by default. Consider marking them noexcept, but only when they are
+compiled by the C++ compiler (and not by a pure C compiler). NOTE: MSVC compiler accepts
+[`/EHsc`](https://docs.microsoft.com/en-us/cpp/build/reference/eh-exception-handling-model?view=msvc-160)
+flag to treat extern "C" functions as not throwing.
```
kuzminrobin, line 397: I'd suggest the following change (does the code example illustrate what you mean?):
> The `extern "C"` functions are not `noexcept` by default. In the exposed headers consider marking them `noexcept` if they are compiled with the C++ compiler (but not marked as `noexcept` if compiled with C compiler):
```c++
#ifdef __cplusplus
#define NOEXCEPT noexcept     // `noexcept` in C++ only.
extern "C"
{
#else
#define NOEXCEPT      // In C this macro is empty.
#endif

.. // `#include`s
void f() NOEXCEPT;    // Exposed function. It is `noexcept` if the header is included to the C++ source file.

#ifdef __cplusplus
} // extern "C"
#endif
```
If the change above is ok then the fragment below probably stops being applicable (and can be excluded):
> NOTE: MSVC compiler accepts[`/EHsc`](https://docs.microsoft.com/en-us/cpp/build/reference/eh-exception-handling-model?view=msvc-160)flag to treat extern "C" functions as not throwing.

src/Qir/QIR-RT-API-Design-Guidelines.md
```diff
@@ -353,103 +353,57 @@ The plan was to replace the exception throw with the [`exit()`](https://www.cplu
...
+* [Vishal Chovatiya Guidelines](http://www.vishalchovatiya.com/7-best-practices-for-exception-handling-in-cpp-with-example/), which may be easier to read.
+This section will describe only exceptions with regards to the runtime API.
+
+* Do not handle hardware exceptions and out-of-memory exception
+
+It is not recommended to catch asynchronous harware exceptions, this is sometimes called
+[Structured Exception Handling](https://docs.microsoft.com/en-us/cpp/cpp/structured-exception-handling-c-cpp?view=msvc-160).
+An example of a hardware exception is a segmentation fault. If such exceptional situation arises, the application should
+fail fast and quit. Typically hardware exceptions are handled in a system level code, and our runtime is an application with
+regards to the host OS. NOTE: MSVC compiler accepts the [`/EHs`](https://docs.microsoft.com/en-us/cpp/build/reference/eh-exception-handling-model?view=msvc-160) switch to disable hardware exception handing in `catch` blocks.
```
kuzminrobin, line 376: In general I would prefer to refrain from mentioning any particular compiler. But this case seems special. What is your opinion about the following change? (replacing the fragment starting with "NOTE:")
> NOTE: The [MSVC's default exception handling behavior](https://docs.microsoft.com/en-us/cpp/build/reference/eh-exception-handling-model?view=msvc-160#default-exception-handling-behavior) does not conform to the C++ standard. Use the `/EHsc` flag to [conform to the standard](https://docs.microsoft.com/en-us/cpp/build/reference/eh-exception-handling-model?view=msvc-160#standard-c-exception-handling).

src/Qir/QIR-RT-API-Design-Guidelines.md
```diff
@@ -353,103 +353,57 @@ The plan was to replace the exception throw with the [`exit()`](https://www.cplu
...
+* [Exceptions and Error Handling](https://isocpp.org/wiki/faq/exceptions) As part of Modern C++ FAQ.
+* Google guidelines recommend against exceptions, but [allow them in Windows code, especially STL](https://google.github.io/styleguide/cppguide.html#Windows_Code).
+* [Vishal Chovatiya Guidelines](http://www.vishalchovatiya.com/7-best-practices-for-exception-handling-in-cpp-with-example/), which may be easier to read.
+This section will describe only exceptions with regards to the runtime API.
+
+* Do not handle hardware exceptions and out-of-memory exception
+
+It is not recommended to catch asynchronous harware exceptions, this is sometimes called
+[Structured Exception Handling](https://docs.microsoft.com/en-us/cpp/cpp/structured-exception-handling-c-cpp?view=msvc-160).
+An example of a hardware exception is a segmentation fault. If such exceptional situation arises, the application should
```
kuzminrobin, line 374: I would add a clarification:
```suggestion
An example of a hardware exception is a segmentation fault (access to a forbidden memory location as a result of null pointer access, stack overflow, corrupt pointer dereferencing, etc.). If such exceptional situation arises, the application should
```

src/Qir/QIR-RT-API-Design-Guidelines.md
DmitryVasilevsky: I think it's OK the way it is mentioned here. Historically Microsoft version of both C and C++ had a way to handle these exceptions because the languages were mostly used for systems programing. These are not the modern C++ exceptions we use now, and that's what I want to describe here. I'll try to rewrite.

src/Qir/QIR-RT-API-Design-Guidelines.md
DmitryVasilevsky: Tried to rewrite.

src/Qir/QIR-RT-API-Design-Guidelines.md
DmitryVasilevsky: OK, they are not always called segmentation faults so I tried to rewrite.

src/Qir/QIR-RT-API-Design-Guidelines.md
DmitryVasilevsky: OK.

src/Qir/QIR-RT-API-Design-Guidelines.md
DmitryVasilevsky: It's applicable but is less relevant. As we already have links to that page I agree it can be excluded.

src/Qir/QIR-RT-API-Design-Guidelines.md
DmitryVasilevsky: Tried to reword.

kuzminrobin (approved):
Looks good. Thank you for the help, Dmitry.

Issue #714, opened 2021-06-07
Experimental support for noisy, stabilizer simulation


cgranade:
This issue tracks development of experimental simulators for open quantum systems and CHP / stabilizer simulation, along with what work needs to be completed as part of this feature work.

## User feedback

(**TODO**: expand)

- https://github.com/microsoft/qsharp-runtime/issues/504

## Child issues

- [ ] https://github.com/microsoft/qsharp-runtime/issues/713

PR #720, opened 2021-06-09
Experimental simulators feature

cgranade:
This PR (currently in draft pending work highlighted below) adds the experimental simulators tracked at #714 to main, allowing for the simulators to be used without additional installation while development proceeds (+@bettinaheim, @efratshabtai, and @bromeg). As an experimental feature, these simulators may undergo breaking changes or may be removed  entirely in future releases without prior notice.

TODO before taking out of draft:
- [x] Finish decompositions of all discrete intrinsic operations
- [x] Expand test coverage of new simulators


build/test.ps1
```diff
@@ -57,6 +57,17 @@ if ($Env:ENABLE_QIRRUNTIME -ne "false") {
     Write-Host "Skipping test of qir runtime because ENABLE_QIRRUNTIME variable is set to: $Env:ENABLE_QIRRUNTIME."
 }
 
+
+if ($Env:ENABLE_EXPERIMENTALSIM -ne "false") {
```
swernli, line 61: I missed this in the previous review, but this script should warn if cargo is missing and skip the test like the build script does. That way test.ps1 won't fail for folks who haven't built opensim due to missing cargo install.

src/Simulation/qdk_sim_rs/test-qdk-sim-rs.ps1
```diff
@@ -0,0 +1,50 @@
...
+    if (Get-Item -ErrorAction SilentlyContinue $reportPath) {
+        New-Item -Type Directory -Force -Path $perfDest;
+        Copy-Item -Recurse -Force -Path $reportPath -Destination $perfDest;
+    }
+
+    # Free disk space by cleaning up.
+    # Note that this takes longer, but saves ~1 GB of space, which is
+    # exceptionally helpful in CI builds.
+        # When building in CI, free disk space by cleaning up.
+    # Note that this takes longer, but saves ~1 GB of space.
```
swernli, line 41: nitpick: I think this comment got accidentally duplicated by a merge at some point.

swernli (approved):
Two minor comments below, otherwise everything looks good (and matches previously reviewed content)!

msoeken (approved):
Approving previously approved PRs from feature branch to main. Merge looks good to me. Thanks @cgranade

src/Simulation/qdk_sim_rs/test-qdk-sim-rs.ps1
cgranade: Ah, good catch, will go on and fix, then. Thanks!

build/test.ps1
cgranade: Sounds good, let me fix that quickly, then.

src/Simulation/Simulators/OpenSystemsSimulator/NativeInterface.cs
```diff
@@ -0,0 +1,265 @@
...
+        {
+            if (errorCode != 0)
+            {
+                var error = _LastError();
+                throw new SimulationException($"Exception in native open systems simulator runtime: {error}", DLL_NAME);
+            }
+        }
+
+
+        public const string DLL_NAME = "Microsoft.Quantum.Experimental.Simulators.Runtime.dll";
```
kuzminrobin, line 55: FYI: I assume that the current support covers Windows only.

src/Simulation/Simulators/OpenSystemsSimulator/README.md
```diff
@@ -0,0 +1,65 @@
...
+pwsh ./bootstrap.ps1
+cd src/Simulation/qdk_sim_rs
+cargo +nightly doc --features python --open
+```
+
+For more details on using these simulators from Q# programs called from Python hosts, please see <https://github.com/microsoft/iqsharp>.
+
+## Native interface
+
+The core interoperability boundary between the C# runtime for the Quantum Development Kit and the `qdk_sim` native library is defined in the [`NativeInterface` static class](./NativeInterface.cs). This class provides P/Invoke declarations for each function exposed by the C API for the `qdk_sim` crate, as well as methods that call into these C API methods. The managed methods that correspond to each C API function check error codes reuturned by C API functions and convert them into .NET exceptions, allowing for C API errors to be easily caught by managed code.
```
kuzminrobin, line 18:
```suggestion
The core interoperability boundary between the C# runtime for the Quantum Development Kit and the `qdk_sim` native library is defined in the [`NativeInterface` static class](./NativeInterface.cs). This class provides P/Invoke declarations for each function exposed by the C API for the `qdk_sim` crate, as well as methods that call into these C API methods. The managed methods that correspond to each C API function check error codes returned by C API functions and convert them into .NET exceptions, allowing for C API errors to be easily caught by managed code.
```

src/Simulation/qdk_sim_rs/docs/c-api.md
```diff
@@ -0,0 +1,147 @@
...
+
+The [`qdk_sim`](..) crate has enabled the use of [`cbindgen`](https://crates.io/crates/cbindgen), such that C-language header files are generated automatically as part of the build for this crate.
+
+```bash
+cargo install --force cbindgen
+cbindgen --language C --output include/qdk_sim.h
+cbindgen --language C++ --output include/qdk_sim.hpp
+```
+
+This will generate `include/qdk_sim.h` and `include/qdk_sim.hpp, which can then be used from C and C++ callers, respectively. For example, to call from C:
```
kuzminrobin, line 25:
```suggestion
This will generate `include/qdk_sim.h` and `include/qdk_sim.hpp`, which can then be used from C and C++ callers, respectively. For example, to call from C:
```

src/Simulation/Simulators/OpenSystemsSimulator/NativeInterface.cs
cgranade: No, this is cross-platform, as is tested by e2e tests in our cross-repo builds. As with the existing `QuantumSimulator` class, however, we use the extension `*.dll` on all platforms when interoperating with .NET to avoid having to write a custom DLL loader that correctly redirects to OS-specific shared library names.

src/Simulation/qdk_sim_rs/docs/c-api.md
```diff
@@ -0,0 +1,147 @@
...
+    h(sim_id, 0);
+    h(sim_id, 1);
+
+    m(sim_id, 0, &result0);
+    m(sim_id, 1, &result1);
+
+    printf("got %llu %llu", result0, result1);
+    destroy(sim_id);
+}
+```
```
kuzminrobin, line 45: I may not know the full picture, and my note can be not applicable, but just in case I'll mention that:
* `uintptr_t` is an [optional type](https://www.cplusplus.com/reference/cstdint/).
* `uintptr_t` has architecture-dependent size, same as pointers (32-bit on 32-bit architecture, 64-bit on 64-bit architecture)

I would not recommend to use `uintptr_t` across the ABI boundary.

src/Simulation/qdk_sim_rs/docs/c-api.md
```diff
@@ -0,0 +1,147 @@
...
+
+To build and run the above example using Clang on Windows:
+
+```bash
+$ clang example.c -Iinclude -Ltarget/debug -lqdk_sim -lws2_32 -lAdvapi32 -lUserenv
+got 1 1
+```
```
kuzminrobin, line 52: Nit: The run command is skipped (I see the build command. Then the output of the run command is shown).

cgranade: Good point, thanks for catching that!

```suggestion
```bash
$ clang example.c -Iinclude -Ltarget/debug -lqdk_sim -lws2_32 -lAdvapi32 -lUserenv
$ ./a.exe
got 1 1
```
```

File: src/Simulation/qdk_sim_rs/docs/c-api.md [hunk 8]
```diff
@@ -0,0 +1,147 @@
...
+    uintptr_t sim_id;
+    uintptr_t result0, result1;
+
+    if (init(2, "invalid", &sim_id) != 0) {
+        printf("Got an error message: %s", lasterr());
+    }
+
+    destroy(sim_id);
+}
+```
```
kuzminrobin, line 72: Nit: I assume that `init()` and `destroy()` are a pair of corresponding functions (the same way as `malloc()` and `free()`, or constructor and destructor). If the first function of the pair (`malloc()`, constructor) _fails_, then the second function of the pair (`free()`, destructor) _is not called_. 
In your example the `destroy()` is called unconditionally (even if `init()` fails).

src/Simulation/qdk_sim_rs/docs/c-api.md
cgranade: In this case, the example uses the types in `qdk_sim.h`, as generated by cbindgen. E.g., the autogenerated `qdk_sim.h` contains the following declaration:

```C
#include <stdarg.h>
#include <stdbool.h>
#include <stdint.h>
#include <stdlib.h>

// ...

int64_t init(uintptr_t initial_capacity, const char *representation, uintptr_t *sim_id_out);
```

My suggestion as to how to best fix that would be to change `usize` declarations in `c_api.rs` to `u64`. Since that could affect our e2e builds, I'd prefer to not make that change this close to release but to defer to improvements in the feature post-release.

kuzminrobin: Whatever you choose to do is totally fine with me. I just want to make sure that you are fully aware of the risks if I see any.

src/Simulation/qdk_sim_rs/docs/c-api.md
cgranade: Fair, agreed. My point with this was to show a minimal example to demonstrate conventions, but agreed that the example could be improved. That said, calling `destroy` is safe even if `init` has not been called:

```rust
#[no_mangle]
pub extern "C" fn destroy(sim_id: usize) -> i64 {
    as_capi_err(|| {
        let state = &mut *STATE.lock().unwrap();
        if state.contains_key(&sim_id) {
            state.remove(&sim_id);
            Ok(())
        } else {
            Err(format!("No simulator with id {} exists.", sim_id))
        }
    })
}
```

If `init` has not been called for a given `sim_id`, then the `state.contains_key` check fails, such that an error code is returned to the C caller, and an actionable error message is set to `lasterr`. On the other hand, using an uninitialized `sim_id` on the C side could cause problems, such that I still agree with improving the example here.

cgranade:
```suggestion
```c
#include <stdio.h>
#include "qdk_sim.h"

int main() {
    uintptr_t sim_id;
    uintptr_t result0, result1;

    if (init(2, "invalid", &sim_id) != 0) {
        printf("Got an error message: %s", lasterr());
    } else {
        destroy(sim_id);
    }
}
```
```

Comment (kuzminrobin) [reply src/Simulation/qdk_sim_rs/docs/c-api.md:72]: Looks better.
Consider also:
```c
    if (init(2, "invalid", &amp;sim_id) != 0) {
        printf("Got an error message: %s", lasterr());
        return 1;
    }
    destroy(sim_id);
    return 0;
}
```
In the earlier examples the `main()` was not returning the value (I've just noticed).

My example may clutter the minimalistic illustration of the idea with the unrelated fragments. Feel free to ignore and stay minimalistic.

src/Simulation/qdk_sim_rs/docs/c-api.md
```diff
@@ -0,0 +1,147 @@
...
+
+For example:
+
+```c
+uintptr_t result;
+if (m(sim_id, 0, &result) != 0) {
+    printf("Got an error message: %s", lasterr());
+}
+printf("Got a measurement result: %llu", result);
+```
```
kuzminrobin, line 86: FYI: (in general case, without looking into the implementation of `m()`)

If `m()` fails then the `result` can stay _uninitialized_. In that case the last unconditional `printf()` can print a misleading message.
Consider
```c
else
{
    printf("Got a measurement result: %llu", result);
}
```

src/Simulation/qdk_sim_rs/docs/c-api.md
```diff
@@ -0,0 +1,147 @@
...
+
+To build and run the above example using Clang on Windows:
+
+```bash
+$ clang example.c -Iinclude -Ltarget/debug -lqdk_sim -lws2_32 -lAdvapi32 -lUserenv
+$ ./a.exe
+got 1 1
+
+## Error Handling and Return Values
```
kuzminrobin, line 54: After `got 1 1` the triple tick is missing (to close the code fragment).

src/Simulation/qdk_sim_rs/src/c_api.rs
```diff
@@ -0,0 +1,443 @@
...
+
+/// Returns the last error message raised by a call to a C function.
+/// If no error message has been raised, returns a null pointer.
+#[no_mangle]
+pub extern "C" fn lasterr() -> *const c_char {
+    match &*LAST_ERROR.lock().unwrap() {
+        None => ptr::null(),
+        Some(msg) => CString::new(msg.as_str()).unwrap().into_raw(),
+    }
+}
```
kuzminrobin, line 92: (Not necessarily in this PR, just something to keep in mind, _if applicable_. I don't know Rust)

I have an impression that `lasterr()` returns a pointer to the _heap-allocated block_. If that's the case then the users are likely to expect a mechanism (in the same binary that made an allocation) to deallocate such blocks (and documentation about that).

src/Simulation/qdk_sim_rs/src/c_api.rs
```diff
@@ -0,0 +1,443 @@
...
+///
+/// After this call completes, this function guarantees that either of the two
+/// conditions below holds:
+///
+/// - The return value is negative, in which case calling `lasterr` will return
+///   an actionable error message, or
+/// - The return value is `0`, and `*noise_model_json` is a valid pointer to a
+///   null-terminated string of Unicode characters, encoded as UTF-8. In this
+///   case, the caller is considered to own the memory allocated for this
+///   string.
```
kuzminrobin, line 282: > the caller is considered to own the memory allocated for this string

For the future, just in case, I would like to refer you to [Allocate and Deallocate In The Same Binary](https://github.com/microsoft/qsharp-runtime/blob/main/src/Qir/QIR-RT-API-Design-Guidelines.md#allocate-and-deallocate-in-the-same-binary) recommendation.
(something to keep in mind)

cgranade: Good point; in that case, I would expect that the caller can deallocate by passing their owned memory back to the Rust library, such that allocation and deallocation occur in the same binary. To make C# interoperability work, we would need to then call [`Marshal.PtrToStringUTF8`](https://docs.microsoft.com/dotnet/api/system.runtime.interopservices.marshal.ptrtostringutf8) to copy out that string before calling the deallocator. I'll open an issue to track improving string handling across this API. Thanks!

src/Simulation/qdk_sim_rs/src/c_api.rs
```diff
@@ -0,0 +1,443 @@
...
+/// calling into this function. The caller is responsible for ensuring that:
+///
+/// - `noise_model_json` is a valid pointer whose lifetime extends for the
+///    duration of this function call.
+///
+/// After this call completes, this function guarantees that either of the two
+/// conditions below holds:
+///
+/// - The return value is negative, in which case calling `lasterr` will return
+///   an actionable error message, or
```
kuzminrobin, line 315: Nit: Returning an arbitrary negative value as an indication of an error, limits the space for the subsequent maneuver. 

At some point in the future we may want to distinguish one group of failures from the other group, by returning `-1` vs. `-2`. You may want to consider limiting the error indication to one value (e.g. `-1`).
(to keep in mind for the future)

cgranade: Fair; in this case, all failures get translated to the same C# exception, but I agree, we can consider providing more granular failure codes in the future.

src/Simulation/qdk_sim_rs/docs/c-api.md
cgranade: Makes sense! I'll add that note to the same issue for tracking C API improvements along with deallocation handling and granular return codes.

src/Simulation/qdk_sim_rs/src/c_api.rs
```diff
@@ -0,0 +1,443 @@
...
+///   encoded as UTF-8, and that the pointer remains valid for the lifetime of
+///   this call.
+/// - `noise_model_json` is a valid pointer whose lifetime extends for the
+///    duration of this function call.
+///
+/// After this call completes, this function guarantees that either of the two
+/// conditions below holds:
+///
+/// - The return value is negative, in which case calling `lasterr` will return
+///   an actionable error message, or
```
kuzminrobin, line 278: (FYI for the future) In the `c-api.md` there is a fragment:
> Most C API functions for this crate return an integer, with `0` indicating success and **any other value indicating failure**.

The examples in `c-api.md` treat any non-zero return value (even _positive one_) as a failure.

If at some point in the future one wants to distinguish one group of success from the other group of success (e.g. "success with warning") by returning `0` vs. `1` (positive), then such a change can become a breaking change for a huge error-handling codebase.

You may want to consider one value for success (e.g. `0`) and one value for failure (e.g. `-1`), making _all other values reserved for the future_.

src/Simulation/qdk_sim_rs/docs/c-api.md
cgranade: https://github.com/microsoft/qsharp-runtime/issues/732

src/Simulation/qdk_sim_rs/src/c_api.rs
cgranade: https://github.com/microsoft/qsharp-runtime/issues/732

src/Simulation/qdk_sim_rs/src/c_api.rs
cgranade: https://github.com/microsoft/qsharp-runtime/issues/732

src/Simulation/qdk_sim_rs/src/c_api.rs
cgranade: https://github.com/microsoft/qsharp-runtime/issues/732

kuzminrobin (approved):
Incredible work! Large number of technologies used.
Unreachably high pilotage for me.
I admire your contributions.

PR #721, opened 2021-06-10
Credentials parameter

anpaz:
Based on the guideline we received from the Azure SDK, automatically using `DefaultCredential` is not the recommended approach, instead the user should create their own credentials and pass it the API, as such we changed the `Workspace` interface to make `credential` a required parameter.
To accommodate for this change, the Q# executables will now accept a new `credential` parameter, that takes an enumeration of possible credential classes to use to authenticate; this gives the user a lot more control and help us mitigates any problems the user might find with a specific credential type.
We also change it to create a custom TokenCredential so we can consume the AAD token if received as parameter. This mostly to keep it consistent with the current `az cli` usage.


vxfield (changes requested):
As discussed, we should at least allow the user to specify the TenantId.

vxfield (approved):
Looks good to me

Issue #618, opened 2021-04-08
Move CSharpGeneration to the qsharp-compiler repo


swernli:
CSharpGeneration is a bit exceptional in this repo as the only project that doesn't use the QDK but takes a dependency on a specific compiler package version. If this version is not kept in sync with the rest of the repo it can cause strange test failures when the rewrite step is loaded even compilation appeared to succeed. This rewrite step probably better fits in the qsharp-compiler repo along with other compiler rewrite steps.

swernli:
This suggestion has been discussed for various reasons, one of which stemmed from #530. 
@bettinaheim FYI

PR #722, opened 2021-06-10
Remove CSharpGeneration package

swernli:
This change removes the CSharpGeneration package in favor of having it built from within compiler (see https://github.com/microsoft/qsharp-compiler/pull/1054). This is part of fixing #618. This change is not expected to build properly until the PR to qsharp-compiler is completed.

Also of note, this removes the old Microsoft.Quantum.Development.Kit which is no longer used and all of the issues in the readme there with custom .NET SDKs have been resolved.


src/Simulation/AutoSubstitution.Integration.Tests/Tests.Microsoft.Quantum.AutoSubstitution.Integration.csproj
```diff
@@ -29,7 +29,10 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <ProjectReference Include="..\CSharpGeneration\Microsoft.Quantum.CSharpGeneration.fsproj" PrivateAssets="All" IsQscReference="true" />
+    <PackageReference Include="Microsoft.Quantum.CSharpGeneration" Version="0.17.2106147878-beta" IsQscReference="true" />
```
samarsha, line 32: If we remove `<CSharpGeneration>false</CSharpGeneration>`, can we remove this `PackageReference`?

src/Simulation/AutoSubstitution/Microsoft.Quantum.AutoSubstitution.csproj
```diff
@@ -20,11 +20,11 @@
   
   <ItemGroup>
     <PackageReference Include="Microsoft.Quantum.Compiler" Version="0.17.2106147878-beta" />
+    <PackageReference Include="Microsoft.Quantum.CSharpGeneration" Version="0.17.2106147878-beta" IsQscReference="true" />
```
samarsha, line 23: If we remove `<CSharpGeneration>false</CSharpGeneration>`, can we remove this `PackageReference`?

src/Simulation/Simulators.Tests/TestProjects/QSharpExe/QSharpExe.csproj
```diff
@@ -11,7 +11,11 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <ProjectReference Include="..\..\..\CSharpGeneration\Microsoft.Quantum.CSharpGeneration.fsproj" PrivateAssets="All" IsQscReference="true" />
+    <PackageReference Include="Microsoft.Quantum.CSharpGeneration" Version="0.17.2106147878-beta" IsQscReference="true" />
```
samarsha, line 14: If we remove `<CSharpGeneration>false</CSharpGeneration>`, can we remove this `PackageReference`?

src/Simulation/Simulators.Tests/TestProjects/TargetedExe/TargetedExe.csproj
```diff
@@ -12,7 +12,11 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <ProjectReference Include="..\..\..\CSharpGeneration\Microsoft.Quantum.CSharpGeneration.fsproj" PrivateAssets="All" IsQscReference="true" />
+    <PackageReference Include="Microsoft.Quantum.CSharpGeneration" Version="0.17.2106147878-beta" IsQscReference="true" />
```
samarsha, line 15: If we remove `<CSharpGeneration>false</CSharpGeneration>`, can we remove this `PackageReference`?

src/Simulation/TargetDefinitions/TargetPackages/Common.Package.props
```diff
@@ -23,7 +22,7 @@
   <ItemGroup>
     <ProjectReference Include="..\TargetDefinitions\Interfaces\Microsoft.Quantum.Targets.Interfaces.csproj" />
     <ProjectReference Include="..\QSharpFoundation\Microsoft.Quantum.QSharp.Foundation.csproj" />
-    <ProjectReference Include="..\CSharpGeneration\Microsoft.Quantum.CSharpGeneration.fsproj" PrivateAssets="All" IsQscReference="true" />
+    <ProjectReference Include="..\Core\Microsoft.Quantum.Runtime.Core.csproj" />
```
samarsha, line 25: How come the extra reference to Core is needed in some of these project files?

samarsha (approved):
Might want to double check with @bettinaheim if it's ok to remove the Microsoft.Quantum.Development.Kit and Microsoft.Quantum.CSharpGeneration.App packages (even though they're very old).

src/Simulation/AutoSubstitution.Integration.Tests/Tests.Microsoft.Quantum.AutoSubstitution.Integration.csproj
swernli: Ah, that's a good idea! I hadn't noticed that. I'll do a pass through and fix up the ones you pointed out and see if there any others that could benefit from that.

build/manifest.ps1
```diff
@@ -37,8 +35,6 @@ $artifacts = @{
     Assemblies = @(
         ".\src\Azure\Azure.Quantum.Client\bin\$Env:BUILD_CONFIGURATION\netstandard2.1\Microsoft.Azure.Quantum.Client.dll",
         ".\src\Simulation\AutoSubstitution\bin\$Env:BUILD_CONFIGURATION\netstandard2.1\Microsoft.Quantum.AutoSubstitution.dll",
-        ".\src\Simulation\CSharpGeneration\bin\$Env:BUILD_CONFIGURATION\netstandard2.1\Microsoft.Quantum.CSharpGeneration.dll",
-        ".\src\Simulation\CSharpGeneration.App\bin\$Env:BUILD_CONFIGURATION\netcoreapp3.1\Microsoft.Quantum.CSharpGeneration.App.dll",
         ".\src\Simulation\RoslynWrapper\bin\$Env:BUILD_CONFIGURATION\netstandard2.1\Microsoft.Quantum.RoslynWrapper.dll",
```
samarsha, line 38:
```suggestion
```

src/Simulation/AutoSubstitution.Integration.Tests/Tests.Microsoft.Quantum.AutoSubstitution.Integration.csproj
swernli: Darn... that doesn't work that well, because of the fact that CSharpGeneration will cause the include of the package Runtime.Core: 
https://github.com/microsoft/qsharp-compiler/blob/6bf4159e57b11e6046ebc68f297e939637d53641/src/QuantumSdk/Sdk/Sdk.props#L71-L72

src/Simulation/Simulators.Tests/TestProjects/QSharpExe/QSharpExe.csproj
swernli: Because this needs to include the locally compiled `Microsoft.Quantum.Runtime.Core.csproj` I don't have a choice and need to use `<CSharpGeneration>false</CSharpGeneration>`, which then requires that I use a package reference to CSharpGeneration. I'm looking into how to centralize these version numbers so I don't start introducing the same problem we worked to remove with the QDK version numbers.

src/Simulation/AutoSubstitution.Integration.Tests/Tests.Microsoft.Quantum.AutoSubstitution.Integration.csproj
swernli: Maybe if we add a property flag that lets us suppress the inclusion of the Runtime.Core package? It would simplify some of this usage here...

swernli: I'm attempting to add this functionality with https://github.com/microsoft/qsharp-compiler/commit/fc1d0ae0158c6eecdcb5daaed900c56b8cbd136f

src/Simulation/TargetDefinitions/TargetPackages/Common.Package.props
swernli: They need that to resolve some of the types defined in the core, but it's not clear to me how they were getting it before, since CSharpGeneration doesn't include that package. After making the other updates I'll see if this is still needed...

swernli: The errors end up looking like this:
```
obj/qsharp/__AutoSubstitution__.g.cs(33,49): error CS0012: The type 'Qubit' is defined in an assembly that is not referenced. You must add a reference to assembly 'Microsoft.Quantum.Runtime.Core, Version=0.16.2106.2312, Culture=neutral, PublicKeyToken=null'. [/home/vsts/work/1/s/src/Simulation/AutoSubstitution.Integration.Tests/Tests.Microsoft.Quantum.AutoSubstitution.Integration.csproj]
```

src/Simulation/AutoSubstitution.Integration.Tests/Tests.Microsoft.Quantum.AutoSubstitution.Integration.csproj
swernli: Ok, I was finally able to resolve this (and fix a bug in the process). The required package inclusions are now much simpler.

src/Simulation/Simulators.Tests/TestProjects/QSharpExe/QSharpExe.csproj
swernli: This is now resolved and uses the right packages from the right places.

src/Simulation/TargetDefinitions/TargetPackages/Common.Package.props
swernli: After shuffling some things around, the explicit reference to the core is no longer needed, and this should now compile without it.

PR #724, opened 2021-06-12
Finish remaining decompositions for initial version of experimental simulators.

cgranade:
This PR adds remaining decompositions for experimental simulators, excluding those operations which will not be supported in the initial experimental release.


src/Simulation/Simulators.Tests/OpenSystemsSimulatorTests/TestOperations.qs
```diff
@@ -0,0 +1,197 @@
...
+            ResetAll([left, right]);
+        }
+    }
+
+    internal function Xor(a : Bool, b : Bool) : Bool {
+        return (a or b) and ((not a) or (not b));
+    }
+
+    @Test("Microsoft.Quantum.Experimental.OpenSystemsSimulator")
+    @Test("QuantumSimulator") // validate against full-state simulator.
```
swernli, line 126: This is the only test case that also gets run against the full-state simulator. Besides overall pipeline perf, is there any reason the other tests shouldn't be run against the full-state simulator too?

src/Simulation/Simulators/OpenSystemsSimulator/Decompositions/IntrinsicInterface.qs
```diff
@@ -0,0 +1,210 @@
...
+        controlled (controls, ...) {
+            if Length(controls) == 0 {
+                Z(target);
+            } elif Length(controls) == 1 {
+                within {
+                    H(target);
+                } apply {
+                    Native.CNOT(controls[0], target);
+                }
+            } elif Length(controls) == 1 {
```
swernli, line 95: This should be `Length(controls) == 2`.

src/Simulation/Simulators/OpenSystemsSimulator/Decompositions/IntrinsicInterface.qs
```diff
@@ -0,0 +1,210 @@
...
+            if bases[idxBasis] != PauliI {
+                set newBases += [bases[idxBasis]];
+                set newQubits += [register[idxBasis]];
+            }
+        }
+
+        if Length(newBases) == 0 {
+            return Zero;
+        } else {
+            return Measure(newBases, register);
```
swernli, line 173: I think this needs to be `Measure(newBases, newQubits)` to ensure the matching sub-register is used.

src/Simulation/Simulators/OpenSystemsSimulator/Decompositions/IntrinsicInterface.qs
```diff
@@ -0,0 +1,210 @@
...
+                within {
+                    H(target);
+                } apply {
+                    Native.CNOT(controls[0], target);
+                }
+            } elif Length(controls) == 1 {
+                // [Page 15 of arXiv:1206.0758v3](https://arxiv.org/pdf/1206.0758v3.pdf#page=15)
+                Adjoint T(controls[0]);
+                Adjoint T(controls[1]);
+                Controlled X([target], controls[0]);
```
swernli, line 99: It's a very minor optimization, but consider using `Native.CNOT` here instead of `Controlled X` (and in other decompositions) as that can avoid some extra overhead in invoking the specailization code and should be safe to use here since `Native.CNOT` is already explicitly implemented as self-adjoint.

src/Simulation/Simulators/OpenSystemsSimulator/Decompositions/IntrinsicInterface.qs
```diff
@@ -0,0 +1,210 @@
...
+            } else {
+                // TODO: Decompositions of `Controlled T` currently used in Q#
+                //       target packages rely on R1Frac, which is not yet
+                //       implemented in experimental simulation.
+                fail "Controlled T operations are not yet supported.";
+            }
+        }
+    }
+
+    // NB: We separate out this operation to avoid hardware targeting rewrite
```
swernli, line 155: Might be good to include a mention of https://github.com/microsoft/qsharp-compiler/issues/768 here.

src/Simulation/Simulators/OpenSystemsSimulator/Decompositions/IntrinsicInterface.qs
```diff
@@ -0,0 +1,210 @@
...
+            let target = register[0];
+            mutable result = Zero;
+            within {
+                if basis == PauliX {
+                    Native.H(target);
+                } elif basis == PauliY {
+                    Native.H(target);
+                    Native.S(target);
+                } else {
+                }
```
swernli, line 199: nitpick: Putting a comment about Pauli-Z in the empty else-block will make the intent a little clearer and the pattern easier to see.

src/Simulation/Simulators/OpenSystemsSimulator/Decompositions/NativeInterface.qs
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+// This namespace contains those intrinsics supported by the experimental
+// simulators native interface (NativeInterface.cs).
+
+namespace Microsoft.Quantum.Experimental.Native {
+
+    operation H(target : Qubit) : Unit {
```
swernli, line 9: nitpick: Even though it is not strictly necessar, I'm personally in favor of including the `is Adj` in the signature even when the operation is explicitly declared as self-adjoint. It just makes it easier to see at a glance that the callable indeed supports the specialization.

src/Simulation/Simulators/OpenSystemsSimulator/QubitManager.cs
```diff
@@ -0,0 +1,49 @@
...
+                    // TODO: There are better ways to do this; increment on the
+                    //       design and make it customizable.
+                    // FIXME: In particular, this implementation uses a lot of
+                    //        extraneous measurements.
+                    if ((this.Parent as IIntrinsicMeasure).Body(new QArray<Pauli>(Pauli.PauliZ), new QArray<Qubit>(qubit)) == Result.One)
+                    {
+                        (this.Parent as IIntrinsicX).Body(qubit);
+                    }
+                }
+                base.Release(qubit, wasUsedOnlyForBorrowing);
```
swernli, line 44: I'm a little bit confused here... for qubits that have been measured the logic above will reset them from the One state to the Zero state (or leave them in Zero if they are already there). Otherwise, the qubit goes straight to release without any further simulator calls. Doesn't this mean that a qubit that is entangled may get "leaked" such that the simlator still tracks it but the qubit manager thinks it's available for reuse?

swernli (changes requested):
I found a few bugs in decompositions, and have a question about qubit release as well. Also, I can't comment on the binary file directly, but I noticed this payload includes the addition of Microsoft.Quantum.Experimental.Simulators.Runtime.dll to the src/Simulation/Native/win10 folder, and was wondering if that committed binary was intentional or accidental?

src/Simulation/Simulators.Tests/OpenSystemsSimulatorTests/TestOperations.qs
cgranade: Was indeed thinking of overall pipeline performance, given we routinely hit up against the limit for job time. That said, adding more validations against full-state simulator shouldn't cost that much time, so let me go on and add that.

src/Simulation/Simulators/OpenSystemsSimulator/Decompositions/NativeInterface.qs
cgranade: Good point, added.

src/Simulation/Simulators/OpenSystemsSimulator/Decompositions/IntrinsicInterface.qs
cgranade: Yes, thank you for catching that!

src/Simulation/Simulators/OpenSystemsSimulator/Decompositions/IntrinsicInterface.qs
cgranade: Done.

src/Simulation/Simulators/OpenSystemsSimulator/Decompositions/IntrinsicInterface.qs
cgranade: Good point, fixed.

src/Simulation/Simulators/OpenSystemsSimulator/Decompositions/IntrinsicInterface.qs
cgranade: Good point, yeah. I'll go on and add that.

src/Simulation/Simulators/OpenSystemsSimulator/Decompositions/IntrinsicInterface.qs
cgranade: Good point, went on and applied that optimization.

src/Simulation/Simulators/OpenSystemsSimulator/QubitManager.cs
cgranade: Right, if I understand correctly that's the correct behavior here. If the qubit has not been measured, then the developer is implicitly promising that they have coherently reset it. On a full-state simulator, we can check that that condition holds, as per your PRs to that effect, but on a noisy simulator, a qubit could validly be entangled with something as a result of noise causing a coherent unpreparation to not work correctly. In that case, we explicitly want to track and simulate the errors caused by that leakage.

swernli: Ah, right! That does make sense, thanks for explaining.

src/Simulation/Simulators/OpenSystemsSimulator/Decompositions/IntrinsicInterface.qs
```diff
@@ -0,0 +1,212 @@
...
+                within {
+                    H(target);
+                } apply {
+                    Native.CNOT(controls[0], target);
+                }
+            } elif Length(controls) == 2 {
+                // [Page 15 of arXiv:1206.0758v3](https://arxiv.org/pdf/1206.0758v3.pdf#page=15)
+                Adjoint Native.T(controls[0]);
+                Adjoint Native.T(controls[1]);
+                Controlled Native.X([target], controls[0]);
```
swernli, line 98: Whoops, looks like this one got partially updated.

src/Simulation/Simulators/OpenSystemsSimulator/Decompositions/IntrinsicInterface.qs
```diff
@@ -0,0 +1,212 @@
...
+        }
+
+        controlled (controls, ...) {
+            if Length(controls) == 0 {
+                Native.S(target);
+            } elif (Length(controls) == 1) {
+                Native.T(controls[0]);
+                Native.T(target);
+                within {
+                    Controlled X([controls[0]], target);
```
swernli, line 127: This spot could also become `Native.CNOT`.

src/Simulation/Simulators/OpenSystemsSimulator/QubitManager.cs
cgranade: No worries, let me go on and add a comment to that effect... it took me a moment to remember the reasoning here, so at the least it's a good reminder to myself!

src/Simulation/Simulators/OpenSystemsSimulator/Decompositions/IntrinsicInterface.qs
cgranade: Just noticed that as well, updated in 3591151c7c3ba48ec98ffbdd6ecfde9a254e4f33.,

src/Simulation/Simulators/OpenSystemsSimulator/Decompositions/IntrinsicInterface.qs
cgranade: Also fixed in 3591151c7c3ba48ec98ffbdd6ecfde9a254e4f33.

swernli (approved):
Changes look good!

cgranade:
e2e build is passing at https://dev.azure.com/ms-quantum-public/Microsoft%20Quantum%20(public)/_build/results?buildId=27967&view=results. It's marked as failing above, as the changes to the cross-repo build definitions aren't being picked up by the GitHub Checks integration feature. Merging to feature branch since e2e and local builds both pass.

PR #725, opened 2021-06-12
Added compiler warning flags -Wall, -Wextra

kuzminrobin:


kuzminrobin:
Ready to merge, please review.

src/Qir/Common/Include/SimulatorStub.hpp
```diff
@@ -15,113 +15,113 @@ namespace Quantum
...
+            Qubit* /*controls*/,
+            long /*numTargets*/,
+            PauliId* /*paulis*/,
+            Qubit* /*targets*/,
+            double /*theta*/) override
         {
             throw std::logic_error("not_implemented: ControlledExp");
         }
-        void AdjointS(Qubit target) override
+        void AdjointS(Qubit) override
```
swernli, line 96: For most of these the parameter name was commented out, but for `AdjointS` here and a few other functions throughout this file it was simply removed. I don't have a strong preference to either style, but do like following one pattern. For what it's worth, leaving the parameter names is helpful for understanding their intended use while on the other hand removing it entirely avoids the overhead of keeping them up to date if the names change in the future.

src/Qir/Runtime/lib/QIR/callables.cpp
```diff
@@ -370,7 +370,8 @@ QirTupleHeader* FlattenControlArrays(QirTupleHeader* tuple, int depth)
 
         QirArray* controls = current->controls;
         const size_t blockSize = qubitSize * controls->count;
-        assert(dst + blockSize <= dstEnd);
+        assert(dst + blockSize <= dstEnd); 
+        (void)dstEnd;   // Calm down the "Unused Var" warning in Release build (`assert()` is Debug only).
```
swernli, line 374: Food for thought: A pattern I've seen in other codebases is to create a macro named `UNUSED` or `UNREFERENCED_ARG` or something similar that does this, just to make the intent clear. Your comment here is clear enough such that I don't think this is really a problem and there aren't that many places we need this pattern (and I'm not certain they all share a convenient header to define that macro in), so feel free to ignore.

src/Qir/Runtime/lib/Simulators/FullstateSimulator.cpp
```diff
@@ -59,14 +59,14 @@ QUANTUM_SIMULATOR LoadQuantumSimulator()
...
 
-bool UnloadQuantumSimulator(QUANTUM_SIMULATOR handle)
-{
-#ifdef _WIN32
-    return ::FreeLibrary(handle);
-#else // not _WIN32
-    return ::dlclose(handle);
-#endif
-}
+// bool UnloadQuantumSimulator(QUANTUM_SIMULATOR handle)
```
swernli, line 62: I think it would be fine to remove this rather than leave it commented out. I don't foresee us ever really supporting this.

swernli (approved):
A few minor things for you to consider, but overall the change looks fine!

src/Qir/Common/Include/SimulatorStub.hpp
kuzminrobin: Made them all commented out.

src/Qir/Runtime/lib/QIR/callables.cpp
kuzminrobin: Good point, thank you! 
I'm adding `UNUSED()`.

src/Qir/Runtime/lib/Simulators/FullstateSimulator.cpp
kuzminrobin: Removing.

PR #737, opened 2021-06-18
Fix broken links

anjbur:


src/Simulation/qdk_sim_rs/docs/c-api.md
```diff
@@ -146,4 +146,4 @@ Noise model:
...
 ```
 
-See [noise model serialization](crate#noise-model-serialization) for more details.
```
cgranade, line 149: This may break the link as published to docs.rs once this crate is published... let me check with a local `cargo doc` build.

cgranade: Confirmed, this change results in a new broken link.

cgranade (changes requested):
Marking as request changes for the link in `c-api.md` but this otherwise looks good to me.

src/Simulation/qdk_sim_rs/docs/c-api.md
anjbur: Got it, reverted the change!

cgranade (approved):

PR #739, opened 2021-06-18
Warn when qdk_sim is missing instead of failing outright.

cgranade:
This PR allows for building the simulators package to proceed even when `qdk_sim_rs` has not yet been built, as per feedback from @anpaz. Simulator packages built this way will fail in unit and integration testing, but can still be useful for local development of other simulation features on machines without `rustup` / `cargo`.

### DIFF
--- a/src/Simulation/Common/Simulators.Dev.props
+++ b/src/Simulation/Common/Simulators.Dev.props
@@ -42,12 +42,19 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <None Include="$(ExperimentalSimDll)" >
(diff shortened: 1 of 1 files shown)


swernli (approved):

kuzminrobin (approved):
For me locally tis PR's changes have solved the issue.

I took the latest from repo. Rebased my branch on top of main. Took to clean dir. The bootstrap failed as expected.
I took to clean dir. Made local changes identical to this PR. The bootstrap and all subsequent build steps succeeded.

The changes of this PR have solved the problem for me locally. Thanks!

PR #742, opened 2021-06-19
Do not include dependency on Standard library

### LINKED ISSUE #741: Autosubstitution depends on packages pulled in by default by the Microsoft.Quantum.Sdk
## Describe the bug
To keep repo dependencies clean, no packages in qsharp-runtime can have dependencies on QuantumLibraries, as the dependencies need to flow just one way. 

By default, though, the Microsoft.Quantum.Sdk adds dependencies to Standard unless turned off by the 
`IncludeQSharpCorePackages` property, so this property is required in all projects on the qsharp-runtime repo.

## To Reproduce
Check the package dependencies in either VS or nuget, and check that Standard is marked as dependency.

anpaz:
Fixes #741

### DIFF
--- a/src/Simulation/AutoSubstitution/AutoSubstitution.csproj
+++ b/src/Simulation/AutoSubstitution/AutoSubstitution.csproj
@@ -11,9 +11,20 @@
     <AssemblyName>Microsoft.Quantum.AutoSubstitution</AssemblyName>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <IncludeQSharpCorePackages>false</IncludeQSharpCorePackages>
+    <CSharpGeneration>false</CSharpGeneration>
+    <QSharpDocsGeneration>false</QSharpDocsGeneration>
+  </PropertyGroup>
+
   <ItemGroup>
     <PackageReference Include="Microsoft.Quantum.Compiler" Version="0.17.2106147878-beta" />
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="3.6.0" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\CSharpGeneration\Microsoft.Quantum.CSharpGeneration.fsproj" />
+    <ProjectReference Include="..\QSharpFoundation\Microsoft.Quantum.QSharp.Foundation.csproj" />
   </ItemGroup>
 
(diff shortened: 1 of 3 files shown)


guenp:
@anpaz @msoeken should we merge this PR first and then resolve naming conflicts in this PR? https://github.com/microsoft/qsharp-runtime/pull/743

src/Simulation/AutoSubstitution/AutoSubstitution.csproj
```diff
@@ -11,11 +11,22 @@
...
+    <QSharpDocsGeneration>false</QSharpDocsGeneration>
+  </PropertyGroup>
+
   <ItemGroup>
     <PackageReference Include="Microsoft.Quantum.Compiler" Version="0.17.2106147878-beta" />
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="3.6.0" />
   </ItemGroup>
 
+  <ItemGroup>
+    <ProjectReference Include="..\CSharpGeneration\Microsoft.Quantum.CSharpGeneration.fsproj" />
```
bettinaheim, line 26: I believe this project also relies on that C# code for is generated.

src/Simulation/AutoSubstitution/AutoSubstitution.csproj
```diff
@@ -11,11 +11,22 @@
...
+    <ProjectReference Include="..\QSharpFoundation\Microsoft.Quantum.QSharp.Foundation.csproj" />
+  </ItemGroup>
+
   <ItemGroup>
```
anpaz, line 30: do you need to add the `BeforeCSharpCompile` target as in the other project?

anpaz (commented):
lgtm (I can't approve since I started the review)

bettinaheim:
@anpaz @msoeken changes were incorporated into #748.

src/Simulation/AutoSubstitution/AutoSubstitution.csproj
msoeken: It did not work without adding this part.

msoeken:
> @anpaz @msoeken changes were incorporated into #748.

@beheim Should we close this PR then?

bettinaheim:
@msoeken I believe it can be, yes.

PR #750, opened 2021-06-22
New example notebook with conceptual overview of open quantum systems.

cgranade:
As per feedback from @tcNickolas, this PR builds on existing documentation of the experimental simulators added in 0.18 by also adding a new notebook with a conceptual overview of open quantum systems, in particular density operators and quantum processes. This notebook presents relevant concepts with a focus on how these concepts are used in the new experimental simulators, and uses the new experimental simulators together with QuTiP to introduce these concepts and relevant notation to help users onboard to using these features.

**Update**: Based on feedback from @efratshabtai and @bromeg, this PR also moves to using the term "preview" instead of "experimental."


cgranade:
This seems to be failing on unit tests unrelated to the changes in this PR; @bettinaheim, @swernli have you seen this failure before, or is it new?

efratshabtai (approved):

swernli:
> This seems to be failing on unit tests unrelated to the changes in this PR; @bettinaheim, @swernli have you seen this failure before, or is it new?

From the logs, this appears to be another instances of the intermittent failure we see in simulator synchronization:
`The active test run was aborted. Reason: Test host process crashed`
Other platforms succeeded and later test runs past, so I think it is safe to ignore this failure. We do have a possible fix but it hasn't been merged because it's currently tied in with https://github.com/microsoft/qsharp-runtime/pull/710.

cgranade:
> From the logs, this appears to be another instances of the intermittent failure we see in simulator synchronization:
> `The active test run was aborted. Reason: Test host process crashed`
> Other platforms succeeded and later test runs past, so I think it is safe to ignore this failure. We do have a possible fix but it hasn't been merged because it's currently tied in with #710.

Got it, thank you! I hadn't noticed that was the upstream root cause, sorry, as I'd not seen that particular effect from the test host crash, but that makes a lot of sense now that you point it out.

cgranade:
The failure above is due to a build agent not downloading artifacts correctly and thus failing validations. Re-ran the validation step since it appears to be a transient network failure.

PR #753, opened 2021-06-24
Resolved some more compiler warnings.

kuzminrobin:


kuzminrobin:
I added a final change to this PR, please review.

src/Qir/Runtime/lib/QIR/QubitManager.cpp
```diff
@@ -34,54 +34,54 @@ static void FailIf(bool condition, const char* message)
 CQubitManager::QubitListInSharedArray::QubitListInSharedArray(
     QubitIdType startId,
     QubitIdType endId,
-    QubitIdType* sharedQubitStatusArray):
+    QubitIdType* sharedQbitStatusArray):
```
cesarzc, line 37: I believe the name of this parameter has a typo. Should it be `sharedQubitStatusArray`?

src/Qir/Runtime/lib/QIR/QubitManager.cpp
```diff
@@ -92,15 +92,15 @@ CQubitManager::QubitIdType CQubitManager::QubitListInSharedArray::TakeQubitFromF
...
     {
-        sharedQubitStatusArray[result] = AllocatedMarker;
+        sharedQbitStatusArray[result] = AllocatedMarker;
     }
 
     return result;
 }
 
-void CQubitManager::QubitListInSharedArray::MoveAllQubitsFrom(QubitListInSharedArray& source, QubitIdType* sharedQubitStatusArray)
+void CQubitManager::QubitListInSharedArray::MoveAllQubitsFrom(QubitListInSharedArray& source, QubitIdType* sharedQbitStatusArray)
```
cesarzc, line 101: Same here, should it be `sharedQubitStatusArray` instead?

src/Qir/Runtime/lib/QIR/QubitManager.cpp
kuzminrobin: Not quite, this is intentional.

The parameter `sharedQubitStatusArray` was shadowing the member variable with the same name in the outer class `CQubitManager`, thus creating a risk of confusing one with the other. And the compiler was warning about that. So I renamed.

src/Qir/Runtime/lib/QIR/QubitManager.cpp
kuzminrobin: Resolved in a different discussion.

kuzminrobin:
Please review.

swernli (approved):
A few minor questions, but overall looks good!

src/Qir/Runtime/lib/QIR/QubitManager.cpp
cesarzc: Ok, thanks for providing more context Robin! If that's the case then it might be worth it to change the name of this variable slightly more (e.g. `sharedQubitStatus`) because maybe one character could still be confusing to the reader. I don't feel strongly about it so I am ok with leaving it in its current state too.

cesarzc (approved):

src/Qir/Runtime/lib/QIR/strings.cpp
```diff
@@ -11,6 +11,10 @@
 #include "QirTypes.hpp"
 #include "QirRuntime.hpp"
 
+// Exposed to tests only:
+std::unordered_map<std::string, QirString*>& AllocatedStrings();
```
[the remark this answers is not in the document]
kuzminrobin: :-)
ONe of the compiler warning levels hints that a global variable and a global function 
should either be `static` and used in the current translation unit only, 
or should have a declaration (likely in a header) if it needs to be exposed outside of the translation unit and needs to be accessed by the other translation units.

In our case it is neither nor (or partially both). It is a global function that needs not be exposed, but still needs to be used by the tests (in the other translation units). Hence the extra declaration before the definition.

src/Qir/Runtime/lib/Simulators/FullstateSimulator.cpp
```diff
@@ -32,18 +38,18 @@ namespace
 {
 #ifdef _WIN32
```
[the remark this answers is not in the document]
kuzminrobin: Since the PR is ready to merge, I'll merge now and will take care of this single change in a subsequent PR.

src/Qir/Runtime/lib/QIR/QubitManager.cpp
kuzminrobin: I have just noticed this. I'll address in the subsequent PRs.

PR #761, opened 2021-06-30
Merge feature/azure-quantum-simulator to main

cesarzc:
This PR merges the changes made in the feature/azure-quantum-simulator branch to main.


.gitignore
```diff
@@ -336,10 +336,20 @@ ASALocalRun/
...
+
 /src/Simulation/Simulators.Tests/TestProjects/QSharpExe/built
+/src/Simulation/Simulators.Tests/TestProjects/QirExe/qir
 /src/Simulation/Simulators.Tests/TestProjects/TargetedExe/built
+
 dbw_test
 
+# Controller test artifacts
+/src/Qir/Controller/test-artifacts/
+/src/Qir/Controller/src
```
kuzminrobin, line 348: 1. `/src/Qir/Controller/src` is a file or directory? (why one line above ends with a slash `/`, but this line ends not with a slash?)
2. It seems rather suspicious that `git` is instructed to ignore a file/directory named **_`src`_**. If the file/directory named `src` is generated during the build, then may be it makes sense to rename `src` to something more evident as a temporary file/directory?

build/test.ps1
```diff
@@ -43,16 +43,22 @@ if ($Env:ENABLE_QIRRUNTIME -ne "false") {
...
+
     $qirSamples = (Join-Path $PSScriptRoot "../src/Qir/Samples")
     & "$qirSamples/test-qir-samples.ps1"
     if ($LastExitCode -ne 0) {
         $script:all_ok = $False
     }
+
+    if ($LastExitCode -ne 0) {
+        $script:all_ok = $False
+    }
```
kuzminrobin, line 61: FYI: I see a sequence of 2 identical `if`s:
```ps1
if ($LastExitCode -ne 0) { ..
if ($LastExitCode -ne 0) { ..
```

src/Qir/Runtime/lib/Simulators/FullstateSimulator.cpp
```diff
@@ -59,8 +58,12 @@ QUANTUM_SIMULATOR LoadQuantumSimulator()
...
-        throw std::runtime_error(
-            std::string("Failed to load ") + FULLSTATESIMULATORLIB + " (" + ::dlerror() + ")");
+        handle = ::dlopen(XPLATFULLSTATESIMULATORLIB, RTLD_LAZY);
+        if (handle == nullptr)
+        {
+            throw std::runtime_error(
+                std::string("Failed to load ") + XPLATFULLSTATESIMULATORLIB + " (" + ::dlerror() + ")");
+        }
     }
 #endif
```
kuzminrobin, line 68: (Just curiosity, feel free to ignore)
I have an impression that after this change the dynamic library `Microsoft.Quantum.Simulator.Runtime` gets the extension `.dll` _on all the platforms_.
But on non-Win platforms if the `.dll` library opening/loading failed, then 
Mac will try to load/open `.dylib`,
and the platforms other than Win and Mac will try to load/open `.so`.

If that's the case then why not leave it as it was
or why not make a full transition to `.dll` extension?

kuzminrobin (approved):
I have reviewed the .ps1, .cpp, .hpp files. In general look good to me (with minor comments).
Approving those.

src/Qir/Runtime/lib/Simulators/FullstateSimulator.cpp
cesarzc: This change was introduced by this [PR](https://github.com/microsoft/qsharp-runtime/pull/760).

The reason for this change was that in the package that contains the `Microsoft.Quantum.Simulator.Runtime` library, the library is always named as a DLL even if it is a Mac or Linux library, while local development and cross-platform pipelines need to handle the specific library name. For these two to co-exist happily, we first try load a .dll (which handles the case when a package reference to `Microsoft.Quantum.Simulators` is used), and if that fails on non-Windows platforms we try the native library name (which handles cases in local development and cross-platform pipelines).

.gitignore
cesarzc: 1. `/src/Qir/Controller/src` was a directory. It was an error that it didn't end with a slash `/`.
2. The reason why `git` was instructed to ignore a directory named `src` was because the controller (which is no longer hosted in this repo) produced some artifacts (namely the generated C++ entry-point driver file) as part of the process of building a QIR executable.

Since the controller is no longer hosted in this repo, I removed these lines.

Good catch! Thanks Robin.

build/test.ps1
cesarzc: I removed the second one. This had a similar cause that the one in your previous [comment](https://github.com/microsoft/qsharp-runtime/pull/761#discussion_r661687131). The second one was used to check for the result of controller tests.

src/Azure/Azure.Quantum.Client/SubmitterFactory.cs
```diff
@@ -0,0 +1,123 @@
...
+    /// Creates Azure Quantum submitters for specific targets.
+    /// </summary>
+    public static class SubmitterFactory
+    {
+        /// <summary>
+        /// Information about each supported QIR submitter.
+        /// </summary>
+        private static readonly ImmutableList<SubmitterInfo> QirSubmitters = ImmutableList.Create(
+            new SubmitterInfo(
+                new Regex(@"^microsoft\.simulator\.([\w]+\.)*[\w]+$"),
```
DmitryVasilevsky, line 26: Very minor. Consider using \A and \z (instead of ^ and $) as they always match the beginning and the end of the string.

src/Qir/Execution/Tests.Microsoft.Quantum.Qir.Tools/LoggerTests.cs
```diff
@@ -0,0 +1,93 @@
...
+                exception = thrownException;
+            }
+
+            var expectedLog = $"{time} [ERROR]: " +
+                "Exception encountered: System.InvalidOperationException: " +
+                exception.Message + Environment.NewLine + exception.StackTrace + Environment.NewLine;
+            Console.SetOut(consoleOutput);
+            logger.LogException(exception);
+            var actualLog = consoleOutput.ToString();
+            Assert.Equal(expectedLog, actualLog);
```
DmitryVasilevsky, line 90: All tests in this file seems rather strict. It's like writing logger twice. Do we have log format documented? If not, does it make sense to make tests less restrictive? For example, make sure that certain elements appear somewhere in the output.

src/Qir/Execution/Tests.Microsoft.Quantum.Qir.Tools/QirDriverGeneratorTests.cs
```diff
@@ -0,0 +1,184 @@
...
+                            new Parameter{ Name = "IntegerArrayArg", Position = 1, Type = DataType.ArrayType, ArrayType = DataType.IntegerType},
+                            new Parameter{ Name = "RangeArg", Position = 2, Type = DataType.RangeType},
+                            new Parameter{ Name = "StringArg", Position = 3, Type = DataType.StringType}
+                        }
+                    }
+                }
+            };
+
+        private static string RemoveLineEndings(string str) =>
+            str.Replace("\r\n", string.Empty).Replace("\n", string.Empty).Replace("\r", string.Empty);
```
DmitryVasilevsky, line 146: "str.Replace("\r\n", string.Empty)" is redundant. The other two will take care of this case.

src/Qir/Execution/Tests.Microsoft.Quantum.Qir.Tools/QirDriverGeneratorTests.cs
```diff
@@ -0,0 +1,184 @@
...
+        [InlineData("UseResultArrayArg")]
+        [InlineData("UseStringArg")]
+        [InlineData("UseMiscArgs")]
+        public void GenerateFullStateSimulatorDriver(string testCase)
+        {
+            var entryPointOperation = TestCases[testCase];
+            var driverGenerator = new QirFullStateDriverGenerator();
+            var driverFileName = $"{testCase}.cpp";
+            var verificationCppSourceCode = RemoveLineEndings(File.ReadAllText(Path.Combine(TestCasesDirectory, driverFileName)));
+            if (!Directory.Exists(TestArtifactsDirectory))
```
DmitryVasilevsky, line 170: CreateDirectory does nothing if directory already exists. This conditional is not needed.

src/Qir/Execution/Tests.Microsoft.Quantum.Qir.Tools/QirExecutableGeneratorTests.cs
```diff
@@ -0,0 +1,116 @@
...
+            streamWriter.Write(contents);
+            return fileInfo;
+        }
+
+        private static bool FilesWereCopied(FileInfo[] files, DirectoryInfo destinationDirectory)
+        {
+            var destinationDirectoryFiles = destinationDirectory.GetFiles();
+            foreach (var file in files)
+            {
+                var copiedFile = destinationDirectoryFiles.FirstOrDefault(fileInfo => fileInfo.Name == file.Name);
```
DmitryVasilevsky, line 97: This conceals quadratic algorithm. It's OK as number of iterations is still small and reading file content prevails. Consider making comment about it. ex:
// Runtime is (Number of files in destination directory)*(number of FileInfo passed in).
 Also mention that file content is also compared, not just file names.

src/Simulation/Core/Submitters/IQuantumMachine.cs
```diff
@@ -1,34 +1,24 @@
...
     /// </summary>
-    public interface IQuantumMachine
+    [Obsolete("Replaced by IQSharpSubmitter.")]
+    public interface IQuantumMachine : IAzureSubmitter
     {
-
         /// <summary>
         /// Function that configures a job object before submission.
         /// </summary>
         public delegate void ConfigureJob(object job);
```
bettinaheim, line 20: How come ConfigureJob is no longer needed? Or did it move somewhere and I missed it?

src/Qir/Execution/Tests.Microsoft.Quantum.Qir.Tools/QirExecutableTests.cs
```diff
@@ -0,0 +1,119 @@
...
+            {
+                using var streamWriter = new StreamWriter(stream);
+                streamWriter.Write(driverFileContents);
+            });
+
+            // Build the executable.
+            await qirExecutable.Object.BuildAsync(entryPoint, new[] { libraryDirectory }, new[] { includeDirectory });
+
+            // Verify that the "bytecode" file was created correctly.
+            var bytecodeFilePath = new FileInfo(Path.Combine(sourceDirectory.FullName, "qir.bc"));
```
DmitryVasilevsky, line 85: The following may be a bit faster for a typical case (when files are equal) as it doesn't need to invoke assert for every byte.
            byte[] bytecodeFileContent = File.ReadAllBytes(Path.Combine(sourceDirectory.FullName, "qir.bc"));
            Assert.True(qirBytecode.SequenceEqual(bytecodeFileContent));

src/Simulation/Core/ArgumentType.cs
```diff
@@ -0,0 +1,73 @@
...
+// Licensed under the MIT License.
+
+#nullable enable
+
+namespace Microsoft.Quantum.Runtime
+{
+    /// <summary>
+    /// The type of an argument to a QIR callable.
+    /// </summary>
+    public class ArgumentType
```
bettinaheim, line 11: What is the idea behind having this be part of the runtime package rather than the entry point driver (holds for Argument.cs, ArgumentType.cs and ArgumentValue.cs)? I assume they are public (only) so the entry point driver can access them? 
It's confusing to me that this is a data type in the Q# runtime called ArgumentType that does not contain an entry for all possible argument types in Q#.

src/Simulation/Core/ArgumentValue.cs
```diff
@@ -0,0 +1,185 @@
...
+using System.Linq;
+using Microsoft.Quantum.Simulation.Core;
+using Core = Microsoft.Quantum.Simulation.Core;
+
+namespace Microsoft.Quantum.Runtime
+{
+    /// <summary>
+    /// The value of an argument to a QIR callable is a discriminated union of the argument types.
+    /// </summary>
+    public abstract class ArgumentValue
```
bettinaheim, line 16: Are the types listed in this class (and in ArgumentType) related to/dictated by what the bond schema for the entry point arguments looks like? If so, then they will change as part of eliminating the dependency on the Q# compiler data structures and should hence not be public.

src/Qir/Execution/Tests.Microsoft.Quantum.Qir.Tools/TestCases/FullStateDriverGenerator/UseBoolArrayArg.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+// Auxiliary functions for interop with Q# Array type.
+struct InteropArray
+{
+    int64_t Size;
+    void* Data;
+
+    InteropArray(int64_t size, void* data) :
+        Size(size),
+        Data(data){}
```
DmitryVasilevsky, line 30: Not sure why do you need {}.

src/Qir/Execution/Tests.Microsoft.Quantum.Qir.Tools/TestCases/FullStateDriverGenerator/UseBoolArg.cpp
```diff
@@ -0,0 +1,87 @@
+//------------------------------------------------------------------------------
+// This code was generated by a tool.
```
DmitryVasilevsky, line 2: A note from future developers of this code: finding this comment in a file that you want to modify is very frustrating. If possible it would be nice to say which tool generates it and how.

src/Qir/Execution/Tests.Microsoft.Quantum.Qir.Tools/TestCases/FullStateDriverGenerator/UseBoolArrayArg.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+template<typename T>
+unique_ptr<InteropArray> CreateInteropArray(vector<T>& v)
+{
+    unique_ptr<InteropArray> array(new InteropArray(v.size(), v.data()));
+    return array;
+}
+
+template<typename S, typename D>
+void TranslateVector(vector<S>& sourceVector, vector<D>& destinationVector, function<D(S&)> translationFunction)
```
DmitryVasilevsky, line 41: Not used, but I assume it's part of the template that generates this code.

src/Qir/Execution/Tests.Microsoft.Quantum.Qir.Tools/Util.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.IO;
+
+namespace Tests.Microsoft.Quantum.Qir.Tools
+{
+    public static class Util
+    {
+        public static void DeleteDirectory(DirectoryInfo directory)
```
DmitryVasilevsky, line 10: If it's OK to remove directory, all files and subdirectories recursively, consider public static void Delete (string path, bool recursive). It will result in a different behavior than current code: if subdirectories are present, current code will fail, but proposed code will delete them. Not sure if this is a significant difference in our case.

src/Qir/Execution/Tools/Driver/IQirSimulatorInitializer.cs
```diff
@@ -0,0 +1,18 @@
...
+using System.IO;
+using System.Threading.Tasks;
+
+namespace Microsoft.Quantum.Qir.Tools.Driver
+{
+    public interface IQirSimulatorInitializer
+    {
+        public string Generate();
+
+        public IList<string> Headers { get; }
```
DmitryVasilevsky, line 14: If this is only used to iterate elements from beginning to end (i.e. in foreach loop) consider making this IEnumerable<sting>. This will give more flexibility to implementers.

src/Qir/Execution/Tools/Driver/IQirSimulatorInitializer.cs
```diff
@@ -0,0 +1,18 @@
...
+
+        public IList<string> LinkLibraries { get; }
```
DmitryVasilevsky, line 16: Same as Headers.

src/Qir/Execution/Tools/Driver/QirCppDriver.tt
```diff
@@ -0,0 +1,246 @@
+﻿<#@ template language="C#" linePragmas="false" #>
+<#@ import namespace="System.IO" #>
+<#@ import namespace="Microsoft.Quantum.QsCompiler.BondSchemas.Execution" #>
+//------------------------------------------------------------------------------
+// This code was generated by a tool.
```
DmitryVasilevsky, line 5: Hard to understand if this file was generated by a tool or if this the source templated used by a tool and generated files will have this comment. Think like a person who sees this codebase for the first time and this is the first file that person sees.

src/Qir/Execution/Tests.Microsoft.Quantum.Qir.Tools/TestCases/FullStateDriverGenerator/UseBoolArrayArg.cpp
cesarzc: Yes, this one is not used for some cases of array data types but if we handle all the specific data types individually to not include this, the template code becomes more difficult to read.

One of the future things to do is to extract these functions into a header-only library to make the generated file smaller.

src/Qir/Execution/Tools/Executable/QirExecutableGenerator.cs
```diff
@@ -0,0 +1,92 @@
...
+                    {
+                        logger?.LogWarning($"Could not find given directory: {dir.FullName}");
+                    }
+                    else
+                    {
+                        includeDirs.Add(dir.FullName);
+                    }
+                }
+
+                var inputFiles = sourceDirectory.GetFiles().Select(fileInfo => fileInfo.FullName).ToArray();
```
DmitryVasilevsky, line 68: Consider Directory.GetFiles(sourceDirectory.FullName). Shorter and without linq and lambdas.

src/Qir/Execution/Tools/Executable/QirExecutableGenerator.cs
```diff
@@ -0,0 +1,92 @@
...
+        {
+            var newPath = Path.Combine(destinationDirectory.FullName, fileToCopy.Name);
+            if (!File.Exists(newPath))
+            {
+                var newFile = fileToCopy.CopyTo(newPath);
+                logger?.LogInfo($"Copied file {fileToCopy.FullName} to {newFile.FullName}");
+            }
+        }
+
+        private void CopyDirectoryContents(DirectoryInfo directoryToCopy, DirectoryInfo destinationDirectory)
```
DmitryVasilevsky, line 83: This uses CopyFileIfNotExists so copies files only if destination file doesn't exist. Suggest reflecting this in the name. Just by the name I'd assume it will overwrite in this situation.

src/Qir/Execution/Tools/Executable/QuantumExecutableRunner.cs
```diff
@@ -0,0 +1,40 @@
...
+            {
+                FileName = executableFile.FullName,
+                Arguments = arguments,
+                RedirectStandardOutput = true,
+            };
+
+            process.EnableRaisingEvents = true;
+            process.Start();
+            var output = await process.StandardOutput.ReadToEndAsync();
+            process.WaitForExit();
```
DmitryVasilevsky, line 34: Theoretically there's Process.WaitForExitAsync, but the standard output is closed at this time so the process should be done.

src/Qir/Execution/Tools/QirTools.cs
```diff
@@ -0,0 +1,56 @@
...
+        {
+            using var qirContentStream = new MemoryStream();
+            if (!AssemblyLoader.LoadQirByteCode(qsharpDll, qirContentStream))
+            {
+                throw new ArgumentException("The given DLL does not contain QIR byte code.");
+            }
+            
+            if (!executablesDirectory.Exists)
+            {
+                executablesDirectory.Create();
```
DmitryVasilevsky, line 41: Conditional isn't needed, Create should be a no op if directory exists.

src/Qir/Execution/Tools/QirTools.cs
```diff
@@ -0,0 +1,56 @@
...
+            }
+            
+            var tasks = EntryPointOperationLoader.LoadEntryPointOperations(qsharpDll).Select(entryPointOp =>
```
DmitryVasilevsky: Not immediately clear if all the builds will run at the same time.

src/Qir/Execution/Tools/Microsoft.Quantum.Qir.Tools.csproj
```diff
@@ -0,0 +1,94 @@
...
+    <PlatformTarget>x64</PlatformTarget>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <Description>Tools for interacting with QIR.</Description>
+    <PackageTags>Quantum QIR</PackageTags>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Quantum.QirGeneration" Version="0.16.210526594-alpha" />
```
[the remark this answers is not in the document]
bettinaheim: Taken care of when I updated the package version to make the local repo build happy after addressing the compiler review comments.

src/Azure/Azure.Quantum.Client/SubmitterFactory.cs
cesarzc: Changed to \A and \z.

src/Qir/Execution/Tests.Microsoft.Quantum.Qir.Tools/LoggerTests.cs
cesarzc: Made the tests less strict by just making sure that the elements that should be in the log are in the log. No specific format is checked.

src/Qir/Execution/Tests.Microsoft.Quantum.Qir.Tools/QirDriverGeneratorTests.cs
cesarzc: Fixed the redundant part.

src/Qir/Execution/Tests.Microsoft.Quantum.Qir.Tools/QirDriverGeneratorTests.cs
cesarzc: Removed conditional.

src/Qir/Execution/Tests.Microsoft.Quantum.Qir.Tools/QirExecutableGeneratorTests.cs
cesarzc: Added comment and moved file content comparison to a method in Utils.cs.

src/Qir/Execution/Tests.Microsoft.Quantum.Qir.Tools/QirExecutableTests.cs
cesarzc: Created a comparison method in Utils.cs and changed this to `Assert.True(Util.CompareStreams(new MemoryStream(qirBytecode), bytecodeFileStream));`

src/Qir/Execution/Tests.Microsoft.Quantum.Qir.Tools/TestCases/FullStateDriverGenerator/UseBoolArg.cpp
cesarzc: I added information about the tool that generated the file and its purpose.

src/Qir/Execution/Tests.Microsoft.Quantum.Qir.Tools/Util.cs
cesarzc: Remove the `DeleteDirectory` function and replaced its usages with `directory.Delete(true)`.

src/Qir/Execution/Tools/Driver/IQirSimulatorInitializer.cs
cesarzc: Great recommendation! Changed it to `IEnumerable`.

src/Qir/Execution/Tools/Driver/IQirSimulatorInitializer.cs
cesarzc: Changed it to `IEnumerable`.

src/Qir/Execution/Tools/Driver/QirCppDriver.tt
cesarzc: Created a README.md file at the *src/Qir/Execution/Tools* level that explains the *\*.tt* files and the corresponding generated *\*.cs* files. It also describes a bit more in detail the contents of each subfolder.

src/Qir/Execution/Tools/Driver/IQirSimulatorInitializer.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System.Collections.Generic;
+using System.IO;
+using System.Threading.Tasks;
+
+namespace Microsoft.Quantum.Qir.Tools.Driver
+{
+    public interface IQirSimulatorInitializer
```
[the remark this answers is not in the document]
cesarzc: It's not necessarily specific to the simulator. I'll rename this to `IQirRuntimeInitiliazer`. This way it is open to runtimes that are not simulators.

src/Qir/Execution/Tools/Driver/QirCppDriverEx.cs
```diff
@@ -0,0 +1,20 @@
...
+
+using Microsoft.Quantum.Qir.Serialization;
+
+namespace Microsoft.Quantum.Qir.Tools.Driver
+{
+    public partial class QirCppDriver
+    {
+        public readonly EntryPointOperation EntryPoint;
+
+        public readonly IQirSimulatorInitializer SimulatorInitializer;
```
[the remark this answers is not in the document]
cesarzc: Following the same pattern than in this [comment](https://github.com/microsoft/qsharp-runtime/pull/761#discussion_r664704771), I'll rename this to `RuntimeInitializer`.

src/Qir/Execution/Tools/Utility/Clock.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+using System;
+
+namespace Microsoft.Quantum.Qir.Utility
+{
+    public class Clock : IClock
```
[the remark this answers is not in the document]
cesarzc: It is currently leveraged by the controller which gets access to this class by referencing the `Microsoft.Quantum.Qir.Tools` package. Since making this public is not ideal, I'll remove this class as well as the `IClock` interafce and use just `DateTimeOffset`.

src/Qir/Execution/Tools/Utility/IClock.cs
```diff
@@ -0,0 +1,15 @@
...
+// Licensed under the MIT License.
+
+using System;
+
+namespace Microsoft.Quantum.Qir.Utility
+{
+    /// <summary>
+    /// Mockable clock interface.
+    /// </summary>
+    public interface IClock
```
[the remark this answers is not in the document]
cesarzc: I'll remove this interface in favor of `DateTimeOffset`. Related to this other [comment](https://github.com/microsoft/qsharp-runtime/pull/761#discussion_r662699795).

src/Qir/Execution/Tools/Utility/ILogger.cs
```diff
@@ -0,0 +1,38 @@
...
+
+using System;
+
+namespace Microsoft.Quantum.Qir.Utility
+{
+    /// <summary>
+    /// Logger for internal traces and errors.
+    /// </summary>
+    /// <remarks>For now, this thinly wraps console logging. In the future, this might be extended to log to files or other systems.</remarks>
+    public interface ILogger
```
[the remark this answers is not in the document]
cesarzc: Yes, this mainly needed for error handling in the service. However, I believe this can be removed too, and for tests in this repo we can use the `FileLogger` class and the controller in the service can implement the custom logger.

src/Simulation/EntryPointDriver/Azure/AzureSettings.cs
```diff
@@ -0,0 +1,183 @@
...
+using System.Threading.Tasks;
+
+namespace Microsoft.Quantum.EntryPointDriver.Azure
+{
+    using Environment = System.Environment;
+
+    /// <summary>
+    /// Settings for a submission to Azure Quantum.
+    /// </summary>
+    public sealed class AzureSettings
```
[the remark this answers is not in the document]
cesarzc: Makes sense.

src/Qir/Execution/Tests.Microsoft.Quantum.Qir.Tools/Util.cs
```diff
@@ -0,0 +1,58 @@
...
+            using var fileStream = fileInfo.OpenWrite();
+            fileStream.Write(contents);
+            return fileInfo;
+        }
+
+        public static FileInfo CreateTextFile(DirectoryInfo directory, string fileName, string contents)
+        {
+            var filePath = Path.Combine(directory.FullName, fileName);
+            var fileInfo = new FileInfo(filePath);
+            using var fileStream = fileInfo.OpenWrite();
```
DmitryVasilevsky, line 52: Did you consider File.WriteAllText?

src/Qir/Execution/Tests.Microsoft.Quantum.Qir.Tools/Util.cs
```diff
@@ -0,0 +1,58 @@
...
+            }
+
+            return byteA == byteB;
+        }
+
+        public static FileInfo CreateBinaryFile(DirectoryInfo directory, string fileName, byte[] contents)
+        {
+            var filePath = Path.Combine(directory.FullName, fileName);
+            var fileInfo = new FileInfo(filePath);
+            using var fileStream = fileInfo.OpenWrite();
```
DmitryVasilevsky, line 43: Did you consider File.WriteAllBytes?

src/Qir/Execution/Tests.Microsoft.Quantum.Qir.Tools/Util.cs
```diff
@@ -0,0 +1,58 @@
...
+            if (streamA.Length != streamB.Length)
+            {
+                return false;
+            }
+
+            (streamA.Position, streamB.Position) = (0, 0);
+            (int byteA, int byteB) = (0, 0);
+            while ((byteA == byteB) && (byteA != -1))
+            {
+                (byteA, byteB) = (streamA.ReadByte(), streamB.ReadByte());
```
DmitryVasilevsky, line 33: Probably good enough for us, but see discussion here on performance. https://stackoverflow.com/questions/43289/comparing-two-byte-arrays-in-net

src/Simulation/Core/Submitters/IQuantumMachineExecutionContext.cs
```diff
@@ -1,11 +1,14 @@
...
 // Licensed under the MIT License.
 
+using System;
+
 namespace Microsoft.Quantum.Runtime
 {
     /// <summary>
     /// Interface to provide configuration details to manage execution.
     /// </summary>
+    [Obsolete("No longer used.")]
```
DmitryVasilevsky, line 11: I general it would be nice to say what replaces it. And by replacement I don't always mean direct replacement. If we switched to different way of doing things, it's enough to mention this new way of doing things.

src/Qir/Execution/Tools/Microsoft.Quantum.Qir.Tools.csproj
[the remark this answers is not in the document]
bettinaheim: Taken care of with the package updates I pushed.

src/Simulation/EntryPointDriver/Azure/Azure.cs
```diff
@@ -0,0 +1,308 @@
...
+        /// <returns>The exit code.</returns>
+        private static async Task<int> SubmitQir(
+            AzureSettings settings, IQirSubmitter submitter, QirSubmission submission)
+        {
+            Log(settings, "Submitting QIR entry point.");
+
+            if (settings.DryRun)
+            {
+                DisplayError("Dry run is not supported with QIR submission.", null);
+                return 1;
```
DmitryVasilevsky, line 141: Should we have constants for specific exit codes?

src/Simulation/EntryPointDriver/Azure/Azure.cs
```diff
@@ -0,0 +1,308 @@
...
+                Console.Error.WriteLine(Environment.NewLine + message);
+            }
+        }
+
+        /// <summary>
+        /// Logs a message if the verbose setting is enabled.
+        /// </summary>
+        /// <param name="settings">The Azure Quantum submission settings.</param>
+        /// <param name="message">The message.</param>
+        private static void Log(AzureSettings settings, string message)
```
DmitryVasilevsky, line 251: Suggest reflecting condition in the name, i.e. LogIfVerbose or similar.

src/Simulation/EntryPointDriver/Azure/AzureSettings.cs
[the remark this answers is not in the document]
DmitryVasilevsky: This is interesting. In general I also like to avoid redundancy and like concise naming. But always keep in mind that the most important tool to study codebase is grep (text search). If someone who's not familiar with codebase will search for 'settings', it would be a disaster as this is a very common word. AzureSettings is concise enough and rather unique at the same time so search will work well.

src/Simulation/EntryPointDriver/Azure/AzureSettings.cs
```diff
@@ -0,0 +1,183 @@
...
+            $"Verbose: {Verbose}");
+
+        internal static string ExtractLocation(Uri? baseUri)
+        {
+            if (baseUri is null || !baseUri.IsAbsoluteUri)
+            {
+                return "";
+            }
+
+            return baseUri.Host.Substring(0, baseUri.Host.IndexOf('.'));
```
DmitryVasilevsky, line 172: Does Host always contain dot?

src/Simulation/EntryPointDriver/Driver.cs
```diff
@@ -93,8 +94,8 @@ public sealed class Driver
         /// <summary>
         /// The job name option.
         /// </summary>
-        private static readonly OptionInfo<string?> JobNameOption = new OptionInfo<string?>(
-            ImmutableList.Create("--job-name"), default, "The name of the submitted job.");
+        private static readonly OptionInfo<string> JobNameOption = new OptionInfo<string>(
+            ImmutableList.Create("--job-name"), "", "The name of the submitted job.");
```
DmitryVasilevsky, line 98: Doesn't matter much, but String.Empty is considered cleaner.

src/Simulation/EntryPointDriver/Azure/AzureSettings.cs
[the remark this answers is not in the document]
bettinaheim: I am spoilt by often using FindAllReferences opposed to grep... Either one is fine by me - I'll leave it up to you.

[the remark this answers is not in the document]
cesarzc: Looking more closely into this, this was an already existing class. The only thing that happened is that its implementation was moved into its own file. When moving the implementation into its own file, the namespace changed from `Microsoft.Quantum.EntryPointDriver` to `Microsoft.Quantum.EntryPointDriver.Azure`. Given that these were existing public classes, I'll revert to the original namespace and keep the name as is.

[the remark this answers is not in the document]
cesarzc: Fixed the namespace backwards compatibility problem.

src/Qir/Execution/Tools/Driver/IQirSimulatorInitializer.cs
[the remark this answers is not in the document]
cesarzc: Renamed.

src/Qir/Execution/Tools/Driver/QirCppDriverEx.cs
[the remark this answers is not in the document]
cesarzc: Renamed.

src/Simulation/Core/ArgumentType.cs
cesarzc: The reason for having this class in the runtime package is that it can be used by the specific client that submits the program to Azure as well as the entry point driver.

Yes, the `ArgumentType` class supports the subset of Q# types currently supported as entry-point arguments to be submitted to Azure Quantum, so even though it is confusing it is "by design".

cesarzc: It would be a good idea to have a conversation about this before merging since in its current implementation this class needs to be `public`.

src/Simulation/Core/ArgumentValue.cs
cesarzc: They are related to the Bond schema but the schema is dictated by what type of arguments are supported as entry point.

Similarly to this [comment](https://github.com/microsoft/qsharp-runtime/pull/761#discussion_r665195518) we should have a conversation about this being public before merging.

src/Simulation/Core/Submitters/IQuantumMachine.cs
cesarzc: This was introduced originally to be able to set an arbitrary state to the job object before submission. This was mostly used for testing and development since it required knowledge of the internal implementation of the submission and the Job object.

Since this is a public interface, I think it makes sense not to keep providing this option.

src/Simulation/EntryPointDriver/Mock/NoOpQirSubmitter.cs
```diff
@@ -0,0 +1,30 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Threading.Tasks;
+
+namespace Microsoft.Quantum.EntryPointDriver.Mock
+{
+    /// <summary>
+    /// A QIR submitter that does nothing.
+    /// </summary>
+    public class NoOpQirSubmitter : IQirSubmitter
```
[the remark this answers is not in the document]
cesarzc: There was no reason for this to be public. Changed it to internal.

src/Qir/Execution/Tools/Utility/Clock.cs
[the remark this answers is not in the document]
cesarzc: Removed.

src/Qir/Execution/Tools/Utility/IClock.cs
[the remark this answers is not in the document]
cesarzc: Removed.

src/Qir/Execution/Tools/Executable/QirExecutableGenerator.cs
cesarzc: In this case, `Select` is used to transform the `FileInfo` array returned by `GetFiles()` into a `string` array. It doesn't really do anything that requires a search pattern.

src/Qir/Execution/Tools/Executable/QuantumExecutableRunner.cs
cesarzc: I believe in this case it is ok to use `WaitForExit` instead of `WaitForExitAsync` since the latter just provides the option to stop waiting using a cancellation token (documentation [here](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.waitforexitasync?view=net-5.0)). Given that we don't really want the option to stop waiting using a cancellation token, `WaitForExit` seems like the appropriate option here.

src/Qir/Execution/Tests.Microsoft.Quantum.Qir.Tools/Util.cs
cesarzc: Yes, I believe for our use it should be good enough.

src/Simulation/Core/ArgumentType.cs
bettinaheim: Discussed offline. While we know this is going to change in the near future, we don't really have an option but to have the public class matching the bond schema. I am fine with that.

src/Simulation/Core/ArgumentValue.cs
bettinaheim: Resolved per the other thread.

bettinaheim (approved):

src/Simulation/EntryPointDriver/Azure/Azure.cs
cesarzc: I believe it might be worth doing when we have a larger set of error conditions but for now I believe it is ok to return the same exit code regardless of the specific error.

src/Simulation/EntryPointDriver/Azure/AzureSettings.cs
cesarzc: Yes.

src/Qir/Execution/Tools/Executable/QirExecutableGenerator.cs
cesarzc: Changed the name to reflect that it only copies the contents if it does not already exist in the target directory,

src/Qir/Execution/Tools/QirTools.cs
cesarzc: Removed conditional.

src/Qir/Execution/Tools/QirTools.cs
cesarzc: Added a comment about it.

src/Qir/Execution/Tests.Microsoft.Quantum.Qir.Tools/Util.cs
cesarzc: Changed the implementation to use `File.WriteAllText`.

src/Qir/Execution/Tests.Microsoft.Quantum.Qir.Tools/Util.cs
cesarzc: Changed the implementation to use `File.WriteAllBytes`.

src/Simulation/Core/Submitters/IQuantumMachineExecutionContext.cs
cesarzc: Added more specific information about a possible replacement.

src/Simulation/EntryPointDriver/Azure/Azure.cs
cesarzc: Changed name to reflect the condition.

src/Simulation/EntryPointDriver/Driver.cs
cesarzc: Changed to `string.Empty`.

PR #763, opened 2021-07-01
QIR Validation Script - Runtime

ScottCarda-MS:
Changes the location the Simulator dll is copied to. This avoids an error on ADO builds that causes the Simulator dll that is appropriate for the given os platform from being overwritten by the dll that is for Windows.

See also: https://github.com/microsoft/Quantum/pull/519

### DIFF
--- a/src/Qir/Execution/Tools/Executable/QirFullStateExecutable.cs
+++ b/src/Qir/Execution/Tools/Executable/QirFullStateExecutable.cs
@@ -51,7 +51,6 @@ namespace Microsoft.Quantum.Qir.Tools.Executable
 
             LibraryDirectories.Add(new DirectoryInfo(Path.Combine(thisModulePath, "runtimes", osID, "native")));
             LibraryDirectories.Add(new DirectoryInfo(Path.Combine(thisModulePath, "Libraries", osID)));
-            LibraryDirectories.Add(new DirectoryInfo(Path.Combine(thisModulePath, "Libraries")));
         }
     }
 }
--- a/src/Qir/Execution/Tools/Microsoft.Quantum.Qir.Tools.csproj
+++ b/src/Qir/Execution/Tools/Microsoft.Quantum.Qir.Tools.csproj
@@ -68,7 +68,9 @@
 
   <ItemGroup>
     <None Include="$(QSimDll)">
-      <Link>Libraries\%(FileName)%(Extension)</Link>
+      <Link Condition="$([MSBuild]::IsOsPlatform('Windows'))">Libraries\win-x64\%(FileName)%(Extension)</Link>
+      <Link Condition="$([MSBuild]::IsOsPlatform('OSX'))">Libraries\osx-x64\%(FileName)%(Extension)</Link>
+      <Link Condition="$([MSBuild]::IsOsPlatform('Linux'))">Libraries\linux-x64\%(FileName)%(Extension)</Link>
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
       <Visible>false</Visible>
     </None>


src/Qir/Execution/Tools/Microsoft.Quantum.Qir.Tools.csproj
```diff
@@ -67,7 +67,7 @@
 
   <ItemGroup>
     <None Include="$(QSimDll)">
-      <Link>Libraries\%(FileName)%(Extension)</Link>
+      <Link>Libraries\win-x64\%(FileName)%(Extension)</Link>
```
bettinaheim, line 70: Why is this Windows specific? This looks wrong. What is testing this behavior?

src/Qir/Execution/Tools/Executable/QirFullStateExecutable.cs
```diff
@@ -49,7 +49,6 @@ public QirFullStateExecutable(FileInfo executableFile, byte[] qirBytecode, ILogg
 
             LibraryDirectories.Add(new DirectoryInfo(Path.Combine(thisModulePath, "runtimes", osID, "native")));
             LibraryDirectories.Add(new DirectoryInfo(Path.Combine(thisModulePath, "Libraries", osID)));
-            LibraryDirectories.Add(new DirectoryInfo(Path.Combine(thisModulePath, "Libraries")));
```
bettinaheim, line 52: What was the reason for removing this?

bettinaheim (changes requested):
No changes requested yet, but blocking the PR for now because we are in the process of pulling in the feature branch.

src/Qir/Execution/Tools/Microsoft.Quantum.Qir.Tools.csproj
cgranade: Simulator.Dev.Props has some good examples of how to handle simulator DLLs in a cross-platform way, may be interesting to look at as a reference.

src/Qir/Execution/Tools/Executable/QirFullStateExecutable.cs
ScottCarda-MS: The dll for the Simulator used to be located directly in the Libraries folder. It has been move in this PR for reasons explained in the other comment thread. Let's consolidate the conversation to be in that thread.

src/Qir/Execution/Tools/Microsoft.Quantum.Qir.Tools.csproj
ScottCarda-MS: Since the dll being move is Windows specific, it needs to go in the Windows specific folder. If the dll is also copied in from the Native folder as performed by the following ItemGroup, it will override the dll copied in this ItemGroup. That is only the case on the ADO builds. However, on local builds we need this ItemGroup's copy to make sure that we have the dll in a place the tool can find. Previously the dll was placed in directly in the Libraries folder, which was non-specific to a platform. This caused errors during Linux and Mac ADO builds because it was overriding the platform-appropriate dll with the one for Windows.

cgranade: I think that makes sense, but that then leaves the problem of getting the native shared libraries for the macOS and Linux runtimes into the tool package. You should be able to pull those out from the `xplat` artifact during e2e builds, with per-repo builds falling back to Windows-only.

ScottCarda-MS: I'm not sure I understand the problem you're suggesting. The natvie shared libraries for the macOS and Linux runtimes are in the tool package.

ScottCarda-MS: Resolved offline. The itemgroup now is specific to the OS being build on.

bettinaheim (approved):

bettinaheim:
@ScottCarda-MS If you merge this to the feature branch, please update the cesarzc/merge-feature-azure-simulator-to-main branch accordingly.

PR #764, opened 2021-07-01
Prevent circular references when serializing Qubit values.

cgranade:
This PR adds the `[JsonIgnore]` attribute to `Qubit.Value` in the C# runtime, preventing circular references when serializing values of type `Qubit`. This in turn allows for `DisplayableUnitaryOperator` from Microsoft.Quantum.Standard and Microsoft.Quantum.Standard.Visualization to be sent from the IQ# kernel through to Python hosts correctly, such that Python host programs can capture operations dumped from Q# programs.


bettinaheim:
@cgranade Does this need to go into main urgently? If not, could we please first pull in the simulator feature before having other merges to the runtime.

cgranade:
Not urgent, no. Just a small bugfix I found while prototyping some Python features.

cgranade:
image

Sigh, another CI failure due to the simulator crashing.

swernli:
Sigh indeed. We did have a potential fix for this, but it's tied up with #710. Perhaps it's worth pulling that part of the change out into its own PR?

cgranade:
@bettinaheim, @swernli: Now that `main` is open again, would it be possible to merge this fix in for the July release? Thank you!

swernli:
Closing as out of date, as there is now a workaround in IQ#.

PR #765, opened 2021-07-01
Added Undefined Behavior Sanitizer, fixed a few issues.

kuzminrobin:
Added the Undefined Behavior Sanitizer. Fixed some of the issues.
Updated the catch.hpp.
Tried to add Memory Sanitizer but failed. Requires instrumenting the libstdc++, which is impractical for CI builds.

### DIFF
--- a/src/Qir/Runtime/public/QirTypes.hpp
+++ b/src/Qir/Runtime/public/QirTypes.hpp
@@ -194,6 +194,5 @@ struct QIR_SHARED_API QirRange
     int64_t step    = 0;
     int64_t end     = 0;
 
-    QirRange() = default;
     QirRange(int64_t start, int64_t step, int64_t end);
 };
--- a/src/Qir/Runtime/public/QirTypes.hpp
+++ b/src/Qir/Runtime/public/QirTypes.hpp
@@ -70,7 +70,10 @@ struct QIR_SHARED_API QirString
     a header that contains the relevant data. The header immediately precedes the tuple's buffer in memory when the
     tuple is created.
 ======================================================================================================================*/
-using PTuple = char*;   // TODO: consider replacing `char*` with `void*` in order to block the accidental {dereferencing and pointer arithmtic}.
+// TODO: Move these types to inside of `QirTupleHeader`.
+using PTuplePointedType = uint8_t;
+using PTuple = PTuplePointedType*;  // TODO: consider replacing `uint8_t*` with `void*` in order to block the accidental {dereferencing and pointer arithmtic}.
+                                    //       Much pointer arithmetic in tests. GetHeader() uses the pointer arithmetic.
 struct QIR_SHARED_API QirTupleHeader
 {
     using TBufSize = size_t;  // Type of the buffer size.
@@ -80,7 +83,7 @@ struct QIR_SHARED_API QirTupleHeader
     TBufSize tupleSize = 0; // when creating the tuple, must be set to the size of the tuple's data buffer (in bytes)
 
     // flexible array member, must be last in the struct
-    char data[];
+    PTuplePointedType data[];
 
     PTuple AsTuple()
     {
--- a/src/Qir/Runtime/public/QirTypes.hpp
+++ b/src/Qir/Runtime/public/QirTypes.hpp
@@ -190,10 +190,10 @@ struct QIR_SHARED_API QirCallable
 
 struct QIR_SHARED_API QirRange
 {
-    int64_t start;
-    int64_t step;
-    int64_t end;
+    int64_t start   = 0;
+    int64_t step    = 0;
+    int64_t end     = 0;
 
-    QirRange();
+    QirRange() = default;
     QirRange(int64_t start, int64_t step, int64_t end);
 };
--- a/src/Qir/qir-utils.ps1
+++ b/src/Qir/qir-utils.ps1
@@ -99,28 +99,51 @@ function Build-CMakeProject {
 
     # Sanitizers (https://clang.llvm.org/docs/UsersManual.html#controlling-code-generation):
 
-    $env:CFLAGS   += " "
-    $env:CXXFLAGS += " "
+    $sanitizeFlags = "" 
+    if (-not ($IsWindows))
+    {
+        # Undefined Behavior Sanitizer (https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html)
+        # WSL:
+        #   unsigned-integer-overflow   Causes error reports in "qir-static-tests" in random number generator standard library (the test still succeeds):
+        #       /usr/bin/../lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/bits/random.h:552:58: runtime error: unsigned integer overflow: 0 - 1 cannot be represented in type 'unsigned long'
+        #   Likely -fsanitize=implicit-integer-sign-change causes error reports in "qir-static-tests" in random number generator standard library (the test still succeeds):
+        #       /usr/bin/../lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/bits/uniform_int_dist.h:284:17: runtime error: implicit conversion from type \
+        #           'std::uniform_int_distribution<long>::result_type' (aka 'long') of value -9223372036854775808 (64-bit, signed) to type 'unsigned long' changed the value \
+        #           to 9223372036854775808 (64-bit, unsigned)
+        # Win:
+        #   [19/35] cmd.exe /C "cd . && C:\PROGRA~1\LLVM12\bin\CLANG_~1.EXE -fuse-ld=lld-link -nostartfiles -nostdlib \
+        #       -Werror -Weverything -Wno-c++98-compat-pedantic -Wno-old-style-cast -Wno-covered-switch-default -Wno-c99-extensions \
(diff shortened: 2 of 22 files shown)


bettinaheim (changes requested):
No changes requested, but blocking the PR for now because we are in the process of pulling in the feature branch.

kuzminrobin:
> No changes requested, but blocking the PR for now because we are in the process of pulling in the feature branch.

FYI: This is not actual PR. It is just to test the CI build.

kuzminrobin:
Please review.

src/Qir/qir-utils.ps1
```diff
@@ -93,10 +93,60 @@ function Build-CMakeProject {
...
 
-
     $env:CFLAGS   += $warningFlags
     $env:CXXFLAGS += $warningFlags
 
+
+    # Sanitizers (https://clang.llvm.org/docs/UsersManual.html#controlling-code-generation):
+
+    $sanitizeFlags = "" 
+    if (-not ($IsWindows))
```
cgranade, line 103: Is this script intended for use with Windows PowerShell (5.1 and earlier, and as opposed to PowerShell for Windows, 6.0 and later)? If so, then the `$IsWindows` automatic variable was added in 6.0 such that a Windows user running this on 5.1 will fall through to this branch (`$null` coerces to `$false` under the `-not` operator). I'd suggest making sure that the `$IsWindows` variable exists unconditionally:

```powershell
if ($null -eq $IsWindows) {
    # PSEdition is "Desktop" for Windows PowerShell (5.1 and earlier),
    # and is "Core" for PowerShell Core (6.0) and for PowerShell
    # (7.0 and later).
    $IsWindows = $PSVersionTable.PSEdition -eq "Desktop";
}
```

src/Qir/qir-utils.ps1
```diff
@@ -93,10 +93,60 @@ function Build-CMakeProject {
...
+        ##      Suppressing of the errors in the updated `catch2/catch.hpp` and standard library headers eventually bumps into errors reported in `memcmp`,
+        ##      suppressing of which does not work (https://github.com/google/sanitizers/issues/1429#issuecomment-876799463).
+        ##      Looks like MSan will not work until the libstdc++ is recompiled to be instrumented (https://clang.llvm.org/docs/MemorySanitizer.html#handling-external-code).
+        ##      Instrumenting libstdc++ during CI builds seems impractical (https://stackoverflow.com/a/22301584/6362941).
+        #$sanitizeFlags += " -fsanitize=memory -fsanitize-memory-track-origins=2"
+
+        # Common for all sanitizers:
+        $sanitizeFlags += " -fsanitize-blacklist="      # https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html#suppressing-errors-in-recompiled-code-ignorelist
+                                                        # https://releases.llvm.org/11.0.1/tools/clang/docs/SanitizerSpecialCaseList.html
+        $sanitizeFlags += (Join-Path $Path .. UBSan.ignore)
```
cgranade, line 140: `Join-Path` with three or more inputs is only supported on 6.0 or later. I'd suggest a `#Requires` statement at the top of this script to make the dependency on 6.0 clear and actionable to users.

src/Qir/qir-utils.ps1
kuzminrobin: Our "[src/Qir/Runtime/README.md](https://github.com/microsoft/qsharp-runtime/blob/main/src/Qir/Runtime/README.md#prerequisites)" specifies as a prerequisite the "PowerShell Core or PowerShell 7+". That is why the `$IsWindows` automatic variable is always expected to be set (to `True` on Windows and to `False` otherwise).

The code inside of `if (-not ($IsWindows))` block is expected to be executed on non-Windows platforms (Linux and Mac).

I assume your question was based on assumption that my line `if (-not ($IsWindows))` was trying to work around different PowerShell versions, _all on Windows_. Thank you for the question. It reminded me how little I know about the PowerShell and encouraged me to read through the article [about_PowerShell_Editions](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_powershell_editions?view=powershell-7.1) and play a bit with the automatic variables on Win and WSL.

cgranade: If PowerShell 6.0 or later are already required, then please disregard my comment. We may separately want to make sure that there's a `#Requires` statement to communicate that to the user in an actionable way, but there's no issue with the code here as written if 6.0 or later are already required.

src/Qir/qir-utils.ps1
kuzminrobin: Adding `#Requires -PSEdition Core` in the beginning of the script.

kuzminrobin:
Please review.

src/Qir/Runtime/public/QirTypes.hpp
```diff
@@ -190,10 +190,10 @@ struct QIR_SHARED_API QirCallable
...
 {
-    int64_t start;
-    int64_t step;
-    int64_t end;
+    int64_t start   = 0;
+    int64_t step    = 0;
+    int64_t end     = 0;
 
-    QirRange();
+    QirRange() = default;
```
swernli, line 197: Does this allow for creation of a range with step zero, bypassing the assert in the non-default constructor?

src/Qir/qir-utils.ps1
```diff
@@ -93,10 +95,60 @@ function Build-CMakeProject {
...
+
+        $sanitizeFlags += " -fno-omit-frame-pointer"            # https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html
```
swernli, line 144: Shouldn't we only enable these flags for debug or for specific pipelines? If sanitizer is enabled during compilation always on these platforms, those hooks get added to the resulting binary, which means we'd ship a binary with sanitizer enabled, making it much slower. We only want sanitizer during specific testing. We might need to make the decision on whether or not to run sanitizer based on which pipeline this is running on.

swernli (changes requested):
One minor question, but one big one about how/when we want sanitizer to be enabled. We should not be shipping a sanitizer enabled binary.

src/Qir/Runtime/public/QirTypes.hpp
kuzminrobin: Looks like this default constructor is never used. Removing.

src/Qir/qir-utils.ps1
kuzminrobin: Good point. What do you think about enabling sanitizers in Debug version only?

The Sanitizers documentation announces the slow-down of a factor of 2 - 3 per major sanitizer. If multiple major sanitizers are on (e.g. [`-fsanitize=undefined`](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html), [`-fsanitize=address`](https://clang.llvm.org/docs/AddressSanitizer.html), ...), then I expect the total slow-down of a factor of 8 - 10. I'm not sure how critical that is at the current stage of QIR RT development.

After I ship the sanitizers, all our tests and samples will be running clean and the sanitizers will only prevent us from the changes that break the clean tests/samples run.
But the users (like PNNL) may be running our QIR RT in the ways we never tested. Their runs may trigger new sanitizers errors/warnings, thus helping us disclose new problem fragments. E.g. recently there was an issue of growing memory use in one of the solvers/algorithms. If the sanitizers were enabled, then in case of a memory leak the [leak sanitizer](https://clang.llvm.org/docs/LeakSanitizer.html) could report something actionalble.

To summarize, I fell that we are interested in getting our users run the _sanitized_ QIR RT. But on the other hand we don't want them to run frustratingly slow. If we enable the sanitizers in the debug version only then they will be able to choose.

swernli: I think these two sanitizers enabled by default on debug builds makes sense for now. As has come up in discussion, we'll need to figure out what it looks like to run sanitizers on the CI pipeline, as right now CI always builds only the release packages that eventually get published to nuget and runs the tests on those. Furthermore, the Q# packages only build on Windows, such that any tests that are part of that build (such as all of the ones under src/Qir/Tests) are not run on Linux and won't include sanitizers. Taking this together with the fact that not all sanitizers can be enabled together, and I think it makes sense for sanitizer validation to become a specific e2e build validation step (akin to the "Valide end-to-end Linux" step) that tries running tests with different sanitizers enabled in parallel.

As far as customer scenarios go, my two cents is that we should never ship/publish a sanitizer enabled runtime. It's not common nor easy to publish two versions of the same package for the same platform (we'd have to do so under a different name), and even then only whatever sanitizer we pick as default would be enabled. I think it makes sense to compile a debug version with the relevant sanitizer enabled as needed for investigating specific issues rather than try to have a setup that catches everything all the time. There, we may need to compile the release version with sanitizers enabled specifically because debug builds can often change timing or relationship between code due to differing optimizations. Overall, my view is that sanitizers is something we can and should have a dedicated step for in the CI for validation, but otherwise are best used as an on-demand tool rather than an on-by-default tool.

kuzminrobin: Two more issues have been added internally to track the items:
* Ensure that we ship a runtime version built in release without sanitizers enabled
* Ensure all sanitizers and additional validation runs during CI builds

kuzminrobin:
Please review.

swernli (approved):

PR #767, opened 2021-07-05
Sparse simulator

sam-jaques:
This PR adds a new simulator that uses memory-efficient representations of sparse quantum states.


src/Simulation/Simulators/SparseSimulator/README.md
```diff
@@ -0,0 +1,155 @@
+# Introduction
+This is a an alternative quantum simulator, compatible with Q\# and Microsoft's quantum development kit. It is efficient for algorithms with a smaller number of states in superposition, regardless of the number of qubits.
+
+# Repository Structure
+
+/SparseQuantumSimulator/
+|--/Native/: C++ code for the simulator
+|--/SparseSimulatorCS/: C# library for the simulator 
+|--/SparseSimQSharpTests/: Several Q# tests of the simulator
+|--/SparseSimulatorTests/: C++ tests of the simulator
```
kuzminrobin, line 10: You may want to add double space `  ` in the end of every line. Otherwise currently in the rendered version all the lines are merged into a single line. 

You can see the rendered version if you press "Display the reach diff" link in the top-right corner of this file's diff (third from the right, after `...` and "Viewed").

src/Simulation/Simulators/SparseSimulator/README.md
```diff
@@ -0,0 +1,155 @@
...
+To add a gate to the simulator, you will need to add it to:
+- SparseSimulatorCS/SparseSimulator.cs
+- Native/capi.cpp
+- Native/quantum_state.hpp
+- Native/basic_quantum_state.hpp
+- Native/SparseSimulator.h
+
+Optionally, you may need to add a Q\# and C\# file to the SparseSimulatorCS folder, to create a Q\# operation to call the gate if it does not already exist.
+
+## Quantum State (quantum_state.hpp and basec_quantum_state.hpp)
```
kuzminrobin, line 78: Did you mean 
bas**_i_**c_quantum_state.hpp

instead of 
bas**_e_**c_quantum_state.hpp
?

src/Simulation/Simulators/SparseSimulator/README.md
```diff
@@ -0,0 +1,155 @@
...
+        private SparseSimulator sim = null;
+        public Native(IOperationFactory m) : base(m)
+        {
+            sim = m as SparseSimulator;
+        }
+        public override Func<\#input types from Q\#\#, \#output types to Q\#\#> __Body__ => sim == null ? base.__Body__ : (args) => {
+            return sim.Your_new_gate(args.Item1, args.Item2,...);
+        };
+    }
+}`
```
kuzminrobin, line 120: (Use triple tick to start and end the code block, and nothing else in the triple-tick lines.
On the right of the opening triple tick you can specify the language name to use the syntax highlighting.
See details in the [Fenced Code Blocks](https://www.markdownguide.org/extended-syntax/#fenced-code-blocks) and [Syntax Highlighting](https://www.markdownguide.org/extended-syntax/#syntax-highlighting))
```suggestion
```c#
public partial class YourNewGate
{
    public class Native : YourNewGate
    {
        private SparseSimulator sim = null;
        public Native(IOperationFactory m) : base(m)
        {
            sim = m as SparseSimulator;
        }
        public override Func<\#input types from Q\#\#, \#output types to Q\#\#> __Body__ => sim == null ? base.__Body__ : (args) => {
            return sim.Your_new_gate(args.Item1, args.Item2,...);
        };
    }
}
\```
```

src/Simulation/Simulators/SparseSimulator/Native/SparseSimulator.h
```diff
@@ -0,0 +1,846 @@
...
+    	// These indicate whether there are any H, Rx, or Ry gates
+    	// that have yet to be applied to the wavefunction. 
+    	// Since HH=I and Rx(theta_1)Rx(theta_2) = Rx(theta_1+theta_2)
+    	// it only needs a boolean to track them.
+    	std::vector<bool> _H_queue;
+    	std::vector<bool> _Rx_queue;
+    	std::vector<bool> _Ry_queue;
+
+    	std::vector<double> _Rx_angles;
+    	std::vector<double> _Ry_angles;
```
kuzminrobin, line 682: FYI: The identifiers whose names start with underscore `_` followed by a capital letter (e.g. the identifier starting with `_H`) are reserved by the C++ standard. You might want ot rename such identifiers.

src/Simulation/Simulators/SparseSimulator/Native/SparseSimulator.h
```diff
@@ -0,0 +1,846 @@
...
+			}
+			if (isEmpty) {
+				return;
+			}
+			// Z assertions are like phase gates
+			// If it's in release mode, it will queue them
+			// as a phase/permutation gate
+			// This means if an assert fails, it will fail
+			// at some future point, not at the point of failure
+			#if NDEBUG
```
kuzminrobin, line 591: FYI: I assume you meant to write `#if defined(NDEBUG)` or `#ifdef NDEBUG`.

src/Simulation/Simulators/SparseSimulator/Native/gates.h
```diff
@@ -0,0 +1,284 @@
...
+                MCProj,
+                Allocate,
+                Release,
+                NUM_OPS // counts gate types; do not add gates after this!
+            };
+
+            // Strings for the names of the various gate types
+            const static std::string op_name(OP gate_type) {
+                switch (gate_type) {
+                    case OP::MCPhase:
```
kuzminrobin, line 54: FYI: May be I don't know something, or maybe some particular compilers allow that, but I don't expect the fragment `OP::MCPhase` to compile successfully (as opposed to fragment `MCPhase`).
If `OP` has been declared `enum class OP` (as opposed to `enum OP`) then I would expect the `OP::MCPhase` to compile successfully.
See details in [Item 10: Prefer scoped `enum`s to unscoped `enum`s](https://moodle.ufsc.br/pluginfile.php/2377667/mod_resource/content/0/Effective_Modern_C__.pdf).

src/Simulation/Simulators/SparseSimulator/Native/gates.h
```diff
@@ -0,0 +1,284 @@
...
+                std::vector<logical_qubit_id> controls;
+                operation(OP gate_type_arg, 
+                    logical_qubit_id target_arg, 
+                    std::vector<logical_qubit_id> controls_arg
+                    ) : gate_type(gate_type_arg),
+                        target(target_arg),
+                        controls(controls_arg){}
+
+                logical_qubit_id shift;
+                logical_qubit_id target_2;
```
kuzminrobin, line 150: FYI: A number of data members stay **un**initialized in a number of constructors.
You may find useful the [Non-static data member initializers](https://github.com/AnthonyCalandra/modern-cpp-features#non-static-data-member-initializers).

src/Simulation/Simulators/SparseSimulator/Native/SparseSimulator.h
```diff
@@ -0,0 +1,846 @@
+#pragma once
```
kuzminrobin, line 1: (Up to you)
This is a C++ file. 

You may want to change the extension from .h to .hpp.
You may want to break up the file into .hpp and .cpp.

src/Simulation/Simulators/SparseSimulator/Native/SparseSimulator.h
```diff
@@ -0,0 +1,846 @@
...
+#include <algorithm>
+#include <iomanip>
+#include <iostream>
+#include <list>
+
+#include <set>
+
+
+#define MAX_QUBITS 1024
+#define MIN_QUBITS 64
```
kuzminrobin, line 19: FYI: [Item 2: Prefer `const`s, `enum`s, and `inline`s to `#defines`](http://wavelino.coffeecup.com/pdf/Effective%20Cshapshap.pdf)

src/Simulation/Simulators/SparseSimulator/Native/SparseSimulator.h
```diff
@@ -0,0 +1,846 @@
...
+
+using namespace std::literals::complex_literals;
+
+namespace Microsoft
+{
+namespace Quantum
+{
+namespace SPARSESIMULATOR
+{
```
cgranade, line 28: Since C++17 is enabled, you should be able to collapse these definitions.
```suggestion
namespace Microsoft::Quantum::SPARSESIMULATOR
{
```

cgranade (commented):
Thanks for opening this and for your contribution, @sam-jaques! I took a quick look through and left some comments, but happy to expand on anything that would be helpful. I think my overall feedback, though, is that it would be good to integrate this into the build pipelines and to make sure that your new simulator could be used from Python.

src/Simulation/Simulators/SparseSimulator/Native/factory.cpp
```diff
@@ -0,0 +1,68 @@
...
+        }
+    }
+    if (emptySlot == -1)
+    {
+        _simulators.push_back(std::shared_ptr<SparseSimulator>(new SparseSimulator(num_qubits)));
+        emptySlot = _simulators.size() - 1;
+    }
+    else
+    {
+        _simulators[emptySlot] = std::shared_ptr<SparseSimulator>(new SparseSimulator(num_qubits));
```
kuzminrobin, line 44: FYI: [Item 21: Prefer `std::make_unique` and `std::make_shared` to direct use of `new`](https://moodle.ufsc.br/pluginfile.php/2377667/mod_resource/content/0/Effective_Modern_C__.pdf).

src/Simulation/Simulators/SparseSimulator/Native/factory.cpp
```diff
@@ -0,0 +1,68 @@
...
+    _simulators[id].reset();
+}
+
+// Returns a simulator at some id (used for the C++/C# API)
+std::shared_ptr<SparseSimulator>& getSimulator(unsigned id)
+{
+    std::shared_lock<std::shared_mutex> shared_lock(_mutex);
+
+    return _simulators[id];
+}
```
kuzminrobin, line 64: FYI: Returns a _non-const reference_ to the element of a container, thus enabling the update of the element in a non-thread-safe manner. E.g. someone can call `getSimulator(..).reset()` (the `reset()` call is not protected with `_mutex`) while some other thread can be accessing that same element (access protected with `_mutex`).

Consider returning _by value_:
`std::shared_ptr<SparseSimulator> getSimulator(unsigned id)` (no ampersand `&`).

src/Simulation/Simulators/SparseSimulator/Native/capi.cpp
```diff
@@ -0,0 +1,303 @@
...
+#include <iostream>
+#include <string>
+
+using namespace Microsoft::Quantum::SPARSESIMULATOR;
+
+
+#include<chrono>
+#include<thread>
+
+std::string sample_string;
```
kuzminrobin, line 21: Consider making `sample_string` a local variable (and a const reference to `std::string`). It is accessed in one function only.

bettinaheim:
@sam-jaques Thanks for putting this together!
I'll create a feature branch shortly for this to be merged into, and @DmitryVasilevsky will coordinate and work with you on updating this branch with the latest and processing the feedback.

kuzminrobin:
I made a quick run through the C/C++ code (but not C# code). I'm done with the first review round.

src/Simulation/Simulators/SparseSimulator/Native/flat_hash_map.hpp
```diff
@@ -0,0 +1,1502 @@
+//          Copyright Malte Skarupke 2017.
```
DmitryVasilevsky, line 1: Is there an STL version of this? Can it be used?

src/Simulation/Simulators/SparseSimulator/SparseSimQSharpTests/Program.qs
```diff
@@ -0,0 +1,643 @@
+﻿namespace Microsoft.Quantum.SparseSimulatorTests {
```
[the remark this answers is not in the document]
DmitryVasilevsky: Addressing it here: https://github.com/microsoft/qsharp-runtime/pull/769

src/Simulation/Simulators/SparseSimulator/SparseSimulatorCS/Probes.qs
```diff
@@ -0,0 +1,33 @@
+﻿namespace Microsoft.Quantum.SparseSimulation {
```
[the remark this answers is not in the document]
DmitryVasilevsky: Addressing copyright here: https://github.com/microsoft/qsharp-runtime/pull/769

src/Simulation/Simulators/SparseSimulator/SparseSimulatorCS/SparseSimulator.cs
```diff
@@ -0,0 +1,385 @@
+﻿using System;
```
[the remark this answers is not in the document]
DmitryVasilevsky: Addressing it here: https://github.com/microsoft/qsharp-runtime/pull/769

src/Simulation/Simulators/SparseSimulator/SparseSimulatorTests/TestHelpers.cpp
```diff
@@ -0,0 +1,159 @@
+#pragma once
```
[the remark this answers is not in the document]
DmitryVasilevsky: Addressing it here: https://github.com/microsoft/qsharp-runtime/pull/769

src/Simulation/Simulators/SparseSimulator/SparseSimulatorCS/Probes.cs
```diff
@@ -0,0 +1,357 @@
+﻿// Functions that access the internal wavefunction for diagnostic purposes
```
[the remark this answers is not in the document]
DmitryVasilevsky: Addressing it here: https://github.com/microsoft/qsharp-runtime/pull/769

src/Simulation/Simulators/SparseSimulator/Native/SparseSimulator.h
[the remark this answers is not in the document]
DmitryVasilevsky: Addressing it here: https://github.com/microsoft/qsharp-runtime/pull/769

src/Simulation/Simulators/SparseSimulator/Native/basic_quantum_state.hpp
```diff
@@ -0,0 +1,82 @@
+#pragma once
```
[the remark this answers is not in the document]
DmitryVasilevsky: Addressing it here: https://github.com/microsoft/qsharp-runtime/pull/769

src/Simulation/Simulators/SparseSimulator/Native/capi.cpp
```diff
@@ -0,0 +1,303 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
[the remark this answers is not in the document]
DmitryVasilevsky: Addressing it here: https://github.com/microsoft/qsharp-runtime/pull/769

src/Simulation/Simulators/SparseSimulator/Native/capi.hpp
```diff
@@ -0,0 +1,146 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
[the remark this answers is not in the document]
DmitryVasilevsky: Addressing it here: https://github.com/microsoft/qsharp-runtime/pull/769

src/Simulation/Simulators/SparseSimulator/Native/factory.cpp
```diff
@@ -0,0 +1,68 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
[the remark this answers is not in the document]
DmitryVasilevsky: Addressing it here: https://github.com/microsoft/qsharp-runtime/pull/769

src/Simulation/Simulators/SparseSimulator/Native/factory.hpp
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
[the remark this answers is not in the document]
DmitryVasilevsky: Addressing it here: https://github.com/microsoft/qsharp-runtime/pull/769

src/Simulation/Simulators/SparseSimulator/Native/types.h
```diff
@@ -0,0 +1,51 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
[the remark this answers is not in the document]
DmitryVasilevsky: Addressing it here: https://github.com/microsoft/qsharp-runtime/pull/769

src/Simulation/Simulators/SparseSimulator/README.md
DmitryVasilevsky: Addressed it here: https://github.com/microsoft/qsharp-runtime/pull/769

src/Simulation/Simulators/SparseSimulator/README.md
DmitryVasilevsky: Yes. Addressed it in https://github.com/microsoft/qsharp-runtime/pull/769

src/Simulation/Simulators/SparseSimulator/README.md
DmitryVasilevsky: Addressed it in https://github.com/microsoft/qsharp-runtime/pull/769

src/Simulation/Simulators/SparseSimulator/Native/gates.h
DmitryVasilevsky: This does compile with every C++ compiler on godbolt.org. Making it scoped is a good idea however. 'enum class' will also require explicit conversion to int if it is ever used.

src/Simulation/Simulators/SparseSimulator/README.md
DmitryVasilevsky: Addressed in https://github.com/sam-jaques/qsharp-runtime/pull/1

src/Simulation/Simulators/SparseSimulator/SparseSimulatorTests/TestHelpers.cpp
DmitryVasilevsky, line 1: #pragma once is likely not needed in .cpp files. (Unless they are textually included.)

DmitryVasilevsky: Addressed here: https://github.com/sam-jaques/qsharp-runtime/pull/1

src/Simulation/Simulators/SparseSimulator/Native/SparseSimulator.h
DmitryVasilevsky: Addressed here: https://github.com/sam-jaques/qsharp-runtime/pull/1

src/Simulation/Simulators/SparseSimulator/Native/gates.h
DmitryVasilevsky: Addressed here: https://github.com/sam-jaques/qsharp-runtime/pull/1

src/Simulation/Simulators/SparseSimulator/Native/SparseSimulator.h
DmitryVasilevsky: Yes, defines should be avoided. Replaced with constexpr here: https://github.com/sam-jaques/qsharp-runtime/pull/1

src/Simulation/Simulators/SparseSimulator/Native/SparseSimulator.h
```diff
@@ -0,0 +1,846 @@
...
+    	// Since HH=I and Rx(theta_1)Rx(theta_2) = Rx(theta_1+theta_2)
+    	// it only needs a boolean to track them.
+    	std::vector<bool> _H_queue;
+    	std::vector<bool> _Rx_queue;
+    	std::vector<bool> _Ry_queue;
+
+    	std::vector<double> _Rx_angles;
+    	std::vector<double> _Ry_angles;
+
+		// Store which qubits are non-zero as a bitstring
```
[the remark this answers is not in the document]
DmitryVasilevsky: Yes. Replaced tabs with spaces here: https://github.com/sam-jaques/qsharp-runtime/pull/1

src/Simulation/Simulators/SparseSimulator/Native/factory.cpp
```diff
@@ -0,0 +1,68 @@
...
+namespace SPARSESIMULATOR
+{
+// Ensures exclusive access to _simulators, the vector of simulators
+std::shared_mutex _mutex;
+std::vector<std::shared_ptr<SparseSimulator>> _simulators;
+
+unsigned createSimulator(logical_qubit_id num_qubits)
+{
+    if (num_qubits > MAX_QUBITS)
+        throw std::runtime_error("Max number of qubits is 1024!");
```
DmitryVasilevsky, line 26: Probably not a good idea to include 1024 in the message when MAX_QUBITS is defined elsewhere. Whoever would change MAX_QUBITS will likely forget to update the message.

DmitryVasilevsky: Removed 1024 from message here: https://github.com/sam-jaques/qsharp-runtime/pull/1

src/Simulation/Simulators/SparseSimulator/Native/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
+cmake_minimum_required(VERSION 3.10)
+project(SparseQuantumSimulator)
+
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_CXX_EXTENSIONS OFF)
+
+# Set OpenMp if it is available
```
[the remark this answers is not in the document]
DmitryVasilevsky: Fixed here: https://github.com/sam-jaques/qsharp-runtime/pull/1

src/Simulation/Simulators/SparseSimulator/Native/SparseSimulator.h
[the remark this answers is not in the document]
DmitryVasilevsky: Yes, defines should be avoided. Replace with constexpr here: https://github.com/sam-jaques/qsharp-runtime/pull/1

src/Simulation/Simulators/SparseSimulator/SparseSimulatorTests/CSharpIntegrationTests.cpp
```diff
@@ -0,0 +1,306 @@
+#pragma once
+
+#include "../SparseQuantumSimulator/SparseSimulator.h"
+#include "../SparseQuantumSimulator/capi.hpp"
+#include "../SparseQuantumSimulator/capi.cpp" // yes really
+#include "../SparseQuantumSimulator/factory.hpp"
```
DmitryVasilevsky, line 6: I believe this should be ../Native rather than ../SparseQuantumSimulator

DmitryVasilevsky: Addressed here: https://github.com/sam-jaques/qsharp-runtime/pull/1

src/Simulation/Simulators/SparseSimulator/Native/SparseSimulator.h
```diff
@@ -0,0 +1,846 @@
+#pragma once
+#include <string>
+#include "types.h"
+#include "gates.h"
+#include <random>
```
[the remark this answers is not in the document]
DmitryVasilevsky: Addressed here: https://github.com/sam-jaques/qsharp-runtime/pull/1

src/Simulation/Simulators/SparseSimulator/Native/SparseSimulator.h
```diff
@@ -0,0 +1,846 @@
...
+			_execute_queued_ops();
+			return _quantum_state->probe(label);
+		}
+
+		std::string Sample() {
+			_execute_queued_ops();
+			return _quantum_state->Sample();
+		}
+
+		// Dumps the state of a subspace of particular qubits, if they are not entangled
```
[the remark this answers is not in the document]
sam-jaques: It passes back "false" to C#, which outputs the error message [here](https://github.com/sam-jaques/qsharp-runtime/blob/5470276e43b31dae7f5d6e20d03a2f8a8695942e/src/Simulation/Simulators/SparseSimulator/SparseSimulatorCS/Probes.cs#L122) to the output channel where it would otherwise print an unentangled state.

src/Simulation/Simulators/SparseSimulator/Native/flat_hash_map.hpp
sam-jaques: This hash map has better performance than the STL hash map, so it would be better if we can use this one. If we can't, the only functional difference from the STL hash map is the "jump_forward" function [defined here](https://github.com/sam-jaques/qsharp-runtime/blob/748672b5664be9cb3111dc6bbb36d2b9e1620615/src/Simulation/Simulators/SparseSimulator/Native/quantum_hash_map.hpp#L364), which makes the multi-threading faster.

src/Simulation/Simulators/SparseSimulator/Native/capi.hpp
```diff
@@ -0,0 +1,146 @@
...
+#define MICROSOFT_QUANTUM_DECL
+#endif
+#define MICROSOFT_QUANTUM_DECL_IMPORT __declspec(dllimport)
+
+using namespace Microsoft::Quantum::SPARSESIMULATOR;
+
+// All of these are called by the C# SparseSimulator class
+extern "C"
+{
+    MICROSOFT_QUANTUM_DECL unsigned init_cpp(logical_qubit_id num_qubits);
```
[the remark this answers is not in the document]
sam-jaques: When the C# class imports these functions, the suffix distinguishes them from the C# wrappers.

src/Simulation/Simulators/SparseSimulator/Native/quantum_state.hpp
```diff
@@ -0,0 +1,1511 @@
...
+	}
+	return 0;
+}
+
+template<size_t num_qubits>
+bool poppar(std::bitset<num_qubits> bitstring){
+	return bitstring.count() % 2;
+}
+
+// Compares two bitsets as through they were bitstrings
```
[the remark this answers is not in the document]
sam-jaques: Interpret the bitset as an integer, and then it uses the usual ordering of integers.

src/Simulation/Simulators/SparseSimulator/Native/quantum_state.hpp
```diff
@@ -0,0 +1,1511 @@
...
+	QuantumState() {
+		_qubit_data = wavefunction();
+		_qubit_data.max_load_factor(_load_factor);
+		// Create an initial all-zeros state
+		_qubit_data.emplace((logical_qubit_id)0, 1);
+		// Initialize randomness
+		std::random_device rd;
+		std::mt19937 gen(rd());
+		std::uniform_real_distribution<double> dist(0, 1);
+		_rng = [gen, dist]() mutable { return dist(gen); };
```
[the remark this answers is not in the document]
sam-jaques: It's local to each `QuantumState` object, which generally isn't threadsafe anyway.  All the functions within `QuantumState` that use `_rng` are single-threaded. So I think it is fine right now, but fragile.

src/Simulation/Simulators/SparseSimulator/Native/SparseSimulator.h
```diff
@@ -0,0 +1,846 @@
...
+			return true;	
+		}
+
+
+		void X(logical_qubit_id index) {
+ 			// XY = - YX
+			if (_Ry_queue[index]){
+				_Ry_angles[index] *= -1.0;
+			}
+			// Rx trivially commutes
```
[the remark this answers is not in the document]
sam-jaques: Rz are part of the phase/permutation queue, so they will be executed in-order with the rest of the phase/permutation.

src/Simulation/Simulators/SparseSimulator/Native/quantum_state.hpp
```diff
@@ -0,0 +1,1506 @@
...
+    case 3:
+        return -1i;
+    default:
+        return 0;
+    }
+    return 0;
+}
+
+template<size_t num_qubits>
+bool get_parity(std::bitset<num_qubits> bitstring){
```
thomashaener, line 52:
```suggestion
bool get_parity(const std::bitset<num_qubits>& bitstring){
```

src/Simulation/Simulators/SparseSimulator/Native/SparseSimulator.h
```diff
@@ -0,0 +1,846 @@
...
+				}
+				return;
+			}
+			// Queue the operation at this point
+			_queued_operations.push_back(operation(OP::MCX, target, controls));
+			_set_qubit_to_nonzero(target);
+		}
+
+		// Same as MCX, but we assert that the target is 0 before execution
+		void AND(std::vector<logical_qubit_id> const& controls, logical_qubit_id  target) {
```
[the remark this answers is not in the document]
sam-jaques: In this file: https://github.com/sam-jaques/qsharp-runtime/blob/sjaques/sparse-simulator/src/Simulation/Simulators/SparseSimulator/SparseSimulatorCS/Simulator.ApplyAnd.cs, it overrides the `ApplyAnd`  and calls this function.

The naming is inconsistent: Here it's called `AND`, but in the C api it's called `MCAnd`, and then C# calls it `ApplyAnd`.

src/Simulation/Simulators/SparseSimulator/Native/quantum_state.hpp
```diff
@@ -0,0 +1,1511 @@
...
+	case 3:
+		return -1i;
+	default:
+		return 0;
+	}
+	return 0;
+}
+
+template<size_t num_qubits>
+bool poppar(std::bitset<num_qubits> bitstring){
```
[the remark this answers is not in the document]
sam-jaques: It meant "parity". I renamed it to `get_parity` in this commit: https://github.com/sam-jaques/qsharp-runtime/commit/a53cbb001614e3ed7fbe9003f51c6069a7b71b03

src/Simulation/Simulators/SparseSimulator/Native/quantum_state.hpp
```diff
@@ -0,0 +1,1511 @@
...
+
+	QuantumState() {
+		_qubit_data = wavefunction();
+		_qubit_data.max_load_factor(_load_factor);
+		// Create an initial all-zeros state
+		_qubit_data.emplace((logical_qubit_id)0, 1);
+		// Initialize randomness
+		std::random_device rd;
+		std::mt19937 gen(rd());
+		std::uniform_real_distribution<double> dist(0, 1);
```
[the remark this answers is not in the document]
sam-jaques: Yes, there is a `SetSeed` function of the `SparseSimulator` class in C# which passes it (eventually) to the quantum state object.

src/Simulation/Simulators/SparseSimulator/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+# Special Operations
+The sparse simulator has some extra C\# Q\# operations for specific purposes.
+
+## Member functions of `SparseSimulator`
+`SetLogging(bool)`: Logging is set to `false` by default; if set to true, the simulator will output the name of each operation as it executes them. 
+
+## Q\# Functions
+
+`GetAmplitudeFromInt(qubit : Qubit[], label : Int)`: Uses `label` as a bitstring to index the qubits in `qubit`. It returns the amplitude of the state with that label, where the value of all other qubits is 0.
```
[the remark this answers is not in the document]
sam-jaques: That's a good idea. I think we only ever used this to dump the amplitude anyway, so replacing this with a diagnostic dump should be just fine.

src/Simulation/Simulators/SparseSimulator/Native/SparseSimulator.h
DmitryVasilevsky: Indeed. https://en.cppreference.com/w/cpp/language/identifiers

src/Simulation/Simulators/SparseSimulator/Native/SparseSimulator.h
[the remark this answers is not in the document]
sam-jaques: I made the naming consistent in this commit: https://github.com/sam-jaques/qsharp-runtime/commit/3375a20d62acf2193849ac3ee9942aca9bf665d6

src/Simulation/Simulators/SparseSimulator/SparseSimQSharpTests/Program.qs
```diff
@@ -0,0 +1,643 @@
...
+            H(qubits[idx]);
+        }
+        for idy in 1..128 {
+            CCNOT(qubits[idy % nqubits], qubits[(idy+1) % nqubits], qubits[(idy+2) % nqubits]);
+        }
+        test();
+        ResetAll(qubits);
+    }
+
+    operation MCXTime(nqubits : Int, ngates : Int) : Unit {
```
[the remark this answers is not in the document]
sam-jaques: This was used to test timing functionality for benchmarking, which was removed for this repo. I removed this operation in this commit: https://github.com/sam-jaques/qsharp-runtime/commit/3375a20d62acf2193849ac3ee9942aca9bf665d6

src/Simulation/Simulators/SparseSimulator/Native/factory.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+unsigned createSimulator(logical_qubit_id num_qubits)
+{
+    if (num_qubits > MAX_QUBITS)
+        throw std::runtime_error("Max number of qubits is exceeded!");
+    std::lock_guard<std::shared_mutex> lock(_mutex);
+    size_t emptySlot = -1;
+    for (auto const& s : _simulators)
+    {
+        if (s == NULL)
```
kuzminrobin, line 28:
```suggestion
        if (s == nullptr)
```

src/Simulation/Simulators/SparseSimulator/Native/gates.h
```diff
@@ -0,0 +1,279 @@
...
+    logical_qubit_id target;
+    operation(OP gate_type_arg,
+        logical_qubit_id target_arg) :
+        gate_type(gate_type_arg),
+        target(target_arg) {}
+
+    std::vector<logical_qubit_id> controls;
+    operation(OP gate_type_arg, 
+        logical_qubit_id target_arg, 
+        std::vector<logical_qubit_id> controls_arg
```
kuzminrobin, line 143:
```suggestion
        const std::vector<logical_qubit_id>& controls_arg
```

src/Simulation/Simulators/SparseSimulator/Native/quantum_state.hpp
```diff
@@ -0,0 +1,1506 @@
...
+
+namespace Microsoft::Quantum::SPARSESIMULATOR
+{
+
+// power of square root of -1
+inline amplitude iExp(int power)
+{
+
+    int p = ((power % 4) + 8) % 4;
+    switch (p)
```
kuzminrobin, line 35: I have an impression that this can be simplified to 
```suggestion
    switch (power % 4)
```

src/Simulation/Simulators/SparseSimulator/Native/quantum_state.hpp
```diff
@@ -0,0 +1,1506 @@
...
+std::bitset<num_qubits> get_mask(std::vector<logical_qubit_id> const& indices){
+    std::bitset<num_qubits> mask = std::bitset<num_qubits>();
+    for (logical_qubit_id index : indices) {
+        mask.set(index);
+    }
+    return mask;
+}
+
+template<size_t num_qubits>
+class QuantumState : public BasicQuantumState
```
kuzminrobin, line 82: FYI: 
See _Item 44: Factor parameter-independent code out of templates_ in [Effective C++: 55 Specific Ways to Improve Your Programs](https://www.google.com/search?q=effective+c%2B%2B+third+edition+pdf&rlz=1C1GCEU_enUS858US858&oq=effective+c%2B%2B+&aqs=chrome.2.69i57j69i59j35i39j0i512l2j69i60l2j69i65.6965j0j7&sourceid=chrome&ie=UTF-8), 3rd edition, by Scott Meyers.

kuzminrobin, line 82: This file (and this class in particular) will definitely benefit from breaking-up into multiple files.
Or at least it would be good to separate the definition of the class from the definition of the member functions:
```c++
// Definition of the class:
template<size_t num_qubits>
class QuantumState : public BasicQuantumState
{
  public:
    QuantumState();  // The member functions are declared only, but not defined here.
    . . . 
  private:
    // Member variables.
};

// Definition of the member functions (if it is a template class then in the same .hpp file):
template<size_t num_qubits>
QuantumState::QuantumState() {
    . . .
}
```

src/Simulation/Simulators/SparseSimulator/SparseSimulatorCS/SparseSimulator.cs
```diff
@@ -0,0 +1,388 @@
...
+                stackDepth += " ";
+            }
+        }
+
+        public override void OnOperationEnd(ICallable operation, IApplyData arguments)
+        {
+            
+            if (isLogging)
+            {
+                stackDepth = stackDepth[0..(stackDepth.Length - 1)];
```
kuzminrobin, line 381: I don't know C# but I have an impression that this line does not change anything.

src/Simulation/Simulators/SparseSimulator/Native/SparseSimulator.h
DmitryVasilevsky: Addressed here: https://github.com/sam-jaques/qsharp-runtime/pull/2

src/Simulation/Simulators/SparseSimulator/Native/SparseSimulator.h
DmitryVasilevsky: Addressed here: https://github.com/sam-jaques/qsharp-runtime/pull/3

src/Simulation/Simulators/SparseSimulator/Native/SparseSimulator.h
```diff
@@ -0,0 +1,846 @@
...
+#define MAX_QUBITS 1024
+#define MIN_QUBITS 64
+
+using namespace std::literals::complex_literals;
+
+namespace Microsoft
+{
+namespace Quantum
+{
+namespace SPARSESIMULATOR
```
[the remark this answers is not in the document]
DmitryVasilevsky: The same capitalization is used in our shipped native full state simulator. If we want to change it we probably need a work item to track it.

src/Simulation/Simulators/SparseSimulator/Native/capi.cpp
```diff
@@ -0,0 +1,302 @@
...
+
+    MICROSOFT_QUANTUM_DECL void MCSWAP_cpp(unsigned sim_id, _In_ int n, _In_reads_(n) logical_qubit_id* c, _In_ logical_qubit_id q1, _In_ logical_qubit_id q2){
+        getSimulator(sim_id)->CSWAP(std::vector<logical_qubit_id>(c, c + n), q1, q2);
+    }
+
+    MICROSOFT_QUANTUM_DECL void MCApplyAnd_cpp(unsigned sim_id,_In_ int length, _In_reads_(length) logical_qubit_id* controls, _In_ logical_qubit_id target){
+        getSimulator(sim_id)->MCApplyAnd(std::vector<logical_qubit_id>(controls, controls + length), target);
+    }
+
+    MICROSOFT_QUANTUM_DECL void MCAdjointApplyAnd(unsigned sim_id,_In_ int length, _In_reads_(length) logical_qubit_id* controls, _In_ logical_qubit_id target){
```
thomashaener, line 107:
```suggestion
    MICROSOFT_QUANTUM_DECL void MCAdjointApplyAnd_cpp(unsigned sim_id,_In_ int length, _In_reads_(length) logical_qubit_id* controls, _In_ logical_qubit_id target){
```

src/Simulation/Simulators/SparseSimulator/Native/quantum_state.hpp
thomashaener: `power` can also be negative, so one would still have to take care of the sign.

src/Simulation/Simulators/SparseSimulator/SparseSimQSharpTests/Program.qs
```diff
@@ -0,0 +1,643 @@
...
+            H(qubits[2]);
+            CNOT(qubits[0], qubits[2]);
+            DumpRegister("Test_file_3", qubits[0..1]);
+            DumpRegister("Test_file_4", qubits);
+            ResetAll(qubits);
+        }
+    }
+
+
+    operation _FakeR1Frac(numerator : Int, denominator : Int, qubit : Qubit[]) : Unit is Adj + Ctl {
```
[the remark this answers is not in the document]
sam-jaques: It does not call the R1 function internally, but simulates the effect with two R rotations. The idea is to test that the two types of rotation are the same.

src/Simulation/Simulators/SparseSimulator/README.md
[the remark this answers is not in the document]
sam-jaques: I replaced them with assertions in this commit: https://github.com/sam-jaques/qsharp-runtime/commit/a8729e7eea600bd8de9de1935200bd2600778bb0. Dumping specific amplitudes or a probabilistic sample to the console might be nice at some future point.

src/Simulation/Simulators/SparseSimulator/SparseSimulatorCS/SparseSimulator.cs
```diff
@@ -0,0 +1,385 @@
...
+using System.Collections.Generic;
+using Microsoft.Quantum.Simulation.QuantumProcessor;
+using Microsoft.Quantum.Simulation.Common;
+using Microsoft.Quantum.Simulation.Core;
+using System.Runtime.InteropServices;
+using System.Runtime.ExceptionServices;
+using System.Threading.Tasks;
+using Microsoft.Quantum.Simulation.QCTraceSimulatorRuntime;
+using Newtonsoft.Json.Serialization;
+using System.Reflection.Metadata.Ecma335;
```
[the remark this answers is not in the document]
sam-jaques: I don't remember. I cleared out a lot of the dependencies in this commit: https://github.com/sam-jaques/qsharp-runtime/commit/a8729e7eea600bd8de9de1935200bd2600778bb0

src/Simulation/Simulators/SparseSimulator/SparseSimQSharpTests/Program.qs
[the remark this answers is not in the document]
cgranade: Understood, and I see you're using it to call the `AssertOperationsEqualReferenced` diagnostic operation, makes sense. In this case, I wouldn't think of this as "fake," then, I guess, but a decomposition of the `R1Frac` operation into `RFrac` calls. That said, since the `RFrac(PauliI, _, _, _)` call doesn't do anything other than impact a global phase, I'm not sure that as written this tests that the global phase imparted is correct?

src/Simulation/Simulators/SparseSimulator/README.md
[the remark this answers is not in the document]
cgranade: In general, we've handled dumping by using `DumpMachine` and `DumpRegister` and then relying on the structure of that output from the host program. In particular, IQ# and Python are able to access JSON serializations of objects emitted via the `SimulatorBase.MaybeDisplayDiagnostic` event, such that emitting a dump to that event from inside a simulator allows host programs to pick out specific amplitudes that way without specializing the Q# program to a specific simulator.

src/Simulation/Simulators/SparseSimulator/SparseSimQSharpTests/Program.qs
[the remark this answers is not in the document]
sam-jaques: That's all true. The sparse simulator doesn't have a well-defined global phase anyway, so it shouldn't be able to check global phase. It would probably be more useful if it also ran a test of controlled versions of the same gates.

src/Simulation/Simulators/SparseSimulator/README.md
```diff
@@ -0,0 +1,155 @@
...
+ - CMake v1.2.0
+ - Q\# v0.15.2101126940
+ - C++11
+ - Dotnet core v 3.1
+ - Dotnet cake v1.1.0
+
+# Setup
+To build the Sparse Simulator, call `dotnet restore` from the main `SparseSimulator` folder, then `dotnet cake`. This builds the C++ backend and the C\# interface. 
+
+To use the SparseSimulator in a Q\# project, ensure that it includes
```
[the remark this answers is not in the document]
kuzminrobin: Yes, there is a plan to make SparseSimulator a part of the Microsoft.Quantum.Simulators NuGet package.

src/Simulation/Simulators/SparseSimulator/Native/factory.cpp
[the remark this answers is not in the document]
kuzminrobin: @sam-jaques, some more info is [here](https://github.com/microsoft/qsharp-runtime/blob/main/src/Qir/QIR-RT-API-Design-Guidelines.md#dont-let-the-c-exceptions-cross-the-abi-boundary).

src/Simulation/Simulators/SparseSimulator/Native/types.h
```diff
@@ -0,0 +1,51 @@
...
+		#ifndef USE_SINGLE_PRECISION
+					using RealType = double;
+		#else
+					using RealType = float;
+		#endif
+
+
+
+		// Logical qubit id is visible to the clients and is immutable during the lifetime of the qubit.
+		using logical_qubit_id = unsigned;
```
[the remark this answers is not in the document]
kuzminrobin: @sam-jaques, some more info is here:
[Be Careful When Exposing Certain Data Types](https://github.com/microsoft/qsharp-runtime/blob/main/src/Qir/QIR-RT-API-Design-Guidelines.md#be-careful-when-exposing-certain-data-types)
[Integer Types](https://github.com/microsoft/qsharp-runtime/blob/main/src/Qir/QIR-RT-API-Design-Guidelines.md#integer-types)

---
Strategically we plan to refactor our QIR Runtime according to [those guidelines](https://github.com/microsoft/qsharp-runtime/blob/main/src/Qir/QIR-RT-API-Design-Guidelines.md). There is also a chance of refactoring the [full state simulator](https://github.com/microsoft/qsharp-runtime/tree/main/src/Simulation/Native/src/simulator) also. And eventually your SparseSimulator also. So it would be beneficial if you could take into account those guidelines when making fixes to your SparseSimulator.

src/Simulation/Simulators/SparseSimulator/README.md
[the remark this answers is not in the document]
kuzminrobin: @sam-jaques, FYI, I will be working in pair with Dmitry on adding your SparseSimulator to our set of simulators.

src/Simulation/Simulators/SparseSimulator/Native/basic_quantum_state.hpp
```diff
@@ -0,0 +1,77 @@
...
+namespace Microsoft::Quantum::SPARSESIMULATOR
+{
+
+// Virtual class for QuantumState
+// This is not templated, so it allows SparseSimulator types to avoid templates
+class BasicQuantumState 
+{
+public:
+
+    BasicQuantumState() {}
```
kuzminrobin, line 20: @DmitryVasilevsky, FYI, I assume this constructor was planned to be virtual.

src/Simulation/Simulators/SparseSimulator/Native/types.h
```diff
@@ -0,0 +1,45 @@
...
+
+#include "quantum_hash_map.hpp"
+
+namespace Microsoft::Quantum::SPARSESIMULATOR
+{
+
+using mutex_type = std::mutex;
+using recursive_mutex_type = std::recursive_mutex;
+using lock_type = std::lock_guard<mutex_type>;
+using recursive_lock_type = std::lock_guard<recursive_mutex_type>;
```
kuzminrobin, line 21: Never used?

DmitryVasilevsky: It was used for multithreading, but now I removed multithreading and these types.

PR #770, opened 2021-07-08
Linux development fixes

swernli:
This includes a mix of fixes for successful development within the devcontianer:

 - Update Dockerfile to include updated version of CMake.
 - Update prerequisites.ps1 to actually call experimental simlator prereq script.
 - Update experimental sim prereq script to include nightly toolchain.
 - Check for sudo before using in QIR prereq script.
 - Use correct environment vars for experimental sim in build script output.
 - Fix native sim OMP thread count set to only occur if OMP is enabled.


.devcontainer/Dockerfile
```diff
@@ -1,22 +1,30 @@
...
-RUN apt update \
-    && apt-get install -y cmake \
+# Install Mono, NuGet, CMake, Ninja, Clang-11, build-essential
+RUN apt-get -y update \
+    && apt-get -y install dirmngr gnupg apt-transport-https ca-certificates \
+    && apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF \
+    && sh -c 'echo "deb https://download.mono-project.com/repo/ubuntu stable-focal main" > /etc/apt/sources.list.d/mono-official-stable.list' \
+    && apt-get -y update \
+    && apt-get -y install --no-install-recommends mono-complete nuget \
+    && apt-get -y install jq \
```
cgranade, line 10: Good idea to combine these two layers!

.devcontainer/Dockerfile
```diff
@@ -1,22 +1,30 @@
...
                           ninja-build \
                           clang-11 \
                           clang-tidy-11 \
                           build-essential \
-    && apt-get clean \
-    && rm -rf /var/lib/apt/lists/
+    && apt-get clean && rm -rf /var/lib/apt/lists/
+
+# Install latest CMake (required since apt-get uses 3.16 and repo requires 3.20)
+RUN json=$(curl -s "https://api.github.com/repos/Kitware/CMake/releases") \
```
cgranade, line 19: This somewhat horrifies me, but I confess that I don't have any suggestions on how to improve...!

cgranade (approved):

.devcontainer/Dockerfile
swernli: Yeah, it's a odd thing to include, but I copied it from the scripts that configure the Azure DevOps/GitHub Actions pipeline VMs (see https://github.com/actions/virtual-environments/blob/main/images/linux/scripts/installers/cmake.sh). The good news is that when apt gets the option to install CMake 3.20 (either by default or as an explicit opt-in) then we can remove this.

cgranade: Makes perfect sense, my slight horror notwithstanding!

.devcontainer/Dockerfile
```diff
@@ -1,22 +1,30 @@
...
+    && latest_tag=$(echo $json | jq -r '.[] | select(.prerelease==false).tag_name' | sort --unique --version-sort | grep -v "rc" | tail -1) \
```
idavis, line 20: This will always install the latest. I think we should control the version more.
```bash
    curl -SsL https://github.com/Kitware/CMake/releases/download/v3.20.5/cmake-3.20.5-linux-x86_64.sh -o cmakeinstall.sh \
    && echo "f582e02696ceee81818dc3378531804b2213ed41c2a8bc566253d16d894cefab cmakeinstall.sh" | sha256sum -c --strict - && \
    && chmod +x cmakeinstall.sh \
    && ./cmakeinstall.sh --prefix=/usr/local --exclude-subdir \
    && rm cmakeinstall.sh
```

idavis (changes requested):

.devcontainer/Dockerfile
swernli: Oh, I like this suggestion. It also removes the dependency on jq that was added above. I'll try this out locally, and if it works I'll push an update that uses this approach.

swernli: Ok, I made the update to use this mechanism and verified the behavior locally. Thanks for the suggestion!

.devcontainer/Dockerfile
```diff
@@ -1,22 +1,27 @@
...
+
+# Install CMake 3.20 (required since apt-get uses 3.16 and repo requires 3.20)
+RUN curl -SsL https://github.com/Kitware/CMake/releases/download/v3.20.5/cmake-3.20.5-linux-x86_64.sh -o cmakeinstall.sh \
+    && echo "f582e02696ceee81818dc3378531804b2213ed41c2a8bc566253d16d894cefab cmakeinstall.sh" | sha256sum -c --strict - \
+    && chmod +x cmakeinstall.sh \
+    && ./cmakeinstall.sh --prefix=/usr/local --exclude-subdir \
+    && rm cmakeinstall.sh
 
+# Install Rust
 RUN curl https://sh.rustup.rs -sSf | sh -s -- -y
```
idavis, line 25: I don't think this will create a usable rust env for dev containers. See [rust-debian.sh](https://github.com/microsoft/qsharp-compiler/blob/26eded8e4938a4fc9b636ff01d523c12d7ac7f4f/src/QirTools/pyqir/.devcontainer/library-scripts/rust-debian.sh) and [Dockerfile](https://github.com/microsoft/qsharp-compiler/blob/26eded8e4938a4fc9b636ff01d523c12d7ac7f4f/src/QirTools/pyqir/.devcontainer/Dockerfile#L3-L26).

idavis (changes requested):

.devcontainer/Dockerfile
swernli: I'm not sure I understand. This works to compile the experimental open system simulator in Rust, and seems sufficient for other related Rust projects I've done. Why would additional handling need to be done?

swernli: To clarify, the entire runtime repo is built via the `build/build.ps1` script, after having initially executed `bootstrap.ps1` the first time. Creating this development container, running bootstrap, build and test scripts succeeds, including Rust components.

swernli: @idavis Any concerns still remaining here? This installation mechanism has been working for all of my container dev environments, and was originally added by @cgranade when we first added Rust code to this repo.

swernli: We landed at using the universal devcontainer approach as used by several Azure SDKs and other projects.

idavis (approved):

PR #772, opened 2021-07-10
Remove reference to props

cesarzc:
This PR removes a reference to `Simulators.Dev.props` which inadvertently caused the `Microsoft.Quantum.Qir.Tools` package to have reference on  a `Microsoft.Quantum.Simulation.Common` package that does not exist.


src/Qir/Execution/Tools/Microsoft.Quantum.Qir.Tools.csproj
```diff
@@ -53,45 +53,4 @@
     </Compile>
   </ItemGroup>
 
-  <ItemGroup>
```
ScottCarda-MS, line 56: This logic was important for the qir-tools to work correctly. Did it get moved somewhere?

PR #773, opened 2021-07-12
Update target packages for QIR emission

swernli:
This updates the internal callables in target packages to use TargetInstruction attribute for easier override/implementation in generated QIR. It also cleans up some leftover adjoint support on uncontrolled Rx, Ry, Rz callables that is no longer used or needed.

In addition, this change updates the use of adjoint S and adjoint T so that they are considered stand-alone operations instead of relying on functor generation. This allows for better use of the TargetInstruction attribute and corresponds better to the way targets define those operations on hardware. It also introduces a more explicit CNOT decomposition that better maps to underlying intrinsics, specifically in that it will allow calls to `CNOT` directly to be translated into `ApplyControlledX` without having to go through the controlled functor and incur the creation of an array of qubits.


src/Simulation/TargetDefinitions/Intrinsic/ApplyControlledX.qs
```diff
@@ -31,6 +32,7 @@ namespace Microsoft.Quantum.Intrinsic {
     /// ```qsharp
     /// CNOT(control, target);
     /// ```
+    @TargetInstruction("cnot")
```
swernli, line 35: This repo doesn't have any sample target package QIR output, so I can explain the effect of this change here. Without the `TargetInstruction` attribute, this callable would translate into a QIR declaration names something like `__quantum__qis____qsref1__applycontrolledx____body`. I say "named something like" because being `internal` means the name of the function automatically gets updated with `qsrefN` for some numeric value `N` in the Q# syntax tree to avoid collisions. With the `TargetInstruction` attribute defined, the provided string "cnot" overrides the name-mangled declaration and instead the QIR will have `__quantum__qis__cnot` which is both more simply named and deterministic such that implementing the definition of the declaration is possible.

kuzminrobin (approved):
At a high level looks good to me.
I would recommend at least one more approval (if it doesn't take too long) before merging.

PR #774, opened 2021-07-13
Adding option to provider UserAgent when submitting jobs to Azure Quantum

anpaz:
The user-agent can be provided either as an extra argument to the executable, or as an environment variable.


ricardo-espinoza (approved):

ScottCarda-MS (approved):
From the logs it looks like there are still a couple of tests that need updating to expect the new parameter. Other than that, it looks good to me.

PR #776, opened 2021-07-13
Sanitizing CI

kuzminrobin:
Adding the sanitizer runs to the PR pipeline builds.


kuzminrobin:
This PR was not intended to be merged. It was making sure that the Debug configuration is not built for QIR RT and the sanitizer errors reported in Debug built are not caught by the CI builds.

kuzminrobin:
I'm reusing this PR for testing a new CI build pipeline.

build/.map
```diff
@@ -0,0 +1,79 @@
+# Scripts Map (Scripts Call Tree, shows the relations between the scripts).
+# Azure Pipelines documentation: https://docs.microsoft.com/en-us/azure/devops/pipelines/?view=azure-devops
+# See also: 
+#   QDK build pipeline wiki https://ms-quantum.visualstudio.com/Quantum%20Program/_wiki/wikis/Quantum-OKRs.wiki/328/QDK-build-pipeline
```
swernli, line 4: This is not a public path and should not appear in this repo.

build/.map
```diff
@@ -0,0 +1,79 @@
...
+# Pipeline: 
+#   microsoft.qsharp-runtime            https://dev.azure.com/ms-quantum-public/Microsoft%20Quantum%20(public)/_build?definitionId=15&_a=summary
+#   microsoft.qsharp-runtime.sanitized  https://dev.azure.com/ms-quantum-public/Microsoft%20Quantum%20(public)/_build?definitionId=34&_a=summary
+ci.yml (Build.Configuration = (microsoft.qsharp-runtime           ? Release : 
+                               microsoft.qsharp-runtime.sanitized ? Debug   :
+                               <?>), 
+        Build.Verbosity = n, Skip.Tests = false)
+    steps.yml
+        steps-init.yml
+            'Use NuGet 5.6.0' NuGetToolInstaller@0
```
swernli, line 15: is this file generated or manually written? If generated, then I would not expect it to be checked in. If manually written, it's going to be a pain point keeping it up to date. Do we really need to check this into the repo? It seems like it is all duplicate information.

swernli (changes requested):

build/set-env.ps1
```diff
@@ -6,38 +6,67 @@ $ErrorActionPreference = 'Stop'
 Write-Host "Setting up build environment variables"
 
 If ($Env:BUILD_BUILDNUMBER -eq $null) { $Env:BUILD_BUILDNUMBER ="0.0.1.0" }
+Write-Host "BUILD_BUILDNUMBER: $Env:BUILD_BUILDNUMBER"
```
cgranade, line 9: To reduce duplicating env variable names and `Write-Host` calls, it's probably easier to use `Get-ChildItem` to write out a list of environment variables:

```pwsh
Get-ChildItem -Path Env:/* -Include BUILD_BUILDNUMBER, BUILD_CONFIGURATION, PYTHON_VERSION  | Format-Table | Out-String | Write-Host
```

For readability in source, you can break up the above using `@()` to write out the array of environment variables instead:

```pwsh
Get-ChildItem -Path Env:/* -Include @(
    "BUILD_BUILDNUMBER",
    "BUILD_CONFIGURATION",
    "PYTHON_VERSION"
) | Format-Table | Out-String | Write-Host
```

build/set-env.ps1
```diff
@@ -6,38 +6,67 @@ $ErrorActionPreference = 'Stop'
...
+Write-Host "PYTHON_VERSION: $Env:PYTHON_VERSION"
+
 If ($Env:NUGET_VERSION -eq $null) { $Env:NUGET_VERSION ="0.0.1-alpha" }
+Write-Host "NUGET_VERSION: $Env:NUGET_VERSION"
 
-If (($Env:ENABLE_NATIVE -ne "false") -and ($Env:NATIVE_SIMULATOR -eq $null) ) {
-    $Env:NATIVE_SIMULATOR = (Join-Path $PSScriptRoot "..\src\Simulation\Native\build\drop")
+
+If ($Env:ENABLE_NATIVE -ne "false") {
+    If ($Env:NATIVE_SIMULATOR -eq $null) {
```
cgranade, line 28: Testing with string interpolation also allows for catching when the environment variable is defined, but is empty:
```suggestion
    If ("$Env:NATIVE_SIMULATOR" -eq "") {
```

build/.map
kuzminrobin: This is manually written. I have written this for myself when understanding how the pipelines work, and I decided to share this with the team (and with myself in the future).

I was assuming that this file does not need to be kept up to date. But when one needs to learn the pipelines again, they will have some starting point and update the file upon need.

Also if we have a problem in the pipeline, then looking at this map file we can quickly get an idea in which .yml or .ps1 to look for a problem.

swernli: I think putting something like this on the wiki might make sense, but having this in repo feels like it could get confusing. I'd personally prefer to not check it in here.

build/.map
kuzminrobin: Thanks, removing.

build/set-env.ps1
kuzminrobin: Thanks! Changing.

build/set-env.ps1
kuzminrobin: Hm.. Interesting.
As I understand, the case `If ("$Env:NATIVE_SIMULATOR" -eq "") {` does not catch the undefined variables. If we want to catch the undefined vars and empty vars then we need to write 
`($Env:NATIVE_SIMULATOR -eq $null) -or ("$Env:NATIVE_SIMULATOR" -eq "")`
Is that a correct understanding?

build/.map
kuzminrobin: OK, Removing.

build/set-env.ps1
cgranade: I believe that in the context of string interpolation, undefined environment variables expand to empty strings, such that `"$Env:FOOBAR" -eq ""` evaluates to `$true` if the environment variable `FOOBAR` is either undefined or is defined as a blank string.

kuzminrobin:
I re-enabled the microsoft.qsharp-runtime.**sanitized** pipeline (after another [PR](https://github.com/microsoft/qsharp-runtime/pull/785)), and now I'm checking how the new pipeline works.

kuzminrobin:
Please review.

kuzminrobin:
Please review

swernli (approved):

Issue #187, opened 2020-05-04
Easy access of attribute values in custom simulator


msoeken:
**Please describe what you would like the feature to accomplish.**
Here is a characteristic example in which the use of attributes and the access of their values are helpful.

```q#
@Test("ResourcesEstimator")
@TSVOutput("operation.tsv")
operation MyOperation() : Unit {
  // ...
}
```

The operation is simulated using `ResourcesEstimator` in a Q# test project. The result of the resource estimation in TSV format are dumped into a file.

**Describe the solution you'd like**
A method or field to access all callable's attributes and their values, which can be used in the implementation of the simulator.

**Describe alternatives you've considered**
I accessed the attributes from the assemblies `CallableDeclarationAttribute` information.

```c#
public class ResourcesEstimator : Microsoft.Quantum.Simulation.Simulators.ResourcesEstimator
{
    public override Task<O> Run<T, I, O>(I args)
    {
        var result = base.Run<T, I, O>(args).Result;

        var filename = (from attrib in typeof(T).Assembly.CustomAttributes
                        where attrib.AttributeType == typeof(CallableDeclarationAttribute)
                        let header = CallableDeclarationHeader.FromJson(attrib.ConstructorArguments[0].Value.ToString()).Item2
                        where header.QualifiedName.Equals(typeof(T).Namespace, typeof(T).Name)
                        from declAttrib in header.Attributes
                        where declAttrib.TypeId.Item.Equals("Some.Namespace", "TSVOutput")
                        let arguments = declAttrib.Argument.Expression as ExpressionKind.ValueTuple
                        let argument = arguments.Item[0]
                        let stringLiteral = argument.Expression as ExpressionKind.StringLiteral
                        select stringLiteral.Item1.Value).FirstOrDefault();
            
        if (!string.IsNullOrEmpty(filename))
        {
            File.WriteAllText(filename, ToTSV());
        }
            
        return Task.Run(() => result);
    }
}
```

**Additional context**
N/A

bettinaheim:
Hi @msoeken,
In general this seems reasonable, though we'll have to flesh out the details. For this particular case, it would probably make sense for DumpMachine to actually dump the tsv. @cgranade, @anpaz-msft Thoughts?

msoeken:
[...]

PR #779, opened 2021-07-16
Update Simulator default random seed

swernli:
This updates the default seed used by the simulator when initializing the random number generator used within the wave function. Previously it used `std::clock()` which is time since process start, but that is implementation and platform dependent (see https://en.cppreference.com/w/cpp/chrono/c/clock). This is changed to instead use current time, and was tested to show sufficient randomness even with rapid repeated invocation.

### DIFF
--- a/src/Qir/Runtime/lib/Simulators/FullstateSimulator.cpp
+++ b/src/Qir/Runtime/lib/Simulators/FullstateSimulator.cpp
@@ -10,7 +10,6 @@
 #include <vector>
 #include <fstream>
 #include <climits>
-#include <chrono>
 
 #pragma clang diagnostic push
     // Ignore warnings for reserved macro names `_In_`, `_In_reads_(n)`:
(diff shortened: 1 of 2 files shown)


cesarzc (approved):
LGTM.

Interesting bug fix :)

kuzminrobin (approved):

idavis (changes requested):
My historical assumptions are that seed is always set to current instant to create "random" behavior. But that if someone wants to create determinism, then, and only then they override

swernli:
@idavis That's a good point... The current behavior is the other way around. I think the right fix to get safe-by-default randomness is to replace the use of `std::clock()` in the simulator wave function itself. I switch to that pattern.

kuzminrobin:
What do you guys think about this PR - #780?
It allows the user to choose whether the random sequence will be deterministic or not.

swernli:
@kuzminrobin I think that PR makes sense, and would even be reasonable to include along with this one. After the change, this one is more about making sure the native fullstate simulator is safe-by-default. But since the C# wrapper already explicitly sets the seed, there is nothing wrong with the QIR wrapper doing the same. Having it be parametrizable is also a step toward addressing feedback we've received about wanting a way to make the simulator deterministic for testing (see https://github.com/microsoft/qsharp-runtime/issues/187#issuecomment-715626082 for example).

idavis (approved):

PR #780, opened 2021-07-16
Added user provided seed to CFullStateSinulator

kuzminrobin:
Based on the early commit in this PR - #779.


swernli (approved):
Looks good! I think this can go in along with the updated #779, both change are worth making.

idavis (approved):

PR #782, opened 2021-07-22
Combining QIR build tools and runtime into one package

bettinaheim:
For the QIR build tools to work properly, that package also needs to include the runtime (libs and headers). Rather than keeping the build tools a separate package that also includes the runtime, this PR consolidates them both into one NuGet package called Microsoft.Quantum.Qir.Runtime that contains
- The compiled libs for the native runtime (built in release config)
- The headers for the native runtime
- The .NET build tools

The .NET build tools include:
- tools for C++ entry point generation
- tools for invoking the clang compiler out of process and building an executable (assuming clang is installed)
- tools for running a built executable

Also, properly signs the runtime package.


swernli (approved):
Only minor feedback is that I'm not personally a fan of when a project and the package it produces have different names (`Microsoft.Quantum.Qir.Runtime.Tools.csproj` produces `Microsoft.Quantum.Qir.Runtime.nupkg`), but that's more of a personal preference than an issue to be resolved. Otherwise everything looks good!

PR #784, opened 2021-07-30
Use Directory.Build.props to suppress test docgen

swernli:
We incorrectly generate documentation for a couple of test projects that we forgot to add `<QSharpDocsGeneration>false</QSharpDocsGeneration>` to. So that we don't have to rely on remembering to include this in each new project, we can use a Directory.Build.props to enforce this setting in the whole test directory.

### DIFF
--- a/src/Simulation/Simulators.Tests/TestProjects/Directory.Build.props
+++ b/src/Simulation/Simulators.Tests/TestProjects/Directory.Build.props
@@ -1,0 +1,9 @@
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Build.props', '$(MSBuildThisFileDirectory)../'))" Condition="'$([MSBuild]::GetPathOfFileAbove(Directory.Build.props, $(MSBuildThisFileDirectory)../))' != ''"/>
+
+  <PropertyGroup>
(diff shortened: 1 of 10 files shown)


src/Simulation/Simulators.Tests/TestProjects/Directory.Build.props
```diff
@@ -0,0 +1,9 @@
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
cgranade, line 1: This is a great idea! Would it perhaps make sense to apply this Directory.Build.props one more level up still, so that it applies to everything in Simulators.Tests?

cgranade (approved):

src/Simulation/Simulators.Tests/TestProjects/Directory.Build.props
swernli: I considered that, but looking at the rest of the tests they all use Simulators.Tests.props, so already have documentation suppressed. I realize this leaves us open to making this mistake again if a new folder with projects that don't use a common props file is introduced, but hopefully that is easier to spot. Overall, I think accidentally publishing documentation for test structures is the lesser evil compared with accidentally skipping publication of library/runtime documentation, so I like the publish by default model in this repo, and we just need to watch out for places in the tree that are known to have multiple projects that are test-only and follow one of the two patterns (common props file or directory props file).

PR #786, opened 2021-08-05
Remove dependency to Microsoft.Quantum.Simulation.Common from Microsoft.Quantum.Qir.Runtime package

cesarzc:
This change removes the dependency that the `Microsoft.Quantum.Qir.Runtime` package currently has on `Microsoft.Quantum.Simulation.Common` which is a DLL that is not distributed as its own package.

### DIFF
--- a/src/Qir/Tools/Microsoft.Quantum.Qir.Runtime.Tools.csproj
+++ b/src/Qir/Tools/Microsoft.Quantum.Qir.Runtime.Tools.csproj
@@ -1,8 +1,5 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
-  <Import Project="..\..\Simulation\Common\Simulators.Dev.props" />
-  <Import Project="..\..\Simulation\Common\AssemblyCommon.props" />
-
   <PropertyGroup>
     <TargetFramework>netstandard2.1</TargetFramework>
     <PlatformTarget>x64</PlatformTarget>
@@ -21,6 +18,11 @@
 
   <ItemGroup>
     <None Include="..\..\..\build\assets\qdk-nuget-icon.png" Pack="true" Visible="false" PackagePath="" />
+    <None Include="Externals\CLI11\*" Pack="true" Visible="false" PackagePath="externals/CLI11">
+      <Link>externals\CLI11\%(RecursiveDir)%(FileName)%(Extension)</Link>
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      <Visible>false</Visible>
+    </None>
     <None Include="..\..\Qir\drops\bin\**\*" Pack="true" Visible="false" PackagePath="runtimes">
       <Link>runtimes\%(RecursiveDir)%(FileName)%(Extension)</Link>
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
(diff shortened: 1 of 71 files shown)


src/Qir/Tools/Microsoft.Quantum.Qir.Runtime.Tools.csproj
```diff
@@ -74,32 +65,4 @@
     </Compile>
   </ItemGroup>
 
-  <ItemGroup>
```
ScottCarda-MS, line 77: Is this logic no longer needed? It seemed pretty important before to make sure the QSimDll got into the right folder. And to get the native resources as well.

cesarzc: You're right. The e2e tests fail because of this. Even though this logic is important, it shouldn't be part of this project file. The problem with this is that it only works when the project is referenced directly and not when the package produced by this project is being referenced.

I'll fix this problem by adding a reference to the `Microsoft.Quantum.Simulators` package to the command line tool project and change some logic there if needed.

cesarzc: Done.

src/Simulation/Simulators/Microsoft.Quantum.Simulators.csproj
```diff
@@ -29,6 +29,15 @@
...
   <ItemGroup>
+    <!-- This is required for simulators' libraries to be copied to the output folder when this project is directly referenced. -->
+    <None Include="$(QSimDll)">
+      <Link Condition="$([MSBuild]::IsOsPlatform('Windows'))">simulators\win-x64\%(FileName)%(Extension)</Link>
+      <Link Condition="$([MSBuild]::IsOsPlatform('OSX'))">simulators\osx-x64\%(FileName)%(Extension)</Link>
+      <Link Condition="$([MSBuild]::IsOsPlatform('Linux'))">simulators\linux-x64\%(FileName)%(Extension)</Link>
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      <Visible>false</Visible>
+    </None>
+    <!-- This is required to be able to invoke the native simulator through .NET. -->
```
swernli, line 40: I don't think this is true, and we shouldn't need to duplicate the file with two different names into the same project/package. If we use the right names and the right folder structure, then .NET will handle the platform specific loading logic for us.

swernli (changes requested):
I think we need to update the e2e builds to skip the rename of files and use the platform specific library names throughout. That will eliminate the need to duplicate the file in the project/package.

src/Simulation/Simulators/Microsoft.Quantum.Simulators.csproj
cesarzc: Removed the duplication.

swernli (approved):
Looks great, thanks for working on this!

swernli:
Whoops... @cesarzc looks like there may be one more spot that needs to be updated:
https://github.com/microsoft/qsharp-runtime/blob/78d68dcfc7f3b6f08a5492e7a22ef11fbca658d7/src/Simulation/Simulators/QuantumSimulator/QuantumSimulator.cs#L17
This line needs to updated to not use `.dll` in the name, otherwise the automatic search will be bypassed.

cesarzc:
> Whoops... @cesarzc looks like there may be one more spot that needs to be updated:
> https://github.com/microsoft/qsharp-runtime/blob/78d68dcfc7f3b6f08a5492e7a22ef11fbca658d7/src/Simulation/Simulators/QuantumSimulator/QuantumSimulator.cs#L17
> 
> 
> This line needs to updated to not use `.dll` in the name, otherwise the automatic search will be bypassed.

Updated. Great catch! Thanks @swernli :)

PR #788, opened 2021-08-10
Reformatted the C++ sources according to clang-format rules.

kuzminrobin:
It is a step two of the three mentioned below:
* Update the `.clang-format` configuration file for `clang-format` tool (enforcing the code style).
* Format the C++ sources according to the `.clang-format` file and update the file itself if needed.
* Enable the `clang-format` tool as a build step.


swernli:
I don't see which part of the PR adds clang-format as a build step... did that not make it into this change?

kuzminrobin:
Adding clang-format as a build step will be done in a separate PR.

swernli (approved):

PR #790, opened 2021-08-10
Qubit management in QIR.

DmitryVasilevsky:
Qubit management functions in the QIR. Function declarations are up to date with design discussion. Functions are hooked up to backing code.


src/Qir/Runtime/public/QirRuntime.hpp
```diff
@@ -12,33 +12,59 @@
...
+    //
     // Return a borrowed qubit.
     // TODO QIR_SHARED_API void quantum__rt__qubit_return(QUBIT*); // NOLINT
-
+    //
     // Return an array of borrowed qubits.
     // TODO QIR_SHARED_API void quantum__rt__qubit_return_array(QirArray*); // NOLINT
 
+
+
```
kuzminrobin, line 67: I feel that this section can be removed. I don't see the API changes in this commented out section and the non-commented-out section "Qubit Management".

kuzminrobin (approved):

src/Qir/Runtime/public/QirRuntime.hpp
DmitryVasilevsky: Of course. I will remove it when we fully settle on the new API.

DmitryVasilevsky: Updated.

src/Qir/Runtime/lib/QIR/delegated.cpp
```diff
@@ -25,27 +25,61 @@ static std::unordered_map<RESULT*, int>& AllocatedResults()
...
+    {
+        // Currently we implement borrowing as allocation.
+        return quantum__rt__qubit_allocate();
+    }
+
+    void quantum__rt__qubit_return(QUBIT* qubit)
+    {
+        // Currently we implement borrowing as allocation.
+        quantum__rt__qubit_release(qubit);
+    }
```
swernli, line 66: I'm a little confused as to why these are implemented here this way... the QIR spec doesn't describe borrowing, our QIR bridge has those functions commented out (they should probably be removed), and QirGeneration rewrite step in the compiler automatically turns any borrows into allocates. Since there are no callers for these and there is no infrastructure to generate calls to these and they aren't in the spec, why include them here as forwarders to allocate?

DmitryVasilevsky: Great question. From our last discussion with @bettinaheim I got under impression that eventually we would like to implement true borrowing. Forwarding in the rewrite step seems useful when runtime doesn't support it. But for the true borrowing implementation we need to support it in the runtime. First, we will forward on the level C interface. Then we will forward in qubit manager and simulators, and finally we will implement true support in qubit manager. I believe this is the plan at the moment.

PR #794, opened 2021-08-12
Enabled the clang-format check

kuzminrobin:


src/Qir/check-sources-formatted.ps1
```diff
@@ -0,0 +1,47 @@
...
+#Write-Host "3"
+#$filesRequireFormatting = get-content $tmpFile | ?{$_ -like "*: warning:*"} `
+#                            | %{[string]::join(":",($_.split("warning:")[0].split(":") | select -SkipLast 3))} `
+#                            | sort | unique
+#Remove-Item $tmpFile
+
+Write-Host "A"
+&{
+   Write-Warning "warning"
+   Write-Error "error"
```
cgranade, line 30: @kuzminrobin: If I recall, the build pipeline is currently configured to interpret any text written to stderr as a failure (similar to `-o pipefail` in bash). The trick to avoid that is to turn stderr into stdout, but then that can also be interpreted as an error unless `$ErrorPreference` is set to `"SilentlyContinue"` (https://stackoverflow.com/a/66726535). As detailed in that answer, you can also use `#Requires` to depend on 7.1, and use the `PSNotApplyErrorActionToStderr` experimental feature to turn off interpreting stderr as an exception stream.

src/Qir/check-sources-formatted.ps1
```diff
@@ -0,0 +1,31 @@
...
+    | ?{$_ -notlike "*/Externals/*"} | ?{$_ -notlike "*/drops/*"} | ?{$_ -notlike "*/bin/*"} `
+    | %{clang-format -n -style=file $_} 2>format.log
+$ErrorActionPreference=$OldErrorActionPreference
+
+$filesRequireFormatting = get-content $tmpFile | ?{$_ -like "*: warning:*"} `
+                            | %{[string]::join(":",($_.split("warning:")[0].split(":") | select -SkipLast 3))} `
+                            | sort | unique
+Remove-Item $tmpFile
+
+if (! ("$filesRequireFormatting" -eq ""))
```
cgranade, line 24:
```suggestion
if (-not ("$filesRequireFormatting" -eq ""))
```

src/Qir/check-sources-formatted.ps1
```diff
@@ -0,0 +1,31 @@
...
+    $DirPath = "$PSScriptRoot"
+)
+
+$tmpFile = "format.log"
+
+$OldErrorActionPreference = $ErrorActionPreference
+$ErrorActionPreference='Continue'
+"$DirPath/*.cpp","$DirPath/*.c","$DirPath/*.h","$DirPath/*.hpp" | get-childitem -Recurse | %{$_.fullname -replace "\\", "/"} `
+    | ?{$_ -notlike "*/Externals/*"} | ?{$_ -notlike "*/drops/*"} | ?{$_ -notlike "*/bin/*"} `
+    | %{clang-format -n -style=file $_} 2>format.log
```
cgranade, line 16: A couple suggestions here:
- short-hand such as `?` and `%` is generally not advised in production scripts, since that shorthand can be overridden.
- can refactor out the join with `$DirPath` into a new `ForEach-Object` and use `Join-Path` for more robust join handling cross-platform
- can use `` ` `` to wrap lines for readability

```suggestion
"*.cpp", "*.c", "*.h", "*.hpp" `
    | ForEach-Object { Join-Path $DirPath $_ } `
    | Get-ChildItem -Recurse `
    <# clang-format accepts path names with Unix-style separators. #> `
    | ForEach-Object {$_.FullName -replace "\\", "/"} `
    | Where-Object { $_ -notlike "*/Externals/*" } `
    | Where-Object { $_ -notlike "*/drops/*" } `
    | Where-Object { $_ -notlike "*/bin/*" } `
    | ForEach-Object {
        clang-format -n -style=file $_
    } 2>format.log
```

src/Qir/check-sources-formatted.ps1
```diff
@@ -0,0 +1,31 @@
+#Requires -PSEdition Core
+#Requires -Version 7.1
+
+param (
+    [Parameter()]
+    [String]
+    $DirPath = "$PSScriptRoot"
```
cgranade, line 7: `$Path` is a pretty common name for this parameter in most other PowerShell cmdlets (e.g.: `Get-ChildItem`, `Get-Content`, and similar).
```suggestion
    $Path = "$PSScriptRoot"
```

src/Qir/check-sources-formatted.ps1
cgranade: It also looks like `$tmpFile` isn't used here, but rather `format.log` is hard-coded?

src/Qir/check-sources-formatted.ps1
```diff
@@ -0,0 +1,31 @@
...
+)
+
+$tmpFile = "format.log"
```
cgranade, line 10: You can explicitly get a temporary file name with ` [System.IO.Path]::GetTempFileName()`.
```suggestion
$tmpFile = [System.IO.Path]::GetTempFileName();
```

src/Qir/check-sources-formatted.ps1
cgranade: As an alternative:
```suggestion
if ("$filesRequireFormatting" -ne "")
```

src/Qir/check-sources-formatted.ps1
```diff
@@ -0,0 +1,31 @@
...
+
+$filesRequireFormatting = get-content $tmpFile | ?{$_ -like "*: warning:*"} `
+                            | %{[string]::join(":",($_.split("warning:")[0].split(":") | select -SkipLast 3))} `
+                            | sort | unique
+Remove-Item $tmpFile
+
+if (! ("$filesRequireFormatting" -eq ""))
+{
+    Write-Host "##vso[task.logissue type=error;]Formatting check failed. The following files need to be formatted before compiling: "
+    Write-Host "(You may use Clang-Format extension in VSCode or clang-format in command line or see https://clang.llvm.org/docs/ClangFormat.html)"
```
cgranade, line 27:
```suggestion
    Write-Host "(You may use the Clang-Format extension in VSCode, clang-format in command line, or see https://clang.llvm.org/docs/ClangFormat.html)"
```

src/Qir/check-sources-formatted.ps1
```diff
@@ -0,0 +1,31 @@
+#Requires -PSEdition Core
```
cgranade, line 1: Copyright header.

src/Qir/Tests/build-qir-tests.ps1
```diff
@@ -8,9 +8,12 @@ param (
     $SkipQSharpBuild
 )
 
+Write-Host "##[info]Compile Q# Test Projects into QIR"
+
 . (Join-Path $PSScriptRoot .. qir-utils.ps1)
 
-Write-Host "##[info]Compile Q# Projects into QIR"
+& "$PSScriptRoot/../check-sources-formatted.ps1" $PSScriptRoot
```
cgranade, line 15:
```suggestion
& (Join-Path $PSScriptRoot ".." check-sources-formatted.ps1) -Path $PSScriptRoot
```

src/Qir/check-sources-formatted.ps1
kuzminrobin: Just curiosity:
May I know the advantages of `-not` over `!`?
In [about_Logical_Operators](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_logical_operators?view=powershell-7.1) I don't see any mentioning of advantages of one of them over the other of them.
In our code base I saw that both are used.

cgranade: My own 2¢, but consistency, I guess? `-eq`, `-ne`, `-gt`, and `-lt` are all formatted as flags (similar to how bash uses flags to the `[` and `[[` commands to represent equality tests), such that `!` feels a bit out of place to me next to flag-style operators. In this case, though, changing from `!`/`-eq` to `-ne` also allows for removing a level of parentheses and making the code slightly more readable as a result.

src/Qir/check-sources-formatted.ps1
kuzminrobin: Thank you, that's good to know.
But for me locally the `[System.IO.Path]::GetTempFileName()` returns `C:\Users\rokuzmin\AppData\Local\Temp\tmp3C56.tmp`
(a path in a different directory, and an unpredictable file name). 

For debugging and failure analysis purposes I'd prefer the file to have a predictable/known path and file name.
I can change the filename to a more unique name (e.g. add a const GUID to the name).

kuzminrobin:
Please review.

src/Qir/Runtime/build-qir-runtime.ps1
```diff
@@ -3,6 +3,11 @@
 
 . (Join-Path $PSScriptRoot .. qir-utils.ps1)
 
+Write-Host "##[info]Compile QIR Runtime"
+
+& (Join-Path $PSScriptRoot ".." check-sources-formatted.ps1) -Path $PSScriptRoot
+& (Join-Path $PSScriptRoot ".." check-sources-formatted.ps1) -Path $PSScriptRoot/../Common
```
cgranade, line 9: [xplat safety nit; please feel free to ignore]

```suggestion
& (Join-Path $PSScriptRoot ".." check-sources-formatted.ps1) -Path (Join-Path $PSScriptRoot ".." "Common")
```

src/Qir/Runtime/prerequisites.ps1
```diff
@@ -3,21 +3,27 @@
 
 if ($Env:ENABLE_QIRRUNTIME -ne "false") {
     if (($IsWindows) -or ((Test-Path Env:AGENT_OS) -and ($Env:AGENT_OS.StartsWith("Win")))) {
```
cgranade, line 5: Minor, but on earlier versions of PowerShell, `$IsWindows` isn't defined. Can either use `#Requires -Version 6.0` or check `$PSVersionTable.PSEdition -eq "Desktop"` to guard against that.

cgranade (commented):
Looks good to me, but I'll leave it to code owners to approve.

build/build.ps1
```diff
@@ -6,24 +6,24 @@ $ErrorActionPreference = 'Stop'
 & "$PSScriptRoot/set-env.ps1"
 $all_ok = $True
 
-if ($Env:ENABLE_NATIVE -ne "false") {
-    $nativeSimulator = (Join-Path $PSScriptRoot "../src/Simulation/Native")
-    & "$nativeSimulator/build-native-simulator.ps1"
+if ($Env:ENABLE_QIRRUNTIME -ne "false") {
```
cesarzc, line 9: I am curious, why is QIR runtime built before the native simulator now?

build/set-env.ps1
```diff
@@ -64,4 +64,7 @@ Get-ChildItem @(
     "Env:\CRATE_OUTDIR",
     "Env:\WHEEL_OUTDIR",
     "Env:\DOCS_OUTDIR"
- ) | Format-Table
\ No newline at end of file
+ ) | Format-Table
+
+Write-Host "PATH:"
```
cesarzc, line 69: Is this for diagnostics during builds?

build/build.ps1
kuzminrobin: We update QIR RT more frequently, and it is more likely to fail the build (than the native simulator). So if the build fails then it is more efficient to fail earlier than later. Some more info is in [Build the Most Failing Files First](https://github.com/kuzminrobin/code_review_notes/blob/master/misc_posts.md#build-the-most-failing-files-first).

build/set-env.ps1
kuzminrobin: Yes, for diagnostic and debugging purposes, to analyze the build failure reasons.

src/Qir/Runtime/unittests/CMakeLists.txt
```diff
@@ -17,7 +17,7 @@ add_executable(qir-runtime-unittests
 )
 
 target_include_directories(qir-runtime-unittests PUBLIC 
-  "${PROJECT_SOURCE_DIR}/../Common/externals/catch2"
+  "${PROJECT_SOURCE_DIR}/../Common/Externals/catch2"
```
swernli, line 20: Changing the casing of the folder like this will break anyone who is using Linux devcontainers, forcing them to manually rename the folder, which may not be obvious. Why is this change being made now? The paths in the CMakeLists.txt and paths in the repo agreed, so there was no bug here before. What motivated the change?

src/Qir/check-sources-formatted.ps1
```diff
@@ -0,0 +1,45 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+#Requires -PSEdition Core
+#Requires -Version 7.1
```
swernli, line 5: These two are redundant, such that only either `#Requires -PSEdition Core` or `#Requires -Version 7.1` are needed, but not both. The version requirement is stricter, since any version after 6.0 counts as "Core", so I would suggest just keeping that one.

swernli (changes requested):
I would suggest not making the capitalization change to the externals folder, as the paths in the repo were already consistent with the name in the repo (all lowercase). If that was causing problems in your local environment, please fix your local environment rather than pushing the casing change into the repo, since that can break Linux devcontainers for everyone else.

src/Qir/Runtime/unittests/CMakeLists.txt
kuzminrobin: We have 2 folders, 
qsharp-runtime/src/Qir/Common/**e**xternals (starts with lower case `e`),
qsharp-runtime/src/Qir/Tools/**E**xternals (starts with upper cased `E`).
Both need to be excluded from clang-formatting. It is more efficient to exclude both with a [single line](https://github.com/microsoft/qsharp-runtime/blob/10c22457fc27dd05206f9746b7fbc1e33ef8b07d/src/Qir/check-sources-formatted.ps1#L23). But on Linux the paths are case-sensitive. So the single line does not exclude both dirs. 
To exclude both with a single line I unified those two directory names.

kuzminrobin: Can you please show what is about to be broken with this change?

swernli: Actually, never mind, I tried this locally and it was odd but didn't cause a problem or break anything. While the rename gets ignored on Windows, which can lead to some weird corner cases, in this case you avoid the break by updating everywhere at the same time. I've gotten overly suspicious of capitalization changes after getting burned by them before, because git on Windows ignores this rename (my local repo when merging in this branch keeps the lowercase "external" folder name) and that can lead to confusion in the future when it looks like paths checked into the repo are wrong. But for Linux devcontainers, the source code is shared from a Windows volume, meaning even then the tools will find the right folder because despite being case  sensitive the files are being served into Linux by the case insensitive Windows file system. Anyone in this mixed mode who thinks they need to update the file paths in the CMakeLists.txt will hit build failures, so that should protect other users, and folks who develop directly in Linux will get the right names from case sensitive git behavior.

Sorry for the confusion, I was wrong to block for this!

swernli (approved):

cesarzc (approved):

PR #795, opened 2021-08-12
Peform post M preparation in target package

swernli:
The Type1 target package requires post measurement preparation of the measured value in order to ensure the API guarantee of `M` is preserved. While this was happening correctly for calls to `Measure` the `M` callabe itself also needs the post measurement preparation. This introduces a new internal callable `MZ` in the target packages so that the post measurement preparation can be correctly handled.


src/Simulation/Simulators/QuantumSimulator/MZ.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
cgranade, line 1:
```suggestion
﻿// Copyright (c) Microsoft Corporation.
```

src/Simulation/TargetDefinitions/Decompositions/MWithPostPrep.qs
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
cgranade, line 1:
```suggestion
// Copyright (c) Microsoft Corporation.
```

src/Simulation/TargetDefinitions/Intrinsic/MZ.qs
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
cgranade, line 1:
```suggestion
// Copyright (c) Microsoft Corporation.
```

cgranade (approved):
Looks good to me! It may be helpful to include a note about what you raised, that this helps avoid an array allocation that would otherwise be implied by using the expansion of `M(q)` into `Measure([PauliZ], [q])` without necessitating an API change, perhaps?

PR #796, opened 2021-08-13
Check qubit release/measurement status on release

### LINKED ISSUE #552: QIR simulator should check result of qubit Release

swernli:
This mimicks the same behavior in the QIR Runtime wrapper for the fullstate simulator as what we have for C#, namely that a qubit is falid for release if and only if it is either in the ground state or the last operation on that qubit was measure.

This fixes #552, and supercedes #710.

### DIFF
--- a/src/Qir/Runtime/lib/Simulators/FullstateSimulator.cpp
+++ b/src/Qir/Runtime/lib/Simulators/FullstateSimulator.cpp
@@ -165,14 +165,14 @@ namespace Quantum
             return proc;
         }
 
-        void UnmarkSingleQubit(Qubit q)
+        void UnmarkAsMeasuredSingleQubit(Qubit q)
         {
             isMeasured[GetQubitId(q)] = false;
         }
 
-        void UnmarkQubitList(long num, Qubit* qubit)
+        void UnmarkAsMeasuredQubitList(long num, Qubit* qubit)
         {
-            for (const auto& id : GetQubitIds(num ,qubit))
+            for (const auto& id : GetQubitIds(num, qubit))
             {
                 isMeasured[id] = false;
             }
@@ -250,9 +250,12 @@ namespace Quantum
             auto id = GetQubitId(q);
             if (!releaseQubit(this->simulatorId, id) && !isMeasured[id])
             {
+                // We reject the release of a qubit that is not in the ground state (releaseQubit returns false),
+                // and was not recently measured (ie: the last operation was not measurement). This means the
+                // state is not well known, and therefore the safety of release is not guaranteed.
                 quantum__rt__fail_cstr("Released qubit neither measured nor in ground state.");
             }
-            qubitManager->Release(q);                       // Release it in the qubit manager.
+            qubitManager->Release(q); // Release it in the qubit manager.
         }
 
         Result Measure(long numBases, PauliId bases[], long numTargets, Qubit targets[]) override
@@ -300,7 +303,7 @@ namespace Quantum
         {
             static TSingleQubitGate op = reinterpret_cast<TSingleQubitGate>(this->GetProc("X"));
             op(this->simulatorId, GetQubitId(q));
-            UnmarkSingleQubit(q);
+            UnmarkAsMeasuredSingleQubit(q);
         }
 
         void ControlledX(long numControls, Qubit controls[], Qubit target) override
@@ -308,15 +311,15 @@ namespace Quantum
             static TSingleQubitControlledGate op = reinterpret_cast<TSingleQubitControlledGate>(this->GetProc("MCX"));
             std::vector<unsigned> ids            = GetQubitIds(numControls, controls);
             op(this->simulatorId, (unsigned)numControls, ids.data(), GetQubitId(target));
-            UnmarkSingleQubit(target);
-            UnmarkQubitList(numControls, controls);
+            UnmarkAsMeasuredSingleQubit(target);
+            UnmarkAsMeasuredQubitList(numControls, controls);
         }
 
         void Y(Qubit q) override
         {
             static TSingleQubitGate op = reinterpret_cast<TSingleQubitGate>(this->GetProc("Y"));
             op(this->simulatorId, GetQubitId(q));
-            UnmarkSingleQubit(q);
+            UnmarkAsMeasuredSingleQubit(q);
         }
 
         void ControlledY(long numControls, Qubit controls[], Qubit target) override
@@ -324,15 +327,15 @@ namespace Quantum
             static TSingleQubitControlledGate op = reinterpret_cast<TSingleQubitControlledGate>(this->GetProc("MCY"));
             std::vector<unsigned> ids            = GetQubitIds(numControls, controls);
             op(this->simulatorId, (unsigned)numControls, ids.data(), GetQubitId(target));
-            UnmarkSingleQubit(target);
-            UnmarkQubitList(numControls, controls);
+            UnmarkAsMeasuredSingleQubit(target);
+            UnmarkAsMeasuredQubitList(numControls, controls);
         }
 
         void Z(Qubit q) override
         {
             static TSingleQubitGate op = reinterpret_cast<TSingleQubitGate>(this->GetProc("Z"));
             op(this->simulatorId, GetQubitId(q));
-            UnmarkSingleQubit(q);
+            UnmarkAsMeasuredSingleQubit(q);
         }
 
         void ControlledZ(long numControls, Qubit controls[], Qubit target) override
@@ -340,15 +343,15 @@ namespace Quantum
             static TSingleQubitControlledGate op = reinterpret_cast<TSingleQubitControlledGate>(this->GetProc("MCZ"));
             std::vector<unsigned> ids            = GetQubitIds(numControls, controls);
             op(this->simulatorId, (unsigned)numControls, ids.data(), GetQubitId(target));
-            UnmarkSingleQubit(target);
-            UnmarkQubitList(numControls, controls);
+            UnmarkAsMeasuredSingleQubit(target);
+            UnmarkAsMeasuredQubitList(numControls, controls);
         }
 
         void H(Qubit q) override
         {
             static TSingleQubitGate op = reinterpret_cast<TSingleQubitGate>(this->GetProc("H"));
             op(this->simulatorId, GetQubitId(q));
-            UnmarkSingleQubit(q);
+            UnmarkAsMeasuredSingleQubit(q);
         }
 
         void ControlledH(long numControls, Qubit controls[], Qubit target) override
@@ -356,15 +359,15 @@ namespace Quantum
             static TSingleQubitControlledGate op = reinterpret_cast<TSingleQubitControlledGate>(this->GetProc("MCH"));
             std::vector<unsigned> ids            = GetQubitIds(numControls, controls);
             op(this->simulatorId, (unsigned)numControls, ids.data(), GetQubitId(target));
-            UnmarkSingleQubit(target);
-            UnmarkQubitList(numControls, controls);
+            UnmarkAsMeasuredSingleQubit(target);
+            UnmarkAsMeasuredQubitList(numControls, controls);
         }
 
         void S(Qubit q) override
         {
             static TSingleQubitGate op = reinterpret_cast<TSingleQubitGate>(this->GetProc("S"));
             op(this->simulatorId, GetQubitId(q));
-            UnmarkSingleQubit(q);
+            UnmarkAsMeasuredSingleQubit(q);
         }
 
         void ControlledS(long numControls, Qubit controls[], Qubit target) override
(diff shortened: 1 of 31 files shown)


src/Qir/Runtime/lib/Simulators/FullstateSimulator.cpp
```diff
@@ -272,71 +300,87 @@ namespace Quantum
...
             op(this->simulatorId, GetQubitId(q));
+            UnmarkSingleQubit(q);
         }
 
         void ControlledX(long numControls, Qubit controls[], Qubit target) override
         {
             static TSingleQubitControlledGate op = reinterpret_cast<TSingleQubitControlledGate>(this->GetProc("MCX"));
             std::vector<unsigned> ids            = GetQubitIds(numControls, controls);
             op(this->simulatorId, (unsigned)numControls, ids.data(), GetQubitId(target));
+            UnmarkSingleQubit(target);
```
kuzminrobin, line 311: The `UnmarkSingleQubit()` is a mechanism common for a number of gates. If we add a few more similar common mechanisms, then a number of gates will contain an identical sequence of calls to those mechanisms.
You may want to consider adding an extra member function containing common mechanisms for multiple gates. And all those gates will be calling that extra member function, and that extra member function will be calling `UnmarkSingleQubit()` and all other common mechanisms.

src/Qir/Runtime/lib/Simulators/FullstateSimulator.cpp
```diff
@@ -272,71 +300,87 @@ namespace Quantum
...
+            UnmarkQubitList(numControls, controls);
```
kuzminrobin, line 312: Are you sure that we need to mark the controls as not measured upon the `Controlled..()` operations?

My impression was that controls stay unchanged during the `Controlled..()` operations (they do not collapse). If they are already measured, then they should still be considered measured after the `Controlled..()` operation.

cgranade: In general, controlled operations can affect the state of all qubits that they act on, whether those are controls or targets. As an example, phase kickback is a technique (used in iterative phase estimation and in the Deutsch–Jozsa problem) where the target stays in a fixed state, but the phase imparted by an operation call is associated with a control register instead.

Another way to see it is that these two Q# programs are equivalent when executed on ideal devices:

```qsharp
within {
    H(left);
    H(right);
} apply {
    CNOT(left, right);
}
// and
CNOT(right, left);
```

As a result, when applying a controlled operation to a register whose state is known exactly, it is no longer the case that the state is known unless one spends possibly exponential overhead to track the state.

src/Qir/Tests/FullstateSimulator/FullstateSimulatorTests.cpp
```diff
@@ -145,6 +145,10 @@ TEST_CASE("Fullstate simulator: ZZ measure", "[fullstate_simulator]")
     Result rOne = iqa->Measure(2, paulis, 2, q);
     REQUIRE(Result_One == sim->GetResultValue(rOne));
 
+    iqa->X(q[1]);
+    iqa->ControlledX(1, &q[0], q[1]);
+    iqa->H(q[0]);
```
kuzminrobin, line 150: Looks like some more comments would not harm in this test. I'm not sure my understanding is correct about the added lines here. Here is how I interpret the logic:

After line 138 the state of the two allocated qubits is `|00>` (the qubit `q[0]` is on the left).
After 139 the `q[0]` is |+>, equally probable |0> or |1> in PauliZ basis. The state of two qubits is equally probable |00> or |10> (in PauliZ).
After 140 the state is equally probable |00> or |11> (Bell state) (in PauliZ).
After 141 the state collapses to either |00> or |11>.

After 144 the state is either |01> or |10>. 
After 145 the state stays as is (because the state is already collapsed).

After green line 148 the state is either |00> or |11>.
After green line 149 the state is either |00> or |10>.
After green line 150 the state of q[0] is either |+> or |->, which I interpret as equally probable |0> or |1> in PauliZ basis (non-collapsed, non-measured state).

The subsequent `sim->ReleaseQubit(q[0])` should fail.

src/Qir/Tests/FullstateSimulator/FullstateSimulatorTests.cpp
```diff
@@ -196,6 +202,9 @@ TEST_CASE("Fullstate simulator: toffoli", "[fullstate_simulator]")
     iqa->ControlledX(2, qs, qs[2]);
     REQUIRE(Result_One == sim->GetResultValue(MZ(iqa, qs[2])));
 
+    iqa->X(qs[1]);
+    iqa->X(qs[0]);
+
```
kuzminrobin, line 207: (I'm likely mistaken but to stay on the same page with you guys, just in case, I'll tell about how I interpret this test. Please point for me where exactly I'm wrong)
The line numbers below are from the right column.

After 195 the state is |000> (`qs[0]` is on the left).
.. 197: |100>.
198: |100>.
199: |100>.

201: |110>.
202: |111>.
203: |111>.

205: |101>.
206: |001>.

The subsequent `sim->ReleaseQubit()` should fail for qubit qs[2].

src/Qir/Runtime/lib/Simulators/FullstateSimulator.cpp
```diff
@@ -238,6 +261,11 @@ namespace Quantum
...
             static TMeasure m         = reinterpret_cast<TMeasure>(this->GetProc("Measure"));
             std::vector<unsigned> ids = GetQubitIds(numTargets, targets);
+            if (ids.size() == 1)
+            {
+                // If measuring exactly one qubit, mark it as measured for tracking.
+                isMeasured[ids[0]] = true;
+            }
             return reinterpret_cast<Result>(
                 m(this->simulatorId, (unsigned)numBases, reinterpret_cast<unsigned*>(bases), ids.data()));
         }
```
kuzminrobin, line 271: I have an impression that only one qubit is marked as measured regardless of the number of qubits being measured.

kuzminrobin, line 271: I got an impression that the measurement does not have to be in PauliZ basis, to deallocate the qubit correctly. 

After the measurement in PauliX basis the qubit state collapses to one of the values in PauliX basis (which I interpret as equally probable |0> or |1> in PauliZ basis). And such a qubit can still be correctly deallocated (because it has been measured).

Feel free to correct me if I'm wrong.

cgranade: One classical bit is returned, but that could come from a joint measurement on an arbitrary number of qubits. E.g.: `Measure([PauliZ, PauliZ], [q0, q1])` collapses the state of `[q0, q1]` onto either the +1 or −1 eigenspace of 𝑍 ⊗ 𝑍, spanned by {|00⟩, |11⟩} and {|01⟩, |10⟩}, respectively.

cgranade: If the result of `Measure([PauliX], [q])` is `Zero`, then the state of `q` is known to be |+⟩ ≔ (|0⟩ + |1⟩) / √2, while if the result is `One`, then `q` must be in |−⟩ ≔ (|0⟩ − |1⟩) / √2. Either way, the state of `q` is known precisely and thus can be efficiently reset to the |0⟩ state without additional measurements (in particular, by `H(q);` in the first case, and by `Z(q); H(q);` in the second). A similar argument applies when measuring in the `PauliY` basis.

src/Qir/Runtime/lib/Simulators/FullstateSimulator.cpp
```diff
@@ -164,6 +165,19 @@ namespace Quantum
             return proc;
         }
 
+        void UnmarkSingleQubit(Qubit q)
```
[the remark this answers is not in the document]
swernli: It is private, but I might as well update the name...

src/Qir/Runtime/lib/Simulators/FullstateSimulator.cpp
```diff
@@ -238,6 +261,11 @@ namespace Quantum
             typedef unsigned (*TMeasure)(unsigned, unsigned, unsigned*, unsigned*);
             static TMeasure m         = reinterpret_cast<TMeasure>(this->GetProc("Measure"));
             std::vector<unsigned> ids = GetQubitIds(numTargets, targets);
+            if (ids.size() == 1)
+            {
+                // If measuring exactly one qubit, mark it as measured for tracking.
```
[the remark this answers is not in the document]
swernli: That's a good point, and interesting idea for future improvement. That said, it don't think it falls in the scope of this change, as we'd have to update the C# simulation wrapper to do the same tracking. Maybe file a follow up issue for us to track this idea?

src/Qir/Runtime/lib/Simulators/FullstateSimulator.cpp
swernli: If multiple qubits are measured it doesn't guarantee that the individual qubit state is known. Only when the length of `ids` is 1 (ie: when only a single qubit is being measured and not a joint measurement) do we want to mark that qubit as measured and safe to release.

src/Qir/Runtime/lib/Simulators/FullstateSimulator.cpp
swernli: `UnmarkSingleQubit` and `UnmarkQubitList` are the member functions that encapsulate common functionality. Right now there is no other anticipated or planned common functionality to abstract out into a new function, so the new method would be only a call to the unmark functions, which would be redundant abstraction.

src/Qir/Runtime/lib/Simulators/FullstateSimulator.cpp
swernli: Deja vu ;) It wasn't so long ago that you were explaining this to me (https://github.com/microsoft/qsharp-runtime/pull/231#issuecomment-639585799)

src/Qir/Tests/FullstateSimulator/FullstateSimulatorTests.cpp
swernli: These lines are performing the adjoint of the operations above, such that the qubits are returned to the ground state, ie: |00⟩. Now that the QIR simulator wrapper checks for this condition the tests would fail with an exception; these same tests would already have failed that way if performed in the C# simulator wrapper.

src/Qir/Tests/FullstateSimulator/FullstateSimulatorTests.cpp
swernli: You have the right register state, but note also that the last operation on `qs[2]` was a measurement, meaning it is allowed to be released because it is in a well-known state.

src/Simulation/Native/src/simulator/capi.cpp
```diff
@@ -69,9 +69,9 @@ extern "C"
         Microsoft::Quantum::Simulator::get(id)->allocateQubit(q);
     }
 
-    MICROSOFT_QUANTUM_DECL void release(_In_ unsigned id, _In_ unsigned q)
+    MICROSOFT_QUANTUM_DECL bool release(_In_ unsigned id, _In_ unsigned q)
```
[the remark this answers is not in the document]
swernli: Sure, I'll add that.

src/Simulation/Native/src/simulator/simulator.hpp
```diff
@@ -32,15 +32,15 @@ class Simulator : public Microsoft::Quantum::Simulator::SimulatorInterface
 
     std::size_t random(std::vector<double> const& d)
     {
-        recursive_lock_type l(mutex());
+        recursive_lock_type l(getmutex());
```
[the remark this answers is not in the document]
swernli: At this point, for clarity, I might was separate this out into it's own PR.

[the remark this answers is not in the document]
swernli: https://github.com/microsoft/qsharp-runtime/pull/798

cgranade (commented):
Some of these comments seemed to get stuck in a pending review, sorry. Probably out of date now...

src/Qir/Runtime/lib/Simulators/FullstateSimulator.cpp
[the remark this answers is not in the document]
cgranade: Absolutely out of the scope of this PR, agreed. Was purely a tangential thought, sorry.

src/Qir/Tests/FullstateSimulator/FullstateSimulatorTests.cpp
kuzminrobin: Thank you very much for explaining. 
Looks like I'm the only one who needs comments in this code ;-)

src/Qir/Tests/FullstateSimulator/FullstateSimulatorTests.cpp
kuzminrobin: Thank you for explaining.

src/Qir/Runtime/lib/Simulators/FullstateSimulator.cpp
kuzminrobin: Thanks.

kuzminrobin (approved):

PR #804, opened 2021-08-23
Changed QUBIT* to Ids

DmitryVasilevsky:
Tests are failing at this time.

### DIFF
--- a/src/Qir/Common/Include/SimulatorStub.hpp
+++ b/src/Qir/Common/Include/SimulatorStub.hpp
@@ -13,101 +13,101 @@ namespace Quantum
         : public IRuntimeDriver
         , public IQuantumGateSet
     {
-        Qubit AllocateQubit() override
+        qubitid_t AllocateQubit() override
         {
             throw std::logic_error("not_implemented: AllocateQubit");
         }
-        void ReleaseQubit(Qubit /* qubit */) override
+        void ReleaseQubit(qubitid_t /* qubit */) override
         {
             throw std::logic_error("not_implemented: ReleaseQubit");
         }
-        virtual std::string QubitToString(Qubit /* qubit */) override
+        virtual std::string QubitToString(qubitid_t /* qubit */) override
         {
             throw std::logic_error("not_implemented: QubitToString");
         }
-        void X(Qubit /*target*/) override
+        void X(qubitid_t /*target*/) override
         {
             throw std::logic_error("not_implemented: X");
         }
-        void Y(Qubit /*target*/) override
+        void Y(qubitid_t /*target*/) override
         {
             throw std::logic_error("not_implemented: Y");
         }
-        void Z(Qubit /* target */) override
+        void Z(qubitid_t /* target */) override
         {
             throw std::logic_error("not_implemented: Z");
         }
-        void H(Qubit /*target*/) override
+        void H(qubitid_t /*target*/) override
         {
             throw std::logic_error("not_implemented: H");
         }
-        void S(Qubit /*target*/) override
+        void S(qubitid_t /*target*/) override
         {
             throw std::logic_error("not_implemented: S");
         }
-        void T(Qubit /*target*/) override
+        void T(qubitid_t /*target*/) override
         {
             throw std::logic_error("not_implemented: T");
         }
-        void R(PauliId /* axis */, Qubit /* target */, double /* theta */) override
+        void R(PauliId /* axis */, qubitid_t /* target */, double /* theta */) override
         {
             throw std::logic_error("not_implemented: R");
         }
-        void Exp(long /* numTargets */, PauliId* /* paulis */, Qubit* /* targets */, double /* theta */) override
+        void Exp(long /* numTargets */, PauliId* /* paulis */, qubitid_t* /* targets */, double /* theta */) override
         {
             throw std::logic_error("not_implemented: Exp");
         }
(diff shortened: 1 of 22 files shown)


swernli:
Change looks good to me, but I'll wait to sign off until you are ready to publish!

kuzminrobin:
I believe we need a more systematic approach here. 

I'd propose to come up to a common vision first, in the form of the .md document or similar (not necessarily public to the open-source community). And then to start making concrete steps, in the form of the PRs, towards the chosen direction.

In that document, to tease the appetite, I would provide the _summary of all the identifiers_ (types) that are currently used to handle qubits (starting with .ll bridge files, then running through the C/C++ files of QIR RT, diving into the FullStateSimulator and QubitManager, and ending in the Native Simulator). E.g. currently I see that in .ll files there are identifiers [`%class.QUBIT`](https://github.com/microsoft/qsharp-runtime/blob/7bf27f0b3b1a1752c5789c9ad4c8f454b8c4e046/src/Qir/Runtime/lib/QIR/bridge-rt.ll#L24) and [`%Qubit`](https://github.com/microsoft/qsharp-runtime/blob/7bf27f0b3b1a1752c5789c9ad4c8f454b8c4e046/src/Qir/Runtime/lib/QIR/bridge-rt.ll#L9) that relate to each other _quite differently_ as opposed to the identifiers [`QUBIT` and `Qubit`](https://github.com/microsoft/qsharp-runtime/blob/7bf27f0b3b1a1752c5789c9ad4c8f454b8c4e046/src/Qir/Runtime/public/CoreTypes.hpp#L35-L36) used in C/C++ files. In particular, in C/C++ `Qubit` is a pointer to `QUBIT`, but in .ll `%Qubit` is _not_ a pointer to `%class.QUBIT`. In .ll files `%Qubit` and `%class.QUBIT` are sort of the same thing, pointers to which are freely cast [there](https://github.com/microsoft/qsharp-runtime/blob/7bf27f0b3b1a1752c5789c9ad4c8f454b8c4e046/src/Qir/Runtime/lib/QIR/bridge-rt.ll#L153) and [back](https://github.com/microsoft/qsharp-runtime/blob/7bf27f0b3b1a1752c5789c9ad4c8f454b8c4e046/src/Qir/Runtime/lib/QIR/bridge-rt.ll#L158).
Other curious observations can be found between the types used in Native Simulator and FullStateSimulator.

The summary table or description of all those identifiers, their use, casts between them will clearly illustrate the problem we are trying to solve. The variety of those identifiers will show that some of them can be merged together. And, subjectively speaking, I believe a single identifier (type alias) is sufficient to effectively cover all the cases and eliminate all the qubit type casts in QIR RT. It is OK for the casts to survive inside of the vendor libraries or simulator libraries dealing with the actual qubit implementation. 

Having illustrated the problem with this qubit types variety, the document can propose and describe a brief preliminary idea of a more effective solution. Then the document can list all the restrictions/limitations that we have (e.g. during a recent meeting Allan Geller mentioned that the original idea was to use the _unique type_ as a type for handling the qubits. I don't remember the grounds behind that idea, would be nice to have a list of all the restrictions/limitations in one place). Then the document can develop the brief preliminary idea into the final idea, moving along the path within the restrictions/limitations.

Considering and discussing such a document in our team and with other interested parties, we can come up with a clear common final idea, after which we will be able to move towards that idea in a shortest way. That can be a better moment for the PRs.

The mentioned document explaining the decision and limitations in details, will hopefully answer all the future questions. Since QIR RT is going public, I expect lots of questions from the community. It is good to have all the answers in one file that we can refer to.

swernli:
@kuzminrobin I agree a doc would be helpful here, but there are some fixes and simplifications that can be made even before this doc is compiled. Indeed, I think this would make the doc easier to compile, because it would eliminate a bunch of the types you mention. For example, all the `%class.*` and `%"struct.*"` types listed in the .ll files are preceded by a comment that explains their purpose:
https://github.com/microsoft/qsharp-runtime/blob/7bf27f0b3b1a1752c5789c9ad4c8f454b8c4e046/src/Qir/Runtime/lib/QIR/bridge-rt.ll#L17-L22
I still think we should endeavor to remove the need for these .ll files entirely, in which case we would be relying on having the bridge's C-functions declared with QIR types and using call to perform an implicit bitcast instead of an explicit one. Once we make that change, all of these types go away because the whole file goes away. I wouldn't want to waste time writing up a description on how the current bridge works when we know we want to change that as soon as possible anyway. As an example, take a look at https://github.com/microsoft/qsharp-runtime/commit/ef8985994cec647136b7ae5d36c388988109ebf8, where I started doing the work here but got stock on the few places where the type differences are actually functional: `%Range` being pass by value when C does not support that, and `%Pauli` being `i2` which is difficult to represent as unsigned in C. Both of these problems can be fixed with spec updates such that the right C code can be written.

kuzminrobin:
@swernli, totally makes sense to me. Agreed.

DmitryVasilevsky:
@kuzminrobin I agree that documentation is in order. I created this change to figure out what we really want and what it takes to make such change. The first step was to replace pointer with id in the runtime. And one of the findings is, of course, that different types are used for qubit ids in different parts of the runtime/simulators.  Now the problem is very apparent! Yes, it would be great if we can use the same type for this throughout runtime.

DmitryVasilevsky:
@kuzminrobin added a list of types to this file: https://github.com/microsoft/qsharp-runtime/blob/dmitryv/QubitIdsInsteadOfPointers/src/Qir/Runtime/public/README.md
Probably not exhaustive.

PR #811, opened 2021-09-05
Remove (most) .ll files from QIR Runtime implementation

swernli:
This change follows up on a conversation mentioned here: https://github.com/microsoft/qsharp-language/issues/103. It removes as much of the .ll files as possible, but leaves behind three functions in bridge-rt.ll due to the issue with `%Range` being passed by value (see https://github.com/microsoft/qsharp-language/issues/108). This also updates tests and adds types where needed to enable defining the QIR Runtime functions and quantum instruction functions directly in C++ as `extern "C"`.

As a possible follow up for this change, we could move some of the math test functions that live in the Q#-based tests in src\Qir\Tests\QIR-static\qsharp\qir-test-math.qs into true unit tests that live under src\Qir\Runtime\unittests.


src/Qir/Runtime/lib/QIR/bridge-rt.ll
```diff
@@ -1,284 +1,20 @@
...
-
-define dllexport %Array* @__quantum__rt__array_project(%Array* %.ar, i32 %dim, i64 %index) {
-  %ar = bitcast %Array* %.ar to %"struct.QirArray"*
-  %project = call %"struct.QirArray"* @quantum__rt__array_project(%"struct.QirArray"* %ar, i32 %dim, i64 %index)
-  %.project = bitcast  %"struct.QirArray"* %project to %Array*
-  ret %Array* %.project
-}
+; NOTE: These three functions can be converted to extern C once the spec and compiler are updated to pass %Range by
+; pointer instead of by value (see https://github.com/microsoft/qsharp-language/issues/108). Once that
+; happens, this file can be removed.
```
swernli, line 17: Drawing extra attention to this comment. The only .ll functions left are used to handle `%Range` being passed by value. If we fix the spec and the corresponding code generation in the compiler, then these functions change to something like `define dllexport %String* @__quantum__rt__range_to_string(%Range* %.range)` and can be implemented directly in C++ like the rest of the runtime.

src/Qir/Runtime/public/QirRuntime.hpp
```diff
@@ -16,140 +16,141 @@ extern "C"
     // ------------------------------------------------------------------------
 
     // Allocate one qubit. Qubit is guaranteed to be in |0> state.
-    // Qubit needs to be released via quantum__rt__qubit_release.
-    QIR_SHARED_API QUBIT* quantum__rt__qubit_allocate(); // NOLINT
+    // Qubit needs to be released via __quantum__rt__qubit_release.
+    QIR_SHARED_API QUBIT* __quantum__rt__qubit_allocate(); // NOLINT
```
kuzminrobin, line 20: I believe there can be a more optimal sequence of changes.

Let's consider by example of qubit allocation.
Before this PR we had 2 functions:
* `quantum__rt__qubit_allocate()` implemented in C/C++, I will call it _C implementation_ below;
* `__quantum__rt__qubit_allocate()` implemented in IR, I will call it _IR implementation_ below.

Both are exposed externally. The IR implementation calls the C implementation. 

The Q# compiler generates the IR files that call the IR implementation.
The other language environments (like Python, Rust) capable of using C interface, can call the C implementation directly.
Eventually we want to get rid of the IR implementation and get the Q# compiler to generate the IR that calls the C implementation directly too.

To achieve that, we can make a switch in the Q# compiler from calling the IR implementation to calling the C implementation _first_. And then (when nobody will be calling the IR implementation any more) just to remove the IR implementation. These two steps are done in different repos, that is why they are easy to do. I feel that this is the most optimal sequence of changes.

Now about the changes in this PR.
This PR removes the IR implementation and gives the C implementation the name starting with double underscore `__`. Q#-compiler-generated code still calls the `__` names (`__quantum__rt__qubit_allocate()`). I assume that _this is an intermediate step_, and the next step will be to make a change in two repos in parallel:
* in the Q# compiler to call `quantum__rt__qubit_allocate()` instead of `__quantum__rt__qubit_allocate()`,
* in the qsharp-runtime to rename `__quantum__rt__qubit_allocate()` to `quantum__rt__qubit_allocate()`.

If that's the case (this is an intermediate step) then this approach seems to me less optimal because of the several reasons:
* the identifiers starting with `__` (reserved in the C/C++ standard) are present in the C/C++ code, which is undesirable, at least some of the community may not like it (and the other community members' compilers or static analyzers will complain);
* the next step (parallel renaming in two repos) can be postponed for an indefinite period of time, and we can go public with `__` names;
* the parallel change in two repos is somewhat tricky.

If that's NOT the case (this is NOT an intermediate step) then the first of the reasons (some of the community may not like it) will stay (until the parallel change in the two repos).

src/Qir/Runtime/lib/QSharpCore/qsharp__core__qis.hpp
```diff
@@ -7,6 +7,20 @@
...
+    double angle;
+    QUBIT* target;
+};
+
+struct QirExpTuple // NOLINT
+{
+    QirArray* paulis;
+    double angle;
+    QirArray* targets;
+};
```
kuzminrobin, line 22: Curious: Are you sure the name `QirExpTuple` requires `// NOLINT`? (if yes then I'm surprised why it does, but the names  `QirRTuple` and `QirArray` do not require `// NOLINT`)

src/Qir/Runtime/lib/QSharpCore/intrinsics.cpp
```diff
@@ -37,46 +37,51 @@ static Microsoft::Quantum::IQuantumGateSet* GateSet()
...
+    void __quantum__qis__exp__ctladj(QirArray* ctls, QirExpTuple* args)
     {
-        quantum__qis__exp__ctl(ctls, paulis, -angle, qubits);
+        assert(args->paulis->count == args->targets->count);
+
+        std::vector<PauliId> pauliIds = ExtractPauliIds(args->paulis);
+        GateSet()->ControlledExp((long)(ctls->count), reinterpret_cast<Qubit*>(ctls->buffer),
+                                 (long)(args->paulis->count), reinterpret_cast<PauliId*>(pauliIds.data()),
+                                 reinterpret_cast<Qubit*>(args->targets->buffer), -(args->angle));
     }
```
kuzminrobin, line 72: Nit: Consider creating a local copy of `QirExpTuple`, negating the angle of it, and passing it to `quantum__qis__exp__ctl()`.
Otherwise currently `__quantum__qis__exp__ctladj()` nearly duplicates the whole `quantum__qis__exp__ctl()`.

src/Qir/Runtime/lib/QSharpCore/qsharp__core__qis.hpp
kuzminrobin, line 22: I assume all these structs need to be in the "extern C" section.

src/Qir/Runtime/lib/QSharpFoundation/qsharp__foundation__qis.hpp
```diff
@@ -11,6 +11,16 @@
...
 
+struct QirAssertMeasurementProbabilityTuple
+{
+    QirArray* bases;
+    QirArray* qubits;
+    RESULT* result;
+    double prob;
+    QirString* msg;
+    double tol;
+};
```
kuzminrobin, line 22: (Nit: In "extern C" section, if that is applicable to structs)

src/Qir/Runtime/public/QirTypes.hpp
```diff
@@ -192,11 +192,12 @@ struct QIR_SHARED_API QirCallable
...
-    int64_t end   = 0;
-
-    QirRange(int64_t start, int64_t step, int64_t end);
-};
+    struct QirRange
+    {
+        int64_t start;
+        int64_t step;
+        int64_t end;
+    };
```
kuzminrobin, line 202: `QIR_SHARED_API` is not needed any more?

src/Simulation/QSharpFoundation/Diagnostics/Assert.qs
```diff
@@ -86,6 +86,7 @@ namespace Microsoft.Quantum.Diagnostics {
     operation AssertMeasurementProbability(bases : Pauli[], qubits : Qubit[], result : Result, prob : Double, msg : String, tolerance : Double) : Unit
     is Adj + Ctl {
         body intrinsic;
+        adjoint self;
     }
```
kuzminrobin, line 90: Doesn't fragment `adjoint self` duplicate the fragment `is Adj`?

kuzminrobin (approved):
Approved with some suggestions.

src/Qir/Runtime/public/QirRuntime.hpp
swernli: I think you make some great points here, and it's worth having more discussion about which QIR spec updates we make and how those get reflected into the compiler and the runtime. But I think the question of how long to keep these .ll files and how we would remove them is tangential to all of this.  Here's my two cents...

> The Q# compiler generates the IR files that call the IR implementation.
The other language environments (like Python, Rust) capable of using C interface, can call the C implementation directly.
Eventually we want to get rid of the IR implementation and get the Q# compiler to generate the IR that calls the C implementation directly too.

I don't think this part is correct. We don't have a goal to write in other languages something that calls into the C implementation directly, we have a goal to robustly implement the QIR spec so that any QIR program can be compiled against our runtime regardless of where it was generated from. The QIR spec defines an API, and the Q# compiler produces QIR that uses that API. Then any runtime that wants to support QIR implements that API. IR vs C is an implementation detail of our particular runtime and not something that affects any other runtime someone would want to use. And if someone uses a different front end language, like Python or Rust, to write a quantum program they want to compile into QIR then they will need to use whatever API is defined in the QIR spec, not what our runtime happens to implement in C.

Let's take the same runtime functions you highlighted for qubit allocation:

> `quantum__rt__qubit_allocate()` implemented in C/C++, I will call it C implementation below;
`__quantum__rt__qubit_allocate()` implemented in IR, I will call it IR implementation below.

The second function, `__quantum__rt__qubit_allocate()` matches the API defined in the [QIR spec](https://github.com/microsoft/qsharp-language/blob/main/Specifications/QIR/Quantum-Runtime.md#qubits), such that any QIR program that is compliant with the spec can include calls to this function and would expect it to return a valid qubit. The first function, `quantum__rt__qubit_allocate()` doesn't match the name defined in the spec, so would not be considered valid QIR. It's just an implementation detail of our runtime, which has the second function forward directly to the first:
https://github.com/microsoft/qsharp-runtime/blob/d32e2890688a04cb4bca4436deeefba314bc540a/src/Qir/Runtime/lib/QIR/bridge-rt.ll#L151-L155
Except in the special case of `%Range` none of these forwards provide any real value. The bridge-rt.ll file is not distributed separately such that others can build on it, so it is not an extensibility point. The fact that the C implementation functions are extern is a necessity in order to be able to refer to them in the .ll files but not intended to make them available for use outside of our runtime. So removing this extra layer and defining the C implementation to match exactly the signatures from the QIR spec means we are more directly creating an implementation of QIR API.

I agree with you that we should update the QIR spec to make friendlier names, but I'm not so sure anymore about how to treat underscores specifically. Looking at the [cppreference page on identifiers](https://en.cppreference.com/w/cpp/language/identifiers), it actually states that "identifiers with a double underscore anywhere are reserved" so it's not just the leading underscores that are a problem. But that is a separate discussion that we should pursue along with cleaning up our runtime implementation. I would assert that whatever names and signatures the spec defines we should try to implement as extern C functions, and if that is not possible (like in the case of  passing `%Range` by value), then we should implement it as best we can and file issues against the spec for update. And once the spec does get updated, only then would we change the API surface implemented by the runtime.

You are right that such a spec update would be tricky and require multi-repo changes. I think the pattern should be
1. Update the QIR spec in the language repo.
2. Provide updated API implementation in the runtime repo alongside of the old one.
3. Update the Q# compiler to produce QIR compliant with the new spec and implementation.
4. Remove the old API implementation in the runtime repo when we are confident it should no longer be used.

This change doesn't affect that plan because it doesn't change the QIR API surface we implement, it just implements it in C instead of in IR forwarders.

All that being said, there is no rush to get this change in, especially because it only affects internal implementation details. We can take the time to make sure the team agrees with this and/or adjust the approach before merging it.

src/Qir/Runtime/lib/QSharpCore/qsharp__core__qis.hpp
swernli: nope, this is definitely an oversight leftover from a previous name I was experimenting with. You are right that it can just be removed.

src/Qir/Runtime/lib/QSharpCore/intrinsics.cpp
swernli: Ah, that's a good point. I had started considering that approach and then forgot to follow through after using this copy-paste just to verify behavior. I'll make that change so duplication is reduced.

src/Simulation/QSharpFoundation/Diagnostics/Assert.qs
swernli: Actually, the default directive for the adjoint specialization is `adjoint invert` which will reverse the order of instructions in an operation and uses their adjoint. So if you have an operation defined like this:
```qsharp
operation Foo(target : Qubit) : Unit is Adj {
    A(target);
    B(target);
}
```
That is the equivalent of:
```qsharp
operation Foo(target : Qubit) : Unit is Adj {
    body (...) { 
        A(target);
        B(target);
    }
    adjoint (...) {
        Adjoint B(target);
        Adjoint A(target);
    }
}
```
Whereas `adjoint self` indicates that the body specialization can be used as the adjoint and allows the Q# compiler to skip generating an adjoint at all. So it effectively makes it such that:
```qsharp
operation Foo(target : Qubit) : Unit is Adj {
    body (...) {
        A(target);
        B(target);
    }
    adjoint self;
}
```
is equivalent to:
```qsharp
operation Foo(target : Qubit) : Unit is Adj {
    body (...) {
        A(target);
        B(target);
    }
    adjoint (...) {
        Foo(target);
    }
}
```

For this specific case, declaring `AssertMeasurementProbability` as `adjoint self` means that no adjoint specialization will be generated, each instance of `Adjoint AssertMeasurementProbability` will just forward to the body specialization, and we don't have to implement either the adjoint or controlled adjoint qis functions for it.

src/Qir/Runtime/lib/QSharpCore/qsharp__core__qis.hpp
swernli: I'm actually not sure on this one... on the one hand `QirArray` and `QirString` are not extern C because they are not expected to be used directly by C code or QIR but rather passed around as opaque pointers. In this case, these tuple structs are convenience definitions for what is ultimately a capture tuple, which in QIR will be just an LLVM struct. That seems to imply that it would be the responsibility of every implementer to define a struct that matches their implementation of these tuple contents (or dynamically interpret the memory in the function body). So I don't expect any extern C callers to reuse these and would instead use their own struct definitions or make use of the `__quantum__rt__tuple_create` and related runtime features. But I'm not really sure what the right behavior here is.

src/Qir/Runtime/lib/QSharpFoundation/qsharp__foundation__qis.hpp
swernli: See my comment above. I think either all these capture tuple structs should be extern C together or all should be C++ structs like `QirArray` but I'm not confident which approach makes sense. They aren't exactly like the opaque pointers, but they also aren't fully a part of the API. Maybe I should just forward declare them here and then move the definitions into the cpp file?

src/Qir/Runtime/public/QirTypes.hpp
swernli: This one is a little odd, but the short answer is yes, it no longer needs to be `QIR_SHARED_API`. In the QIR spec, `%Range` is just a struct and not a class. Nothing in QIR ever called the constructor function for `QirRange` because it was just created as a struct of three integers directly. The one thing the constructor did was verify that `step` was not zero at construction, but since the constructor isn't called in QIR that doesn't provide any protection. The real protection comes from the `CheckedRange` struct that gets used everyone we actually operate on ranges, specifically this check:
https://github.com/microsoft/qsharp-runtime/blob/d32e2890688a04cb4bca4436deeefba314bc540a/src/Qir/Runtime/lib/QIR/arrays.cpp#L456-L464
Once the constructor is removed and it is made extern C, it can no longer be declared as dllexport, so the `QIR_SHARED_API` had to be removed.

src/Qir/Runtime/public/QirRuntime.hpp
kuzminrobin: Stefan, thank you very much for a detailed explanation! It has pointed to me that the importance of the QIR Spec has fallen out of my attention, which resulted in some misunderstanding on my side.
Now I clearly understand the reasons behind the changes in this PR and totally agree with each item that you provide in your explanation (including the fragment where you write "I don't think this part is correct" ;-) ).
Your 4 steps to the new API are the right thing to do, as opposed to what I previously considered as "more optimal sequence of changes".
I bookmarked this explanation in the description of the internal deliverable "Refactor QIR RT to expose C interface".

src/Qir/Runtime/lib/QSharpCore/qsharp__core__qis.hpp
kuzminrobin: I feel that eventually we will follow the item [Avoid Exposing Entities With Multiple Layouts](https://github.com/microsoft/qsharp-runtime/blob/main/src/Qir/QIR-RT-API-Design-Guidelines.md#avoid-exposing-entities-with-multiple-layouts). No need to follow that item in this PR (feel free to leave as is), but if you like you can try to make a first step in that direction and "test that item" :-).

src/Qir/Runtime/lib/QSharpFoundation/qsharp__foundation__qis.hpp
kuzminrobin: "Forward declaring them here and moving the definitions into the cpp file" seems to me a quite reasonable future alternative to the item [Avoid Exposing Entities With Multiple Layouts](https://github.com/microsoft/qsharp-runtime/blob/main/src/Qir/QIR-RT-API-Design-Guidelines.md#avoid-exposing-entities-with-multiple-layouts). However, if we expose externally just the forward declaration, then we will have to expose also the functions that 
* are `new`-like (allocate and initialize the instance),
* manipulate the instance (getters and setters),
* are `delete`-like (deinitialize and deallocate the instance)

(feel free to leave as is in this PR).

kuzminrobin:
Re-approved.

src/Qir/Runtime/lib/QSharpCore/qsharp__core__qis.hpp
swernli: That's a really good point, thanks! I think adjusting our use of tuples here (and in other places where argument tuples or capture tuples are used) to properly follow the guidance you put together for QIR APIs will be our best bet for success.

src/Qir/Runtime/lib/QSharpFoundation/qsharp__foundation__qis.hpp
swernli: You are totally right. I think we are in a tricky spot with the `%Tuple` type, because it straddles the boundary between struct and opaque type. The methods for creation, manipulation, and deletion (via reference counting), as described here: https://github.com/microsoft/qsharp-language/blob/main/Specifications/QIR/Data-Types.md#tuples-and-user-defined-types. However, unlike the opaque types we use for arrays, the returned memory is accessed directly by the client, making it "semi-opaque" which is odd. 

I think this might be another place where we need to consider filing feedback against the QIR spec and figuring out the right design to resolve this ambiguity. Let's sync up more on this sometime this week, as I could really use your insight on this one!

kuzminrobin: Sure, let's sync up.

PR #812, opened 2021-09-08
Add generateazurepayload command to standalone executables

cesarzc:
This change adds the generateazurepayload command to Q# standalone executables.


cesarzc:
This should be merged at the same time than this PR in qsharp-compiler: https://github.com/microsoft/qsharp-compiler/pull/1145

cesarzc:
This PR should be followed up by addressing the following issues:
- [Consider supporting generateazurepayload for all targets](https://github.com/microsoft/qsharp-runtime/issues/821)
- [Consider removing the --dry-run option fromthe submit command for standalone executables](https://github.com/microsoft/qsharp-runtime/issues/822)

src/Simulation/Core/Submitters/IQuantumMachine.cs
```diff
@@ -14,8 +14,13 @@ namespace Microsoft.Quantum.Runtime
...
     // TODO: deprecate once IQSharpSubmitter is fully hooked up.
     //[Obsolete("Replaced by IQSharpSubmitter.")]
-    public interface IQuantumMachine : IAzureSubmitter
+    public interface IQuantumMachine
     {
+        /// <summary>
+        /// The name of the execution target.
+        /// </summary>
+        string Target { get; }
+
```
kuzminrobin, line 23: I may be wrong but I have an impression that `IQSharpSubmitter`, `IQirSubmitter`, and `IQuantumMachine` can have a common base class/interface with contents:
```c#
        /// <summary>
        /// The name of the execution target.
        /// </summary>
        string Target { get; }
```

src/Azure/Azure.Quantum.Client/SubmitterFactory.cs
```diff
@@ -27,11 +27,25 @@ public static class SubmitterFactory
                 "Microsoft.Quantum.Providers.Targets.MicrosoftSimulatorSubmitter, Microsoft.Quantum.Providers.Core",
                 "QirSubmitter"));
 
+        private static readonly ImmutableList<SubmitterInfo> QirPayloadGenerators = ImmutableList.Create(
+            new SubmitterInfo(
+                new Regex(@"\Amicrosoft\.simulator\.([\w]+\.)*[\w]+\z"),
+                "Microsoft.Quantum.Providers.Targets.MicrosoftSimulatorSubmitter, Microsoft.Quantum.Providers.Core",
+                "QirPayloadGenerator"));
```
kuzminrobin, line 34: Just curious:
What is `\A` in the fragment `"\Amicrosoft`?

src/Simulation/EntryPointDriver.Tests/Tests.fs
```diff
@@ -989,6 +989,42 @@ let ``Uses documentation`` () =
...
+        |> sprintf "Usage:
+                      %s generateazurepayload [options]
+
+                    Options:
+                      -n <n> (REQUIRED)                                   A number.
+                      --pauli <PauliI|PauliX|PauliY|PauliZ> (REQUIRED)    The name of a Pauli matrix.
+                      --my-cool-bool (REQUIRED)                           A neat bit.
+                      --target <target>                                   The target device ID.
+                      --verbose                                           Show additional information about the submission.
+                      -?, -h, --help                                      Show help and usage information"
```
kuzminrobin, line 1024: I have an impression of the code duplication for the fragment 
```f#
    let name = Path.GetFileNameWithoutExtension (Assembly.GetEntryAssembly().Location)
    let message =
        name
        |> sprintf "Usage:
                      %s generateazurepayload [options]
                    Options:
                      -n <n> (REQUIRED)                                   A number.
                      --pauli <PauliI|PauliX|PauliY|PauliZ> (REQUIRED)    The name of a Pauli matrix.
                      --my-cool-bool (REQUIRED)                           A neat bit.
                      --target <target>                                   The target device ID.
                      --verbose                                           Show additional information about the submission.
                      -?, -h, --help                                      Show help and usage information"
```

src/Simulation/EntryPointDriver.Tests/Tests.fs
```diff
@@ -989,6 +989,42 @@ let ``Uses documentation`` () =
...
+    let name = Path.GetFileNameWithoutExtension (Assembly.GetEntryAssembly().Location)
+    let message =
+        name
+        |> sprintf "Usage:
+                      %s generateazurepayload [options]
+
+                    Options:
+                      -n <n> (REQUIRED)                                   A number.
+                      --pauli <PauliI|PauliX|PauliY|PauliZ> (REQUIRED)    The name of a Pauli matrix.
+                      --my-cool-bool (REQUIRED)                           A neat bit.
```
kuzminrobin, line 1003: The fragment `--my-cool-bool (REQUIRED)` makes an impression that the flag `--my-cool-bool` (without any value following it) always has to be specified in the command line. Why one needs a flag that has to be specified _always_?

src/Simulation/EntryPointDriver.Tests/Tests.fs
```diff
@@ -989,6 +989,42 @@ let ``Uses documentation`` () =
...
+[<Fact>]
+let ``Shows help text for generateazurepayload command`` () =
+    let name = Path.GetFileNameWithoutExtension (Assembly.GetEntryAssembly().Location)
+    let message =
+        name
+        |> sprintf "Usage:
+                      %s generateazurepayload [options]
+
+                    Options:
+                      -n <n> (REQUIRED)                                   A number.
```
kuzminrobin, line 1001: The fragment `A number` is not very informative. Some more info wouldn't harm about the purpose, range of values, type (floating point?)...

src/Simulation/EntryPointDriver/IEntryPoint.cs
```diff
@@ -33,6 +33,14 @@ public interface IEntryPoint
         /// </summary>
         IEnumerable<Option> Options { get; }
 
+        /// <summary>
+        /// Generates payload for Azure Quantum for the entry point.
+        /// </summary>
+        /// <param name="parseResult">The command-line parsing result.</param>
+        /// <param name="settings">The generate Azure payload settings.</param>
+        /// <returns>The exit code.</returns>
+        Task<int> GenerateAzurePayload(ParseResult parseResult, GenerateAzurePayloadSettings settings);
```
kuzminrobin, line 42: I'm curious, is it possible to avoid duplicating the param name in the doc comments? E.g. if one renames `parseResult` in the code,  but forgets to rename in the doc comments, then the doc comments will mislead. 

To summarize, is it possible to write like this?
```c#
        /// <summary>
        /// Generates payload for Azure Quantum for the entry point.
        /// </summary>
        /// <returns>The exit code.</returns>
        Task<int> GenerateAzurePayload(
                ParseResult parseResult, /// <param>The command-line parsing result.</param>
                GenerateAzurePayloadSettings settings); /// <param>The generate Azure payload settings.</param>
```

kuzminrobin, line 42: > /// <returns>The exit code.</returns>

The return type `Task<int>` does not look like exit code. If that's ok then the explanation "The exit code" is still not very informative. Some more details wouldn't harm (which values in which cases).

src/Simulation/EntryPointDriver/Mock/ErrorQuantumMachine.cs
```diff
@@ -15,14 +15,12 @@ namespace Microsoft.Quantum.EntryPointDriver.Mock
     /// </summary>
     internal class ErrorQuantumMachine : IQuantumMachine
     {
-        /// <summary>
+        // <summary>
         /// The target for the error quantum machine.
         /// </summary>
```
kuzminrobin, line 20: What is the reason for this change? Accidentally deleted character? 
The fragments 
`// <summary>`
and 
`/// </summary>`
do not match. 
The former is the ordinary comment (`//`), the latter is the doc comment (`///`).

kuzminrobin (approved):

src/Simulation/Core/Submitters/IQuantumMachine.cs
cesarzc: The plan is to deprecate `IQuantumMachine` and both `IQSharpSubmitter` and `IQirSubmitter` do share `Target` as a member but I think it would be overdoing it just creating an interface with one member in it.

src/Azure/Azure.Quantum.Client/SubmitterFactory.cs
cesarzc: It means that the match must occur at the start of the string. For example `microsoft.simulator.fullstate` would be a match while `Mmicrosoft.simulator.fullstate` would not.

src/Simulation/EntryPointDriver.Tests/Tests.fs
cesarzc: Since these are test cases, I believe code duplication is ok because it is easier to see what is the expected message for any particular test case.

src/Simulation/EntryPointDriver.Tests/Tests.fs
cesarzc: In this case, `--my-cool-bool` comes from the [Help](https://github.com/microsoft/qsharp-runtime/blob/f95b1d4c338234be35cdced88e8f9e6be98e9477/src/Simulation/EntryPointDriver.Tests/Tests.qs#L361) Q# test case, and all arguments for entry-point operations translate into required command line options for standalone executables.

kuzminrobin: I was expecting something like 
> --my-cool-bool **<false|true>** (REQUIRED)

rather than
> --my-cool-bool (REQUIRED)

src/Simulation/EntryPointDriver.Tests/Tests.fs
cesarzc: Updated comment description to be `An integer`.

src/Simulation/EntryPointDriver/Mock/ErrorQuantumMachine.cs
cesarzc: This was an accidentally deleted character. Good catch. Fixed it in the latest push.

src/Simulation/EntryPointDriver/IEntryPoint.cs
cesarzc: I expanded the description of the return value for all methods in this interface.

cesarzc: Regarding the values themselves, it is not really a fixed set of values, hence we can't list them all. And since this is an interface, the implementation details is up to the interface implementer. The integer value is just for callers and implementers to know whether the operation was successful.

cesarzc: Actually, we do have a known value for the exit code, `0` means success. I'll update the documentation to add this detail. Thanks Robin!

cesarzc: Updated.

cesarzc: When that happens, a warning like the following will appear when building the project:
```
C:\Microsoft\ReposA\qsharp-runtime\src\Simulation\EntryPointDriver\IEntryPoint.cs(59,68): warning CS1573: Parameter 'afhak' has no matching param tag in the XML comment for 'IEntryPoint.Simulate(ParseResult, DriverSettings, string)' (but other parameters do) [C:\Microsoft\ReposA\qsharp-runtime\src\Simulation\EntryPointDriver\Microsoft.Quantum.EntryPointDriver.csproj]
```

Additionally, both IDEs like VS and VSCode underline both the param doc comment and the method parameter itself pointing to the lack of doc comment and doc comment for non-existing param.

Additionally, per the [official documentation](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/documentation-comments), documentation comments "must immediately precede a user-defined type (such as a class, delegate, or interface) or a member (such as a field, event, property, or method). "

src/Simulation/EntryPointDriver.Tests/Tests.fs
cesarzc: This is a good point! I created #824 to improve this in a separate PR since it is out of the scope of this one.

kuzminrobin:
Re-approved.

PR #813, opened 2021-09-09
Validate process exist code in QirExecutableRunner

george-moussa:


src/Qir/Tools/Executable/QirExecutableRunner.cs
```diff
@@ -37,6 +38,11 @@ public async Task RunExecutableAsync(FileInfo executableFile, Stream stream, str
             using var streamWriter = new StreamWriter(stream);
             await streamWriter.WriteAsync(output);
             logger?.LogInformation($"Executable has finished running. Result code: {process.ExitCode}");
+
+            if (process.ExitCode != 0)
```
george-moussa, line 42: The common pattern is that success means exist code = 0. @cesarzc let me know if that's not the case for the simulator executable.

cesarzc: Exit code 0 means success in the simulator executable.

cesarzc (approved):

PR #820, opened 2021-09-14
Use single dash for single letter command line options in QIR driver generator

samarsha:
This fixes an issue when using the full state simulator in Azure, where the generated entry point expects `-x` but it is invoked with the argument string `--x`.

### DIFF
--- a/src/Qir/Tools/Driver/QirCppDriverGenerator.cs
+++ b/src/Qir/Tools/Driver/QirCppDriverGenerator.cs
@@ -33,7 +33,7 @@ namespace Microsoft.Quantum.Qir.Runtime.Tools.Driver
 
         public string GetCommandLineArguments(ExecutionInformation executionInformation)
         {
-            string Argument(Parameter param)
+            string ToArgument(Parameter param)
             {
                 var prefix = param.Name.Length > 1 ? "--" : "-";
                 var value = GetArgumentValueString(param, executionInformation.ArgumentValues[param.Name]);
@@ -41,7 +41,7 @@ namespace Microsoft.Quantum.Qir.Runtime.Tools.Driver
             }
 
             var parameters = executionInformation.EntryPoint.Parameters.OrderBy(param => param.Position);
-            return string.Join(" ", parameters.Select(Argument));
+            return string.Join(" ", parameters.Select(ToArgument));
         }
 
         private static string GetArgumentValueString(Parameter argument, ArgumentValue argumentValue)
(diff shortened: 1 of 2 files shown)


src/Qir/Tools/Driver/QirCppDriverGenerator.cs
```diff
@@ -33,20 +33,15 @@ public async Task GenerateAsync(EntryPointOperation entryPointOperation, Stream
 
         public string GetCommandLineArguments(ExecutionInformation executionInformation)
         {
-            // Sort arguments by position.
-            var sortedArguments = executionInformation.EntryPoint.Parameters.OrderBy(arg => arg.Position);
-            var argumentBuilder = new StringBuilder();
-            foreach (var arg in sortedArguments)
+            string Argument(Parameter param)
```
idavis, line 36: nit: naming to something more descriptive such as `{Map/Convert}ToArgument` or something similar.

src/Qir/Tools/Driver/QirCppDriverGenerator.cs
```diff
@@ -33,20 +33,15 @@ public async Task GenerateAsync(EntryPointOperation entryPointOperation, Stream
...
             {
-                if (argumentBuilder.Length != 0)
-                {
-                    argumentBuilder.Append(' ');
-                }
-
-                argumentBuilder.Append($"--{arg.Name}").Append(' ').Append(GetArgumentValueString(arg, executionInformation.ArgumentValues[arg.Name]));
+                var prefix = param.Name.Length > 1 ? "--" : "-";
```
idavis, line 38: Is `param.Name.Length == 0` possible?

samarsha: No. (I wish a type like `NonEmptyString` was more common.)

src/Qir/Tools/Driver/QirCppDriverGenerator.cs
samarsha: Hmm, I'm not sure that that's more descriptive... a function by definition converts its input to its output, so the name `Argument` should be enough to deduce that it converts a parameter to a string argument.

samarsha: I renamed it to `ToArgument`.

idavis (approved):

cesarzc (approved):

PR #826, opened 2021-09-17
Make `QirFullStateExecutable` constructor public again

samarsha:
In #807, I made the constructor for `QirFullStateExecutable` internal, which was a mistake because it needs to be called by the Azure Quantum service.


samarsha:
#825 does the same thing

PR #827, opened 2021-09-21
Fixed the #qsharp-compiler:1086

kuzminrobin:


src/Qir/Runtime/lib/QIR/strings.cpp
```diff
@@ -132,7 +132,8 @@ extern "C"
     // Returns a string representation of the Boolean.
     QirString* __quantum__rt__bool_to_string(bool value) // NOLINT
     {
-        std::string str = value ? "true" : "false";
+        volatile bool tmp = value;
+        std::string str   = (tmp ? "true" : "false");
         return CreateOrReuseAlreadyAllocated(std::move(str));
     }
```
kuzminrobin, line 138: This actual implementation has the `bool` parameter. 
> QirString* __quantum__rt__bool_to_string(**bool** value)

The QIR caller uses the `i1` parameter:
> %String* @__quantum__rt__bool_to_string(**i1**)

(A glance at disassembly makes the following impression)
During the call the argument is likely passed through the CPU register (the bool param is 8 bytes), the argument `true` (with value `1` of type `i1`) is likely implicitly cast to bool param with sign-extension to 8 bytes, resulting in a `bool value` 0xFFFFFFFFFFFFFFFF (not equal to `false` (0) and not equal to `true` (1)).

Then the (red line) fragment `std::string str = value ? "true" : "false"` likely uses some array of 2 pointers: array[0] pointing to C string "false" and array[1] pointing to C string "true". The address of the array is loaded to a reg, `bool value` (expected to be 0 or 1) is added to that address, and we are expected to get the address of either array[0] or array[1], and use a pointer in there to init the `std::string str`.
But the `bool value` 0xFFFFFFFFFFFFFFFF is added instead and we get a wrong pointer to init the `std::string str` which results in access violation in subsequent code.

The fix solves the issue for now.

cesarzc: Is `%String* @__quantum__rt__bool_to_string(i1)` generated QIR? If so, would the eventual fix be to generate `%String* @__quantum__rt__bool_to_string(i18)` instead? I believe this is what is done for interop QIR for entry-points.

If this is a temporary fix, then it might be good to add a comment explaining that. It would also be useful to describe in a comment why a temporary `volatile bool` variable is required.

swernli: This doesn't line up with my experience, but it's possible I'm remembering it wrong, and can't quickly validate at the moment. I did previously try compiling an extern C function with a `bool` in its signature with clang, and telling it to output IR showed that the `bool` was translated as `i1`. I think we should double check this before moving forward with this fix, just to make sure that the analysis of the source of the bug is correct. As it stands, I'm of the opinion that `i1` is actually the correct type for our Boolean values, but it could be my previous analysis was flawed.

swernli: oh, interesting! I had been so busy thinking about the runtime signature I hadn't considered updating the caller in the generated QIR. Good idea! I think that can give us a path forward for how to use types that are smaller than i8, and we should definitely consider explicitly calling that out in the spec as you suggest.

kuzminrobin:
[Resolution](https://github.com/microsoft/qsharp-runtime/pull/827/files#r713452949)

PR #829, opened 2021-09-21
Add `--job-params` option to the entry point driver's `submit` command

samarsha:


ricardo-espinoza (approved):

src/Simulation/EntryPointDriver/Maybe.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System;
+
+namespace Microsoft.Quantum.EntryPointDriver
+{
+    internal abstract class Maybe<T>
```
cesarzc, line 5: I find the name `Maybe` a bit confusing. After reading the code, `Default` might be a good name for this class. Otherwise, I think it's hard to know what is used for until you see the OptionInfo.cs source file.

cgranade: Tangential, but you may find the `OneOf` NuGet package interesting. That's how I often implement `Maybe`, `Result`, and all my other favorite type-parameterized DUs.

https://github.com/mcintyre321/OneOf

samarsha: > I find the name `Maybe` a bit confusing.

It's from Haskell... I would have used the more common name, `Option`, but that would've been confusing since that refers to a command-line option in this context.

> After reading the code, `Default` might be a good name for this class.

Could you explain why you think `Default` is a good name? I don't think it really has anything to do with defaults... it's just a way to have either a value of something, or no value. The `OptionInfo` class interprets it as either having a default value or not, but that's just based on the context and the variable name `defaultValue`. The type itself is more generic.

samarsha: Since we'd like to get this in for this month's release, I'll merge this now but leave this comment open... since `Maybe` is an internal class I can make a follow-up PR to change it.

samarsha: > Tangential, but you may find the `OneOf` NuGet package interesting. That's how I often implement `Maybe`, `Result`, and all my other favorite type-parameterized DUs.

I don't love that they make you write `OneOf<T, None>` instead of giving you a specific named type, but yeah, that would work too, if we're OK with adding a new (small) dependency to EntryPointDriver.

cesarzc: I was looking at it too narrowly, thanks for the explanation! Now I see how the type is more generic. Now that I understand better what is meant to do, I don't think I have a good idea on how to name it. Maybe `OneOf`? (inspired on the link that Chris posted).

samarsha: > Maybe `OneOf`? (inspired on the link that Chris posted).

OneOf is even more general than Maybe - it's an union between any number of types. `Maybe<T>` could be implemented as `OneOf<T, None>`, but I looked and the OneOf package seems to have unsound behavior when nullable reference types are enabled and you use a non-nullable type like `string` as part of a union, so I'd rather avoid using it now...

The only names I've seen used for this type are `Option<T> = None | Some(T)` (F#, Rust, Scala) and `Maybe<T> = Nothing | Just(T)` (Haskell, PureScript). Because of the naming clash with `Option<T>` from System.CommandLine, I'd rather keep using `Maybe<T>` instead of inventing a new name. I can add doc comments to explain it. Would that work @cesarzc?

cesarzc: Yes, I believe just adding doc comments would be enough.

PR #830, opened 2021-09-23
Added missing last param to __quantum__rt__array_slice[_1d]

kuzminrobin:
Relates to https://github.com/microsoft/qsharp-language/issues/102


swernli:
Related issue: https://github.com/microsoft/qsharp-language/issues/102. I'm of the opinion that our runtime should always create copies when using slicing rather than allowing for views into the existing array, essentially always creating the copy even when `forceNewInstance` is false and alias count is 0. The parameter would then be present but ignored.

kuzminrobin:
Changed to always create a copy.

swernli (approved):

PR #833, opened 2021-09-24
Update QIR Runtime to use QubitIdType

swernli:
This updates the QIR Runtime to use a consistent type for qubit ids, `typedef intptr_t QubitIdType;`. All APIs that correspond to the QIR spec continue to use opaque pointer type `QUBIT*` and then immediately cast to `QubitIdType` for use internally. Of note, the Qubit Manager uses negative values for special meaning when tracking qubits in use, so the unsigned `uintptr_t` is not compatible.

This is part 1 of the change. The update of the native simulator and corresponding C# types from `unsigned` to `intptr_t` will be part of a separate change.

### DIFF
--- a/src/Qir/Runtime/lib/QIR/QubitManager.cpp
+++ b/src/Qir/Runtime/lib/QIR/QubitManager.cpp
@@ -70,8 +70,7 @@ namespace Quantum
         firstElement              = id;           // The new element is now the first in the chain.
     }
 
-    CQubitManager::QubitIdType CQubitManager::QubitListInSharedArray::TakeQubitFromFront(
-        QubitIdType* sharedQbitStatusArray)
+    QubitIdType CQubitManager::QubitListInSharedArray::TakeQubitFromFront(QubitIdType* sharedQbitStatusArray)
     {
         FailIf(sharedQbitStatusArray == nullptr, "Shared status array is not provided.");
 
@@ -241,14 +240,14 @@ namespace Quantum
                                                                 sharedQubitStatusArray);
     }
 
-    Qubit CQubitManager::Allocate()
+    QubitIdType CQubitManager::Allocate()
     {
         QubitIdType newQubitId = AllocateQubitId();
         FailIf(newQubitId == NoneMarker, "Not enough qubits.");
-        return CreateQubitObject(newQubitId);
+        return newQubitId;
     }
 
-    void CQubitManager::Allocate(Qubit* qubitsToAllocate, int32_t qubitCountToAllocate)
+    void CQubitManager::Allocate(QubitIdType* qubitsToAllocate, QubitIdType qubitCountToAllocate)
     {
         if (qubitCountToAllocate == 0)
         {
@@ -258,76 +257,73 @@ namespace Quantum
         FailIf(qubitsToAllocate == nullptr, "No array provided for qubits to be allocated.");
 
         // Consider optimization for initial allocation of a large array at once
-        for (int32_t i = 0; i < qubitCountToAllocate; i++)
+        for (QubitIdType i = 0; i < qubitCountToAllocate; i++)
         {
             QubitIdType newQubitId = AllocateQubitId();
             if (newQubitId == NoneMarker)
             {
-                for (int32_t k = 0; k < i; k++)
+                for (QubitIdType k = 0; k < i; k++)
                 {
                     Release(qubitsToAllocate[k]);
                 }
                 FailNow("Not enough qubits.");
             }
-            qubitsToAllocate[i] = CreateQubitObject(newQubitId);
+            qubitsToAllocate[i] = newQubitId;
         }
     }
 
-    void CQubitManager::Release(Qubit qubit)
+    void CQubitManager::Release(QubitIdType qubit)
     {
         FailIf(!IsValidQubit(qubit), "Qubit is not valid.");
-        ReleaseQubitId(QubitToId(qubit));
-        DeleteQubitObject(qubit);
+        ReleaseQubitId(qubit);
     }
 
-    void CQubitManager::Release(Qubit* qubitsToRelease, int32_t qubitCountToRelease)
+    void CQubitManager::Release(QubitIdType* qubitsToRelease, QubitIdType qubitCountToRelease)
     {
         if (qubitCountToRelease == 0)
         {
             return;
         }
-        FailIf(qubitCountToRelease < 0, "Cannot release negative number of qubits.");
         FailIf(qubitsToRelease == nullptr, "No array provided with qubits to be released.");
 
-        for (int32_t i = 0; i < qubitCountToRelease; i++)
+        for (QubitIdType i = 0; i < qubitCountToRelease; i++)
         {
             Release(qubitsToRelease[i]);
-            qubitsToRelease[i] = nullptr;
+            qubitsToRelease[i] = 0;
         }
     }
 
-    Qubit CQubitManager::Borrow()
+    QubitIdType CQubitManager::Borrow()
     {
         // We don't support true borrowing/returning at the moment.
         return Allocate();
     }
 
-    void CQubitManager::Borrow(Qubit* qubitsToBorrow, int32_t qubitCountToBorrow)
+    void CQubitManager::Borrow(QubitIdType* qubitsToBorrow, QubitIdType qubitCountToBorrow)
     {
         // We don't support true borrowing/returning at the moment.
         return Allocate(qubitsToBorrow, qubitCountToBorrow);
     }
 
-    void CQubitManager::Return(Qubit qubit)
+    void CQubitManager::Return(QubitIdType qubit)
     {
         // We don't support true borrowing/returning at the moment.
         Release(qubit);
     }
 
-    void CQubitManager::Return(Qubit* qubitsToReturn, int32_t qubitCountToReturn)
+    void CQubitManager::Return(QubitIdType* qubitsToReturn, QubitIdType qubitCountToReturn)
     {
         // We don't support true borrowing/returning at the moment.
         Release(qubitsToReturn, qubitCountToReturn);
     }
 
-    void CQubitManager::Disable(Qubit qubit)
+    void CQubitManager::Disable(QubitIdType qubit)
     {
         FailIf(!IsValidQubit(qubit), "Qubit is not valid.");
-        QubitIdType id = QubitToId(qubit);
 
         // We can only disable explicitly allocated qubits that were not borrowed.
-        FailIf(!IsExplicitlyAllocatedId(id), "Cannot disable qubit that is not explicitly allocated.");
-        sharedQubitStatusArray[id] = DisabledMarker;
+        FailIf(!IsExplicitlyAllocatedId(qubit), "Cannot disable qubit that is not explicitly allocated.");
+        sharedQubitStatusArray[qubit] = DisabledMarker;
 
         disabledQubitCount++;
         FailIf(disabledQubitCount <= 0, "Incorrect disabled qubit count.");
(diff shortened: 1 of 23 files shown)


src/Qir/Runtime/lib/QIR/QubitManager.cpp
```diff
@@ -491,7 +457,7 @@ namespace Quantum
 
     void CQubitManager::ReleaseQubitId(QubitIdType id)
     {
-        FailIf(id < 0 || id >= qubitCapacity, "Internal Error: Cannot release an invalid qubit.");
+        FailIf(id >= qubitCapacity, "Internal Error: Cannot release an invalid qubit.");
```
kuzminrobin, line 460: Nit? Any details about why the `id < 0 ||` check has been excluded?

swernli: Whoops! I had removed that check when converting to unsigned type, but then found that wouldn't work with the Qubit Manager design since it needs negative numbers to mark certain qubits. I'll put this one back.

src/Qir/Runtime/lib/Simulators/FullstateSimulator.cpp
```diff
@@ -230,30 +230,30 @@ namespace Quantum
             dump(this->simulatorId, callback);
         }
 
-        virtual std::string QubitToString(Qubit q) override
+        virtual std::string QubitToString(QubitIdType q) override
         {
             return std::to_string(GetQubitId(q));
```
kuzminrobin, line 235:
```suggestion
            return std::to_string(q);
```

swernli: Good catch! While the `GetQubitId` is still needed for converting to `unsigned` when calling into the simulator, it doesn't need to be used here.

src/Qir/Runtime/unittests/QirRuntimeTests.cpp
```diff
@@ -979,15 +979,15 @@ TEST_CASE("Tuples: copy elision", "[qir_support]")
 // Adjoints for R and Exp are implemented by qis, so let's check they at least do the angle invertion in adjoints.
 struct AdjointsTestSimulator : public SimulatorStub
 {
-    int lastId           = -1;
+    intptr_t lastId      = -1;
```
kuzminrobin, line 982:
```suggestion
    QubitIdType lastId      = -1;
```

src/Qir/Runtime/unittests/QirRuntimeTests.cpp
```diff
@@ -979,15 +979,15 @@ TEST_CASE("Tuples: copy elision", "[qir_support]")
...
     double rotationAngle = 0.0;
     double exponentAngle = 0.0;
 
-    Qubit AllocateQubit() override
+    QubitIdType AllocateQubit() override
     {
-        return reinterpret_cast<Qubit>(++this->lastId);
+        return static_cast<QubitIdType>(++this->lastId);
```
kuzminrobin, line 988:
```suggestion
        return ++this->lastId;
```

kuzminrobin (approved):
Approved with minor comments.

src/Qir/Runtime/unittests/QirRuntimeTests.cpp
```diff
@@ -750,8 +750,8 @@ TEST_CASE("Qubits: allocate, release, dump", "[qir_support]")
     REQUIRE(qs->count == 3);
     REQUIRE(qs->itemSizeInBytes == sizeof(void*));
 
-    Qubit last = *reinterpret_cast<Qubit*>(__quantum__rt__array_get_element_ptr_1d(qs, 2));
-    qstr       = __quantum__rt__qubit_to_string(last);
+    QUBIT* last = *reinterpret_cast<QUBIT**>(__quantum__rt__array_get_element_ptr_1d(qs, 2));
+    qstr        = __quantum__rt__qubit_to_string(last);
```
kuzminrobin, line 754: For the future to keep in mind:
I got a feeling that there potentially can be different string representation of qubit ID generated 
* by the `QubitToString()` member functions (e.g. `FullStateSimulator::QubitToString()`)
* and the `__quantum__rt__qubit_to_string()` function. 

To avoid that ambiguity it probably makes sense for them all to call the same shared function to get a string representation of a qubit ID (such that they all generate the same string representation for the same qubit ID).

src/Qir/Runtime/unittests/QirRuntimeTests.cpp
```diff
@@ -979,7 +979,7 @@ TEST_CASE("Tuples: copy elision", "[qir_support]")
 // Adjoints for R and Exp are implemented by qis, so let's check they at least do the angle invertion in adjoints.
 struct AdjointsTestSimulator : public SimulatorStub
 {
-    intptr_t lastId      = -1;
+    QubitIdType lastId      = -1;
     double rotationAngle = 0.0;
     double exponentAngle = 0.0;
```
kuzminrobin, line 984: Unaligned consecutive assignment. Does the clang-format let it pass during the compilation?

kuzminrobin:
Done reviewing. Thanks.

src/Qir/Runtime/unittests/QirRuntimeTests.cpp
swernli: Nope, it did indeed fail. Side effect of accepting the updates directly from github ;).

src/Qir/Runtime/unittests/QirRuntimeTests.cpp
swernli: Our runtime is already set up such that `__quantum__rt__qubit_to_string` will forward to `QubitToString` on whichever runtime driver is currently registered:
https://github.com/microsoft/qsharp-runtime/blob/06339b58e644c5559c8fb64ae98c28e90fdf8ffa/src/Qir/Runtime/lib/QIR/delegated.cpp#L150-L154
So those two functions will always produce the same result. However, there is no guarantee that two runtimes or simulators will produce the same result. For example, the full state simulator implements `QubitToString` like this:
https://github.com/microsoft/qsharp-runtime/blob/e61934496d3dbc7763e0013eb7ccda34322bde63/src/Qir/Runtime/lib/Simulators/FullstateSimulator.cpp#L233-L236
while the Toffoli simulator implements it like this:
https://github.com/microsoft/qsharp-runtime/blob/e61934496d3dbc7763e0013eb7ccda34322bde63/src/Qir/Runtime/lib/Simulators/ToffoliSimulator.cpp#L84-L87
so there is already a difference in our own simulators. But because each implementer of `IRuntimeDriver` is free to implement `QubitToString` however they want (and indeed every runtime is free to choose their own qubit representation) there is no guarantee that `__quantum__rt__qubit_to_string` will produce the same result across different runtimes or different simulators.

src/Qir/Runtime/public/CoreTypes.hpp
```diff
@@ -33,7 +33,7 @@
 // is a unique type in the QIR.
```
DmitryVasilevsky, line 33: In the comment above "Qubit" type is referenced. We need to change this comment.

src/Qir/Runtime/lib/QIR/QubitManager.cpp
```diff
@@ -241,14 +240,14 @@ namespace Quantum
...
+    QubitIdType CQubitManager::Allocate()
     {
         QubitIdType newQubitId = AllocateQubitId();
         FailIf(newQubitId == NoneMarker, "Not enough qubits.");
-        return CreateQubitObject(newQubitId);
+        return newQubitId;
     }
 
-    void CQubitManager::Allocate(Qubit* qubitsToAllocate, int32_t qubitCountToAllocate)
+    void CQubitManager::Allocate(QubitIdType* qubitsToAllocate, QubitIdType qubitCountToAllocate)
```
DmitryVasilevsky, line 250: qubitCountToAllocate should still be int32_t.

src/Qir/Runtime/lib/QIR/QubitManager.cpp
```diff
@@ -258,93 +257,89 @@ namespace Quantum
...
+    void CQubitManager::Release(QubitIdType qubit)
     {
         FailIf(!IsValidQubit(qubit), "Qubit is not valid.");
-        ReleaseQubitId(QubitToId(qubit));
-        DeleteQubitObject(qubit);
+        ReleaseQubitId(qubit);
     }
 
-    void CQubitManager::Release(Qubit* qubitsToRelease, int32_t qubitCountToRelease)
+    void CQubitManager::Release(QubitIdType* qubitsToRelease, QubitIdType qubitCountToRelease)
```
DmitryVasilevsky, line 281: qubitCountToRelease should still be int32_t and variables to iterate over it.

src/Qir/Runtime/lib/QIR/QubitManager.cpp
```diff
@@ -258,93 +257,89 @@ namespace Quantum
...
         }
-        FailIf(qubitCountToRelease < 0, "Cannot release negative number of qubits.");
         FailIf(qubitsToRelease == nullptr, "No array provided with qubits to be released.");
 
-        for (int32_t i = 0; i < qubitCountToRelease; i++)
+        for (QubitIdType i = 0; i < qubitCountToRelease; i++)
         {
             Release(qubitsToRelease[i]);
-            qubitsToRelease[i] = nullptr;
+            qubitsToRelease[i] = 0;
```
DmitryVasilevsky, line 292: As the global type is now signed int, a special value such as NoneMarker can be used.

src/Qir/Runtime/lib/QIR/QubitManager.cpp
```diff
@@ -258,93 +257,89 @@ namespace Quantum
...
 
-    Qubit CQubitManager::Borrow()
+    QubitIdType CQubitManager::Borrow()
     {
         // We don't support true borrowing/returning at the moment.
         return Allocate();
     }
 
-    void CQubitManager::Borrow(Qubit* qubitsToBorrow, int32_t qubitCountToBorrow)
+    void CQubitManager::Borrow(QubitIdType* qubitsToBorrow, QubitIdType qubitCountToBorrow)
```
DmitryVasilevsky, line 302: qubitCountToBorrow should still be int32_t.

src/Qir/Runtime/lib/QIR/QubitManager.cpp
```diff
@@ -258,93 +257,89 @@ namespace Quantum
...
 
-    void CQubitManager::Return(Qubit qubit)
+    void CQubitManager::Return(QubitIdType qubit)
     {
         // We don't support true borrowing/returning at the moment.
         Release(qubit);
     }
 
-    void CQubitManager::Return(Qubit* qubitsToReturn, int32_t qubitCountToReturn)
+    void CQubitManager::Return(QubitIdType* qubitsToReturn, QubitIdType qubitCountToReturn)
```
DmitryVasilevsky, line 314: qubitCountToReturn should still be int32_t.

src/Qir/Runtime/lib/QIR/QubitManager.cpp
```diff
@@ -258,93 +257,89 @@ namespace Quantum
...
+        sharedQubitStatusArray[qubit] = DisabledMarker;
 
         disabledQubitCount++;
         FailIf(disabledQubitCount <= 0, "Incorrect disabled qubit count.");
         allocatedQubitCount--;
         FailIf(allocatedQubitCount < 0, "Incorrect allocated qubit count.");
     }
 
-    void CQubitManager::Disable(Qubit* qubitsToDisable, int32_t qubitCountToDisable)
+    void CQubitManager::Disable(QubitIdType* qubitsToDisable, QubitIdType qubitCountToDisable)
```
DmitryVasilevsky, line 334: qubitCountToDisable should still be int32_t  (and iterator vairables)

src/Qir/Runtime/public/QubitManager.hpp
```diff
@@ -268,9 +244,9 @@ namespace Quantum
         CRestrictedReuseAreaStack freeQubitsInAreas;
 
         // Counts:
-        int32_t disabledQubitCount  = 0;
```
DmitryVasilevsky, line 271: These aren't Ids and should stay int32_t.

src/Qir/Runtime/lib/QIR/QubitManager.cpp
swernli: In theory changing this back to `int32_t` would mean a caller could never allocate all the qubits at once, since max qubit id is larger than max value for `int32_t`. In practice this will likely never be a problem, but it's worth noting that this is another of places where there is a mismatch between the max the type can support and the max that the code can actually accommodate. The types below or similar, where changing them back to smaller integers would imply a restriction. If I can get it compile that way I can make the shift and put it up for PR so we can discuss, but I figured it was worth calling out this implied restriction.

PR #834, opened 2021-09-25
Swernli/qubitid-fix-2

swernli:


swernli:
Creating a PR here to have something concrete to discuss. This builds but fails testing.

After working on this a bit I think I'm not in favor of this change. The native simulator uses `unsigned` both in the C API it exposes and internally when calling into the avx kernels. We could change the C API type to `intptr_t` to match the change I have in #833, but then we'd have to either do the conversions inside of the native sim to go back to `unsigned` before calling into the kernels or update all the avx kernel types as well.

Instead, I think we can do some minor clean-up to the native simulator (for example, using `logical_qubit_id` and `positional_qubit_id` to signal intent of the code, even though they are defined to the same type), but leave the C API with `unsigned` as the qubit type. This avoids a breaking change on that API, and leaves us in what I think is a reasonable state: QIR API uses the opaque `QUBIT*` type, QIR Runtime internally uses `typedef intptr_t QubitIdType`, and the QIR wrapper around our fullstate simulator converts to `unsigned` before calling into that C API. I don't feel strongly that the native simulator API necessarily needs to be changed to match our QIR Runtime internal representation, especially since there doesn't seem to be a big benefit to doing so.

src/Simulation/Native/src/simulator/capi.cpp
```diff
@@ -35,20 +35,21 @@ extern "C"
...
-        _In_reads_(n) unsigned* q)
+        _In_reads_(n) intptr_t* q)
     {
         std::vector<Gates::Basis> bv;
         bv.reserve(n);
         for (unsigned i = 0; i < n; ++i)
             bv.push_back(static_cast<Gates::Basis>(*(b + i)));
-        std::vector<unsigned> qv(q, q + n);
+        std::vector<logical_qubit_id> qv(
+            reinterpret_cast<logical_qubit_id*>(q), reinterpret_cast<logical_qubit_id*>(q + n));
```
swernli, line 45: These kinds of `reinterpret_cast` invocations are sufficient to compile but do not actually work, because `logical_qubit_id` (which is just `unsigned`) and `intptr_t` are not the same size. To convert an array of one to the other, we'd need a helper function akin to what the QIR wrapper around the fullstate simulator uses to ensure the ids are copied and converted one at a time:
https://github.com/microsoft/qsharp-runtime/blob/e61934496d3dbc7763e0013eb7ccda34322bde63/src/Qir/Runtime/lib/Simulators/FullstateSimulator.cpp#L145-L154

swernli: @bettinaheim This is the array copy you were hoping to avoid, but it would either have to happen in the runtime, or have to happen in the fullstate simulator, or we'd have to update the AVX kernels to also take `intptr_t` instead of their current types.

PR #838, opened 2021-09-28
Clean up QubitManager types

swernli:
This is a follow-up to #833, where some internal types in the Qubit Manager were updated to `QubitTypeId` that didn't need to be. See comments there for context.

### DIFF
--- a/src/Qir/Runtime/lib/QIR/QubitManager.cpp
+++ b/src/Qir/Runtime/lib/QIR/QubitManager.cpp
@@ -247,7 +247,7 @@ namespace Quantum
         return newQubitId;
     }
 
-    void CQubitManager::Allocate(QubitIdType* qubitsToAllocate, QubitIdType qubitCountToAllocate)
+    void CQubitManager::Allocate(QubitIdType* qubitsToAllocate, int32_t qubitCountToAllocate)
     {
         if (qubitCountToAllocate == 0)
         {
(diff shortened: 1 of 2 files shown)


swernli:
These fixes made in response to Dmitry's feedback on #833. One thing to note is that changing these internal counts to use `int32_t` means they will not be able to accommodate the highest qubit id of type `intptr_t`. It's incredibly unlikely we will hit a problem with the difference between the maximum values of these types, but it's worth calling out as a place where there is a disconnect between two indirectly linked types.

kuzminrobin:
(Sorry if my comment adds even more ambiguity ;-) )

Looking at #833 and this PR I feel that in a broad sense the _qubit identifier_ and the _number of (supported) qubits_ are different concepts, and it is reasonable for those concepts to use different types (similar to how in the operating systems we have a concept of a file descriptor (handle) and a number of files that can be opened at the same time - completely different types are used).
Another illustration, as has been mentioned earlier, could be a long-running quantum computing cloud service, that allocates and deallocates qubits endlessly. And upon every allocation the qubit id increments until the type overflows, and id enumeration starts over. Whereas the number of qubits used at the same time can be relatively small, e.g. 1000.

We may want to consider a new type to express the number of qubits, e.g. `MaxQubitQuantity` (in common case not equal to `QubitIdType` and not equal to `int32_t`, but in particular case can be an alias of any of those types or of the other type, e.g. `uint16_t`). After this introduction of a new type we will be able to change that type easily, if needed, with a small fix in one place.

kuzminrobin (approved):
I don't state that I agree or disagree with this change. I approve just to unblock moving forward and it is up to you guys to move or to not move, forward or backwards. ;-)

PR #841, opened 2021-09-29
Force use of Chocolatey installed clang-11

swernli:
This will force the install of Clang 11.1.0 by Chocolatey, and then put that version first in the path to ensure it gets used by later build steps.

### DIFF
--- a/src/Qir/Runtime/prerequisites.ps1
+++ b/src/Qir/Runtime/prerequisites.ps1
@@ -9,7 +9,7 @@ if ($Env:ENABLE_QIRRUNTIME -ne "false") {
             !(Get-Command clang-format -ErrorAction SilentlyContinue) -or `
             (Test-Path Env:/AGENT_OS)) {
             choco install llvm --version=11.1.0 --allow-downgrade
-            Write-Host "##vso[task.setvariable variable=PATH;]C:\Program Files\LLVM\bin;$Env:Path"
+            Write-Host "##vso[task.setvariable variable=PATH;]$($env:SystemDrive)\Program Files\LLVM\bin;$Env:PATH"
         }
         if (!(Get-Command ninja -ErrorAction SilentlyContinue)) {
             choco install ninja
--- a/src/Qir/Runtime/prerequisites.ps1
+++ b/src/Qir/Runtime/prerequisites.ps1
@@ -6,9 +6,10 @@
 if ($Env:ENABLE_QIRRUNTIME -ne "false") {
     if (($IsWindows) -or ((Test-Path Env:/AGENT_OS) -and ($Env:AGENT_OS.StartsWith("Win")))) {
         if (!(Get-Command clang        -ErrorAction SilentlyContinue) -or `
-            !(Get-Command clang-format -ErrorAction SilentlyContinue)) {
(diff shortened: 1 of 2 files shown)


src/Qir/Runtime/prerequisites.ps1
```diff
@@ -6,9 +6,10 @@
 if ($Env:ENABLE_QIRRUNTIME -ne "false") {
     if (($IsWindows) -or ((Test-Path Env:/AGENT_OS) -and ($Env:AGENT_OS.StartsWith("Win")))) {
         if (!(Get-Command clang        -ErrorAction SilentlyContinue) -or `
-            !(Get-Command clang-format -ErrorAction SilentlyContinue)) {
-            choco install llvm --version=11.1.0
-            Write-Host "##vso[task.setvariable variable=PATH;]$Env:Path;C:\Program Files\LLVM\bin"
+            !(Get-Command clang-format -ErrorAction SilentlyContinue) -or `
+            (Test-Path Env:/AGENT_OS)) {
+            choco install llvm --version=11.1.0 --allow-downgrade
+            Write-Host "##vso[task.setvariable variable=PATH;]C:\Program Files\LLVM\bin;$Env:Path"
```
kuzminrobin, line 12: Just in case I would recommend the following:
```suggestion
            Write-Host "##vso[task.setvariable variable=PATH;]C:\Program Files\LLVM\bin;$Env:PATH"
```
(it makes no difference for Win, but some time ago I copy-pasted the fragment `$Env:Path` to the scope common for all platforms, and on Linux the behavior was unexpected)

src/Qir/qir-utils.ps1
```diff
@@ -192,9 +192,10 @@ function Build-CMakeProject {
         $env:CXX = "clang++.exe"
         $env:RC = "clang++.exe"
 
-        if (!(Get-Command clang -ErrorAction SilentlyContinue) -and (choco find --idonly -l llvm) -contains "llvm") {
+        if ((!(Get-Command clang -ErrorAction SilentlyContinue) -and (choco find --idonly -l llvm) -contains "llvm") -or `
+            (Test-Path Env:AGENT_OS)) {
```
kuzminrobin, line 196: In one file the fragment is 
`Env:/AGENT_OS`    (with `/`)
in the other - 
`Env:AGENT_OS`      (without `/`).

It probably makes sense to use `Env:/AGENT_OS` in both.

src/Qir/qir-utils.ps1
```diff
@@ -192,9 +192,10 @@ function Build-CMakeProject {
...
             # LLVM was installed by Chocolatey, so add the install location to the path.
-            $env:PATH += ";$($env:SystemDrive)\Program Files\LLVM\bin"
+            $env:PATH = "$($env:SystemDrive)\Program Files\LLVM\bin;$env:Path"
```
kuzminrobin, line 198: If you have time and desire (to unify the heritage):

In one file the fragment was 
`C:\Program Files`
in the other 
`$($env:SystemDrive)\Program Files`.
It probably makes sense to use identical in both.

kuzminrobin (approved):

PR #842, opened 2021-09-30
Add `CX`, `CY` and `CZ` to Type1, Type3 target packages.

swernli:
This simplifies the use of controlled Pauli gates to allow for programs that make use of singly controlled gates (or CCNOT) to have reasonable decompositoins without incurring array creation or using the controlled specialization. This leads to better compatibility with the draft spec for the QIR base profile.

### DIFF
--- a/src/Simulation/TargetDefinitions/Decompositions/CX.qs
+++ b/src/Simulation/TargetDefinitions/Decompositions/CX.qs
@@ -1,0 +1,39 @@
+namespace Microsoft.Quantum.Canon {
+    open Microsoft.Quantum.Intrinsic;
+
+    /// # Summary
+    /// Applies the controlled-X (CX) gate to a pair of qubits.
+    ///
+    /// # Description
+    /// This operation can be simulated by the unitary matrix
+    /// $$
+    /// \begin{align}
+    ///     \left(\begin{matrix}
+    ///         1 & 0 & 0 & 0 \\\\
+    ///         0 & 1 & 0 & 0 \\\\
+    ///         0 & 0 & 0 & 1 \\\\
+    ///         0 & 0 & 1 & 0
+    ///      \end{matrix}\right)
+    /// \end{align},
+    /// $$
+    /// where rows and columns are organized as in the quantum concepts guide.
+    ///
+    /// # Input
+    /// ## control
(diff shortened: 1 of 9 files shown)


swernli:
Note that this is going into a feature branch for now as part of a larger effort to apply simplifications to the target packages and ease their use across different QIR profiles.

This is related to https://github.com/microsoft/QuantumLibraries/issues/477, but does not necessarily remove the need for considering that proposal.

src/Simulation/TargetDefinitions/Decompositions/CX.qs
```diff
@@ -0,0 +1,39 @@
+namespace Microsoft.Quantum.Canon {
```
cgranade, line 1: Do we need a copyright header here?

src/Simulation/TargetDefinitions/Decompositions/CYFromCNOT.qs
```diff
@@ -0,0 +1,43 @@
+namespace Microsoft.Quantum.Canon {
```
cgranade, line 1: Similarly here.

src/Simulation/TargetDefinitions/Decompositions/CZFromSinglyControlled.qs
```diff
@@ -0,0 +1,47 @@
...
+    /// Equivalent to:
+    /// ```qsharp
+    /// Controlled Z([control], target);
+    /// ```
+    operation CZ(control : Qubit, target : Qubit) : Unit {
+        body (...) {
+            ApplyControlledZ(control, target);
+        }
+        controlled (ctls, ...) {
+            if (Length(ctls) == 0) {
```
cgranade, line 35: Nit, but I think the Q# language server diagnostics now recommend dropping parens around conditionals in `if` blocks.
```suggestion
            if Length(ctls) == 0 {
```

cgranade (approved):
LGTM, no substantiative comments.

src/Simulation/TargetDefinitions/Decompositions/CX.qs
swernli: Good call, I totally spaced on that!

PR #847, opened 2021-10-05
Added some C APIs for Llvm Jit Compilation

kuzminrobin:


bettinaheim (approved):
lgtm. Once the build finishes, I'll update the QIR execution tests on the compiler to use this package.

PR #863, opened 2021-11-04
Moving QIR runtime build config into CMakeLists.

idavis:
To aid in consumption of the QIR Runtime, most configuration has been moved into the `CMakeLists.txt`.


src/Qir/Common/cmake/qir_cmake_include.cmake
```diff
@@ -39,3 +39,79 @@ macro(target_source_from_qir target_name source_file)
...
+
+
+#===============================================================================
+# Common flags
+
+set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -DDEBUG")
+set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -DDEBUG")
+
+if (NOT WIN32)
+    set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -fsanitize=undefined -fsanitize=float-divide-by-zero -fsanitize=unsigned-integer-overflow -fsanitize=implicit-conversion -fsanitize=local-bounds -fsanitize=nullability")
```
swernli, line 51: nitpick: Is there a reason to not just combine this and the lines below into a single long string with newlines? Something like:
```
set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} \
  -fsanitize=undefined \
  -fsanitize=float-divide-by-zero \
 ...
")
```

src/Qir/qir-utils.ps1
```diff
@@ -34,163 +34,41 @@ function Build-CMakeProject {
...
-            #         Any allocations made between the call and the exception throw (caught by `REQUIRE_THROWS()`) are leaking.
-            #         Extract such tests to a separate .cpp file or executable and compile with leak check off (or suppress leaks in that .cpp or executable only).
-
+        if (-not ($IsWindows)) {
             # Common for all sanitizers:
-            $sanitizeFlags += " -fsanitize-blacklist="      # https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html#suppressing-errors-in-recompiled-code-ignorelist
-                                                            # https://releases.llvm.org/11.0.1/tools/clang/docs/SanitizerSpecialCaseList.html
+            $sanitizeFlags = " -fsanitize-blacklist="      # https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html#suppressing-errors-in-recompiled-code-ignorelist
+            # https://releases.llvm.org/11.0.1/tools/clang/docs/SanitizerSpecialCaseList.html
             $sanitizeFlags += (Join-Path $Path .. UBSan.ignore)
```
swernli, line 51: Is not having this flag in the CMakeLists.txt going to interfere with the sanitizer behavior when compiling directly through there? Would it be worth putting this path into a define variable that is then used in the CMakeLists.txt?

swernli (approved):
This looks good to me! One question below on the UBSan.ignore file and how that might be moved into the CMakeLists.txt too. But as is I think this is a good change to make!

src/Qir/Common/cmake/qir_cmake_include.cmake
```diff
@@ -39,3 +39,79 @@ macro(target_source_from_qir target_name source_file)
...
+
+    set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -fsanitize=address")
+    set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -fsanitize=address")
+
+    set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -fno-omit-frame-pointer")
+    set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -fno-omit-frame-pointer")
+
+    set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -fno-optimize-sibling-calls")
+    set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -fno-optimize-sibling-calls")
+endif()
```
kuzminrobin, line 62: Consider avoiding duplication of the fragments, something like this:
```suggestion
if (NOT WIN32)
    set(SANITIZE_FLAGS \
       "-fsanitize=undefined \
        -fsanitize=float-divide-by-zero \
        -fsanitize=unsigned-integer-overflow \
        -fsanitize=implicit-conversion \
        -fsanitize=local-bounds \
        -fsanitize=nullability \
        -fsanitize=address \
        -fno-omit-frame-pointer \
        -fno-optimize-sibling-calls")
        
    set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} ${SANITIZE_FLAGS}")
    set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} ${SANITIZE_FLAGS}")
endif()
```

src/Qir/Common/cmake/qir_cmake_include.cmake
```diff
@@ -39,3 +39,79 @@ macro(target_source_from_qir target_name source_file)
...
+# https://clang.llvm.org/docs/UsersManual.html#enabling-all-diagnostics
+# https://clang.llvm.org/docs/DiagnosticsReference.html
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Weverything")
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Weverything")
+
+# Disable these warnings:
+
+# https://clang.llvm.org/docs/DiagnosticsReference.html#wc-98-compat-pedantic
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-c++98-compat-pedantic")
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-c++98-compat-pedantic")
```
kuzminrobin, line 79: Same for warning flags, something like this:
```cmake
# Treat warnings as errors:
# https://clang.llvm.org/docs/UsersManual.html#options-to-control-error-and-warning-messages
set(WARNING_FLAGS "-Werror")

# https://clang.llvm.org/docs/UsersManual.html#enabling-all-diagnostics
# https://clang.llvm.org/docs/DiagnosticsReference.html
set(WARNING_FLAGS "${WARNING_FLAGS} -Weverything")

# Disable these warnings:

# https://clang.llvm.org/docs/DiagnosticsReference.html#wc-98-compat-pedantic
set(WARNING_FLAGS "${WARNING_FLAGS} -Wno-c++98-compat-pedantic")
. . .

set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${WARNING_FLAGS}")
set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${WARNING_FLAGS}")
```

src/Qir/qir-utils.ps1
```diff
@@ -34,163 +34,41 @@ function Build-CMakeProject {
...
-            # Undefined Behavior Sanitizer (https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html)
-            # Win:
-            #   FAILED: lib/QIR/Microsoft.Quantum.Qir.Runtime.dll lib/QIR/Microsoft.Quantum.Qir.Runtime.lib
-            #   lld-link: error: /failifmismatch: mismatch detected for 'RuntimeLibrary':
-            #   >>> lib/QIR/CMakeFiles/qir-rt-support-obj.dir/QubitManager.cpp.obj has value MD_DynamicRelease
-            #   >>> clang_rt.ubsan_standalone_cxx-x86_64.lib(ubsan_type_hash_win.cc.obj) has value MT_StaticRelease
-            #   clang++: error: linker command failed with exit code 1 (use -v to see invocation)
-            $sanitizeFlags += " -fsanitize=undefined -fsanitize=float-divide-by-zero -fsanitize=unsigned-integer-overflow -fsanitize=implicit-conversion -fsanitize=local-bounds -fsanitize=nullability"
-            # TODO: 
-            #     For Win consider extra build configuration linking all libs statically, enable `-fsanitize=undefined`, run the statically linked tests.
```
kuzminrobin, line 118: I see that after this PR most of the comments (of the sanitizer options) will not survive.
I would strongly recommend to keep the comments. They are a result of a number of hours spent, and they will save from spending those hours again in the future. I'll explain.

Enabling the sanitizers is an incomplete work. At the moment of enabling, certain things were not supported on specific platforms or some sanitizers were conflicting with each other. The sanitizers will continue being developed and some time later we will get back to them and enable more. 
The comments describe certain stage of achievement on our way to enable the sanitizers. The comments say that certain directions are dead ends, etc. If the comments will not survive, then in the future, when we get back to sanitizers, we will have to repeat gaining all that knowledge saved in the comments.

src/Qir/qir-utils.ps1
```diff
@@ -34,163 +34,41 @@ function Build-CMakeProject {
...
-            #       * Some tests verify the failure behavior, i.e. they cause `Fail()` to be called and return to the caller with the exception. 
-            #         Any allocations made between the call and the exception throw (caught by `REQUIRE_THROWS()`) are leaking.
-            #         Extract such tests to a separate .cpp file or executable and compile with leak check off (or suppress leaks in that .cpp or executable only).
-
+        if (-not ($IsWindows)) {
             # Common for all sanitizers:
-            $sanitizeFlags += " -fsanitize-blacklist="      # https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html#suppressing-errors-in-recompiled-code-ignorelist
-                                                            # https://releases.llvm.org/11.0.1/tools/clang/docs/SanitizerSpecialCaseList.html
+            $sanitizeFlags = " -fsanitize-blacklist="      # https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html#suppressing-errors-in-recompiled-code-ignorelist
+            # https://releases.llvm.org/11.0.1/tools/clang/docs/SanitizerSpecialCaseList.html
```
kuzminrobin, line 50: I would suggest to preserve the indentation of this comment line.
If it is indented then it is clear (at least to me) that the comment relates to the previous code line.
If it is not indented then it becomes ambiguous wither the comment relates to the previous or to the subsequent code line.

src/Qir/qir-utils.ps1
```diff
@@ -34,163 +34,41 @@ function Build-CMakeProject {
...
         } # if (-not ($IsWindows))
 
-        $env:CFLAGS   += $sanitizeFlags
-        $env:CXXFLAGS += $sanitizeFlags
     } # if ($Env:BUILD_CONFIGURATION -eq "Debug") 
 
 
-    if (($IsMacOS) -or ((Test-Path Env:/AGENT_OS) -and ($Env:AGENT_OS.StartsWith("Darwin"))))
-    {
+    if (($IsMacOS) -or ((Test-Path Env:/AGENT_OS) -and ($Env:AGENT_OS.StartsWith("Darwin")))) {
```
kuzminrobin, line 59: I would recommend to do the unrelated changes like this one in a separate small preparatory PR. 

The current PR has certain logic relocation. And it takes some time to double-check that the correctness of the logic is preserved. The multiple small unrelated changes like this one complicate the analysis of the correctness.

kuzminrobin (commented):
I'm done with the CR pass.
I would recommend to preserve the sanitizer comments. Other notes are up to you.

idavis:
@kuzminrobin @swernli Thanks for taking a look. I left the flags separate rather than consolidate as I was unsure as to which way to go with the comments. It sounds like the preferred path is to migrate the comments into the `CMakeLists.txt` to preserve more historical knowledge. I'll rework and push. Thanks for your help.

kuzminrobin (approved):
LGTM. Approving.
I would consider handling WARNING_FLAGS before the SANITIZE_FLAGS. 
Also the fragment 
```cmake
# Always use available Spectre mitigations where available
if (NOT APPLE)
    set(WARNING_FLAGS "${WARNING_FLAGS} -mspeculative-load-hardening -mretpoline")
endif()
```
does not seem applicable to warnings. So you might want to use a different var instead of `WARNING_FLAGS`.
Up to you.

kuzminrobin:
LGTM, re-approving.

idavis:
@kuzminrobin Nice catch on the Spectre mitigations. Pushed what you requested.

PR #867, opened 2021-11-10
Preparing the QuantumSimulator for unifying the Dump().

kuzminrobin:
This PR is a stage one of the three below:
* Preparing the QuantumSimulator for unifying the Dump(). Providing the new dump functionality on the `qsharp-runtime` repo side.
* Redirecting the `IQSharp` repo to the new dump functionality in `qsharp-runtime` repo.
* Cleaning-up the commented out code on the `qsharp-runtime` repo side.


src/Simulation/Simulators/QuantumSimulator/StateDumper.cs
```diff
@@ -155,25 +323,163 @@ public virtual string Format(uint idx, double real, double img)
...
+                {
+                    // We cast here as we don't support a large enough number
+                    // of qubits to saturate an int.
+                    QubitIds = qubits?.Select(q => q.Id) ?? Simulator.QubitIds.Select(q => (int)q) ?? Enumerable.Empty<int>(),
+                    NQubits = (int)_count,
+                    Amplitudes = _data,
+                    DivId = $"dump-machine-div-{id}" 
+                };
+
+                if(this.FileWriter != null)
```
cgranade, line 453: [nit: code style]
```suggestion
                if (this.FileWriter != null)
```

src/Simulation/Simulators/QuantumSimulator/StateDumper.cs
```diff
@@ -155,25 +323,163 @@ public virtual string Format(uint idx, double real, double img)
...
+            public override string ToString()
+            {
+                if(Amplitudes == null)
+                {
+                    return "";
+                }
+
+                string retVal = "";
+                uint idx = 0;
+                foreach (Complex complexNum in Amplitudes)
```
cgranade, line 335: This may need to be changed to significant amplitudes to allow the new truncation option to propagate out to C# and command-line users.

src/Simulation/Simulators/QuantumSimulator/StateDumper.cs
```diff
@@ -155,25 +323,163 @@ public virtual string Format(uint idx, double real, double img)
 
             }
 
+            public override string ToString()
```
cgranade, line 326: For the eventual IQ# PR, we may want to refactor its plain-text display encoder to use the new implementation here so that it can reuse `FormatCartesian` and `FormatPolar` instead of re-implementing. If we do that, it may be worth making `ToString` call a new overload of `ToString` that allows passing more options. That could likely be deferred until we have the matching IQ# PR closer to being ready to merge in.

cgranade (commented):
This looks like a really good start, thanks for tagging me in on this! There's some minor suggestions that I could make as well as call forwards to an eventual IQ# PR, but I think this should be great for unblocking progress on the IQ# side of the refactoring.

src/Simulation/Simulators/QuantumSimulator/Dump.cs
```diff
@@ -23,48 +23,34 @@ protected virtual QVoid Dump<T>(T target, IQArray<Qubit>? qubits = null)
...
                     logMessage.Apply($"[warning] Unable to write state to '{filename}' ({e.Message})");
-                    return QVoid.Instance;
                 }
             }
+            return QVoid.Instance;
         }
 
-        public class QsimDumpMachine<T> : Quantum.Diagnostics.DumpMachine<T>
-        {
+        public class QsimDumpMachine<T> : Quantum.Diagnostics.DumpMachine<T> // Is inherited (and replaced at runtime)
```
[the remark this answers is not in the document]
kuzminrobin: I plan the work in 3 stages:
* This PR (prepare the qsharp-runtime side for unifying the dump)
* On IQ# side to redirect the functionality to the unified dump (a separate next PR)
* Clean-up the qsharp-runtime side.

During the last stage I plan to remove all such comments and commented-out code.

src/Simulation/Simulators/QuantumSimulator/StateDumper.cs
```diff
@@ -2,7 +2,13 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Collections.Generic;   // IComparer
```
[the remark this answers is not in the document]
kuzminrobin: This is based on `main`. I don't know any details about the .NET 6 refactoring (anything related to .NET 6 is a black box for me at the moment).
I'm removing the comments in these `using` lines to preserve the code style (I was planning to do that any way). But I would like to refrain right now from proceeding to `global using` since that change is unrelated to dump and may be distracting.

src/Simulation/Simulators/QuantumSimulator/StateDumper.cs
```diff
@@ -60,29 +67,190 @@ public virtual bool Dump(IQArray<Qubit>? qubits = null)
...
+            /// </summary>
+            [JsonProperty("diagnostic_kind")]
+            private string DiagnosticKind => "state-vector";
+
+            /// <summary>
+            ///     ID for an HTML element where the vertical measurement probability histogram
+            ///     will be displayed.
+            /// </summary>
+            [JsonProperty("div_id")]
+            public string DivId { get; set; } = string.Empty;
```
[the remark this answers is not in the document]
kuzminrobin: (My familiarity with this functionality is just intuitive)
What is your suggestion?
Is there anything related to this topic that I need to address in this PR? (anything that will prevent the next PR on IQ# side to run smoothly in one attempt/step)
I'm ready to discuss this offline, feel free to give me a call whenever I'm available.

src/Simulation/Simulators/QuantumSimulator/StateDumper.cs
```diff
@@ -60,29 +67,190 @@ public virtual bool Dump(IQArray<Qubit>? qubits = null)
...
+            [JsonProperty("n_qubits")]
+            public int NQubits { get; set; }
+
+            /// <remarks>
+            ///     These amplitudes represent the computational basis states
+            ///     labeled in little-endian order, as per the behavior of
+            ///     <see cref="Microsoft.Quantum.Simulation.Simulators.QuantumSimulator.StateDumper.Dump" />.
+            /// </remarks>
+            [JsonProperty("amplitudes")]
+            public Complex[]? Amplitudes { get; set; }
```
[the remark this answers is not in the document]
kuzminrobin: The attempt to use `init;` results in the compiler error 
> Error	[CS0518](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs0518)	Predefined type 'System.Runtime.CompilerServices.IsExternalInit' is not defined or imported	

The [`init` documentation](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/init) starts with "_In C# 9 and later_". Are we using an earlier C#?

src/Simulation/Simulators/QuantumSimulator/StateDumper.cs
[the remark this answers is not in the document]
cgranade: Makes sense. By the time the third PR in your plan above hits, `main` will likely be on .NET 6 so maybe this will fit best as part of that cleanup?

src/Simulation/Simulators/QuantumSimulator/StateDumper.cs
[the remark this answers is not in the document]
cgranade: Right, this was also scoped on the question of 3.1 vs 6, as 3.1 doesn't support `init` yet. That landed in 5.0 and hence will be made available with the 6.0 port coming soon.

src/Simulation/Simulators/QuantumSimulator/StateDumper.cs
```diff
@@ -60,29 +67,190 @@ public virtual bool Dump(IQArray<Qubit>? qubits = null)
...
+                    _ => throw new ArgumentException($"Invalid basis state labeling convention {convention}.")
+                };
+
 
             /// <summary>
             /// Returns a string that represents the label for the given base state.
             /// </summary>
-            public virtual string FormatBaseState(uint idx) =>
-                $"∣{idx.ToString().PadLeft(_maxCharsStateId, ' ')}❭";
+            public virtual string FormatBaseState(uint idx)
```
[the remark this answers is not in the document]
kuzminrobin: FYI: 
The diff of this file does not display the change properly.

The whole `SimpleDumper` with its `Format..()` logic (consisting of 
+ string Format(uint idx, double real, double img)
+ string FormatBaseState(uint idx)
+ string FormatCartesian(double real, double img)
+ string FormatPolar(double magnitude, double angle)
+ string FormatMagnitude(double magnitude, double phase)
+ string FormatAngle(double magnitude, double angle)

) has been removed. 
The `enum BasisStateLabelingConvention` and `class DisplayableState` have been added. The `Format..()` logic has been moved from the removed `SimpleDumper` to the `class DisplayableState`. In that `Format..()` logic the `FormatBaseState()` has been updated slightly. That is why in that `Format..()` logic only the `FormatBaseState()` is shown (by `git diff`) as new, but in reality the whole `Format..()` logic is new in this `class DisplayableState`.

I assume that with your words 
"Since **it**'s public we may not be able to remove it" you mean 
"Since **the `FormatBaseState()`** is public we may not be able to remove it". 
I'd like to admit that the `FormatBaseState()` is totally new for `class DisplayableState`, that is why _we can remove it_.

I was also thinking may be I should encapsulate this `Format..()` logic into a class (probably defined inside of `DisplayableState`) to keep the `Format..()` logic "as a single piece". In which case it would make sense to preserve the `FormatBaseState()`.
Or I can remove the `FormatBaseState()` and replace the call to it with `BasisStateLabel(BasisStateLabelingConvention.LittleEndian, idx)`.

What would you recommend?

src/Simulation/Simulators/QuantumSimulator/StateDumper.cs
```diff
@@ -155,25 +323,163 @@ public virtual string Format(uint idx, double real, double img)
...
+            /// </summary>
+            // TODO(rokuzmin): What does it return?
             public override bool Dump(IQArray<Qubit>? qubits = null)
             {
-                var count = qubits == null
-                    ? this.Simulator.QubitManager.AllocatedQubitsCount
-                    : qubits.Length;
-                this._maxCharsStateId = ((1 << (int)count) - 1).ToString().Length;
+                _count = qubits == null
+                            ? this.Simulator.QubitManager?.AllocatedQubitsCount ?? 0
```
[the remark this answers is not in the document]
kuzminrobin: I assume that with fragment 
> I'm not sure I agree with `?? 0` here, since that would mean `_data` has a length of 1, causing the logic below to necessarily fail

you mean that 
"If at the moment of dumping 
* the `QubitManager` is not yet assigned then it should result in an exception or an assert (a bug in our code is detected);
* none of the qubits is allocated (`0` qubits are allocated) then dumping should still succeed but should display nothing (or a message like "No qubits allocated").

I'm changing to 
```c#
                System.Diagnostics.Debug.Assert(this.Simulator.QubitManager != null,   // Added.
                    "Internal logic error, QubitManager must be assigned");

                _count = qubits == null
                            ? this.Simulator.QubitManager.AllocatedQubitsCount  // Removed `?? 0`.
                            : qubits.Length;
                _data = new Complex[(_count != 0) ? 1 << ((int)_count) : 0]; // `new Complex[0]` if no qubits are allocated.
                var result = base.Dump(qubits);
```
After this, if no qubits are allocated, then a 0-length array is assigned below:
```c#
var state = new DisplayableState
{
    . . .
    Amplitudes = _data,
};
```
and all the subsequent fragments `Amplitudes.Select()` have nothing to select (and return an empty array).

Is my interpretation and fix correct?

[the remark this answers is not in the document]
cgranade: > the QubitManager is not yet assigned then it should result in an exception or an assert (a bug in our code is detected);

Right, I don't think there's a valid set of conditions under which `QubitManager` has not yet been initialized, but `DumpMachine` has been called by a user. Catching that condition as a bug seems like the most robust thing to do there instead of deferring the problem to later.

> none of the qubits is allocated (0 qubits are allocated) then dumping should still succeed but should display nothing (or a message like "No qubits allocated").

Ah, that's a good point. I guess technically in that case the Hilbert space of the system is ℂ¹ (that is, complex-valued scalars), such that the definition of `_data` below makes sense in that case — namely, as an array of length 1. In particular, `new Complex[1]` makes more sense to me than `new Complex[0]`, since the Hilbert space is one-dimensional rather than zero-dimensional. 

---

Up to a global phase, there's one valid vector in ℂ¹, such that `Amplitudes` should be a length 1 array containing a complex number of the form 𝑒^{𝑖θ} for some θ ∈ ℝ.

src/Simulation/Simulators/QuantumSimulator/StateDumper.cs
kuzminrobin: The fragment is already changed to the following:
```c#
            public override string ToString()
            {
                return string.Join('\n', Amplitudes.Select((amplitude, idx) => 
                    Format((uint)idx, amplitude.Real, amplitude.Imaginary)));
            }
```
Based on that what change do you propose? Something like this?
```c#
            public override string ToString()
            {
                return string.Join('\n', 
                    SignificantAmplitudes(BasisStateLabelingConvention.LittleEndian, true, 0.8)  // Fixed settings!!!
                        .Select(
                            item =>
                            {
                                var (cmplx, basisLabel) = item;
                                var amplitude = (cmplx.Real * cmplx.Real) + (cmplx.Imaginary * cmplx.Imaginary);
                                var angle = System.Math.Atan2(cmplx.Imaginary, cmplx.Real);

                                return $"|{basisLabel}⟩\t{FormatCartesian(cmplx.Real, cmplx.Imaginary)}\t == \t" +
                                                    $"{FormatPolar(amplitude, angle)}";
                            }));
            }
```

src/Simulation/Simulators/QuantumSimulator/StateDumper.cs
kuzminrobin: I assume you mean something like this:
```c#
            public string ToString(BasisStateLabelingConvention convention,   // Non-override. Parameterized.
                                   bool truncateSmallAmplitudes, 
                                   double truncationThreshold)
            {
                return string.Join('\n', 
                    SignificantAmplitudes(convention, truncateSmallAmplitudes, truncationThreshold)
                        .Select(
                            item =>
                            {
                                var (cmplx, basisLabel) = item;
                                var amplitude = (cmplx.Real * cmplx.Real) + (cmplx.Imaginary * cmplx.Imaginary);
                                var angle = System.Math.Atan2(cmplx.Imaginary, cmplx.Real);

                                return $"|{basisLabel}⟩\t{FormatCartesian(cmplx.Real, cmplx.Imaginary)}\t == \t" +
                                                    $"{FormatPolar(amplitude, angle)}";
                            }));
            }

            public override string ToString()   // An override of the `object.ToString()`.
            {
                return ToString(BasisStateLabelingConvention.LittleEndian, true, 0.8);   // Fixed values!!!
            }
```
Would you recommend to make this change now in this PR?

src/Simulation/Simulators/QuantumSimulator/StateDumper.cs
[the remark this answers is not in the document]
kuzminrobin: Thank you for the explanation. Your words make me believe that the line below should stay as it was (if 0 qubits are allocated then the array has a single element)
```c#
_data = new Complex[1 << ((int)_count)];
```

src/Simulation/Simulators/QuantumSimulator/StateDumper.cs
cgranade: As mentioned, I don't think it's urgent to include in this PR. At the same there's also no harm in doing so.

```csharp
var amplitude = (cmplx.Real * cmplx.Real) + (cmplx.Imaginary * cmplx.Imaginary);
var angle = System.Math.Atan2(cmplx.Imaginary, cmplx.Real);
```

As a side note, I believe that System.Numerics.Complex already has properties for that, such that the `Select` expression can be significantly simplified.

cgranade:
```csharp
public override string ToString() => // An override of the `object.ToString()`.
    ToString(BasisStateLabelingConvention.LittleEndian, false, 0.0);   // Fixed values!!!
```

Setting truncation to `false` should reproduce previous behavior; the threshold is ignored in that case.

src/Simulation/Simulators/QuantumSimulator/StateDumper.cs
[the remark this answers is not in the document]
cgranade: I think that's correct, yes. I was under the misunderstanding that if no qubits had been allocated by users, there was still at least one available via the qubit manager. Based on your clarifications above, I think that line is correct as-is. Thanks for your patience!

src/Simulation/Simulators/QuantumSimulator/StateDumper.cs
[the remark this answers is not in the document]
kuzminrobin: (FYI: I'm not familiar with `DisplayableHtmlElement`, IQ# code base in general, very fresh to C#. At the moment it is easier for me to follow particular instructions about what and how to change)
For now I'm commenting out all the fragments related to `DivId` with a note to get back to this when refactoring IQ# side.

src/Simulation/Simulators/QuantumSimulator/StateDumper.cs
kuzminrobin: > As a side note, I believe that System.Numerics.Complex already has properties for that, such that the Select expression can be significantly simplified.

Based on [System.Numerics Methods](https://docs.microsoft.com/en-us/dotnet/api/system.numerics.complex?view=net-5.0#methods), here is what I've come up with, not sure if it is absolutely right (already forgot a lot of Complex arithmetic):
```c#
var amplitude =  Abs(cmplx) * Abs(cmplx);   // (cmplx.Real * cmplx.Real) + (cmplx.Imaginary * cmplx.Imaginary);
var angle = Atan(cmplx);                    // System.Math.Atan2(cmplx.Imaginary, cmplx.Real);
```

src/Simulation/Simulators/QuantumSimulator/StateDumper.cs
```diff
@@ -60,29 +67,190 @@ public virtual bool Dump(IQArray<Qubit>? qubits = null)
         }
 
         /// <summary>
-        /// A simple implementation of a <see cref="StateDumper"/>. It outputs the
-        /// a string representation of the state to the given channel.
+        ///     The convention to be used in labeling computational basis states
+        ///     given their representations as strings of classical bits.
         /// </summary>
-        public class SimpleDumper : StateDumper
+        [JsonConverter(typeof(StringEnumConverter))]
```
DmitryVasilevsky, line 73: Can you explain the meaning of this in a comment?

src/Simulation/Simulators/QuantumSimulator/StateDumper.cs
```diff
@@ -60,29 +67,190 @@ public virtual bool Dump(IQArray<Qubit>? qubits = null)
...
+                .Select(
+                    item => (item.amplitude, BasisStateLabel(convention, item.idx))
+                )
+                .OrderBy(
+                    item => item.Item2,
+                    // If a basis state label is numeric, we want to compare
+                    // numerically rather than lexographically.
+                    convention switch {
+                        BasisStateLabelingConvention.BigEndian => ToIntComparer,
+                        BasisStateLabelingConvention.LittleEndian => ToIntComparer,
```
DmitryVasilevsky, line 208: Would be nice to explain why both conventions result in the same comparer.

src/Simulation/Simulators/QuantumSimulator/StateDumper.cs
```diff
@@ -60,29 +67,190 @@ public virtual bool Dump(IQArray<Qubit>? qubits = null)
...
+                            System
+                                .Convert
+                                .ToString(index, 2)
+                                .PadLeft(NQubits, '0')
+                                .Reverse()
+                        ),
+                    BasisStateLabelingConvention.BigEndian =>
+                        System.Convert.ToInt64(
+                            String.Concat(
+                                System.Convert.ToString(index, 2).PadLeft(NQubits, '0').Reverse()
```
DmitryVasilevsky, line 233: The case above should probably also be written in one line.

DmitryVasilevsky (approved):

src/Simulation/Simulators/QuantumSimulator/StateDumper.cs
kuzminrobin: This code has been moved from IQSharp repo. 
@cgranade, can you tell what that fragment means?

src/Simulation/Simulators/QuantumSimulator/StateDumper.cs
kuzminrobin: This code has been moved from IQSharp repo too.
Based on the comments I assume that for numerical values (regardless of the endianness) the numerical comparer is used, but for the string values - the lexicographical comparer.
(@cgranade, please correct me if I'm wrong)

src/Simulation/Simulators/QuantumSimulator/StateDumper.cs
kuzminrobin: Makes sense. If I'll be making more changes (with subsequent CI rebuild) then I'll make this change.
But if the build succeeds then I'd prefer to merge.

src/Simulation/Simulators/QuantumSimulator/StateDumper.cs
cgranade: This fragment is a fairly standard JSON serialization attribute that specifies enum values are converted as strings, not as ints.

src/Simulation/Simulators/QuantumSimulator/StateDumper.cs
cgranade: Big-endian and little-endian are both used as different conventions for producing an int from a bitstring, but once you have an int, natural ordering on integers applies from there in both cases.

src/Simulation/Simulators/QuantumSimulator/StateDumper.cs
cgranade: `amplitude` can be replaced by the `cmplx.Amplitude` property and `angle` can be replaced by `Amplitude.Phase`.

kuzminrobin: FYI (I have just noticed your last note in this discussion):
```c#
var amplitude = cmplx.Amplitude;
```
Compiler error:
'Complex' does not contain a definition for 'Amplitude' and no accessible extension method 'Amplitude' accepting a first argument of type 'Complex' could be found (are you missing a using directive or an assembly reference?) [Microsoft.Quantum.Simulators]csharp(CS1061)

I also fail to find "Amplitude" in [Complex](https://docs.microsoft.com/en-us/dotnet/api/system.numerics.complex?view=netcore-3.1), its [Properties](https://docs.microsoft.com/en-us/dotnet/api/system.numerics.complex?view=netcore-3.1#properties), etc.

cgranade: Sorry, I should have written `Magnitude` instead of `Amplitude`.

kuzminrobin: Ah, now I understand. Thank you.
From the very beginning, prior to my change, the [line](https://github.com/microsoft/qsharp-runtime/blob/5ea517ccd953bfea1609b6ebe176bde66576830c/src/Simulation/Simulators/QuantumSimulator/StateDumper.cs#L149) 
```c#
var amplitude = (real * real) + (img * img);
```
should have been
```c#
var magnitude = Math.Sqrt(real * real + img * img); // `Math.Sqrt()` was missing, and named magitude.
```
I'll simplify during stage 3.

src/Simulation/Simulators/QuantumSimulator/StateDumper.cs
DmitryVasilevsky: And having this in a comment would be nice.

kuzminrobin: I'm leaving this and a few more items as non-resolved. I will address those at stage 3 (see PR description).

PR #870, opened 2021-11-11
Use cargo-edit instead of dotnet-script to inject versions.

cgranade:
This PR changes version injection into Rust crates to use `cargo-edit` instead of `dotnet-script`, removing the need to install .NET 5.0–specific tooling.


ricardo-espinoza:
It seems that we'd need to add a toml for this, since we're getting:

`Command failed due to unhandled error: Unable to find Cargo.toml`

cgranade:
> It seems that we'd need to add a toml for this, since we're getting:
> 
> `Command failed due to unhandled error: Unable to find Cargo.toml`

Ah, I thought I had, sorry, but I forgot it was on gitignore from the previous approach, such that it didn't get added to the commit right. Thanks for catching that!

ricardo-espinoza:
> > It seems that we'd need to add a toml for this, since we're getting:
> > `Command failed due to unhandled error: Unable to find Cargo.toml`
> 
> Ah, I thought I had, sorry, but I forgot it was on gitignore from the previous approach, such that it didn't get added to the commit right. Thanks for catching that!

I believe that there are still some issues with that file, judging by the build error:

```
error: 1 files in the working directory contain changes that were not yet committed into git:
       Cargo.toml

to proceed despite this and include the uncommitted changes, pass the `--allow-dirty` flag
```

cgranade:
> I believe that there are still some issues with that file, judging by the build error:
> 
> ```
> error: 1 files in the working directory contain changes that were not yet committed into git:
>        Cargo.toml
> 
> to proceed despite this and include the uncommitted changes, pass the `--allow-dirty` flag
> ```

Makes sense that would only come up in CI, yeah... not trying to publish during most local development. Let me see if I can fix by adding `--allow-dirty` as they suggest. In general, Rust practice would be to have the version number committed into the repo, but in general that runs counter to our CI strategy leading to this kind of minor friction.

cgranade:
Build failures seem to be related to macOS failures more generally, and not to changes in this PR; let's rerun once those are fixed, perhaps?

ricardo-espinoza (approved):

raghavthind2005 (approved):

PR #878, opened 2021-11-19
Create Rust QIR Runtime Wrapper Library

idavis:
- QIR Runtime build is now done through cmake instead of PowerShell configuration.
- microsoft-quantum-qir-runtime-sys package compiles and sets up embedding/linkage
- CC/CXX/RC env vars are no longer needed
- Adds a basic runtime driver for nonadaptive jit execution.

### DIFF
--- a/src/Qir/microsoft-quantum-qir-runtime-sys/build.rs
+++ b/src/Qir/microsoft-quantum-qir-runtime-sys/build.rs
@@ -3,25 +3,38 @@
 
 use cmake::Config;
 use std::boxed::Box;
+use std::env;
 use std::error::Error;
 
 fn main() -> Result<(), Box<dyn Error>> {
     println!("cargo:rerun-if-env-changed=TARGET");
     println!("cargo:rerun-if-changed=build.rs");
 
-    let path_to_runtime_src = "../Runtime";
-    compile_runtime_libraries(path_to_runtime_src)?;
+    for (key, value) in env::vars() {
+        println!("{}: {}", key, value);
+    }
+    if cfg!(target_os = "windows") {
+        let path_to_runtime_src = "..\\Runtime";
+        compile_runtime_libraries(path_to_runtime_src)?;     
+    } else {
+        let path_to_runtime_src = "../Runtime";
+        compile_runtime_libraries(path_to_runtime_src)?;
+    }
 
     Ok(())
 }
 
 fn compile_runtime_libraries(path_to_runtime_src: &str) -> Result<(), Box<dyn Error>> {
     let mut config = Config::new(path_to_runtime_src);
+    
+    if cfg!(target_os = "windows") {
+        config.static_crt(true);
+    }
 
     set_compiler(&mut config);
     set_profile(&mut config)?;
 
-    config.generator("Ninja").no_build_target(true);
+    config.generator("Ninja");
 
     let _ = config.build();
     Ok(())
@@ -31,11 +44,21 @@ fn compile_runtime_libraries(path_to_runtime_src: &str) -> Result<(), Box<dyn Er
 // We set this here as setting it in the cmakefile is discouraged
 fn set_compiler(config: &mut Config) {
     if cfg!(target_os = "linux") {
-        config.define("CMAKE_C_COMPILER", "clang-11");
-        config.define("CMAKE_CXX_COMPILER", "clang++-11");
+        let mut c_cfg = cc::Build::new();
+        c_cfg.compiler("clang-11");
+        config.init_c_cfg(c_cfg);
+
+        let mut cxx_cfg = cc::Build::new();
+        cxx_cfg.compiler("clang++-11");
+        config.init_cxx_cfg(cxx_cfg);
     } else if cfg!(target_os = "windows") {
-        config.define("CMAKE_C_COMPILER", "clang");
-        config.define("CMAKE_CXX_COMPILER", "clang++");
+        let mut c_cfg = cc::Build::new();
+        c_cfg.compiler("clang.exe");
+        config.init_c_cfg(c_cfg);
+        
+        let mut cxx_cfg = cc::Build::new();
+        cxx_cfg.compiler("clang++.exe");
+        config.init_cxx_cfg(cxx_cfg);
     } else if cfg!(target_os = "macos") {
         // Use macos default
     } else {
@@ -45,5 +68,7 @@ fn set_compiler(config: &mut Config) {
 
 fn set_profile(config: &mut Config) -> Result<(), Box<dyn Error>> {
     config.define("CMAKE_BUILD_TYPE", "RelWithDebInfo");
+    config.define("CMAKE_C_COMPILER_WORKS", "1");
+    config.define("CMAKE_CXX_COMPILER_WORKS", "1");
     Ok(())
 }
--- a/src/Qir/microsoft-quantum-qir-runtime-sys/build.rs
+++ b/src/Qir/microsoft-quantum-qir-runtime-sys/build.rs
@@ -1,0 +1,49 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use cmake::Config;
+use std::boxed::Box;
+use std::error::Error;
+
+fn main() -> Result<(), Box<dyn Error>> {
+    println!("cargo:rerun-if-env-changed=TARGET");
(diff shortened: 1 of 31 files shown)


kuzminrobin:
(Feel free to ignore, I don't know Rust and the background of this effort)

> src/Qir/microsoft-quantum-qir-runtime-sys

The "sys" part seems to me not very informative (does not tell me the purpose or the structure/contents). I would consider a more informative name instead of "sys".
The name "microsoft-quantum-qir-runtime-sys" seems to me somewhat redundantly long name for a dir, where the part "microsoft-quantum-qir" does not add any info. However if you want to keep the dir name in sync with the package name, then that totally makes sense.

kuzminrobin:
(Feel free to ignore on the same reasons as in the previous comment)
In the dir "src" there seem to be the files of different purpose. 
Some files (or parts of the files) seem to be a part of the build infrastructure, 
some files (or parts) seem to be a regular source code, 
and some parts seem to be the tests.
Placing the regular source code and the tests for it in the same file (and the same dir) may be a common practice in Rust (I don't know), but as for the files that are a part of the build infra, I would consider separating/moving them to a different dir from sources and tests.

src/Qir/microsoft-quantum-qir-runtime-sys/build.rs
```diff
@@ -0,0 +1,79 @@
...
+    for (key, value) in env::vars() {
+        println!("{}: {}", key, value);
+    }
+    if cfg!(target_os = "windows") {
+        let path_to_runtime_src = "..\\Runtime";
+        compile_runtime_libraries(path_to_runtime_src)?;     
+    } else {
+        let path_to_runtime_src = "../Runtime";
+        compile_runtime_libraries(path_to_runtime_src)?;
+    }
```
kuzminrobin, line 22: (Nit)
Consider deduplicating as much as possible, such that only the differing part is placed in `if .. else ..`. Something like this (pseudocode):
```suggestion
    let path_separator = (cfg!(target_os = "windows") ? "\\" : "/");  // Either \ or /
    let path_to_runtime_src = ".." + path_separator + "Runtime";   // "..\Runtime" or "../Runtime"
    compile_runtime_libraries(path_to_runtime_src)?;
```

src/Qir/microsoft-quantum-qir-runtime-sys/build.rs
```diff
@@ -0,0 +1,79 @@
...
+
+    set_compiler(&mut config)?;
+    set_profile(&mut config)?;
+
+    config.generator("Ninja");
+
+    let _ = config.build();
+    Ok(())
+}
```
kuzminrobin, line 41: (Nit. I don't know Rust. Talking in C/C++ terminology)
I have an impression that the function always returns `Ok(())`. If true, then doesn't it make sense to return `void`?

src/Qir/microsoft-quantum-qir-runtime-sys/build.rs
```diff
@@ -0,0 +1,79 @@
...
+        
+        let mut cxx_cfg = cc::Build::new();
+        let clangpp = which::which("clang++.exe")?;
+        cxx_cfg.compiler(clangpp);
+        config.init_cxx_cfg(cxx_cfg);
+    } else if cfg!(target_os = "macos") {
+        // Use macos default
+    } else {
+        panic!("Unsupported platform")
+    }
```
kuzminrobin, line 70: (Nit. . . .)
Consider dedulplicating (such that only the difference is in `if .. else..`), something like (Rust/C/C++ pseudocode)
```suggestion
if cfg!(target_os != "macos") {
  if cfg!(target_os != "linux") && cfg!(target_os != "windows"){
        panic!("Unsupported platform")
  }
  let clang = which::which(
      cfg!(target_os = "linux") ? "clang-11" : "clang.exe")?;
  let mut c_cfg = cc::Build::new();
  c_cfg.compiler(clang);
  config.init_c_cfg(c_cfg);
  
  let clangpp = which::which(
      cfg!(target_os = "linux") ? "clang++-11" : "clang++.exe")?;
  let mut cxx_cfg = cc::Build::new();
  cxx_cfg.compiler(clangpp);
  config.init_cxx_cfg(cxx_cfg);
}
```

src/Qir/microsoft-quantum-qir-runtime-sys/src/foundation.rs
```diff
@@ -0,0 +1,59 @@
...
+const FOUNDATION_BYTES: &'static [u8] = include_bytes!(concat!(
+    env!("OUT_DIR"),
+    "/build/lib/QSharpFoundation/libMicrosoft.Quantum.Qir.QSharp.Foundation.dylib"
+));
+
+#[cfg(target_os = "windows")]
+const FOUNDATION_BYTES: &'static [u8] = include_bytes!(concat!(
+    env!("OUT_DIR"),
+    "/build/lib/QSharpFoundation/Microsoft.Quantum.Qir.QSharp.Foundation.dll"
+));
```
kuzminrobin, line 24: Just FYI (I'm not saying whether it is good or bad):
I have seen a case where a single name for the dynamic library (written in Rust?) is used for all 3 platforms (in C#). See [here](https://github.com/microsoft/qsharp-runtime/blob/1334dc8cefb447e65feca66c463bcd77421bd5a2/src/Simulation/Simulators/OpenSystemsSimulator/NativeInterface.cs#L55), [here](https://github.com/microsoft/qsharp-runtime/blob/f6d9cf0292cf38bffff48d83fc1151207ed34429/build/pack.ps1#L32-L40), and [here](https://github.com/microsoft/qsharp-runtime/blob/5604fc6bdf31eb36e3e8ec577f416fc2426fcdee/src/Simulation/Common/Simulators.Dev.props#L35-L50). Written by Chris.

If that is not applicable to Rust then is it possible to write like this? (deduplicated)
```rust
const FOUNDATION_BYTES: &'static [u8] = include_bytes!(concat!(
    env!("OUT_DIR"),
#[cfg(target_os = "linux")]
    "/build/lib/QSharpFoundation/libMicrosoft.Quantum.Qir.QSharp.Foundation.so"
#[cfg(target_os = "macos")]
    "/build/lib/QSharpFoundation/libMicrosoft.Quantum.Qir.QSharp.Foundation.dylib"
#[cfg(target_os = "windows")]
    "/build/lib/QSharpFoundation/Microsoft.Quantum.Qir.QSharp.Foundation.dll"
));
```

kuzminrobin (approved):

src/Qir/microsoft-quantum-qir-runtime-sys/src/foundation.rs
cgranade: To clarify, I did that due to requirements in our CI and packaging pipeline for .NET projects; I wouldn't recommend following my approach in this case.

src/Qir/microsoft-quantum-qir-runtime-sys/build.rs
cgranade: In Rust, expressions and functions always must return a value (by contrast with `void`, which is the absence of any value). Typically, a function with no meaningful return value will be marked as returning `()` (similar to Q#'s `Unit` or F#'s `unit`). Here, the return type is `Result<(), Box<dyn Error>>`, meaning it can either be `Ok(())` or `Err(t)` for some `t` of type `Box<dyn Error>`.

All that leads up to that `Ok(())` _is_ the Rust equivalent to a `void` return type in C++.

As for why it seems like it always returns `Ok(())`, the `?` operator at lines 34 and 35 cause early `return`s when either value is an `Error`. That's somewhat similar to exceptions, but has the benefit of being much more strongly typed, as it's not possible for a caller to ignore exceptions and as there's no use of exceptional control flow needed.

cgranade:
> (Feel free to ignore, I don't know Rust and the background of this effort)
> 
> > src/Qir/microsoft-quantum-qir-runtime-sys
> 
> The "sys" part seems to me not very informative (does not tell me the purpose or the structure/contents). I would consider a more informative name instead of "sys". The name "microsoft-quantum-qir-runtime-sys" seems to me somewhat redundantly long name for a dir, where the part "microsoft-quantum-qir" does not add any info. However if you want to keep the dir name in sync with the package name, then that totally makes sense.

The suffix `-sys` in crate names is somewhat of a Rust idiom, and carries with it the meaning that the crate acts to wrap a non-Rust library in a Rust package.

Issue #876, opened 2021-11-18
Upgrade GCC to a version compatible with latest Mac


kuzminrobin:
The latest Mac image (11.6 ?) conflicts with g++-7 that we use currently.
If we just update the g++ to g++-11 then the debug version of the full state simulator, compiled with g++-11, crashes on Mac (`microsoft.qsharp-runtime.sanitized (Build mac)` pipeline) when running these 2 QIR tests (they and only they use the full state simulator):
```
1: Test command: /Users/runner/work/1/s/src/Qir/Tests/bin/Debug/FullstateSimulator/fullstate-simulator-tests "~[skip]" "-o" "fullstate-simulator-tests_results.xml" "-r" "junit"
1: Environment variables: 
1:  LD_LIBRARY_PATH=/Users/runner/work/1/s/build/../src/Simulation/Native/build/drop:/Users/runner/work/1/s/src/Qir/Tests/bin/Debug/bin:/Users/runner/work/1/s/src/Qir/Tests/../Runtime/bin/Debug/bin:
1:  PATH=/Users/runner/work/1/s/build/../src/Simulation/Native/build/drop;/Users/runner/work/1/s/src/Qir/Tests/bin/Debug/bin;/Users/runner/work/1/s/src/Qir/Tests/../Runtime/bin/Debug/bin;
1:  DYLD_LIBRARY_PATH=/Users/runner/work/1/s/build/../src/Simulation/Native/build/drop:/Users/runner/work/1/s/src/Qir/Tests/bin/Debug/bin:/Users/runner/work/1/s/src/Qir/Tests/../Runtime/bin/Debug/bin:
1: Test timeout computed to be: 1500
1/4 Test #1: fullstate-simulator-tests ........***Exception: Illegal  0.62 sec
test 2
    Start 2: qir-dynamic-tests

2: Test command: /Users/runner/work/1/s/src/Qir/Tests/bin/Debug/QIR-dynamic/qir-dynamic-tests "~[skip]" "-o" "qir-dynamic-tests_results.xml" "-r" "junit"
2: Environment variables: 
2:  LD_LIBRARY_PATH=/Users/runner/work/1/s/build/../src/Simulation/Native/build/drop:/Users/runner/work/1/s/src/Qir/Tests/bin/Debug/bin:/Users/runner/work/1/s/src/Qir/Tests/../Runtime/bin/Debug/bin:
[...]

PR #881, opened 2021-11-24
Migrate native simulator from GCC to Clang.

kuzminrobin:
See #876.
This PR started as upgrading the GCC and ended up in migrating from GCC to Clang.


src/Simulation/Native/build-native-simulator.ps1
```diff
@@ -10,25 +10,67 @@ if (-not (Test-Path $nativeBuild)) {
...
 {
-    Write-Host "On MacOS build native simulator using gcc (needed for OpenMP)"
-    # `gcc`on Darwin seems to be a shim that redirects to AppleClang, to get real gcc, must point to the specific
-    # version of gcc we've installed.
-    cmake -D BUILD_SHARED_LIBS:BOOL="1" -D CMAKE_C_COMPILER=gcc-7 -D CMAKE_CXX_COMPILER=g++-7 -D CMAKE_BUILD_TYPE="$Env:BUILD_CONFIGURATION" ..
+    Write-Host "On MacOS build using the default C/C++ compiler (should be AppleClang)"
+
+    $OPENMP_PATH="/usr/local/opt/libomp"
+    $OPENMP_COMPILER_FLAGS="-Xpreprocessor -fopenmp -I$OPENMP_PATH/include -lomp -L$OPENMP_PATH/lib"
+    $OPENMP_LIB_NAME="omp"
```
swernli, line 66: Not necessary for this change, but @idavis did some work on the QIR CMakeLists.txt to move these kinds of flags from the powershell script into the cmake, such that it becomes easier to compile without needing the powershell script (like from cargo or cmake directly). We don't need that immediately, but when we do want to enable that we should move these options into an `if (APPLE)` section in CMakeLists.txt.

src/Simulation/Native/prerequisites.ps1
```diff
@@ -2,15 +2,29 @@
...
+    }
+    if (!(Get-Command cmake -ErrorAction SilentlyContinue)) {
+        choco install cmake
+    }
+    refreshenv
+}
+else {
+    if (Get-Command sudo -ErrorAction SilentlyContinue) {
+        sudo apt update
+        sudo apt-get install -y clang-11
```
swernli, line 23: I think this should include `ninja-build` for completeness (like https://github.com/microsoft/qsharp-runtime/blob/main/src/Qir/Runtime/prerequisites.ps1#L33)

swernli: same below in the other branch.

swernli (commented):
One minor point of feedback, so I'm not approving yet to avoid triggering the auto-merge.

src/Qir/Runtime/public/QubitManager.hpp
```diff
@@ -40,7 +40,7 @@ namespace Quantum
         // No complex scenarios for now. Don't need to support copying/moving.
         CQubitManager(const CQubitManager&) = delete;
         CQubitManager& operator=(const CQubitManager&) = delete;
-        virtual ~CQubitManager();
+        ~CQubitManager();
```
DmitryVasilevsky, line 43: This is very suspicious and against recommendations. Can you write your experience with crashes in comments or file it as an issue.

src/Qir/Samples/StandaloneInputReference/CMakeLists.txt
```diff
@@ -34,7 +34,7 @@ add_test(
 if(DEFINED ENV{NATIVE_SIMULATOR})
     set(TEST_DEPS1 $ENV{NATIVE_SIMULATOR})
 else()
-    set(TEST_DEPS1 "${PROJECT_SOURCE_DIR}/../../Simulation/native/build/$ENV{BUILD_CONFIGURATION}")
+    set(TEST_DEPS1 "${PROJECT_SOURCE_DIR}/../../Simulation/native/build/drop")
```
DmitryVasilevsky, line 37: Will this result in multiple build configurations mixed together?

src/Simulation/Native/prerequisites.ps1
kuzminrobin: I noticed that on Linux and Mac the CMake uses Clang (either by default on Mac or the one specified in the command line args on Linux), without Ninja.
But on Windows, without Ninja the CMake continues to use MSVC even if the compiler is specified as a command line arg. The only way I found to switch the CMake on Windows from MSVC to Clang is by using the Ninja generator (if you know a better way then feel free to share). That is why _for native simulator_ Ninja is only used on Win.
I believe it also makes sense to exclude Ninja _for QIR RT and its tests_ on Mac and Linux. I opened #882 for that.

swernli: I'll comment on the issue so we can discuss more there, but I don't think I agree. I prefer to use the same generator across all platforms so we can have a more uniform configuration experience. I'll elaborate on the issue itself and then once we determine what course we'd prefer to take either remove Ninja for the QIR Runtime or add it here. Either way, no changes for this PR are required.

src/Qir/Samples/StandaloneInputReference/CMakeLists.txt
kuzminrobin: No.
Previously the native simulator (NS) was built with GCC on Mac and Linux, the GCC was putting the binaries (of NS and its tests) to 
`Simulation/native/build`, the subsequent script lines were copying the binaries to 
`Simulation/native/build/drop`.
But on Windows the MSVC was putting the binaries to 
`Simulation/native/build/Release` (all the CI pipelines build the _Release config_ of all, the only pipeline that builds Debug config is the [`microsoft.qsharp-runtime.sanitized`](https://dev.azure.com/ms-quantum-public/Microsoft%20Quantum%20(public)/_build?definitionId=34&_a=summary) pipeline (running on Mac and Linux only), but still the Release and Debug binaries are never generated on the same machine).

Now the NS is built with Clang on all platforms, and Clang puts the binaries to `Simulation/native/build` on all platforms (and the subsequent script lines still copy the binaries to `Simulation/native/build/drop`).

src/Qir/Runtime/public/QubitManager.hpp
kuzminrobin: As for recommendations. Yes, I know there is a recommendation that encourages making the destructors _virtual_. But there is also a refinement (to that recommendation) that instructs to avoid making the destructor virtual if there is no any other virtual function, see [Item 7: Declare destructors virtual in polymorphic base classes](https://aovst.cs.msu.ru/morda/effective_cpp.pdf) (search for "If a class does not contain virtual functions" on paper page 41).
On the other hand, yes, I agree that there is nothing wrong with the code. There is nothing resulting in a crash if the destructor is the only virtual member function of the class.
I'll add the comment in the code.

kuzminrobin:
```c++
        ~CQubitManager();   // If this dtor is made _virtual_ then the QIR RT tests crash (at least in Debug config)
          // if the native simulator is compiled with Clang (as opposed to GCC). Nothing wrong found in the code,
          // probably is the compiler bug.
```

src/Simulation/Native/build-native-simulator.ps1
kuzminrobin: Transferred the OpneMP flags to CMake.

swernli (approved):

bettinaheim:
@kuzminrobin Please take a look at the performance tests to double check that the change didn't have any negative performance impact.

PR #887, opened 2021-12-03
Adding ability to provide content type and encoding for Azure blobs

anpaz:
The `StorageHelper` classes for the `Azure.Client` package does not provide a way for clients to specify the content type of the payload and/or when the payload is compressed, specify its encoding as `gzip`.
This is going to be problematic soon as there are plans on the service to start enforcing that both of these settings are correctly set on the blob and be smarter about a Job's payload.
With these changes, a client can now provide the Job's payload content type and to upload it compress, in which case the encoding will be correctly set automatically.


src/Azure/Azure.Quantum.Client/Storage/IStorageHelper.cs
```diff
@@ -32,14 +32,32 @@ public interface IStorageHelper
...
+        /// <param name="contentType">The MIME type indicating the content of the payload.</param>
+        /// <param name="contentEncoding">The blob encoding.</param>
         /// <param name="cancellationToken">The cancellation token.</param>
         /// <returns>async task.</returns>
         Task UploadBlobAsync(
             BlobContainerClient containerClient,
             string blobName,
             Stream input,
+            string contentType,
+            string contentEncoding,
```
cesarzc, line 44:
```suggestion
            string? contentType,
            string? contentEncoding,
```

cesarzc: Given that these two parameters can be null, it would be good to make them nullable.

cesarzc (approved):

PR #889, opened 2021-12-09
Re-trying the PR #881: Migrating native sim from GCC to Clang.

kuzminrobin:
A set of 3 PRs: [Q#RT](https://github.com/microsoft/qsharp-runtime/pull/889), [QDK](https://github.com/microsoft/qdk/pull/194), [IQ#](https://github.com/microsoft/iqsharp/pull/553) repos (see also [quantum-docs PR](https://github.com/MicrosoftDocs/quantum-docs/pull/427)).
A set of the latter 2 are expected to require a branch-specific e2e build or a merge of the former one first.


anpaz (commented):
lgtm.

kuzminrobin:
Please review.

src/Qir/Common/cmake/unit_test_include.cmake
```diff
@@ -9,7 +9,7 @@ macro(add_unit_test target)
     if(DEFINED ENV{NATIVE_SIMULATOR})
         set(TEST_DEPS1 $ENV{NATIVE_SIMULATOR})
     else()
-        set(TEST_DEPS1 "${PROJECT_SOURCE_DIR}/../../Simulation/native/build/$ENV{BUILD_CONFIGURATION}")
+        set(TEST_DEPS1 "${PROJECT_SOURCE_DIR}/../../Simulation/native/build/drop")
```
idavis, line 12: @kuzminrobin Why did this path change? Seems like the CI should have had issues before or after.

kuzminrobin: When the native simulator was built with MSVC (on Win), the compiler was putting the binary to
`Simulation/native/build/Release` (whereas the GCC on other platforms was putting the binary to 
`Simulation/native/build`). Then the scripts were copying the binary to 
`Simulation/native/build/drop` (on all platforms), but the dependency in the tests was different for win and other platforms. 
After the native simulator has been migrated to Clang, on all the platforms the binary started being put 
`Simulation/native/build` and copied to 
`Simulation/native/build/drop`. This change has unified the dependency.

kuzminrobin:
Any more feedback please? 
Any approvals?

src/Simulation/Native/build-native-simulator.ps1
```diff
@@ -10,25 +10,62 @@ if (-not (Test-Path $nativeBuild)) {
...
+    cmake -G Ninja -D BUILD_SHARED_LIBS:BOOL="1" $CMAKE_C_COMPILER $CMAKE_CXX_COMPILER `
+        -D CMAKE_BUILD_TYPE="$Env:BUILD_CONFIGURATION"  ..
+        # Without `-G Ninja` we fail to switch from MSVC to Clang.
+        # Sanitizers are not supported on Windows at the moment. Check again after migrating from Clang-11.
 }
 elseif (($IsLinux) -or ((Test-Path Env:AGENT_OS) -and ($Env:AGENT_OS.StartsWith("Lin"))))
 {
-    Write-Host "On Linux build native simulator using gcc (needed for OpenMP)"
-    cmake -D BUILD_SHARED_LIBS:BOOL="1" -D CMAKE_C_COMPILER=gcc -D CMAKE_CXX_COMPILER=g++ -D CMAKE_BUILD_TYPE="$Env:BUILD_CONFIGURATION" ..
+    cmake -D BUILD_SHARED_LIBS:BOOL="1" -D CMAKE_C_COMPILER=clang-11 -D CMAKE_CXX_COMPILER=clang++-11 `
```
swernli, line 56: Looks like sanitizer flags are being added even for release builds, which we don't want. The sanitizers should only be enabled for debug builds or only in the sanitizer pipeline, we don't want to ship binaries on linux and mac that have sanitizers enabled.

swernli (changes requested):

src/Simulation/Native/build-native-simulator.ps1
kuzminrobin: Not exactly, The sanitizer flags are only added for DEBUG, see below:
CMAKE_C_FLAGS_**DEBUG**="$SANITIZE_FLAGS"
CMAKE_CXX_FLAGS_**DEBUG**="$SANITIZE_FLAGS"

swernli: Ah, I misread, you are correct!

kuzminrobin: Can you please uncheck the "Change requested" flag?

swernli (approved):

PR #895, opened 2022-01-10
Cherry-Picked: Broken up QuantumSimulator into CommonNativeSimulator and QuantumSimulator

kuzminrobin:
Cherry-picked to `main` the feature-branch changes of these 2 PRs:
* #853: Broken up QuantumSimulator into CommonNativeSimulator and QuantumSimulator,
* #857: Reverting the Allocate() change.


swernli:
@kuzminrobin What is the intent of this change? Does the additional layer of indirection make the infrastructure more compatible with the sparse simulator? The changes seem reasonable, it's just hard to tell from the files what this enables.

kuzminrobin:
@swernli wrote:
> @kuzminrobin What is the intent of this change? Does the additional layer of indirection make the infrastructure more compatible with the sparse simulator? The changes seem reasonable, it's just hard to tell from the files what this enables.

Brief extract from an email that I have written to Bettina and Cassandra:
I would like to cherry-pick from the feature branch `feature/sparse-simulator` to the `main` the change that extracts from the QuantumSimulator the part common for QuantumSimulator and SparseSimulator – CommonNativeSimulator (see this [PR](https://github.com/microsoft/qsharp-runtime/pull/853)). It just seems more natural to simplify the Dump _on top of that change_, because some of the code fragments need to relocate from IQSharp to the class common for QuantumSimulator and SparseSimulator (rather than to QuantumSimulator).
Then I plan to re-do the Dump changes on the QuantumSimulator side (with improvements that move some code to CommonNativeSimulator).

kuzminrobin:
@swernli wrote:
> @kuzminrobin What is the intent of this change? Does the additional layer of indirection make the infrastructure more compatible with the sparse simulator? The changes seem reasonable, it's just hard to tell from the files what this enables.

This is preparation for SparseSimulator, which has common part with QuantumSimulator. That common part is being extracted into a common parent - CommonNativeSimulator.

swernli (approved):
Thanks for the additional context!

src/Simulation/Simulators/CommonNativeSimulator/QubitManager.cs
```diff
@@ -8,46 +8,48 @@
...
-
             /// <summary>
             ///  The max number used as qubit id so far.
             /// </summary>
             public long MaxId { get; private set; }
 
             public override Qubit CreateQubitObject(long id)
             {
-                Debug.Assert(id < 50, "Using a qubit id > 50. This is a full-state simulator! Validating ids uniquenes might start becoming slow.");
+                Debug.Assert(id < 50, "Using a qubit id > 50. This is a full-state simulator! Validating ids uniqueness might start becoming slow.");
```
DmitryVasilevsky, line 34: Sparse simulator may handle more than 50 qubits.

src/Simulation/Simulators/CommonNativeSimulator/CommonNativeSimulator.cs
```diff
@@ -0,0 +1,191 @@
...
+        ///         - none of the qubits are null
+        ///         - there are no duplicated qubits
+        ///     Also sets the isMeasured flag to false for each qubit
+        /// </summary>
+        bool[] CheckQubits(IQArray<Qubit> targets)
+        {
+            if (targets == null) throw new ArgumentNullException(nameof(targets), "Trying to perform an intrinsic operation on a null Qubit array.");
+            if (targets.Length == 0) throw new ArgumentNullException(nameof(targets), "Trying to perform an intrinsic operation on an empty Qubit array.");
+
+            bool[] used = new bool[((QSimQubitManager)QubitManager).MaxId];
```
DmitryVasilevsky, line 118: Here and in a few other places. For the full state simulator the maximum number of qubits were small. Sparse simulator can handle quite a few more in certain circumstances. It may not be always desirable to allocate array with the size of the number of qubits. I would consider array for a full state simulator and use HashSet for the sparse simulator.

src/Simulation/Simulators/CommonNativeSimulator/QubitManager.cs
kuzminrobin: So what should be the number here instead of 50?

src/Simulation/Simulators/CommonNativeSimulator/CommonNativeSimulator.cs
kuzminrobin: I didn't touch the SparseSimulator for a few months and I don't remember exactly its difference from the QuantumSimulator. And before I get back to the SparseSimulator I also need to resolve the Dump functionality issue. So paying significant attention to the SparseSim at this stage seems to me less efficient than if I move forward now. I will get back to SparseSim later any way, and that's when I plan to finalize the changes.
If there are no critical objections, I would prefer to move forward with this PR at the moment.

src/Simulation/Simulators/CommonNativeSimulator/QubitManager.cs
DmitryVasilevsky: Theoretically there's no artificial limit. You may have millions of qubits as long as they are all in zero state. However, it's hard to imagine a meaningful program that would have that many in the zero state. I believe there're tests that check functionality of the simulator with a few thousand qubits. But the point is - the check for full state simulator and sparse simulator should be substantially different.

src/Simulation/Simulators/CommonNativeSimulator/CommonNativeSimulator.cs
DmitryVasilevsky: I'm OK leaving this as is, especially given that sparse simulator won't be used on more than a few thousand qubits. Maybe we should note this somewhere - comments or work item.

PR #898, opened 2022-01-12
Refactored the Dump functionality on Q#RT side.

kuzminrobin:
Simplified the Dump functionality for QuantumSimulator and the functionality use in IQSharp.
This is a part of the set of 3 PRs:
[Q#RT side](https://github.com/microsoft/qsharp-runtime/pull/898),
[IQ#](https://github.com/microsoft/iqsharp/pull/562),
[Katas](https://github.com/microsoft/QuantumKatas/pull/727).


src/Simulation/Simulators/CommonNativeSimulator/CommonNativeSimulator.cs
```diff
@@ -35,6 +35,8 @@ public abstract partial class CommonNativeSimulator : SimulatorBase, IQSharpCore
 
         public uint Id { get; protected set; }
 
+        public abstract uint[] GetQubitIds();
```
cgranade, line 38: Does this need to be specifically a `uint[]`, or can it be an `IEnumerable<uint>`? May also be worth adding an API documentation comment to indicate what's returned by this method.

src/Simulation/Simulators/CommonNativeSimulator/Dump.cs
```diff
@@ -96,5 +92,393 @@ public QSimDumpRegister(CommonNativeSimulator m) : base(m)
                 return Simulator.Dump(location, qubits);
             };
         }
+
+        /// <summary>
+        ///     The convention to be used in labeling computational basis states
+        ///     given their representations as strings of classical bits.
+        /// </summary>
+        [JsonConverter(typeof(StringEnumConverter))]
+        public enum BasisStateLabelingConvention
```
cgranade, line 101: [nit] As a minor organizational thing, this file is getting a bit large; since the containing class is `partial` anyway, it may be worth splitting this and `DisplayableState` into a new file.

src/Simulation/Simulators/CommonNativeSimulator/Dump.cs
```diff
@@ -96,5 +92,393 @@ public QSimDumpRegister(CommonNativeSimulator m) : base(m)
...
+            [JsonProperty("n_qubits")]
+            public int NQubits { get; set; }
+
+            /// <remarks>
+            ///     These amplitudes represent the computational basis states
+            ///     labeled in little-endian order, as per the behavior of
+            ///     <see cref="Microsoft.Quantum.Simulation.Simulators.CommonNativeSimulator.StateDumper.Dump" />.
+            /// </remarks>
+            [JsonProperty("amplitudes")]
+            public Complex[]? Amplitudes { get; set; }
```
cgranade, line 173: No suggested action at the moment, but when adding the sparse simulator, this will need to be generalized to a dictionary from indices to amplitudes.

src/Simulation/Simulators/CommonNativeSimulator/Dump.cs
```diff
@@ -96,5 +92,393 @@ public QSimDumpRegister(CommonNativeSimulator m) : base(m)
...
+            /// <summary>
+            /// Returns a string that represents the label for the given base state.
+            /// </summary>
+            public virtual string FormatBaseState(uint idx) 
+            {
+                int qubitCount = Amplitudes?.Length ?? 0;
+                int maxCharsStateId = ((1 << qubitCount) - 1).ToString().Length;
+
+                return $"∣{idx.ToString().PadLeft(maxCharsStateId, ' ')}❭";
+            }
```
cgranade, line 257: Should this be `FormatBasisState`? Also, it looks like this method doesn't use the basis state labeling method above, but is fixed to using `BasisStateLabelingConvention.LittleEndian`. I'd actually recommend removing this method entirely, since it duplicates `BasisStateLabel` but is less configurable.

src/Simulation/Simulators/CommonNativeSimulator/Dump.cs
```diff
@@ -96,5 +92,393 @@ public QSimDumpRegister(CommonNativeSimulator m) : base(m)
...
+            /// <summary>
+            /// Returns a string for the amplitude's cartesian representation (real + imagnary).
+            /// </summary>
+            public virtual string FormatCartesian(double real, double img) =>
+                $"{real,9:F6} + {img,9:F6} i";
+
+            /// <summary>
+            /// The method to use to format the amplitude into a string.
+            /// </summary>
+            public virtual string Format(uint idx, double real, double img)
```
cgranade, line 319:
```suggestion
            public virtual string Format(uint idx, double real, double img, BasisStateLabelingConvention convention: BasisStateLabelingConvention.LittleEndian)
```

src/Simulation/Simulators/CommonNativeSimulator/Dump.cs
```diff
@@ -96,5 +92,393 @@ public QSimDumpRegister(CommonNativeSimulator m) : base(m)
...
+            {
+                var amplitude = (real * real) + (img * img);
+                var angle = System.Math.Atan2(img, real);
+
+                return $"{FormatBaseState(idx)}:\t" +
```
cgranade, line 324:
```suggestion
                return $"{BasisStateLabel(convention, idx)}:\t" +
```

src/Simulation/Simulators/CommonNativeSimulator/Dump.cs
cgranade, line 319: Is this method still used? It looks like it's been superseded by `ToString` below?

src/Simulation/Simulators/CommonNativeSimulator/Dump.cs
```diff
@@ -96,5 +92,393 @@ public QSimDumpRegister(CommonNativeSimulator m) : base(m)
...
+            // TODO(rokuzmin): What does it return?
+            public override bool Dump(IQArray<Qubit>? qubits = null)
+            {
+                System.Diagnostics.Debug.Assert(this.Simulator.QubitManager != null, 
+                    "Internal logic error, QubitManager must be assigned");
+
+                _count = qubits == null
+                            ? this.Simulator.QubitManager.AllocatedQubitsCount
+                            : qubits.Length;
+                _data = new Complex[1 << ((int)_count)];    // If 0 qubits are allocated then the array has
```
cgranade, line 448: No suggested action at the moment, but this will likely need to be changed to a dictionary ass well once the sparse simulator is added.

src/Simulation/Simulators/CommonNativeSimulator/Dump.cs
```diff
@@ -96,5 +92,393 @@ public QSimDumpRegister(CommonNativeSimulator m) : base(m)
...
+                _data = new Complex[1 << ((int)_count)];    // If 0 qubits are allocated then the array has 
+                                                            // a single element. The Hilbert space of the system is 
+                                                            // ℂ¹ (that is, complex-valued scalars).
+                var result = base.Dump(qubits);
+
+                // At this point, _data should be filled with the full state
+                // vector, so let's display it, counting on the right display
+                // encoder to be there to pack it into a table.
+
+                // var id = System.Guid.NewGuid();
```
cgranade, line 457: Should this be removed now?

src/Simulation/Simulators/QuantumSimulator/Dump.cs
```diff
@@ -26,19 +26,9 @@ public uint[] QubitIds
             }
         }
 
-        protected override QVoid process(Action<string> channel, IQArray<Qubit>? qubits)
+        public override uint[] GetQubitIds()
```
cgranade, line 29: I'd suggest an expression-body method here:

```csharp
public override uint[] GetQubitIds() =>
    QubitIds;
```

More to the point, though, it looks like this method duplicates `QubitIds { get; }`. Would it be better to make the abstract class define `QubitIds { get; }` directly?

src/Simulation/Simulators/CommonNativeSimulator/CommonNativeSimulator.cs
cgranade: As mentioned below, it may also be reasonable to make the abstract class define a property directly rather than a method:

```suggestion
        public abstract uint[] QubitIds { get; }
```

src/Simulation/Simulators/CommonNativeSimulator/Dump.cs
```diff
@@ -96,5 +92,393 @@ public QSimDumpRegister(CommonNativeSimulator m) : base(m)
...
+            {
+                if (_data == null) throw new Exception("Expected data buffer to be initialized before callback, but it was null.");
+                _data[idx] = new Complex(real, img);
+                return true;
+            }
+
+            /// <summary>
+            ///     Dumps the state of a register of qubits as a displayable object.
+            /// </summary>
+            // TODO(rokuzmin): What does it return?
```
cgranade, line 439: Should this TODO still be here?

cgranade (commented):
Looks good to me! Had a last few comments, mostly about duplicated functionality in the string output for displayable states, but I think this is getting pretty close. Please let me know how I can help!

cgranade (approved):
Sorry, accidentally hit comment instead of approve.

src/Simulation/Simulators/CommonNativeSimulator/Dump.cs
kuzminrobin: I'm removing `Format()` and I assume that other notes about changing the signature or structure of the `Format()` are not applicable any more.

src/Simulation/Simulators/CommonNativeSimulator/Dump.cs
kuzminrobin: Removing.

src/Simulation/Simulators/CommonNativeSimulator/CommonNativeSimulator.cs
kuzminrobin: Thank you for helping me to master the abstract properties (I'm still new to C# and I'm still thinking in C++). Updating.
As for `uint[]` vs. `IEnumerable<uint>`, I would prefer to postpone the decision until later.

src/Simulation/Simulators/CommonNativeSimulator/Dump.cs
kuzminrobin: Extracted the `DisplayableState` into a separate file.

src/Simulation/Simulators/CommonNativeSimulator/Dump.cs
kuzminrobin: Changed to a dictionary.

src/Simulation/Simulators/CommonNativeSimulator/Dump.cs
kuzminrobin: `Format()` has been removed.

src/Simulation/Simulators/CommonNativeSimulator/Dump.cs
kuzminrobin: `Format()` has been removed.

src/Simulation/Simulators/CommonNativeSimulator/Dump.cs
kuzminrobin: Changed to a dictionary.

src/Simulation/Simulators/CommonNativeSimulator/Dump.cs
kuzminrobin: Done.

src/Simulation/Simulators/QuantumSimulator/Dump.cs
kuzminrobin: Restructured a bit. Thanks!

src/Simulation/Simulators/CommonNativeSimulator/Dump.cs
kuzminrobin: Removed.

src/Simulation/Simulators/CommonNativeSimulator/DisplayableState.cs
```diff
@@ -0,0 +1,243 @@
...
+            ///     threshold are suppressed.
+            /// </param>
+            public IEnumerable<(Complex, string)> SignificantAmplitudes(
+                BasisStateLabelingConvention convention,
+                bool truncateSmallAmplitudes, double truncationThreshold
+            ) =>
+                (
+                    truncateSmallAmplitudes
+                    ? Amplitudes
+                        .Select(idx_amplitude => idx_amplitude)
```
cgranade, line 113: I think this line may not be needed now that `Amplitudes` is a dictionary?

```suggestion
```

In any case, `idx_amplitude` is a `KeyValuePair` with the new definition of `Amplitudes`, such that it's no longer an index:

image

[nit / code style: the name `idx_amplitude` is in `snake_case` but should be in `camelCase`.]

src/Simulation/Simulators/CommonNativeSimulator/DisplayableState.cs
```diff
@@ -0,0 +1,243 @@
...
+                        .Where(item =>
+                            System.Math.Pow(item.Value.Magnitude, 2.0) >= truncationThreshold
+                        )
+                    : Amplitudes.Select(idx_amplitude => idx_amplitude)
```
cgranade, line 117: Similarly here, the call to `Select` that was previously used to attach indices shouldn't be needed any more:

```suggestion
                    : Amplitudes
```

cgranade (approved):
Ah, you went on and modified `DisplayableState` to use a dictionary already! That should make adding in the sparse simulator much easier — your most recent changes should be precisely the logic needed to handle an incomplete sequence of callback calls.

Left a couple last comments about adapting some of `SignificantAmplitudes` to the new property, but otherwise looks good to me.

PR #901, opened 2022-01-14
Ship-room: Added `libomp` instructions.

kuzminrobin:


README.md
```diff
@@ -58,7 +58,7 @@ To build on other platforms:
     * Install [CMake](https://cmake.org/install/)
     * Install [.NET Core 3 SDK](https://dotnet.microsoft.com/download)
     * On [WSL](https://docs.microsoft.com/en-us/windows/wsl/)/Linux:
-      * Install `g++` (e.g. in Ubuntu 20.04 `sudo apt-get install g++`).
+      * Install [`libomp`](https://openmp.llvm.org) needed for the native (C++) full-state simulator.
```
kuzminrobin, line 61: For other platforms the libomp is installed automatically:
Mac: [`src\Simulation\Native\prerequisites.ps1`](https://github.com/microsoft/qsharp-runtime/blob/7736de760cf2ae9b3a099ba8df9713fffe453005/src/Simulation/Native/prerequisites.ps1#L5)
Win: As part of MSVC/Clang.

cgranade: To clarify, this is only the build instructions, right? If I understand correctly, since `libomp` is included in the NuGet package, we should have no user-facing dependencies at all.

cgranade (approved):

README.md
kuzminrobin: Correct.

PR #902, opened 2022-01-18
Fix capitalization in xref links

tcNickolas:
xref links in https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.diagnostics.assertqubitwithintolerance and https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.diagnostics.assertqubit don't render as links. I believe docgen doesn't recognize API links if they are given in lowercase, it needs proper capitalization in xref.


tcNickolas:
The build failure is caused by the link https://github.com/microsoft/qsharp-language/tree/main/Specifications/QIR not being active any longer, probably should be replaced with https://github.com/qir-alliance/qir-spec - let me know if you'd like me to help with that

kuzminrobin (approved):

PR #906, opened 2022-01-20
Update macOS build scripts for libomp dependency

swernli:
By adding the macOS libomp headers alongside the checked in binary, we are able to compile directly against that folder without using `brew install libomp`. Combining this with using `install_name_tool` to update the install id in the library itself, it ensures the built simulator library will search for the dependency using the current `rpath` settings, which now includes the current folder of the binary itself.

This should resolve https://github.com/microsoft/iqsharp/issues/568, and corresponding change to the qdk repo to remove explicit installation of libomp on macOS will confirm.


src/Simulation/Native/prerequisites.ps1
```diff
@@ -2,7 +2,8 @@
 # Licensed under the MIT License.
 
 if (($IsMacOS) -or ((Test-Path Env:AGENT_OS) -and ($Env:AGENT_OS.StartsWith("Darwin")))) {
-    brew install libomp
+    # Skip explicit install for testing purposes
+    # brew install libomp
```
kuzminrobin, line 6: FYI, I'm OK leaving these comment as is.

kuzminrobin (approved):

PR #911, opened 2022-01-23
Consolidate native dlls in a single folder to avoid dupe references (fixes NETSDK1152 error)

anpaz:
With the migration to .net6, the dotnet compiler is not reporting an error because the native simulator dll is picked up from two different locations.
These changes consolidate the native simulators in a single folder by explicitly copying them there, and then selects the right folder depending on the platform.


swernli (approved):
LGTM

Issue #915, opened 2022-01-25
Expm support in qdk_sim_experimental crate


cgranade:
**Please describe what you would like the feature to accomplish.**  
As part of https://github.com/microsoft/qsharp-runtime/issues/914, we will need to implement support for the matrix exponential function in the qdk_sim_experimental crate so as to allow computing noise channels from dynamical generators.

PR #920, opened 2022-01-27
Initial support for matrix inverse in qdk_sim_experimental crate

cgranade:
This PR begins work on #915 by adding a pure-Rust implementation of LU decompositions and matrix inverses, drawing from the MIT-licensed examples at https://github.com/cgranade/cursed-linalg. These linear algebra routines will be needed in order to implement `expm` without passing a LAPACK dependency through to end users.

The PR will appear very messy until #979 is merged, as it includes unrelated changes from `main`.

Note that this PR targets a feature branch, and further work on unit tests, integration tests, and perf benchmarks are expected before that feature is merged to main.


kuzminrobin (approved):
I'm learning Rust and still don't understand most of the code here. But at a glance I didn't see any thing to report. 
Approving to unblock you, but I would recommend an approval from someone more knowledgeable than me.

cgranade:
@kuzminrobin: Thanks! If you're curious about any of the Rust-y parts, I'd be happy to sync with you and go over in more detail.

In the meantime, @swernli and @samarsha, if you have any comments on Rust-specific points I'd greatly appreciate your feedback. Thanks!

src/Simulation/qdk_sim_rs/src/linalg/array_ext.rs
```diff
@@ -0,0 +1,113 @@
...
+use crate::error::QdkSimError;
+
+pub(crate) trait ShapeExt {
+    fn require_square(&self) -> Result<usize, QdkSimError>;
+}
+
+impl<S, D> ShapeExt for ArrayBase<S, D>
+where
+    S: RawData,
+    D: Dimension,
```
samarsha, line 20: So for extension traits there's [RFC 445](https://rust-lang.github.io/rfcs/0445-extension-trait-conventions.html). It looks like the traits in this module don't follow those naming conventions.

src/Simulation/qdk_sim_rs/src/linalg/array_ext.rs
```diff
@@ -0,0 +1,113 @@
...
+{
+    fn require_square(&self) -> Result<usize, QdkSimError> {
```
samarsha, line 22: This is a very general error type but there is only one possible error in this case. Maybe `Option<usize>`?

src/Simulation/qdk_sim_rs/src/linalg/decompositions/lu.rs
```diff
@@ -0,0 +1,293 @@
...
+    A: Scalar,
+{
+    pub(self) factors: ArrayBase<S, Ix2>,
+    pub(self) pivots: Array1<usize>,
+}
+
+/// Represents types that represent matrices that are decomposable into lower-
+/// and upper-triangular parts.
+pub trait LUDecomposable {
+    /// The element type for the output matrices.
```
samarsha, line 29: Are `LUDecompasable` and `LUDecomposition` also extension traits? I only see an impl for one type each. Is it needed for crate consumers to write their own impls?

samarsha (commented):
I've only looked through half of the files so far.

src/Simulation/qdk_sim_rs/README.md
```diff
@@ -148,3 +143,5 @@ TODO
 - Stabilizer simulation not yet exposed via C API.
 - Test and microbenchmark coverage still incomplete.
 - Too many APIs `panic!` or `unwrap`, and need replaced with `Result` returns instead.
+
+# Crate features
```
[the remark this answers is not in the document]
cgranade: Yes, the rest is filled in by the `document-features` crate based on documentation comments in `Cargo.toml` itself.

src/Simulation/qdk_sim_rs/src/linalg/array_ext.rs
cgranade: Ah, that it should be `ArrayBaseExt`?

src/Simulation/qdk_sim_rs/src/linalg/decompositions/lu.rs
cgranade: They're note extension traits, no, but an attempt to generalize 𝐿𝑈 decompositions to allow for application over other kinds of scalars and data structures including precomputed decompositions.

src/Simulation/qdk_sim_rs/README.md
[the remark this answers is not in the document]
samarsha: I see, I assume it's appending it to the end of the file (rather than looking for a specific header name)? A comment in the readme could help avoid accidentally deleting the section or adding sections after it.

swernli (approved):
Some minor feedback below, but overall it looks great!

src/Simulation/qdk_sim_rs/README.md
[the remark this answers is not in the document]
cgranade: That's right, yeah. There's two `#![cfg_attr(..., doc = ...)]` attributes at https://github.com/microsoft/qsharp-runtime/pull/920/files#diff-09741599828fb3d312b03bab5767da1f994d179496ea35a281db89a47339aebfR4, which causes rustdoc to append the contents of each attribute together. In particular, the first `#![doc(...)]` attribute uses `include_str` to bring in README.md, while the second one calls `document_features`.

Will add a note to that effect, then, as you suggest.

src/Simulation/qdk_sim_rs/src/linalg/array_ext.rs
samarsha: Yeah, I think so.

- `ShapeExt` -> `ArrayBaseExt`
- `MatrixExt` -> `ArrayBaseMatrixExt`
- `RemoveAxisExt` -> `ArrayBaseRemoveAxisExt`

(Of course, assuming that there will never be any other impls for these - for example, n-d arrays represented by something other than an ArrayBase. Extension traits are kind of weird since the line between a single-impl trait and a many-impl trait seems blurry.)

src/Simulation/qdk_sim_rs/src/linalg/decompositions/lu.rs
samarsha: If it's a generalization, that would imply there should be multiple impls for each trait, right? Are you planning to add some later?

src/Simulation/qdk_sim_rs/src/tableau.rs
```diff
@@ -91,16 +94,18 @@ impl Tableau {
...
+    pub fn assert_meas(&self, idx_target: usize, expected: bool) -> Result<(), QdkSimError> {
+        let actual = self
+            .determinstic_result(idx_target)
+            .ok_or(QdkSimError::MiscError(format!(
+                "Expected {}, but measurement result would be random.",
+                expected
+            )))?;
         if actual != expected {
-            Err(format!(
+            Err(QdkSimError::MiscError(format!(
```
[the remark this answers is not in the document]
cgranade: My goal here had been to offer more specific error types before stabilizing the API, but at this point it wasn't clear if there should be a general-purpose assertion error, or one more focused to the `actual != expected` format.

src/Simulation/qdk_sim_rs/src/utils.rs
```diff
@@ -31,9 +39,9 @@ fn log_message(msg: &str) {
 }
 
 /// Prints a message as an error, and returns it as a [`Result`].
-pub fn log_as_err<T>(msg: String) -> Result<T, String> {
+pub fn log_as_err<T>(msg: String) -> Result<T, QdkSimError> {
     log(&msg);
-    Err(msg)
+    Err(QdkSimError::MiscError(msg))
```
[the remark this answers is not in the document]
cgranade: The logging API as written only supports returning strings as errors; I suppose that's as good a reason as any to keep a `MiscError` type? Maybe I could weaken the comment to something like "consider more specific types if applicable"?

src/Simulation/qdk_sim_rs/src/linalg/array_ext.rs
cgranade: Sounds good, I'll go on and modify, then.

src/Simulation/qdk_sim_rs/src/linalg/decompositions/lu.rs
cgranade: Not necessarily, but the corresponding split for eigendecompositions has proved useful (allows for precomputed decompositions as well, for example) such that leaving it open for consumers to write their own implementations may be useful as well. That does to some level presuppose using the linear algebra functionality outside of the context of the `qdk_sim` crate, but ideally I'd love to split that out into its own crate.

src/Simulation/qdk_sim_rs/Cargo.toml
```diff
@@ -21,72 +24,84 @@ exclude = [
...
+anyhow = "1.0.56"
+# We use built to expose compile-time metadata about how this crate
+# was built to C and Rust callers.
+built = "0.5.0"
+cauchy = "0.4.0"
 cfg-if = "1.0.0"
-num_enum = "0.5.1"
+derive_more = "0.99.10"
+# We use document-features to automatically generate feature documentation from
+# Cargo.toml, following the example at
```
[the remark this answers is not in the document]
cgranade: Good catch, thanks! Fixed.

src/Simulation/qdk_sim_rs/src/linalg/array_ext.rs
```diff
@@ -0,0 +1,113 @@
...
+        //       borrow.
+        let source = self.index_axis(axis, idx_source).clone().to_owned();
+        let dest = self.index_axis(axis, idx_dest).clone().to_owned();
+        self.index_axis_mut(axis, idx_source).assign(&dest);
+        self.index_axis_mut(axis, idx_dest).assign(&source);
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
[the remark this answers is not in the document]
cgranade: Implicitly in functions which use it, which in theory should work since it's not a public API, but I'll go on and add a test case to be sure. Thanks for the suggestion!

src/Simulation/qdk_sim_rs/src/linalg/decompositions/lu.rs
samarsha: Ok, sounds good, thanks for explaining.

cgranade: Absolutely, thanks for the comment! ♥

src/Simulation/qdk_sim_rs/src/linalg/array_ext.rs
cgranade: My thinking here was that this is an internal-only trait/impl pair (`pub(crate)` in particular), and is mainly used to generate the specific error case `NotSquare` as a utility. Were this a public API, I'd agree entirely, but making that change here would necessitate adding a bunch of `.ok_or` calls where the return value for this extension method is used.

src/Simulation/qdk_sim_rs/src/utils.rs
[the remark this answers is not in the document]
cgranade: Modified the comment on `MiscError` to be a bit less agressive. Thanks!

cgranade:
@samarsha: Wanted to double-check that all your feedback has been addressed before merging. Thanks!

samarsha:
Yes, looks good, thanks.

cgranade:
Awesome, thank you for all your detailed feedback! 💕

PR #924, opened 2022-02-02
Migrate toolchain to LLVM 13

swernli:
This updates the tools used by the build scripts, CI pipeline, and linter/formatter to LLVM 13. Notably, the update to the newer version of clang-tidy required cleaning up some C++ patterns that were previously accepted under LLVM 11.


src/Qir/Runtime/lib/QIR/arrays.cpp
```diff
@@ -118,7 +118,7 @@ QirArray::QirArray(TItemCount countItems, TItemSize itemSizeBytes, TDimCount dim
 
     assert(this->count * (TBufSize)itemSizeInBytes < std::numeric_limits<TBufSize>::max());
     // Using `<` rather than `<=` to calm down the compiler on 32-bit arch.
-    const TBufSize bufferSize = this->count * itemSizeInBytes;
+    const TBufSize bufferSize = (TBufSize)this->count * (TBufSize)itemSizeInBytes;
```
kuzminrobin, line 121: (nit) For the future: It is probably more efficient to do the operation first, and then cast the result.
 ```suggestion
    const TBufSize bufferSize = (TBufSize)(this->count * itemSizeInBytes);
```

kuzminrobin (approved):

src/Qir/Runtime/lib/QIR/arrays.cpp
swernli: I think that would work here, yes. But I was a little concerned about a future where these two types would differ. Right now, `this->count` is of type `TItemCount` while `itemSizeInBytes` is of type `TItemSize`. It just so happens that both of these are `uint32_t`, but since they are independently defined they could vary. If we want to tie these values together more closely, we could consider consolidating `TItemCount` and `TItemSize` into one type.

PR #928, opened 2022-02-05
Enabling sparse simulator argument in driver

bettinaheim:
Putting up this as a reminder, but this change needs to be coordinated with the corresponding compiler PR: https://github.com/microsoft/qsharp-compiler/pull/1338

Also, we should add a test for the command line invocation of the sparse simulator.


kuzminrobin (approved):

bettinaheim:
The build failure is expected since the qdk repo is not yet update, and I'll update the package version number for the local repo build in this PR: https://github.com/microsoft/qsharp-runtime/pull/940 
Pulling this onto the feature branch per offline conversation.

PR #936, opened 2022-02-16
Added Telemetry

kuzminrobin:
Preliminary PR.

### DIFF
--- a/src/Simulation/Simulators/Microsoft.Quantum.Simulators.csproj
+++ b/src/Simulation/Simulators/Microsoft.Quantum.Simulators.csproj
@@ -23,6 +23,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.Quantum.Telemetry" Version="0.0.1-alpha" />
     <PackageReference Include="System.Text.Json" Version="4.7.2" />
     <PackageReference Include="System.Text.Encodings.Web" Version="4.7.2" />
     <PackageReference Include="NumSharp" Version="0.20.5" />
--- a/src/Simulation/Simulators/Microsoft.Quantum.Simulators.csproj
+++ b/src/Simulation/Simulators/Microsoft.Quantum.Simulators.csproj
@@ -23,7 +23,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Quantum.Telemetry" Version="0.0.1-alpha" />
+    <PackageReference Include="Microsoft.Quantum.Telemetry" Version="0.0.9-alpha" />
     <PackageReference Include="System.Text.Json" Version="4.7.2" />
     <PackageReference Include="System.Text.Encodings.Web" Version="4.7.2" />
     <PackageReference Include="NumSharp" Version="0.20.5" />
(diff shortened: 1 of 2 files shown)


src/Simulation/Simulators/SparseSimulator/SparseSimulator.cs
```diff
@@ -34,16 +38,51 @@ public partial class SparseSimulator : CommonNativeSimulator
...
+            };
+            TelemetryManagerDisposer = TelemetryManager.Initialize(TelemetryConfig);
+
+            Microsoft.Applications.Events.EventProperties eventProperties = new Applications.Events.EventProperties()
+                {
+                    Name = "Quantum.SparseSim.Constructed",
+                };
+            eventProperties.SetProperty("throwOnReleasingQubitsNotInZeroState", throwOnReleasingQubitsNotInZeroState);
+            // (`SimulatorBase.Seed` is assigned to `randomNumberGeneratorSeed ?? Guid.NewGuid().GetHashCode()`):
+            eventProperties.SetProperty("randomNumberGeneratorSeed", Seed);
```
bettinaheim, line 56: This seems a bit odd to me to collect; the only useful information I believe would be to know whether the seed has been set or not, but not which value.

src/Simulation/Simulators/Microsoft.Quantum.Simulators.csproj
```diff
@@ -23,6 +23,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.Quantum.Telemetry" Version="0.0.9-alpha" />
```
bettinaheim, line 26: If we keep the version number of the telemetry library in sync with the QDK version, it makes most sense to set the same version number here as we set for all other packages produced on the compiler repo. This makes it easier to do a search and replace when we need to update the packages for the local repo build.

src/Simulation/Simulators/SparseSimulator/SparseSimulator.cs
```diff
@@ -34,16 +38,51 @@ public partial class SparseSimulator : CommonNativeSimulator
...
+
+            Microsoft.Applications.Events.EventProperties eventProperties = new Applications.Events.EventProperties()
+                {
+                    Name = "Quantum.SparseSim.Constructed",
+                };
+            eventProperties.SetProperty("throwOnReleasingQubitsNotInZeroState", throwOnReleasingQubitsNotInZeroState);
+            // (`SimulatorBase.Seed` is assigned to `randomNumberGeneratorSeed ?? Guid.NewGuid().GetHashCode()`):
+            eventProperties.SetProperty("randomNumberGeneratorSeed", Seed); 
+            eventProperties.SetProperty("disableBorrowing", disableBorrowing);
+            eventProperties.SetProperty("numQubits", numQubits);
```
bettinaheim, line 58: I believe the qubit manager by default allows to allocate more than numQubits, though it can be configured to fail if more qubits than the initially specified number is ever allocated. If we want to gain an understanding of the maximum number of qubits in use, we would need to track this number in the qubit manager, and send it only after the execution terminates.

src/Simulation/Simulators/Microsoft.Quantum.Simulators.csproj
kuzminrobin: @vxfield, FYI.

PR #937, opened 2022-02-16
Excluded the temporary Rust files in .gitignore.

kuzminrobin:


swernli (approved):
Thanks for doing this! I kept meaning to get around to it but always forgot. Great repo cleanup/hygiene!

idavis (approved):

PR #938, opened 2022-02-16
Excluded temporary copy of SparseQuantumSimulator dynamic library in .gitignore.

kuzminrobin:
The native part of the Sparse Simulator is copied to near the native part of the Full-State Simulator (and native libs like OpenMP) for subsequent packaging. I'm excluding those temporary copies in `.gitignore`.

### DIFF
--- a/src/Simulation/Native/.gitignore
+++ b/src/Simulation/Native/.gitignore
@@ -10,3 +10,6 @@ foo*
 CMakeFiles/
 CMakeCache.txt
 *.so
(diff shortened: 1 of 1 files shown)


kuzminrobin:
Closing in favor of the similar change [here](https://github.com/microsoft/qsharp-runtime/pull/940/files#diff-bc37d034bad564583790a46f19d807abfe519c5671395fd494d8cce506c42947).

PR #943, opened 2022-02-17
Runtime support for Quantinuum targets.

cesarzc:
This change adds runtime support for Quantinuum targets.


vxfield (approved):
I'm assuming that because the target id has not changed, this change will work for customers that are either using the Honeywell or Quantinuum provider, right?
If that's the case, it looks good to me.

swernli (approved):

PR #944, opened 2022-02-17
Updating package version numbers

bettinaheim:
This should restore the repo build (not the e2e pipelines).

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,5 +1,5 @@
 {
     "msbuild-sdks": {
-        "Microsoft.Quantum.Sdk": "0.22.192435-beta"
+        "Microsoft.Quantum.Sdk": "0.22.192949-beta"
     }
 }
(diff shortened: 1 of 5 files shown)


global.json
```diff
@@ -1,5 +1,5 @@
 {
     "msbuild-sdks": {
-        "Microsoft.Quantum.Sdk": "0.22.192435-beta"
+        "Microsoft.Quantum.Sdk": "0.22.192949-beta"
```
kuzminrobin, line 3: Why not [0.22.192977](https://ms-quantum.visualstudio.com/Quantum%20Program/_build/results?buildId=192977&view=results)?

kuzminrobin (approved):

global.json
bettinaheim: This package number works perfectly fine, and was simply a complete set that was ready earlier.

PR #945, opened 2022-02-18
Fixed the SparseSimulator tests.

kuzminrobin:


src/Simulation/Simulators/CommonNativeSimulator/QubitManager.cs
```diff
@@ -31,8 +31,6 @@ public QSimQubitManager(bool throwOnReleasingQubitsNotInZeroState = true, long q
 
             public override Qubit CreateQubitObject(long id)
             {
-                Debug.Assert(id < 50, "Using a qubit id > 50. This is a full-state simulator! Validating ids uniqueness might start becoming slow.");
```
DmitryVasilevsky, line 34: Consider keeping it in the full-state simulator.

kuzminrobin: Since the SparseSimulator and QuantumSimulator share as much as possible, including the QubitManager, it seems to me easier for the shared code to be the same for both simulators.

Of course I can add a parameter to the constructor of the CommonNativeSimulator, QSimQubitManager, QubitManager, telling 
either which simulator has instantiated the QubitManager
or whether the qubit number limitation should be validated,
but that seems to me not worth the effort and not giving much value but obscuring the code.

I can be wrong, I realize that :-). Let me know. For now I would leave as is. Would you still like me to keep it for the QuantumSimulator case?

DmitryVasilevsky: Agreed, it doesn't make sense to put this check in the common code. Is there a way to put it in the full state simulator code that's not shared with the sparse simulator? If it is not possible - I approve.

kuzminrobin: Makes sense. 

I double-checked, unfortunately 
this function is only called from the code common for QuantumSimulator and SparseSimulator, 
and the argument is also calculated in the common code,
and the returned value cannot be easily analyzed by the QuantumSimulator only.

So it seems more efficient to move forward with this change. Can you please actually approve the PR?

PR #956, opened 2022-02-24
Extend IQirSubmitter and IQSharpSubmitter interfaces

cesarzc:
This change extends the `IQirSubmitter` and the `IQSharpSubmitter` to support a new QIR submission format.

### DIFF
--- a/src/Simulation/Core/Submitters/IQirSubmitter.cs
+++ b/src/Simulation/Core/Submitters/IQirSubmitter.cs
@@ -31,6 +31,20 @@ namespace Microsoft.Quantum.Runtime.Submitters
             Stream qir, string entryPoint, IReadOnlyList<Argument> arguments, SubmissionOptions options);
 
         /// <summary>
+        /// Submits a job to execute a QIR program without waiting for execution to complete.
+        /// </summary>
+        /// <param name="qir">The QIR program as a byte stream.</param>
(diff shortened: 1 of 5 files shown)


src/Simulation/Core/Submitters/IQirSubmitter.cs
```diff
@@ -30,6 +30,20 @@ public interface IQirSubmitter
...
+        /// </summary>
+        /// <param name="qir">The QIR program as a byte stream.</param>
+        /// <param name="entryPoint">The fully-qualified name of the entry point to execute.</param>
+        /// <param name="arguments">List of arguments used to execute the entry point.</param>
+        /// <param name="options">Additional options for the submission.</param>
+        /// <returns>The submitted job.</returns>
+        Task<IQuantumMachineJob> SubmitAsync(
+            Stream qir,
+            string entryPoint,
+            IReadOnlyList<EntryPointArgument> arguments,
```
ScottCarda-MS, line 44: I'm not sure why there needs to be a separate interface for using the EntryPointArgument data type instead of the Argument data type. What's the difference?

cesarzc: That's a good point. Originally, I was thinking on doing it this way because I thought the `Argument` data type was too tightly coupled to Q# data types but it is not and it represents QIR types well. Having said that then I think we don't need this new interface.

PR #965, opened 2022-03-09
Clarify BoolArrayAsBigInt behavior for negative numbers

tcNickolas:
Fix https://github.com/microsoft/QuantumLibraries/issues/390.


tcNickolas:
@msoeken Could you please check whether this is what you had in mind? It's slightly different from what the issue says (the padding part) but I double-checked that this matches the actual function behavior.

msoeken (approved):

swernli (approved):

PR #966, opened 2022-03-16
Adopt .NET6 + VS2022 in Quantum Development Kit

ricardo-espinoza:
With this change, we're migrating the main release train of the QDK to the most recent Long Time Support version of the .NET framework. For details about this change, refer to the original issue [Q#C:1224](https://github.com/microsoft/qsharp-compiler/issues/1224).

As part of this change, we're:

- Re-targeting all .NetCoreApp3.1 binaries to .NET6.0
- Updating Docker images, samples and templates.
- Libraries using .NetStandard2.1 are not affected by this change.
- The minimum supported .NET version in the QDK will also be updated from 3.1 to 6.0
- The Visual Studio extension will now target Visual Studio 2022.

**Important: This is a multi-repo change, and all changes will need to be merged concurrently.**


src/Simulation/qdk_sim_rs/.gitignore
```diff
@@ -4,8 +4,6 @@ win10
 target
 drop
 
-# We inject version numbers into Cargo.toml, so don't want them stored in repo.
-Cargo.toml
```
anpaz, line 8: was this removed on purpose?

anpaz (commented):
lgtm.

src/Simulation/qdk_sim_rs/.gitignore
ricardo-espinoza: Yes, this is part of a change in the change done by @cgranade to avoid the .NET5 dependency that had been introduced by Cargo Edit.

swernli (approved):

PR #971, opened 2022-03-21
Fix `Exp` decomposition to Ising gates

### LINKED ISSUE #967: Decomposition of `Exp` operation uses incorrect angle convention on IonQ targets
## Describe the bug
When submitting to IonQ targets, Q# uses a set of decompositions that uses an incorrect angle convention for Exp (https://docs.microsoft.com/qsharp/api/qsharp/microsoft.quantum.intrinsic.exp). This results in incorrect rotations when Exp is used on more than one qubit. The disparity can be identified when comparing the results of Exp on multiple qubits between the QDK full-state simulator and the IonQ full-state simulator target.

## Bug Status
	This bug is **active**. Please subscribe to notifications on this issue to be notified when a fix is available.

## Expected behavior
```qsharp
Exp([PauliX, PauliX], theta, qubits);
```
Should effectively decompose into:
```qsharp
IsingXX(-2.0 * theta, qubits);
```

## Actual Behavior
```qsharp
Exp([PauliX, PauliX], theta, qubits);
```

Effectively decomposes into:
```qsharp
IsingXX(theta / 2.0, qubits);
```

## Affected Versions
QDK version 0 0.23.195983 and earlier. 

## Additional Context
As part of fixing this issue, we are extending our existing decomposition tests to include additional test cases for the Exp operation. We are also reviewing our test suite to increase coverage of the decompositions used when running Q# on Azure Quantum targets.

## Impacted Users
If you believe that you were impacted by this issue, please create a billing ticket with [Azure support]( https://azure.microsoft.com/support/create-ticket) to request a refund or comment email [AzureQuantumInfo@Microsoft.com](mailto:AzureQuantumInfo@Microsoft.com) for more information.

swernli:
This fixes the decomposition of `Exp` in terms of Ising gates in the Type2 decompositions. Additionally, it adds a new test case that verifies 2 qubit `Exp` in terms of decompositions of `SWAP` and `CNOT` using Choi–Jamiołkowski isomorphism (see https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.diagnostics.assertoperationsequalreferenced).

Fixes #967


src/Simulation/Simulators.Tests/Circuits/ExpTest.qs
```diff
@@ -27,6 +28,28 @@ namespace Microsoft.Quantum.Simulation.Simulators.Tests.Circuits {
         }
     }
     
+    operation VerifyExpUsingDecompositions() : Unit {
```
cgranade, line 31: This makes perfect sense to add, fully agreed. For the benefit of future maintainers, would it help to add a comment that this is mainly useful for checking the magnitude of the angle and not its sign?

cgranade (approved):

src/Simulation/Simulators.Tests/Circuits/ExpTest.qs
swernli: Good call, adding comment.

PR #976, opened 2022-03-25
Add tests for evaluation of repeated value in sized array expressions

samarsha:
Runtime tests for the code gen change in microsoft/qsharp-compiler#1394.

### DIFF
--- a/src/Simulation/Simulators.Tests/Circuits/Arrays.qs
+++ b/src/Simulation/Simulators.Tests/Circuits/Arrays.qs
@@ -11,13 +11,13 @@ namespace Microsoft.Quantum.Simulation.Simulators.Tests.Circuits {
         for x in source {
             set result += [mapper(x)];
         }
+
         return result;
     }
 
     operation LengthTest() : Unit {
         let a1 = [One, Zero];
         let a2 = [Zero, Zero, Zero];
-
         AssertEqual(2, Length(a1));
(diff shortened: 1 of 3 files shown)


src/Simulation/Simulators.Tests/Circuits/Arrays.qs
```diff
@@ -77,7 +76,22 @@ namespace Microsoft.Quantum.Simulation.Simulators.Tests.Circuits {
...
     }
+
+    @Test("QuantumSimulator")
+    operation RepeatedValueIsEvaluatedOnceSize0() : Unit {
+        use q = Qubit();
+        let _ = [H(q), size = 0];
+        AssertMeasurement([PauliX], [q], Zero, "");
+        Reset(q);
+    }
+
```
ScottCarda-MS, line 89: For completeness, should we have a test for a size 1 array as well?

ScottCarda-MS, line 89: For completeness, should we have a test for a size 1 array as well?

ScottCarda-MS (approved):
Approved with optional suggestion.

src/Simulation/Simulators.Tests/Circuits/Arrays.qs
ScottCarda-MS: duplicate

PR #987, opened 2022-04-11
Add explicit decomposition testing infrastructure

swernli:
This change adds an explicit decomposition test infrastructure based off of `AssertOperationsEqualReferenced` and wrappers around the native simulator intrinsics as references. This allows for verification tests of decompositions against the simulator written directly in Q#.


swernli:
One downside of assert-based tests is that the test case crashes instead of failing when run via xunit. If we rewrite the reference version of `AssertOperationsEqualReferenced` to use measurement, then the code can use a `fail` statement and react more gracefully in xunit.

kuzminrobin (approved):

PR #991, opened 2022-04-14
Added to QIR RT the QIR ouput handling.

kuzminrobin:


kuzminrobin:
Please review.

src/Qir/Runtime/lib/QIR/QirOutputHandling.cpp
```diff
@@ -0,0 +1,86 @@
...
+    {
+        PrintCStr(RECORD_PREFIX "ARRAY_END");
+    }
+
+
+    // Primitive Result Records
+
+    void __quantum__rt__result_record_output(Result res) // NOLINT
+    {
+        if (!res)
```
swernli, line 50: To make this compatible with other result representations, this will need to use `__quantum__rt__result_get_zero` and `__quantum__rt__result_equal` rather than directly using it as a Boolean.

swernli (changes requested):
You will need to adjust how branching on measurement results works for this to function well with implementers of the `IRuntimeDriver` class that use non-Boolean measurement result handling. Additionally, even though these are fairly straightforward, we should add a test case that exercises these similar to the test cases around `__quantum__rt__message`.

src/Qir/Runtime/lib/QIR/QirOutputHandling.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+#define RECORD_TYPE "RESULT"
+#define RECORD_COLUMN_DELIMITER "\t"
+#define RECORD_PREFIX RECORD_TYPE RECORD_COLUMN_DELIMITER /* "RESULT" "\t" (== "RESULT\t") */
+
+
+static void PrintCStr(const char* cStr)
+{
+    QirString msg(cStr);
+    __quantum__rt__message(&msg);
```
cgranade, line 15: If I understand correctly, `__quantum_rt_message` appends the record prefix `INFO`, such that this function as written would result in program output of the form `INFO RESULT 0` rather than the desired `RESULT 0`?

kuzminrobin: Looking at the `__quantum__rt__message()` [implementation](https://github.com/microsoft/qsharp-runtime/blob/8d305b9345f8f924abf6c5fdc2370a0a17883d06/src/Qir/Runtime/lib/QIR/Output.cpp#L11) and that the [`OutputStream` is just the `std::cout`](https://github.com/microsoft/qsharp-runtime/blob/8d305b9345f8f924abf6c5fdc2370a0a17883d06/src/Qir/Runtime/lib/QIR/OutputStream.cpp#L17-L22) (and that the `std::cout` is supposed to be [replaced](https://github.com/microsoft/qsharp-runtime/blob/8d305b9345f8f924abf6c5fdc2370a0a17883d06/src/Qir/Runtime/lib/QIR/OutputStream.cpp#L24-L43) for testing purposes only), I don't see any place where the mentioned `INFO` can be added.
What do I miss?

kuzminrobin:
> You will need to adjust how branching on measurement results works for this to function well with implementers of the `IRuntimeDriver` class that use non-Boolean measurement result handling. Additionally, even though these are fairly straightforward, we should add a test case that exercises these similar to the test cases around `__quantum__rt__message`.

A. Changing.
B. Yes, I was also thinking about writing the tests for the added functionality. Working on the tests.

src/Qir/Runtime/lib/QIR/QirOutputHandling.cpp
swernli: I think we do eventually want the message intrinsic to show up as `INFO` records in the output, but that change hasn't been made. What will need when we go that way is a more generic output function, perhaps something like `__quantum__rt__record_output(char*, char*)` that takes both the tag and the content to be tagged, and then reimplement both the functions added here and message on top of those functions. I don't know if that's in-scope for this PR though...

kuzminrobin: @cgranade, @swernli, can you please refer me to more details about that `INFO`? (a task/issue may be?) Looks like I really miss some story here.

cgranade: > I don't know if that's in-scope for this PR though...

Fair, yeah. I guess my suggestion, then, would be to anticipate that change by not calling `__quantum__rt_message` directly here, but renaming that implementation to some private thing like `__write_to_current_stream(char*)` and then making `__quantum__rt_message` a one-line function that just calls `__write_to_current_stream`. That could be about a five-line change, and would make it easier to modify `__quantum__rt_message` in the future if needed without having to chase down other dependencies that really are dependencies on the output stream in disguise.

kuzminrobin:
I believe I addressed all the feedback. Can you, guys, please review again?

swernli (approved):
Changes look great, thanks Robin!

PR #992, opened 2022-04-15
Refactoring decompositions for hardware patterns

swernli:
This change refactors the decompositions to avoid specific patterns. Namely recursion, array concatenation and slicing, and complex mathematical operations. It uses the new decomposition tests to validate the refactor across controlled and adjoint specializations with up to 8 controls. This should make the decompositions more compatible with future hardware targeting transformations.

Note that this updatest to a beta version of the QDK to pick up a compiler fix that is required to compile the decompositions.

### DIFF
--- a/src/Simulation/EntryPointDriver.Tests/Tests.Microsoft.Quantum.EntryPointDriver.fsproj
+++ b/src/Simulation/EntryPointDriver.Tests/Tests.Microsoft.Quantum.EntryPointDriver.fsproj
@@ -23,7 +23,7 @@
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.5.0" />
     <PackageReference Include="xunit" Version="2.4.0" />
     <PackageReference Include="xunit.runner.visualstudio" Version="2.4.0" />
-    <PackageReference Include="Microsoft.Quantum.CSharpGeneration" Version="0.24.201332" />
+    <PackageReference Include="Microsoft.Quantum.CSharpGeneration" Version="0.24.205218-beta" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/Simulation/TargetDefinitions/Decompositions/CZFromSinglyControlled.qs
+++ b/src/Simulation/TargetDefinitions/Decompositions/CZFromSinglyControlled.qs
@@ -41,8 +41,95 @@ namespace Microsoft.Quantum.Canon {
             elif Length(ctls) == 1 {
                 CCZ(ctls[0], control, target);
             }
+            elif Length(ctls) == 2 {
+                use temp = Qubit();
+                within {
+                    PhaseCCX(ctls[0], ctls[1], temp);
+                }
+                apply {
+                    CCZ(temp, control, target);
+                }
+            }
+            elif Length(ctls) == 3 {
+                use temps = Qubit[2];
+                within {
+                    PhaseCCX(ctls[0], ctls[1], temps[0]);
+                    PhaseCCX(ctls[2], control, temps[1]);
+                }
+                apply {
+                    CCZ(temps[0], temps[1], target);
+                }
+            }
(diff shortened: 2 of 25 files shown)


src/Simulation/TargetDefinitions/Decompositions/CZFromSinglyControlled.qs
```diff
@@ -41,8 +41,95 @@ namespace Microsoft.Quantum.Canon {
...
+                apply {
+                    CCZ(temps[6], control, target);
+                }
+            }
             else {
-                ApplyWithLessControlsA(Controlled CZ, (ctls, (control, target)));
+                fail "Too many control qubits specified to CZ gate.";
+
+                // Eventually, we can use recursion via callables with the below utility:
+                // ApplyWithLessControlsA(Controlled Z, (ctls, qubit));
```
cgranade, line 132: Weird idea:

```suggestion
                if DecompositionsSupportBoundedRecursion() {
                    ApplyWithLessControlsA(Controlled Z, (ctls, qubit));
                }
```

Assuming a function elsewhere of the form:

```
// Config.qs
function DecompositionsSupportBoundedRecursion() : Bool {
    return false;
}
// AdvancedConfig.qs
function DecompositionsSupportBoundedRecursion() : Bool {
    return true;
}
```

Now a target that supports the advanced case can use `<Compile Include="AdvancedConfig.qs" />` while more limited targets can use `<Compile Include="Config.qs" />`, perhaps?

Not sure if it works, sorry, but an odd thought.

swernli: It's a neat idea we could consider for the future. For now, the motivation behind avoiding recursion is to facilitate inlining, which doesn't know about optimizing across constant branches that cannot be executed (that kind of constant folding and dead code elimination would be done after inlining).

kuzminrobin:
> Note that this updatest to a beta version of the QDK to pick up a compiler fix that is required to compile the decompositions.

I assume that those beta packages are actually published to nuget.org (otherwise they can get removed from the alpha feed after couple of weeks and a number of repos CI build will break. That's what happened to me) .

src/Simulation/EntryPointDriver.Tests/Tests.Microsoft.Quantum.EntryPointDriver.fsproj
```diff
@@ -23,7 +23,7 @@
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.5.0" />
     <PackageReference Include="xunit" Version="2.4.0" />
     <PackageReference Include="xunit.runner.visualstudio" Version="2.4.0" />
-    <PackageReference Include="Microsoft.Quantum.CSharpGeneration" Version="0.24.201332" />
+    <PackageReference Include="Microsoft.Quantum.CSharpGeneration" Version="0.24.205218-beta" />
```
bettinaheim, line 26: Remind me: This is a compiler version that is in main, correct?

src/Simulation/TargetDefinitions/Decompositions/R1FromSinglyControlled.qs
```diff
@@ -0,0 +1,139 @@
...
+                apply {
+                    CR1(theta, temps[6], qubit);
+                }
+            }
+            else {
+                fail "Too many control qubits specified to R1 gate.";
+            }
+        }
+
+        // R(PauliZ, theta, qubit);
```
bettinaheim, line 127: What are these commented out lines?

bettinaheim (approved):
Looks good to me. 
Is there another PR coming up that adds the attribute to force the required runtime capability?
I think it would also be good to add some tests to qat (one for each gate here) to ensure that all of these indeed compile into suitable QIR once the Q# implementation for it is complete.

PR #997, opened 2022-04-24
Add Type4 target package, update `Measure` decomp

swernli:
This change adds a Type4 package that avoids qubit reuse, and updates the Type2 package to use similar patterns. It adds tests, packaging and publishing for Type4.

In addition, this updates the decomposition of `Measure` in terms of single qubit Pauli-Z measurement to be limited to 8 qubits, updating all the target packages in the process.


src/Simulation/TargetDefinitions/Decompositions/Measure.qs
```diff
@@ -40,29 +41,115 @@ namespace Microsoft.Quantum.Intrinsic {
...
+                JointMeasureHelper(bases[6], q, qubits[6]);
+                JointMeasureHelper(bases[7], q, qubits[7]);
+            }
+            return MResetZ(q);
+        }
+        else {
+            fail "Too many qubits specified in call to Measure.";
         }
-        return res;
     }
```
kuzminrobin, line 154: I assume there is a reason why `operation Measure()` cannot to be implemented iteratively or recursively (for any length of the params `bases : Pauli[], qubits : Qubit[]`). If that's the case then it probably makes sense to change either the name of this operation `Measure()` or the doc comments of it to indicate that `Length(bases) <= 8` is supported only.

src/Simulation/TargetDefinitions/Decompositions/MeasureWithNoReuse.qs
```diff
@@ -39,27 +40,120 @@ namespace Microsoft.Quantum.Intrinsic {
...
+                JointMeasureHelper(bases[5], q, qubits[5]);
+                JointMeasureHelper(bases[6], q, qubits[6]);
+                JointMeasureHelper(bases[7], q, qubits[7]);
+            }
+            return MResetZ(q);
+        }
+        else {
+            fail "Too many qubits specified in call to Measure.";
         }
     }
```
kuzminrobin, line 158: (I can be wrong but) I have an impression that the changes in [Measure.qs](https://github.com/microsoft/qsharp-runtime/pull/997/files#diff-f4fcfc529ef943522c75549272dbcec2437ad6803109db9a0ec6c917d3df72b5) and [MeasureWithNoReuse.qs](https://github.com/microsoft/qsharp-runtime/pull/997/files#diff-716e9e692507d2246f5d7a51db7ac8b6c13549bcd9f0ac0da16abf7d9364da12) have a significant common part that can be extracted into a separate file and placed into an operation invoked from both files.

src/Simulation/TargetDefinitions/Interfaces/IType4Core.cs
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
```
kuzminrobin, line 2: Cassandra typically lets me know that the part `All rights reserved` is out-of-date. IIRC, based on [this](https://docs.opensource.microsoft.com/releasing/maintain/copyright-headers/).

src/Simulation/TargetDefinitions/Decompositions/Measure.qs
swernli: Yes, it has to do with avoiding variable indexing, recursion, and callable variables, which also motived the changes in #992. By avoiding those qualities, we can compile away complexity here with higher confidence for execution on current hardware.

Unfortunately, the docs must match across `Measure` implementations because they are essentially all the same function. Or at least, that's the strategy we've been employing thus far. @cgranade do you have any thoughts on documentation here? With the current way target packages are pulled in, callers using the IDE to preview documentation will always see the default QSharp.Core package docs, and I've kept them the same throughout because I didn't want to confuse any of the automatic doc generation logic with two identically named (and namespaced) operations with different docs. Any advice here?

src/Simulation/TargetDefinitions/Decompositions/MeasureWithNoReuse.qs
swernli: Hmm... you're right. As I was working on these initially they had more differences, but now the only difference is in the `Length(bases) == 1` case. I'll pull the rest out to a common utility operation.

src/Simulation/TargetDefinitions/Interfaces/IType4Core.cs
swernli: I'll at least update that in the files I'm changing, but avoid changing otherwise unmodified files for now.

kuzminrobin (approved):
Overall looks good to me. But I didn't dive deep into the code.
Approving to unblock. But I would recommend at least one more approval.

src/Simulation/TargetDefinitions/Decompositions/Measure.qs
cgranade: Let me tag in @msoeken here as well. In any case, I think we should document this restriction in QSharp.Core, noting that the limitation to eight qubits is on some targets (maybe in `/// # Remarks`?). Where we document profiles and targets, we can note the eight-qubit limitation there as well, perhaps?

src/Simulation/TargetDefinitions/Decompositions/Measure.qs
```diff
@@ -40,29 +41,115 @@ namespace Microsoft.Quantum.Intrinsic {
...
+        elif Length(bases) == 2 {
+            use q = Qubit();
             within {
-                MapPauli(qubits[0], PauliZ, bases[0]);
+                H(q);
             }
             apply {
-                set res = M(qubits[0]);
-                PreparePostM(res, qubits[0]);
+                JointMeasureHelper(bases[0], q, qubits[0]);
```
cgranade, line 59: I'll readily admit this is a nitpick, but `JointMeasureHelper` isn't a verb as suggested for Q# operation names.

src/Simulation/TargetDefinitions/Decompositions/Measure.qs
swernli: That seems like a good option. It may be something we should do across the board, since the target packages now also have a limit of 8 control qubits in the controlled operations.

src/Simulation/Simulators.Tests/OperationsTestHelper.cs
```diff
@@ -156,11 +157,11 @@ internal static void ctrlTestShell(SimulatorBase sim, Action<(IQArray<Qubit>, Qu
...
                     {
-                        var value = (i & (1 << j)) == 0
-                            ? Result.Zero
-                            : Result.One;
-
-                        set.Apply((value, ctrls[j]));
+                        reset.__Body__(ctrls[j]);
+                        if ((i & (1 << j)) != 0)
+                        {
+                            x.__Body__(ctrls[j]);
```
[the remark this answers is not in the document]
swernli: Yes, exactly. This was done to minimize dependency on `SetToBasisState` because of some complications that introduced in how I was building the target packages. Because `SetToBasisState` uses measurement, I like the idea of using `Reset` explicitly which may be implemented as `M` and conditional `X` or as an intrinsic dependending on the taget package getting tested. So while those compliciations have since been resolved and the code could be put back, I preferred this `Reset` based approach, personally.

swernli (commented):
Responding to some comments, which somehow showed up as a self-review...

src/Simulation/TargetDefinitions/Decompositions/MeasureWithNoReuse.qs
```diff
@@ -39,27 +40,120 @@ namespace Microsoft.Quantum.Intrinsic {
...
-                }
-                apply {
-                    for k in 0 .. Length(bases) - 1 {
-                        if( bases[k] == PauliX ) { Controlled X ([q], qubits[k]); }
-                        if( bases[k] == PauliZ ) { Controlled Z ([q], qubits[k]); }
-                        if( bases[k] == PauliY ) { Controlled Y ([q], qubits[k]); }
-                    }
-                }
-                return M(q);
+        elif Length(bases) == 1 {
```
cesarzc, line 47: The only difference between this version of measure and the other one seems to be when the length of bases is 1. When one decomposition would be used over the other?

swernli: When qubits can be reused after measurement, `Measure` can operate directly on the target qubit and allow later operations. When the platform does not allow reuse of a qubit after measurement, the alternate decomposition uses an auxiliary and entanglement to get information about the current state (and collapse the target) while still allowing it to be used in later operations. I've added a comment explaining this succinctly, but I'm debating whether or not this needs a longer explanation somewhere...

src/Simulation/TargetDefinitions/Decompositions/Measure.qs
swernli: I updated the name to `EntangleForJointMeasurement` to match style guide.

src/Simulation/TargetDefinitions/Decompositions/MeasureWithNoReuse.qs
cesarzc: I think the explanation you added could be good enough for now, and the name of the source code file also provides a good hint on what that measure decomposition is about :)

cesarzc (approved):

raghavthind2005 (approved):

src/Simulation/TargetDefinitions/Decompositions/Measure.qs
swernli: We should continue the documentation discussion, however I'm going to go ahead and merge the PR to get the code changes in.

PR #1002, opened 2022-04-29
Added target Capability to Entry Point Driver.

kuzminrobin:
3-repo change: [Q#RT](https://github.com/microsoft/qsharp-runtime/pull/1002) + [Q#Compiler](https://github.com/microsoft/qsharp-compiler/pull/1427) + Internal Repo.
The CI build is expected to fail, but the qdk.release with full validation is expected to succeed.

TODO:
* [x] Target capability tests

### DIFF
--- a/src/Azure/Azure.Quantum.Client/SubmitterFactory.cs
+++ b/src/Azure/Azure.Quantum.Client/SubmitterFactory.cs
@@ -62,6 +62,7 @@ namespace Microsoft.Azure.Quantum
         /// <param name="target">The name of the execution target.</param>
         /// <param name="workspace">The workspace used to manage jobs.</param>
         /// <param name="storageConnection">The connection string for the storage account.</param>
+        /// <param name="targetCapability">The target capability.</param>
         /// <returns>A QIR submitter.</returns>
         public static IQirSubmitter? QirSubmitter(string target, IWorkspace workspace, string? storageConnection, string targetCapability) =>
             Submitter<IQirSubmitter>(QirSubmitters, target, workspace, storageConnection, targetCapability);
@@ -145,7 +146,6 @@ namespace Microsoft.Azure.Quantum
         /// Returns the name and the type of a constructor that matches the target.
         /// </summary>
         /// <param name="submitter">Information about the submitter.</param>
-        /// <param name="target">The name of the execution target.</param>
         /// <returns>A name and type tuple.</returns>
         private static (string Name, Type Type) ConstructorInfo(SubmitterInfo submitter)
         {
--- a/src/Azure/Azure.Quantum.Client/SubmitterFactory.cs
+++ b/src/Azure/Azure.Quantum.Client/SubmitterFactory.cs
@@ -30,7 +30,8 @@ namespace Microsoft.Azure.Quantum
             new SubmitterInfo(
                 new Regex(@"\Aquantinuum\.([\w-_]+\.)*[\w-_]+\z"),
                 "Microsoft.Quantum.Providers.Quantinuum.Targets.QuantinuumQirSubmitter, Microsoft.Quantum.Providers.Honeywell",
-                "QirSubmitter"),
+                "QirSubmitter",
+                true),
             new SubmitterInfo(
                 new Regex(@"\Aqci\.([\w-_]+\.)*[\w-_]+\z"),
                 "Microsoft.Quantum.Providers.QCI.Targets.QCIQirSubmitter, Microsoft.Quantum.Providers.QCI",
@@ -95,7 +96,13 @@ namespace Microsoft.Azure.Quantum
         /// <returns>The QIR submitter instance.</returns>
         private static IQirSubmitter? AzurePayloadGenerator(IEnumerable<SubmitterInfo> submitters, string target)
         {
-            var constructorInfo = ConstructorInfo(submitters, target);
+            var submitter = submitters.FirstOrDefault(s => s.TargetPattern.IsMatch(target));
+            if (submitter is null)
+            {
+                return null;
+            }
+
+            var constructorInfo = ConstructorInfo(submitter, target);
             if (constructorInfo is null)
             {
                 return null;
@@ -120,14 +127,25 @@ namespace Microsoft.Azure.Quantum
             IEnumerable<SubmitterInfo> submitters, string target, IWorkspace workspace, string? storageConnection, string targetCapability = "")
             where T : class
         {
-            var constructorInfo = ConstructorInfo(submitters, target);
+            var submitter = submitters.FirstOrDefault(s => s.TargetPattern.IsMatch(target));
+            if (submitter is null)
+            {
+                return null;
+            }
+
+            if (submitter.RequiresTargetCapability && string.IsNullOrWhiteSpace(targetCapability))
+            {
+                return null;
+            }
+
+            var constructorInfo = ConstructorInfo(submitter, target);
             if (constructorInfo is null)
             {
                 return null;
             }
 
             (var constructorName, var constructorType) = constructorInfo.Value;
-            var args = new object?[] { target, workspace, targetCapability, storageConnection };
+            var args = new object?[] { target, workspace, storageConnection };
             return (T)constructorType.InvokeMember(
                 constructorName, BindingFlags.InvokeMethod, Type.DefaultBinder, null, args);
         }
@@ -138,14 +156,8 @@ namespace Microsoft.Azure.Quantum
         /// <param name="submitters">Information about each submitter.</param>
         /// <param name="target">The name of the execution target.</param>
         /// <returns>A name and type tuple.</returns>
-        private static (string Name, Type Type)? ConstructorInfo(IEnumerable<SubmitterInfo> submitters, string target)
+        private static (string Name, Type Type)? ConstructorInfo(SubmitterInfo submitter, string target)
         {
-            var submitter = submitters.FirstOrDefault(s => s.TargetPattern.IsMatch(target));
-            if (submitter is null)
-            {
-                return null;
-            }
-
             var name = submitter.MethodName;
             var type = QdkType(submitter.TypeName);
             return (name, type);
@@ -174,8 +186,10 @@ namespace Microsoft.Azure.Quantum
             /// <param name="targetPattern">The pattern for targets supported by the submitter.</param>
             /// <param name="typeName">The fully-qualified or assembly-qualified name of the submitter type.</param>
             /// <param name="methodName">The name of the static factory method.</param>
-            public SubmitterInfo(Regex targetPattern, string typeName, string methodName) =>
-                (TargetPattern, TypeName, MethodName) = (targetPattern, typeName, methodName);
+            public SubmitterInfo(Regex targetPattern, string typeName, string methodName, 
+                                 bool requiresTargetCapability = false) =>
+                (TargetPattern, TypeName, MethodName, RequiresTargetCapability) = 
+                    (targetPattern, typeName, methodName, requiresTargetCapability);
 
             /// <summary>
             /// The pattern for targets supported by the submitter.
@@ -191,6 +205,11 @@ namespace Microsoft.Azure.Quantum
             /// The name of the static factory method.
             /// </summary>
             public string MethodName { get; }
+
+            /// <summary>
+            /// Flag telling if the target requires the target capability.
+            /// </summary>
+            public bool RequiresTargetCapability { get; }
         }
     }
 }
--- a/src/Azure/Azure.Quantum.Client/SubmitterFactory.cs
+++ b/src/Azure/Azure.Quantum.Client/SubmitterFactory.cs
@@ -62,8 +62,8 @@ namespace Microsoft.Azure.Quantum
         /// <param name="workspace">The workspace used to manage jobs.</param>
         /// <param name="storageConnection">The connection string for the storage account.</param>
         /// <returns>A QIR submitter.</returns>
-        public static IQirSubmitter? QirSubmitter(string target, IWorkspace workspace, string? storageConnection) =>
-            Submitter<IQirSubmitter>(QirSubmitters, target, workspace, storageConnection);
+        public static IQirSubmitter? QirSubmitter(string target, IWorkspace workspace, string? storageConnection, string targetCapability) =>
+            Submitter<IQirSubmitter>(QirSubmitters, target, workspace, storageConnection, targetCapability);
 
         public static IQirSubmitter? QirSubmitterFromConfigFile(
             string target, FileInfo configFile, IWorkspace workspace, string? storageConnection)
@@ -117,7 +117,7 @@ namespace Microsoft.Azure.Quantum
         /// <typeparam name="T">The type of the submitter interface.</typeparam>
         /// <returns>The submitter instance.</returns>
         private static T? Submitter<T>(
-            IEnumerable<SubmitterInfo> submitters, string target, IWorkspace workspace, string? storageConnection)
+            IEnumerable<SubmitterInfo> submitters, string target, IWorkspace workspace, string? storageConnection, string targetCapability = "")
             where T : class
         {
             var constructorInfo = ConstructorInfo(submitters, target);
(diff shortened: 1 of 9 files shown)


kuzminrobin:
The qdk.release pipeline has succeeded (0.24.208317, with an unrelated warning, I'm discussing that warning).
In the meantime please review.

kuzminrobin:
Please review.

src/Simulation/EntryPointDriver/Azure/Azure.cs
```diff
@@ -418,7 +418,9 @@ private static void LogIfVerbose(GenerateAzurePayloadSettings settings, string m
             null => null,
             NoOpQirSubmitter.Target => new NoOpQirSubmitter(),
             NoOpSubmitter.Target => new NoOpSubmitter(),
-            _ => SubmitterFactory.QirSubmitter(settings.Target, settings.CreateWorkspace(), settings.Storage)
+            _ => (settings.TargetCapability != null)
```
samarsha, line 421: `TargetCapability` can never be null.

src/Simulation/EntryPointDriver/Azure/Azure.cs
```diff
@@ -418,7 +418,9 @@ private static void LogIfVerbose(GenerateAzurePayloadSettings settings, string m
             null => null,
             NoOpQirSubmitter.Target => new NoOpQirSubmitter(),
             NoOpSubmitter.Target => new NoOpSubmitter(),
-            _ => SubmitterFactory.QirSubmitter(settings.Target, settings.CreateWorkspace(), settings.Storage)
+            _ => (settings.TargetCapability != null) 
+                ? SubmitterFactory.QirSubmitter(settings.Target, settings.CreateWorkspace(), settings.Storage)
```
samarsha, line 422: I think we should also add the handling for `TargetCapability` to the QIR submitter before merging this to main. `TargetCapability` doesn't do anything until then, and we'd need another PR to qsharp-runtime to change that. Is there a reason this should be merged before the submitter is updated?

src/Simulation/EntryPointDriver.Tests/Tests.fs
```diff
@@ -565,6 +565,7 @@ let ``Submit uses default values`` () =
                Resource Group: myResourceGroup
                Workspace: myWorkspace
                Target: test.machine.noop
+               TargetCapability:
```
samarsha, line 568: We should have tests that check the basic handling of `TargetCapability`: that it uses the default value if the assembly constant if one is provided, and that the `--target-capability` option overrides that default.

samarsha (dismissed):
I have two concerns about merging this: there are no tests for non-empty values of `TargetCapability`, and the Azure submitters still need to be updated to consume `TargetCapability` (until then I think it is dead code).

src/Simulation/EntryPointDriver/Azure/Azure.cs
kuzminrobin, line 421:
```suggestion
            _ => ( ! string.IsNullOrWhitespace(settings.TargetCapability)) 
```

src/Simulation/EntryPointDriver/Driver.cs
```diff
@@ -109,6 +109,12 @@ public sealed class Driver
         private static readonly OptionInfo<string> JobNameOption = new OptionInfo<string>(
             ImmutableList.Create("--job-name"), Maybe.Just(string.Empty), "The name of the submitted job.");
 
+        /// <summary>
+        /// The target capability option.
+        /// </summary>
+        private static readonly OptionInfo<string> TargetCapabilityOption = new OptionInfo<string>(
+            ImmutableList.Create("--target-capability"), Maybe.Just(string.Empty), "The name of the target capability.");
```
bettinaheim, line 116: Come to think of it, I don't think it makes sense to add this to the command line options for the executable; The Q# compiler will (and should) populate what is currently called DefaultTargetCapability in the DriverSettings based on the project information. We will want to pass that information to the AzureSettings. But a user cannot really change what a program has been compiled for after it was compiled, so adding it to the command line for the executable doesn't make sense since it is already built for a specific TargetCapability.

kuzminrobin:
After discussion with Bettina the target capability has stopped being a command line argument, but now is rather a project setting. 
I still need to work on the tests. In the meantime please review the current changes to make sure that I'm on the right track.

src/Simulation/EntryPointDriver/Azure/Azure.cs
kuzminrobin: Fixed. Now the check is against white-space:
```c#
 _ => (!string.IsNullOrWhiteSpace(settings.TargetCapability)) 
```

src/Simulation/EntryPointDriver/Azure/Azure.cs
kuzminrobin: The changes to the C# client were scheduled as a separate DevOps task, but I added them now, to the internal repo. Please see there.

src/Simulation/EntryPointDriver.Tests/Tests.fs
kuzminrobin: I'm working on the tests. In the meantime I would like to make sure I'm on the right track. Please review the current changes.

src/Simulation/EntryPointDriver/Driver.cs
kuzminrobin: Changed.

src/Azure/Azure.Quantum.Client/SubmitterFactory.cs
```diff
@@ -127,7 +127,7 @@ public static class SubmitterFactory
             }
 
             (var constructorName, var constructorType) = constructorInfo.Value;
-            var args = new object?[] { target, workspace, storageConnection };
+            var args = new object?[] { target, workspace, targetCapability, storageConnection };
```
swernli, line 130: It looks like these are in the wrong order... above, the `targetCapabilitiy` is added to the end of the argument list. This ordering would pass `targetCapability` as `storageConnection` and vice versa.

swernli: Oh, never mind. This follows the ordering of a different API from another repo, not the ones above. That's a little confusion/tricky, since they are both strings and can easily be interchanged.

src/Azure/Azure.Quantum.Client/SubmitterFactory.cs
```diff
@@ -62,8 +62,8 @@ public static class SubmitterFactory
         /// <param name="workspace">The workspace used to manage jobs.</param>
         /// <param name="storageConnection">The connection string for the storage account.</param>
         /// <returns>A QIR submitter.</returns>
-        public static IQirSubmitter? QirSubmitter(string target, IWorkspace workspace, string? storageConnection) =>
-            Submitter<IQirSubmitter>(QirSubmitters, target, workspace, storageConnection);
+        public static IQirSubmitter? QirSubmitter(string target, IWorkspace workspace, string? storageConnection, string targetCapability) =>
```
cesarzc, line 65: I believe `targetCapability` shouldn't be part of these factory methods since that information is already part of the `SubmissionOptions` object.

cesarzc: I take this back, we do need `targetCapability` in these methods.

src/Simulation/EntryPointDriver/Azure/Azure.cs
```diff
@@ -432,7 +432,9 @@ private static void LogIfVerbose(GenerateAzurePayloadSettings settings, string m
             {
                 NoOpQirSubmitter.Target => new NoOpQirSubmitter(),
                 NoOpSubmitter.Target => new NoOpSubmitter(),
-                _ => SubmitterFactory.QirSubmitter(settings.Target, settings.CreateWorkspace(), settings.Storage)
+                _ => (!string.IsNullOrWhiteSpace(settings.TargetCapability))
```
cesarzc, line 435: This is slightly more nuanced than just checking whether target capability is null or empty. Submitters for some targets require target capability to be non-null and non-empty, but others don't have that requirement. I suggest adding a property to the `SubmitterInfo` class: `public bool RequiresTargetCapability`. This can be used inside the `SubmitterFactory.QirSubmitter` method to check the value of `settings.TargetCapability` and act accordingly.

@kuzminrobin please let me know if this makes sense.

kuzminrobin:
Any approvals or any more comments?

src/Simulation/EntryPointDriver/DriverSettings.cs
```diff
@@ -44,6 +44,11 @@ public sealed class DriverSettings
         /// </summary>
         internal string DefaultExecutionTarget { get; }
 
+        /// <summary>
+        /// The default execution target capability to use when submitting the entry point to Azure Quantum.
```
samarsha, line 48:
```suggestion
        /// The execution target capability to use when submitting the entry point to Azure Quantum.
```

src/Azure/Azure.Quantum.Client/SubmitterFactory.cs
```diff
@@ -174,8 +186,10 @@ private class SubmitterInfo
             /// <param name="targetPattern">The pattern for targets supported by the submitter.</param>
             /// <param name="typeName">The fully-qualified or assembly-qualified name of the submitter type.</param>
             /// <param name="methodName">The name of the static factory method.</param>
```
cesarzc, line 188: Can you please add `requiresTargetCapability` to the documentation comments?

src/Azure/Azure.Quantum.Client/SubmitterFactory.cs
```diff
@@ -138,14 +156,8 @@ public static class SubmitterFactory
         /// <param name="submitters">Information about each submitter.</param>
         /// <param name="target">The name of the execution target.</param>
         /// <returns>A name and type tuple.</returns>
-        private static (string Name, Type Type)? ConstructorInfo(IEnumerable<SubmitterInfo> submitters, string target)
+        private static (string Name, Type Type)? ConstructorInfo(SubmitterInfo submitter, string target)
```
samarsha, line 159: The `target` parameter is unused now. Also, this method no longer needs to return a nullable type since it never fails.

src/Azure/Azure.Quantum.Client/SubmitterFactory.cs
```diff
@@ -174,8 +186,10 @@ private class SubmitterInfo
             /// <param name="targetPattern">The pattern for targets supported by the submitter.</param>
             /// <param name="typeName">The fully-qualified or assembly-qualified name of the submitter type.</param>
             /// <param name="methodName">The name of the static factory method.</param>
-            public SubmitterInfo(Regex targetPattern, string typeName, string methodName) =>
-                (TargetPattern, TypeName, MethodName) = (targetPattern, typeName, methodName);
+            public SubmitterInfo(Regex targetPattern, string typeName, string methodName, 
+                                 bool requiresTargetCapability = false) =>
```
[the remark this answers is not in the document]
cesarzc: nit: formatting.

src/Simulation/Core/Submitters/SubmissionOptions.cs
```diff
@@ -27,17 +27,24 @@ public class SubmissionOptions
...
         /// <summary>
         /// The default submission options.
         /// </summary>
         public static SubmissionOptions Default { get; } =
-            new SubmissionOptions("", 500, ImmutableDictionary<string, string>.Empty);
+            new SubmissionOptions("", 500, ImmutableDictionary<string, string>.Empty, "");
 
-        private SubmissionOptions(string friendlyName, int shots, ImmutableDictionary<string, string> inputParams)
+        private SubmissionOptions(string friendlyName, int shots, ImmutableDictionary<string, string> inputParams,
+                                  string targetCapability)
```
cesarzc, line 42:
```suggestion
        private SubmissionOptions(
            string friendlyName, int shots, ImmutableDictionary<string, string> inputParams, string targetCapability)
```

cesarzc: nit: formatting.

src/Simulation/Core/Submitters/SubmissionOptions.cs
```diff
@@ -46,11 +53,14 @@ private SubmissionOptions(string friendlyName, int shots, ImmutableDictionary<st
...
         /// <returns>The updated submission options.</returns>
         public SubmissionOptions With(
             string? friendlyName = null,
             int? shots = null,
-            ImmutableDictionary<string, string>? inputParams = null) =>
-            new SubmissionOptions(friendlyName ?? FriendlyName, shots ?? Shots, inputParams ?? InputParams);
+            ImmutableDictionary<string, string>? inputParams = null,
+            string? targetCapability = null) =>
+            new SubmissionOptions(friendlyName ?? FriendlyName, shots ?? Shots, inputParams ?? InputParams, 
+                                  targetCapability ?? TargetCapability);
```
cesarzc, line 64:
```suggestion
            new SubmissionOptions(
                friendlyName ?? FriendlyName, shots ?? Shots, inputParams ?? InputParams, targetCapability ?? TargetCapability);
```

src/Simulation/EntryPointDriver/Azure/Azure.cs
```diff
@@ -432,7 +432,8 @@ private static void LogIfVerbose(GenerateAzurePayloadSettings settings, string m
             {
                 NoOpQirSubmitter.Target => new NoOpQirSubmitter(),
                 NoOpSubmitter.Target => new NoOpSubmitter(),
-                _ => SubmitterFactory.QirSubmitter(settings.Target, settings.CreateWorkspace(), settings.Storage)
+                _ => SubmitterFactory.QirSubmitter(settings.Target, settings.CreateWorkspace(), settings.Storage, 
+                                                   settings.TargetCapability)
```
cesarzc, line 436:
```suggestion
                _ => SubmitterFactory.QirSubmitter(
                    settings.Target, settings.CreateWorkspace(), settings.Storage, settings.TargetCapability)
```

cesarzc: nit: formatting.

src/Azure/Azure.Quantum.Client/SubmitterFactory.cs
```diff
@@ -117,10 +124,21 @@ public static class SubmitterFactory
...
-            var constructorInfo = ConstructorInfo(submitters, target);
+            var submitter = submitters.FirstOrDefault(s => s.TargetPattern.IsMatch(target));
+            if (submitter is null)
+            {
+                return null;
+            }
+
+            if (submitter.RequiresTargetCapability && string.IsNullOrWhiteSpace(targetCapability))
+            {
+                return null;
```
samarsha, line 138: I'm wondering about the error handling here. There are two ways for this function to fail: either no submitter matches the target string, or the required target capability is not present. But I think if a capability is missing, the entry point driver will still say "no submitter found for the target."

Since the Q# compiler itself provides the capability and there's no command line option, I don't think users would run into this, but it could be confusing if a developer adds/changes a submitter and forgets about the capability.

cesarzc (approved):
Added a few more minor suggestions but the implementation looks good to me.

src/Azure/Azure.Quantum.Client/SubmitterFactory.cs
kuzminrobin: Thanks for catching this! Fixing.

src/Azure/Azure.Quantum.Client/SubmitterFactory.cs
kuzminrobin: I'm making the following change in `Microsoft.Quantum.EntryPointDriver.Azure.Submit<TIn, TOut>()`. Will that work?
```diff
@@ -92,19 +92,22 @@ namespace Microsoft.Quantum.EntryPointDriver
             if (qirSubmission is null && !(QirSubmitter(settings) is null))
             {
                 DisplayError(
                     $"The target {settings.Target} requires QIR submission, but the project was built without QIR. "
                     + "Please enable QIR generation in the project settings.",
                     null);
             }
             else
             {
-                DisplayError($"No submitters were found for the target {settings.Target}.", null);
+                DisplayError(
+                    $"No submitters were found for the target \"{settings.Target}\" " + 
+                    $"and target capability \"{settings.TargetCapability}\".",
+                    null);
             }
 
             return Task.FromResult(1);
         }
```

samarsha: Yes, I think that works, thanks.

kuzminrobin:
I assume that I addressed all the feedback but feel free to comment more.

samarsha (approved):

Issue #917, opened 2022-01-26
Expose continuous-angle rotation intrinsics through to open systems simulator


cgranade:
**Please describe what you would like the feature to accomplish.**  
As part of https://github.com/microsoft/qsharp-runtime/issues/914, we will need to add continous-angle intrinsics to the C API for the open systems simulator crate, and to the C# runtime integration.

PR #1009, opened 2022-05-03
Support for continuous-angle rotations in Rust and C# layers of open systems simulator.

cgranade:
This PR supports the `Rx`, `Ry`, and `Rz` intrinsics in the open systems simulator.

Implements #915 and #917.


src/Simulation/qdk_sim_rs/src/linalg/decompositions/eig.rs
```diff
@@ -0,0 +1,100 @@
+use std::ops::Mul;
```
kuzminrobin, line 1: Do the new files require copyright?

src/Simulation/qdk_sim_rs/src/linalg/expm.rs
```diff
@@ -0,0 +1,212 @@
...
+    }
+}
+
+#[cfg(feature = "pade")]
+impl<A, S, E> Expm<A> for ArrayBase<S, Ix2>
+where
+    S: Data + RawData<Elem = A>,
+    A: Scalar + ScalarOperand,
+    E: std::convert::From<QdkSimError>,
+    ArrayBase<S, Ix2>: Inv<Error = E> + Identity,
```
kuzminrobin, line 83: (Nit, my subjective perception)
I'm personally not a fan of the single-letter identifiers. For template parameters either (I don't know Rust and I assume that `A, S, E` are analogous to the C++'s template parameters). I would recommend to consider more informative identifiers instead of `A, S, E`.
(In C++ it is common to use a single-letter template parameter `T` if it is a data type of the container elements, etc. I'm OK with that. But if there is more than one template param, or it is not a data type (but is an integer constant for example), or is a type with some peculiarities, then the single-letter template params make an impression of sub-optimal code).

src/Simulation/qdk_sim_rs/src/linalg/expm.rs
```diff
@@ -0,0 +1,212 @@
...
+{
+    type Error = E;
```
kuzminrobin, line 85: Why not remove this line and in line 78 do
```diff
- impl<A, S, E>
---
+ impl<A, S, Error>
```
?

src/Simulation/qdk_sim_rs/src/linalg/expm.rs
```diff
@@ -0,0 +1,212 @@
...
+    fn expm_scale(&self, scale: A) -> Result<Self::Output, Self::Error> {
+        self.apply_mtx_fn(|x| (*x * scale).exp())
+    }
+}
+
+#[cfg(feature = "pade")]
+impl<A, S, E> Expm<A> for ArrayBase<S, Ix2>
+where
+    S: Data + RawData<Elem = A>,
+    A: Scalar + ScalarOperand,
```
kuzminrobin, line 81: For me the relation is not evident between the letter `A` and the word `Scalar`. If it is scalar then why not use letter `S` instead of `A`?

src/Simulation/qdk_sim_rs/src/linalg/expm.rs
```diff
@@ -0,0 +1,212 @@
...
+
+        Ok(d_inv.dot(&n))
+    }
+
+    fn expm_scale(&self, scale: A) -> Result<Self::Output, Self::Error> {
+        Ok((self * scale).expm()?)
+    }
+}
+
+// TODO: Allow simpler type constraints.
```
kuzminrobin, line 104: (Nit) A few months ago I saw a coding style item that I liked. It was prescribing to complement every `TODO` with, in parentheses, either user alias (who opened the `TODO` or who the `TODO` is assigned to), or the bug number, or both. E.g. `TODO(cgranade): ..`.

src/Simulation/qdk_sim_rs/src/linalg/expm.rs
```diff
@@ -0,0 +1,212 @@
...
+        let coeff = (approximate_factorial::<_, f64>(p + q - j)
+            * approximate_factorial::<_, f64>(p))
+            / (approximate_factorial::<_, f64>(p + q)
+                * approximate_factorial::<_, f64>(j)
+                * approximate_factorial::<_, f64>(q - j));
+        let coeff = A::from(coeff).unwrap();
+        let mtx_pow = (-mtx.to_owned()).matrix_power(j.into())?;
+        let term = mtx_pow * coeff;
+        result = result + term;
+    }
```
kuzminrobin, line 131: I see that `arr` is created, but I fail to see how it is used here in this scope. What shall we lose if we remove lines 120 - 121?

src/Simulation/qdk_sim_rs/src/math/sp_func.rs
```diff
@@ -0,0 +1,61 @@
...
+            actual,
+            epsilon = actual * 1e-8
+        );
+
+        let actual = 1.214630436702533e+205;
+        assert_abs_diff_eq!(
+            approximate_factorial::<f64, f64>(123.0),
+            1.214630436702533e+205,
+            epsilon = actual * 1e-8
+        );
```
kuzminrobin, line 59: A. The fragment `1.214630436702533e+205` occurs more than once. You might want to avoid duplication. Compare the line 57 to line 50.
B. How can the reader make sure that such "magic numbers" as `3628800.0`, `1.214630436702533e+205` are correct? Code comments explaining the value wouldn't harm.

src/Simulation/qdk_sim_rs/src/linalg/expm.rs
```diff
@@ -0,0 +1,212 @@
...
+            vectors: array![
+                [
+                    c64::new(-0.0, -0.7071067811865474),
+                    c64::new(0.7071067811865475, 0.0)
+                ],
+                [
+                    c64::new(0.7071067811865476, 0.0),
+                    c64::new(0.0, -0.7071067811865475)
+                ],
+            ],
```
kuzminrobin, line 200: Fragment "0.7071067811865475" occurs more than once. Consider de-duplicating. Applicable to more than one file.
Consider explaining the values for the "magic numbers" (in the code comments), such that the reader can make sure that the value is correct.

kuzminrobin (approved):

src/Simulation/qdk_sim_rs/src/linalg/decompositions/eig.rs
cgranade: Yes, thanks for catching that!

src/Simulation/qdk_sim_rs/src/linalg/expm.rs
cgranade: Good point, thanks; these were generated by using Python to produce unit test cases, but could now be replaced with [`std::f64::consts::FRAC_1_SQRT_2`](https://doc.rust-lang.org/std/f64/consts/constant.FRAC_1_SQRT_2.html).

(As a side note, 0.707 comes up a _lot_ in quantum applications, such that I'm slightly biased — I no longer really tend to see it as a magic number.)

src/Simulation/qdk_sim_rs/src/math/sp_func.rs
cgranade: In this case, these are testing implementations of the approximate factorial function, such that `1.214630436702533e+205` is a double-precision approximation of 123!. Will clarify that in comments, as you suggest.

src/Simulation/qdk_sim_rs/src/linalg/expm.rs
cgranade: Good catch, thanks; I think that was from older attempts that I forgot to remove while refactoring.

src/Simulation/qdk_sim_rs/src/math/sp_func.rs
```diff
@@ -0,0 +1,64 @@
...
+
+        // We expect approximate_factorial(123.0) to return 123! ≈ 1.214 × 10²⁰⁵.
+        // https://www.wolframalpha.com/input?i=123%21
+        let expected = 1.214630436702533e+205;
+        assert_abs_diff_eq!(
+            approximate_factorial::<f64, f64>(123.0),
+            1.214630436702533e+205,
+            epsilon = expected * 1e-8
+        );
```
kuzminrobin, line 62: Did you mean this? (for consistency between lines 60 and 51)
```suggestion
        let expected = 1.214630436702533e+205;
        assert_abs_diff_eq!(
            approximate_factorial::<f64, f64>(123.0),
            expected,
            epsilon = expected * 1e-8
        );
```

cgranade:
@kuzminrobin: Thanks for your earlier review and approval! The PR has changed a bit now that it's no longer in draft, so please let me know if your previous approval still applies.

@swernli, @idavis, @samarsha: Please let me know if you would like to review, thank you!

src/Simulation/qdk_sim_rs/src/linalg/expm.rs
kuzminrobin: Thanks for the example! Resolving this discussion.

src/Simulation/Simulators.Tests/Circuits/TeleportTest.qs
```diff
@@ -7,8 +7,7 @@ namespace Microsoft.Quantum.Simulation.Simulators.Tests.Circuits {
 
     @Test("QuantumSimulator")
     @Test("SparseSimulator")
-    // TODO: Disabled until we have a noise model for Rz.
-    // @Test("Microsoft.Quantum.Experimental.OpenSystemsSimulator")
+    @Test("Microsoft.Quantum.Experimental.OpenSystemsSimulator")
     operation TestTeleport() : Unit {
```
kuzminrobin, line 11: I assume that some time later there will be a change in the Q# compiler such that the following change below will be understood by the compiler, similar to lines 8 - 9 above. If that's true then either a TODO comment here or a task wouldn't harm, to not forget to make a change in the compiler.
```suggestion
    @Test("OpenSystemsSimulator")
    operation TestTeleport() : Unit {
```
@bettinaheim, FYI.

src/Simulation/Simulators.Tests/OpenSystemsSimulatorTests/NativeInterfaceTests.cs
```diff
@@ -1,39 +1,30 @@
...
-
-        [Fact]
-        public void GetIdealStabilizerNoiseModelByNameWorks()
-        {
-            var idealStabilizer = NativeInterface.GetNoiseModelByName("ideal_stabilizer");
-            // TODO: Add assertions here to check properties of each noise model.
-        }
+        var ideal = NativeInterface.GetNoiseModelByName("ideal");
+        // TODO: Add assertions here to check properties of the above noise model.
+    }
```
kuzminrobin, line 13:
```suggestion
        var ideal = NativeInterface.GetNoiseModelByName("ideal");
        // TODO(cgranade): Add assertions here to check properties of the above noise model.
    }
```

src/Simulation/Simulators.Tests/Circuits/TeleportTest.qs
cgranade: Yes, at this point cross-repo changes are difficult to test with end-to-end build issues, such that I didn't want to take that change as well with this PR. Definitely would be good to do, though, as we remove `Microsoft.Quantum.Experimental` from other parts of this feature.

src/Simulation/qdk_sim_rs/src/linalg/expm.rs
cgranade: No worries, thanks for the question! 💕

src/Simulation/Simulators.Tests/SerializationTests/StateSerializationTests.cs
```diff
@@ -1,70 +1,62 @@
...
+            }
+        ";
+        var reader = new Utf8JsonReader(System.Text.Encoding.UTF8.GetBytes(expectedJson));
+        var array = JsonSerializer.Deserialize<StabilizerState.TableArray>(ref reader);
+        var state = new StabilizerState
+        {
+            NQubits = 2,
+            Table = array
+        };
     }
```
kuzminrobin, line 61: What do the lines 56 - 60 do? They create an instance that is never used? In order to make sure that no exception thrown during `new StabilizerState`?

src/Simulation/Simulators.Tests/GlobalUsings.cs
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+global using System;
+global using Xunit;
+global using Xunit.Abstractions;
```
kuzminrobin, line 6: (I don't know C# and can be wrong)
I have an impression that this new file is never used.
In this PR there are 2 files with such a name. Both are never used?

cgranade: This is a new C# 10 feature, but allows breaking common `using` statements into a separate file that applies to all C# files in the same project. In particular, this is useful for things like `System` or `System.Linq` that are used in every or almost every file, and can help cut down on boilerplate in new C# files.

As it's a compile-time feature, using this new capability does not change anything at runtime, and in particular does not introduce new dependencies for users.

src/Simulation/Simulators.Tests/SerializationTests/StateSerializationTests.cs
cgranade: Let me look into this, I suspect this is a mistake made in the original experimental version of the feature, such that I can likely correct it as part of this PR. Thanks for the catch!

src/Simulation/Simulators/Microsoft.Quantum.Simulators.csproj
```diff
@@ -5,7 +5,7 @@
   <Import Project="..\Common\Simulators.Dev.props" />
 
   <PropertyGroup>
-    <LanguageVersion>8.0</LanguageVersion>
+    <LanguageVersion>10.0</LanguageVersion>
```
kuzminrobin, line 8: I believe this change affects all the simulators, not only the OpenSystemsSimulator.
The CI runs the tests for Release config only.
Just in case I would recommend to build locally the Debug config of the simulators and run the Debug config tests. Something like 
```powershell
dotnet clean Simulation.sln
dotnet build Simulation.sln -c Debug
dotnet test Simulation.sln -c Debug

# Or even more reliable:
$Env:BUILD_CONFIGURATION="Debug"
pwsh ./bootstrap.ps1
dotnet test Simulation.sln
```

src/Simulation/Simulators/OpenSystemsSimulator/DataModel/ArrayConverter.cs
```diff
@@ -3,249 +3,245 @@
...
 using System.Numerics;
 using System.Text.Json;
 using System.Text.Json.Serialization;
 using NumSharp;
 
-namespace Microsoft.Quantum.Experimental
+namespace Microsoft.Quantum.Experimental;
+
+/// <summary>
+///     Converts <see cref="NDArray" /> instances of dtype <c>double</c>
```
kuzminrobin, line 14: I assume that `dtype` is a term that I haven't met before (e.g. a shorthand of "data type"), but not a typo.

In the fragment `<see cref="NDArray" />` is there any special meaning in the space between `"` and `/`? Or is that space just a readability convenience?

src/Simulation/Simulators/OpenSystemsSimulator/DataModel/ArrayConverter.cs
```diff
@@ -3,249 +3,245 @@
...
+///     compatible with the representation used by Rust's <c>ndarray</c>
+///     and <c>serde</c> crates.
+///   </para>
+///   <para>
+///     In particular, each JSON object produced by this converter has
+///     three properties:
+///   </para>
+///   <list type="bullet">
+///     <item><c>"v"</c>: Indicates the <c>ndarray</c> format version being
+///       being used. Always <c>1</c> for this implementation.</item>
```
kuzminrobin, line 30:
```suggestion
///     <item><c>"v"</c>: Indicates the <c>ndarray</c> format version
///       being used. Always <c>1</c> for this implementation.</item>
```

src/Simulation/Simulators/OpenSystemsSimulator/DataModel/ArrayConverter.cs
```diff
@@ -3,249 +3,245 @@
...
     {
-        public override NDArray Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+        // By the end, we need to have read "v", "dim", and "data".
+        int? version = null;
+        List<int>? dims = null;
+        List<Complex>? data = null;
+
+        // We require that the reader be in the right state to read an
+        // object.
+        if (reader.TokenType != JsonTokenType.StartObject)
```
kuzminrobin, line 51: If the `if` (in line 51) is entered then the definitions above (in lines 45 - 47) are useless (never used). Consider moving the code in lines 45 - 47 to after the line 54 (if C# supports that).

src/Simulation/Simulators/OpenSystemsSimulator/DataModel/ArrayConverter.cs
```diff
@@ -3,249 +3,245 @@
...
+
+            switch (propertyName)
             {
-                if (reader.TokenType == JsonTokenType.EndObject)
-                {
-                    // We're at the end of the array, and can break out of the
-                    // read loop.
+                case "v":
+                    reader.Read();
+                    version = reader.GetInt32();
```
kuzminrobin, line 78: (Up to you, feel free to ignore. I can be asking a question that may never be applicable)

Earlier in the lines 29 - 30 there is a fragment 
```c#
///     <item><c>"v"</c>: Indicates the <c>ndarray</c> format version being
///       being used. Always <c>1</c> for this implementation.</item>
```
Is there a chance that "v" can be non-integer? E.g. "1.2.3". If yes then what will the (green) line 78 do? Throw a (hardly actionable) [exception](https://docs.microsoft.com/en-us/dotnet/api/system.text.json.utf8jsonreader.getint32?view=net-6.0#exceptions)?

src/Simulation/Simulators/OpenSystemsSimulator/DataModel/ArrayConverter.cs
```diff
@@ -3,249 +3,245 @@
...
+///     <item><c>"dim"</c>: Lists the dimensions of the array being
+///       serialized. Will always contain one less dimension than the
+///       original array, as one dimension represents the real and imaginary
+///       components of each element.</item>
+///     <item><c>"data"</c>: Lists the elements of the array, with the
+///       right-most dimension varying the fastest (defined by <c>ndarray</c>
+///       as "logical order", and by NumPy as "C-ordered").</item>
+///   </list>
+/// </remarks>
+public class ComplexArrayConverter : JsonConverter<NDArray>
```
kuzminrobin, line 40: (Up to you)
Discrepancy between the file name 
`ArrayConverter` and the class name
`ComplexArrayConverter `.

src/Simulation/Simulators/OpenSystemsSimulator/DataModel/ArrayConverter.cs
```diff
@@ -3,249 +3,245 @@
...
-                    throw new JsonException();
-                }
+                case "dim":
+                    dims = new List<int>();
+                    reader.Require(JsonTokenType.StartArray);
+                    while (reader.Read() && reader.TokenType != JsonTokenType.EndArray)
+                    {
+                        dims.Add(reader.GetInt32());
+                    }
+                    break;
```
kuzminrobin, line 88: (I can be paranoid here, feel free to ignore)

Is there a chance that at some iteration the `reader.Read()` in line 84 succeeds, but after that the `reader.TokenType` is `null` or some non-integer (and not `JsonTokenType.EndArray` in the end of line 84)? If yes then the subsequent `reader.GetInt32()` in line 86 will [throw](https://docs.microsoft.com/en-us/dotnet/api/system.text.json.utf8jsonreader.getint32?view=net-6.0#exceptions), right?

(I assume that this class is a part of a library with a very broad and unpredictable set of use scenarios, i.e. _any input_ can be fed to this function, including the _malicious input_. If this function is used in an online service, then the malicious input can be used to take the online service down)

src/Simulation/Simulators/Microsoft.Quantum.Simulators.csproj
cgranade: Sounds good, I've not run into any issues so far running locally, as C# 10 is back-compat with 8, but I agree with being cautious here.

src/Simulation/Simulators/OpenSystemsSimulator/DataModel/ArrayConverter.cs
cgranade: dtype is a common term in NumPy, the Python library that NumSharp is intended to be compatible with. As a result, NumSharp uses the term as well to mean "type of elements of an array."

For the space before `/>`, that's just the style I've learned to write XML in, but happy if there's another style that is used here.

src/Simulation/Simulators/OpenSystemsSimulator/DataModel/ArrayConverter.cs
cgranade: In this case, `"v"` should always be an integer, as per the design of the serde-json and ndarray crates for Rust.

src/Simulation/Simulators/OpenSystemsSimulator/DataModel/ArrayConverter.cs
cgranade: Good point, let me rename the file. The original intent was to be more general, but wound up having to specialize to just complex arrays due to limitations in NumSharp.

src/Simulation/Simulators/OpenSystemsSimulator/DataModel/ArrayConverter.cs
cgranade: I do not believe the token type can be a C# `null`, as that indicates the absence of a token (as opposed to `JsonTokenType.Null`). In either case, though, the result from malformed input (whether malicious or due to a bug) should be to throw a JSON serialization exception propagating out to the deserialization call site. In general, using any `System.Text.Json` APIs should be expected to possibly raise that exception such that my understanding would be that when used in a cloud service, checking for that exception would be best practice regardless of this line of code in particular.

Regardless, thanks for raising the question. I don't think it's paranoid at all, but a good discussion nonetheless; I think we should be OK as-is in this case, but I appreciate your raising to help make sure of that!

src/Simulation/Simulators/OpenSystemsSimulator/DataModel/ArrayConverter.cs
```diff
@@ -3,249 +3,245 @@
...
-                var propertyName = reader.GetString();
+                case "data":
+                    data = new List<Complex>();
+                    reader.Require(JsonTokenType.StartArray);
+
+                    while (reader.Read() && reader.TokenType != JsonTokenType.EndArray)
+                    {
+                        // We expect an inner array at this point, so
+                        // go on and read the begin array token.
+                        reader.Require(JsonTokenType.StartArray, read: false);
```
kuzminrobin, line 98: (I fail to (quickly) find documentation for `reader.Require()`)

Why is there in line 98 the fragment `read: false`, but in line 83 there is no? 

For now I interpret the lines 83 - 86 like this (see my comment in the code):
```c#
reader.Require(JsonTokenType.StartArray);  // Make sure that there is an array start and **consume that array start**.
while (reader.Read() ..  // Read the next token (containing the integer) after the array start.
{
    dims.Add(reader.GetInt32());   // Interpret the token as an integer.
```
But I interpret the lines 98 - 101 like this:
```c#
reader.Require(JsonTokenType.StartArray, read: false);  // Make sure that there is an array start, but **DO NOT consume that array start**.
reader.Read();   // Read the token CONTAINING THE ARRAY START.
var real = reader.GetDouble();  // Interpret the ARRAY START token as a double.
```
To summarize, I have an impression that the fragment `read: false` in line 98 results in the misinterpretation in line 101.

cgranade: `Require` is an extension method defined at https://github.com/microsoft/qsharp-runtime/blob/1334dc8cefb447e65feca66c463bcd77421bd5a2/src/Simulation/Simulators/OpenSystemsSimulator/DataModel/Extensions.cs#L66, and is used to ensure that a read succeeds with the right token.

In any case, I think your interpretation is correct, such that L103 can be safely dropped if L101 is changed to remove `read: false`.

kuzminrobin: Your refs to the line numbers seem shifted but looking at the definition of `Require()` I understood what you meant.
Consider renaming the param `read` to `preRead` or `readBefore` or similar. Otherwise I interpreted it as a `postRead` or `readAfter`.

src/Simulation/Simulators/OpenSystemsSimulator/DataModel/ArrayConverter.cs
```diff
@@ -3,249 +3,245 @@
...
-
-                            data.Add(new Complex(real, imag));
-                        }
-                        break;
-
-                    default:
-                        throw new JsonException();
-                }
-            }
+                        var real = reader.GetDouble();
```
kuzminrobin, line 101: (Paranoid)
In this line I feel a risk of interpreting the possible `JsonTokenType.EndArray` (immediately following the `JsonTokenType.StartArray` in line 98) as `double`.

Up to you to decide whether that is possible or not.

kuzminrobin: Same for L104.

src/Simulation/Simulators/OpenSystemsSimulator/DataModel/ArrayConverter.cs
```diff
@@ -3,249 +3,245 @@
...
 
-        public override void Write(Utf8JsonWriter writer, NDArray value, JsonSerializerOptions options)
+        // At this point, none of version, dims, or data should be null.
+        if (version == null) throw new JsonException(nameof(version));
+        if (dims == null) throw new JsonException(nameof(dims));
+        if (data == null) throw new JsonException(nameof(data));
+
+        // We now know we have everything we need to make the new array.
+        // In doing so, we allocate an ndarray with of shape (nElements, 2)
+        // where the last index represents real-vs-imag. We'll reshape
```
kuzminrobin, line 124: a. The wording "an ndarray **with of** shape" seems odd to me.
b. FYI: I don't understand the fragment "the last index represents real-vs-imag".

src/Simulation/Simulators/OpenSystemsSimulator/DataModel/ArrayConverter.cs
cgranade: The `GetDouble` method throws an `InvalidOperationException` when the token type is not a JSON number such that if the array contains the wrong number of elements (an array representing a complex should always have length exactly two), that gets propagated out into an exception. It could be safer to write a new method that wraps that in a more actionable error message, but I think that's likely overkill here?

src/Simulation/Simulators/OpenSystemsSimulator/DataModel/ArrayConverter.cs
cgranade: (a) was a typo, thanks for catching that. (b) indicates that the last index is of length two, and indicates whether each element describes the real or imaginary part of a complex number. For example, the 2D array `[[1, i], [-i, 1]]` would get represented by first flattening to `[1, i, -i, 1]` (where `dims` indicates how to unflatten again), then expanding out complex numbers, `[[1, 0], [0, 1], [0, -1], [1, 0]]`.

src/Simulation/Simulators/OpenSystemsSimulator/DataModel/ArrayConverter.cs
cgranade: Done.

src/Simulation/Simulators/OpenSystemsSimulator/DataModel/ArrayConverter.cs
cgranade: Expanded comment.

src/Simulation/Simulators/OpenSystemsSimulator/DataModel/ChpDecomposition.cs
```diff
@@ -3,130 +3,129 @@
...
-using System.Collections.Generic;
 using System.Text.Json;
 using System.Text.Json.Serialization;
 
-namespace Microsoft.Quantum.Experimental
-{
+namespace Microsoft.Quantum.Experimental;
 
-    public class ChpOperationConverter : JsonConverter<ChpOperation>
+public class ChpOperationConverter : JsonConverter<ChpOperation>
```
kuzminrobin, line 11: I'm not sure if the file name `ChpDecomposition` relates to any identifier defined in this file. Up to you.

cgranade: Effectively, a decomposition is a list of operations, such that the file name was intended to reflect the group of classes that work together to represent that, but happy to rename if you prefer.

kuzminrobin: Sounds reasonable, no need to rename. Thanks.

src/Simulation/Simulators/OpenSystemsSimulator/DataModel/ComplexArrayConverter.cs
```diff
@@ -0,0 +1,250 @@
...
+        }
+
+        while (reader.Read())
+        {
+            if (reader.TokenType == JsonTokenType.EndObject)
+            {
+                // We're at the end of the array, and can break out of the
+                // read loop.
+                break;
+            }
```
kuzminrobin, line 204: (Readability nit) In a number of places, consider shortening to
```suggestion
        while (reader.Read() && (reader.TokenType != JsonTokenType.EndObject))
        {
```

src/Simulation/Simulators/OpenSystemsSimulator/DataModel/Extensions.cs
```diff
@@ -6,159 +6,217 @@
...
+                        completion = readData(ref reader, kind);
+
+                        // Finally, require an end to the object.
+                        reader.Require(JsonTokenType.EndObject);
+                    }
+                    else if (reader.TokenType == JsonTokenType.String)
+                    {
+                        var kind = reader.GetString();
+                        // The same logic follows here as well.
+                        Debug.Assert(kind is not null);
```
kuzminrobin, line 170: (Just FYI)
I remember a few months ago Bettina complained that `Debug.Assert()` is undesirable for our test infra.
Later on I observed something like, if the `Debug.Assert()` is executed in the test and the assertion fails, then the test is terminated but it does not result in the CI run failure. The break stays unnoticed until 
someone actually runs the tests locally and observe the test failure, 
or looks at the CI run log.

cgranade: As the comment above notes, this assertion isn't actually used as an assertion per se, but to prove to the C# compiler that `kind` is non-null. That's not an invariant of `reader.GetString()`, such that the compiler believes that `kind` can be null at this point, but we have checked all conditions under which `GetString()` can be null such that we need to tell the compiler that to suppress warnings. This is identical to another case above in the same file, where a comment explains that in a bit more detail.

src/Simulation/Simulators/OpenSystemsSimulator/DataModel/Generator.cs
```diff
@@ -0,0 +1,146 @@
...
+using System.Diagnostics;
+using System.Numerics;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using NumSharp;
+using static System.Math;
+
+namespace Microsoft.Quantum.Experimental;
+
+public record class GeneratorCoset(
```
kuzminrobin, line 13: Does "Coset" mean something like a "co-set" (addendum to a set, complementary set)?

src/Simulation/Simulators/OpenSystemsSimulator/DataModel/Generator.cs
```diff
@@ -0,0 +1,146 @@
...
+                        throw reader.CurrentState.Exception($"Generator kind was ExplicitEigenvalueDecomposition, but no eigenvalues were provided.");
+                    }
+                    Debug.Assert(eigenvalues is not null);
+
+                    if (eigenvectors == null)
+                    {
+                        throw reader.CurrentState.Exception($"Generator kind was ExplicitEigenvalueDecomposition, but no eigenvectors were provided.");
+                    }
+                    Debug.Assert(eigenvectors is not null);
+
```
kuzminrobin, line 108: I have an impression that in both cases the `Debug.Assert()` is redundant after the corresponding `if`.
```suggestion
                    if (eigenvalues == null)
                    {
                        throw reader.CurrentState.Exception($"Generator kind was ExplicitEigenvalueDecomposition, but no eigenvalues were provided.");
                    }

                    if (eigenvectors == null)
                    {
                        throw reader.CurrentState.Exception($"Generator kind was ExplicitEigenvalueDecomposition, but no eigenvectors were provided.");
                    }
```

src/Simulation/Simulators/OpenSystemsSimulator/DataModel/Generator.cs
```diff
@@ -0,0 +1,146 @@
...
+                    }
+                }
+            }
+            else if (value is UnsupportedGenerator _)
+            {
+                writer.WriteStringValue("Unsupported");
+            }
+        }
+    }
+}
```
kuzminrobin, line 146: (May be paranoid, feel free to ignore)
If the param named `value` gains a new value (that is not `ExplicitEigenvalueDecomposition` and is not `UnsupportedGenerator`) then both the `if` (L126) and `else if` (L140) will _fail_ and the line 125 (`writer.WritePropertyName("data")`) will (silently) NOT be followed with the sensible data. Any subsequent output (`writer.Write...`) may result in wrong/unexpected data in the `writer`.
Consider placing something like `else { .. Compile-time failure .. }` after L143.

src/Simulation/Simulators/OpenSystemsSimulator/DataModel/Instrument.cs
```diff
@@ -3,90 +3,95 @@
...
+            case ZMeasurementInstrument zInstrument:
+                writer.WriteStartObject();
+                writer.WritePropertyName("ZMeasurement");
+                JsonSerializer.Serialize(writer, zInstrument);
+                writer.WriteEndObject();
+                break;
+
+            default:
+                throw new JsonException($"Enum variant {value.GetType()} not yet supported.");
+        }
```
kuzminrobin, line 70: (Minor optimization and readability) Consider extracting the parts common for `case` arms to outside of `switch`:
```suggestion
        writer.WriteStartObject();
        switch (value)
        {
            case EffectsInstrument effectsInstrument:
                writer.WritePropertyName("Effects");
                JsonSerializer.Serialize(writer, effectsInstrument.Effects);
                break;

            case ZMeasurementInstrument zInstrument:
                writer.WritePropertyName("ZMeasurement");
                JsonSerializer.Serialize(writer, zInstrument);
                break;

            default:
                throw new JsonException($"Enum variant {value.GetType()} not yet supported.");
        }
        writer.WriteEndObject();
```

src/Simulation/Simulators/OpenSystemsSimulator/DataModel/State.cs
```diff
@@ -1,116 +1,107 @@
...
+    {
+        NQubits = nQubits;
     }
+}
 
-    public class StabilizerState : State
+public class StabilizerState : State
+{
+    // Design note:
+    // We could use the same array converter as used of complex-like arrays
```
kuzminrobin, line 28: I have an impression of some typo in the highlighted fragment
> // We could use the same array converter **as used of complex-like arrays**

src/Simulation/Simulators/OpenSystemsSimulator/DataModel/Generator.cs
cgranade: A "coset" is the set obtained by multiplying all elements of another set by a given element. For example, if 𝑆 = {𝑠₀, 𝑠₁, …}, then the left coset 𝑎𝑆 = {𝑎𝑠₀, 𝑎𝑠₁, …}, the right coset 𝑆𝑏 = {𝑠₀𝑏, 𝑠₁𝑏, …}, and one can consider cosets in the most general form of 𝑎𝑆𝑏 = {𝑎𝑠₀𝑏, 𝑎𝑠₁𝑏, …}.

Here, a generator coset is used to denote that a generator 𝐺 = −𝑖𝐿 + 𝐷 generates a continuous time monoid Λ(𝑡) = 𝑒^{𝑡𝐺} such that noise of the form Φ_post Λ(𝑡) Φ_pre is a coset of the monoid given by the generator 𝐺. Colloquially, this captures noise where some fixed time-independent thing may or may not happen, a time dependent thimg happens, and then a second fixed time-independent thing may or may not happen.

src/Simulation/Simulators/OpenSystemsSimulator/DataModel/Generator.cs
cgranade: Here as well, the extraneous asserts are used to suppress warnings in C# by providing the compiler with appropriate metadata to prove that neither `eigenvectors` nor `eigenvalues` is `null` in the case that no exception is thrown.

src/Simulation/Simulators/OpenSystemsSimulator/DataModel/State.cs
cgranade:
```suggestion
    // We could use the same array converter as used to convert complex-like arrays
```

src/Simulation/Simulators/OpenSystemsSimulator/DataModel/Generator.cs
kuzminrobin: Hm.... OK. I thought the `if` with nested `throw` was enough to convince the compiler that after the `if` the var is not `null`.

src/Simulation/Simulators/OpenSystemsSimulator/DataModel/State.cs
kuzminrobin: Looks more clear, thanks.

src/Simulation/Simulators/OpenSystemsSimulator/DataModel/Generator.cs
cgranade: It almost always is, agreed; I'm honestly not sure why Roslyn failed to conclude nonnull in this case.

src/Simulation/Simulators/OpenSystemsSimulator/DataModel/StateConverter.cs
```diff
@@ -1,112 +1,100 @@
...
+                ),
+                _ => throw new JsonException($"Unknown state kind {kind}.")
+            }
+        );
+    }
+
+    public override void Write(Utf8JsonWriter writer, State value, JsonSerializerOptions options)
+    {
+        var arrayConverter = new ComplexArrayConverter();
+
```
kuzminrobin, line 64: (Paranoid)
Consider moving L63 to between L81 and L82. Otherwise L63 increases the probability of the stack overflow during L65 - L81 (and probably during L83 - L98).

src/Simulation/Simulators/OpenSystemsSimulator/DataModel/StateConverter.cs
```diff
@@ -1,112 +1,100 @@
...
-                        writer.WriteStartObject();
-                            writer.WritePropertyName("n_qubits");
-                            writer.WriteNumberValue(stabilizerState.NQubits);
-                            writer.WritePropertyName("table");
-                            JsonSerializer.Serialize(writer, array);
-                        writer.WriteEndObject();
+                        PureState _ => "Pure",
+                        MixedState _ => "Mixed",
+                        StabilizerState _ => "Stabilizer",
+                        _ => throw new JsonException($"Unknown state type {value.GetType()}.")
```
kuzminrobin, line 76: (Paranoid, up to you)
During writing the `writer.Flush()` immediately before `throw new JsonException()` can probably simplify the failure debugging/analysis.

src/Simulation/Simulators/OpenSystemsSimulator/NativeInterface.cs
```diff
@@ -1,265 +1,302 @@
...
-    ///     Abstracts away calls to and from the experimental simulators DLL.
-    /// </summary>
-    internal static class NativeInterface
+    public const string DLL_NAME = "Microsoft.Quantum.Experimental.Simulators.Runtime.dll";
+
+    [DllImport(DLL_NAME, ExactSpelling=true, CallingConvention=CallingConvention.Cdecl, EntryPoint="lasterr")]
+    private static extern string _LastError();
+
+    [DllImport(DLL_NAME, ExactSpelling=true, CallingConvention=CallingConvention.Cdecl, EntryPoint="lastbacktrace")]
+    private static extern string? _LastBackTrace();
```
kuzminrobin, line 73: (Subjective judgement)
Passing nullable value (or a reference?) across DLL boundary makes me somewhat uncertain... I assume that _nullability_ can be implemented differently on different sides of the DLL boundary. And class instances (`string`), as opposed to primitive types, can also be implemented differently...

cgranade: Here, the `DllImport` attribute causes the .NET runtime to use P/Invoke in marshalling data across the managed/unmanaged DLL boundary — in particular, I believe that P/Invoke should marshal `char* result = nullptr` as a managed null reference of type `System.String`. Since in the .NET runtime, all reference types are nullable, nullability data (that is, `string` vs `string?`) should only affect nullability metadata used by the compiler for type checking at the C# level, such that I don't think there's any issues with it being implemented differently.

That is, if this line was changed to `private static extern string _LastBackTrace()`, the emitted CIL would differ only by nullability metadata used at compile-time when linking against this DLL.

src/Simulation/Simulators/OpenSystemsSimulator/DataModel/Instrument.cs
cgranade: I'd prefer not to in this case for a couple reasons:

- In some cases, enum variants may be serialized as a single string rather than as an object (in particular, that's the case for variants of type unit).
- Doing so would leave the writer in an inconsistent state if an exception occurs (e.g.: the exception thrown by `default`), as an object would then be started but not ended.

global.json
```diff
@@ -1,5 +1,5 @@
 {
     "msbuild-sdks": {
-        "Microsoft.Quantum.Sdk": "0.24.205218-beta"
+        "Microsoft.Quantum.Sdk": "0.24.210350-beta"
```
kuzminrobin, line 3: FYI: the latest release is 0.24.210930.

src/Simulation/qdk_sim_rs/src/linalg/expm.rs
```diff
@@ -0,0 +1,288 @@
...
+//
+//         e^{A} ≈ R_pq(A) ≔ [D_pq(A)]⁻¹ N_pq(A)
+//
+//     where
+//
+//         N_pq(A) ≔ Σ_j=0^p [(p + q - j)! p!] / [(p + q)! j! (p - j)!] A^j
+//
+//     and
+//
+//         D_pq(A) ≔ Σ_j=0^q [(p + q - j)! p!] / [(p + q)! j! (q - j)!] (-A)^j.
```
kuzminrobin, line 33: Consider removing empty lines:
```suggestion
// NB: We use the notation of
//     https://www.cs.jhu.edu/~misha/ReadingSeminar/Papers/Moler03.pdf.
//     In particular:
//         e^{A} ≈ R_pq(A) ≔ [D_pq(A)]⁻¹ N_pq(A)
//     where
//         N_pq(A) ≔ Σ_j=0^p [(p + q - j)! p!] / [(p + q)! j! (p - j)!] A^j
//     and
//         D_pq(A) ≔ Σ_j=0^q [(p + q - j)! p!] / [(p + q)! j! (q - j)!] (-A)^j.
```

src/Simulation/qdk_sim_rs/src/processes/apply.rs
```diff
@@ -244,3 +340,600 @@ pub(crate) fn apply_pauli_channel(
...
+                        c64!(-0.009666069915131954+0.013842256732161827 i),
+                        c64!(0.017373442929624686)
+                    ]
+                ];
+                for (actual, expected) in actual.iter().zip(expected.iter()) {
+                    assert_abs_diff_eq!(actual, expected, epsilon = 1e-6);
+                }
+                Ok(())
+            }
+            _ => panic!("did not return density operator as expected"),
```
kuzminrobin, line 936: The fragment `"did not return density operator as expected"` occurs more than once. Consider defining it as a global constant or a macro. Or event the whole fragment `panic!("did not return density operator as expected")` as a (shorter) macro.

kuzminrobin:
> @kuzminrobin: Thanks for your earlier review and approval! The PR has changed a bit now that it's no longer in draft, so please let me know if your previous approval still applies.

My previous approval still applies.

src/Simulation/qdk_sim_rs/src/linalg/expm.rs
cgranade: I personally tend to prefer blank lines around formula, as that's what I'm used to from technical publishing in mathematics, but happy to make the change if you feel strongly about it.

kuzminrobin: My opinion is not strong, feel free to leave as is.

src/Simulation/qdk_sim_rs/src/processes/apply.rs
kuzminrobin: That's OK.

cgranade:
Changed to target main directly, as the feature wound up being easier to write as one larger PR instead of smaller PRs against the feature.

PR #1013, opened 2022-05-15
Update decompositions for latest compiler

swernli:
This change updates the decompositions for the latest compiler features, allowing for iteration through ranges, allowing extension to arbitrary controls.

### DIFF
--- a/src/Simulation/TargetDefinitions/Decompositions/Utils.qs
+++ b/src/Simulation/TargetDefinitions/Decompositions/Utils.qs
@@ -54,15 +54,29 @@ namespace Microsoft.Quantum.Intrinsic {
         }
     }
 
-    internal operation CollectControls(ctls : Qubit[], aux : Qubit[]) : Unit is Adj {
+    /// Collects the given list of control qubits into one or two of the given auxiliarly qubits, using
+    /// all but the last qubits in the auxiliary list as scratch qubits. The auxiliary list must be
+    /// big enough to accomodate the data, so it is usually smaller than controls list by number of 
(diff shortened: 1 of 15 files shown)


src/Simulation/TargetDefinitions/Decompositions/Utils.qs
```diff
@@ -166,21 +54,18 @@ namespace Microsoft.Quantum.Intrinsic {
...
-            within {
-                for numPair in 0 .. numControlPairs - 1 { // constant depth
-                    PhaseCCX(controls[2*numPair], controls[2*numPair + 1], temps[numPair]);
-                }
-            }
-            apply {
-                let newControls = numControls % 2 == 0 ? temps | temps + [controls[numControls - 1]];
-                op(newControls, arg);
-            }
+    internal operation CollectControls(ctls : Qubit[], aux : Qubit[]) : Unit is Adj {
```
cgranade, line 57: Some comments here may help explain what these two utility operations are intended for.

cgranade (approved):

src/Simulation/TargetDefinitions/Decompositions/Utils.qs
swernli: Writing out the comments made realize the double controlled case worked but was using qubits inefficiently. I have a tweak that I think will correctly optimize those cases.

swernli: Ok, I've added comments and adjusted the logic to correctly use fewer auxiliary qubits for doubly-controlled unitaries.

PR #1015, opened 2022-05-16
Repeated PR: Added target Capability to Entry Point Driver.

kuzminrobin:
Repeated #1002 (after revert).


kuzminrobin:
Force-merging since the overall qdk.release succeeded, and this is a repeated PR after a revert.

Issue #389, opened 2020-09-28
Rework use of `internal` to allow defining of callabes that are not part of public API but implemented outside of original binary


swernli:
If I mark a Q# callable as internal but also as `body intrinsic` it must be implemented in the same binary or in a binary that has been allowed via the [InternalsVisibleTo](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.internalsvisibletoattribute?view=netcore-3.1) attribute. If we want to allow implementations of Q# helper callabes to come from outside of this repo, we need a different way to expose those callables that is neither public nor internal. Perhaps an attribute on a non-internal declaration that marks it as not part of the API surface? Some other trick that makes use of specific patterns in the generated code?

See discussion for this issue in #367, as well as original context in #249.

swernli:
The interface based design introduced when addressing #432 in the feature/decomp branch may be our best path forward to fixing this. Extending that design would allow for `body intrinsic` callables to be generated into concrete classes that expect an implementation that corresponds to a generated interface to be provided. This will allow for any external package to provide the implementation of the callabe, and is conceptually a stepping stone toward the QIR designs. Whether we actually take that incremental step now just depends on how much more we want to invest in the C# generation vs wait to take advantage of the QIR for this purpose.

@bettinaheim for FYI.

PR #1023, opened 2022-05-24
Edits to target packages to make intrinsics accessible in regenerated library code

bettinaheim:
Note that I did not make anything public that is part of the Q# Core, only the callables that are used in the target packages that replace it. This should hence be a non-breaking change. Unfortunately, having callables in target packages be internal makes it inpossible to regenerate the code that calls into it, and hence poses a problem.

### DIFF
--- a/src/Simulation/Simulators.Tests/TestProjects/TargetedExe/TargetedExe.csproj
+++ b/src/Simulation/Simulators.Tests/TestProjects/TargetedExe/TargetedExe.csproj
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>net6.0</TargetFramework>
-	<QscVerbosity>detailed</QscVerbosity>
+    <QscVerbosity>detailed</QscVerbosity>
     <!-- we will provide our own -->
     <IncludeCSharpRuntime>false</IncludeCSharpRuntime>
     <IncludeQSharpCorePackages>false</IncludeQSharpCorePackages>
--- a/src/Simulation/Simulators.Tests/TestProjects/TargetedExe/TargetedExe.csproj
+++ b/src/Simulation/Simulators.Tests/TestProjects/TargetedExe/TargetedExe.csproj
@@ -14,7 +14,7 @@
   <ItemGroup>
     <ProjectReference Include="..\..\..\EntryPointDriver\Microsoft.Quantum.EntryPointDriver.csproj" />
     <ProjectReference Include="..\..\..\QSharpCore\Microsoft.Quantum.QSharp.Core.csproj" />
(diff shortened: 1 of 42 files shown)


swernli:
I don't understand the problem this change is trying to solve, but I would be opposed to merging a change like this into main. Since this is just merging into a working branch it should be fine, so we can defer that discussion to a later date.

bettinaheim:
> I don't understand the problem this change is trying to solve, but I would be opposed to merging a change like this into main. Since this is just merging into a working branch it should be fine, so we can defer that discussion to a later date.

@swernli The issue is that I need to instantiate the callables in the code that is regenerated. I may have missed something in how this was intended to work, but the C# generation as written was incompatible with using the target packages. I hence made the necessary update to have it in a working state. As written, I need to be able to instantiate the class that implements each gate as defined in the target package, hence the internal is interfering with that.

swernli:
Could this just be an issue of regeneration of C# callables not respecting the `GenerateConcreteIntrinsics` property? C# generation with the target packages works in this repo when the target packages are testing with the simulator (see for example https://github.com/microsoft/qsharp-runtime/tree/main/src/Simulation/Simulators.Type1.Tests and corresponding test projects for the other target packages). We've run into quirks with internal intrinsics before (#389) and the concrete intrinsics feature was meant to solve this.

bettinaheim:
> Could this just be an issue of regeneration of C# callables not respecting the `GenerateConcreteIntrinsics` property? C# generation with the target packages works in this repo when the target packages are testing with the simulator (see for example https://github.com/microsoft/qsharp-runtime/tree/main/src/Simulation/Simulators.Type1.Tests and corresponding test projects for the other target packages). We've run into quirks with internal intrinsics before (#389) and the concrete intrinsics feature was meant to solve this.

Setting GenerateConcreteIntrinsic to true in the corresponding project file (i.e. in QCIExe and in QSharpExe) did not solve the issue. You can try it out with package version 0.24.41904-alpha. However, there were a couple of issues I had to fix across repos, and another change in 0.24.41904-alpha may be interfering, so it may have been a case of it getting lost in the weeds. I gave it another try, now that I know what *all* the changes and issues were that needed to be solved: https://dev.azure.com/ms-quantum-public/Microsoft%20Quantum%20(public)/_build/results?buildId=41944&view=results

src/Simulation/Simulators.Tests/TestProjects/TargetedExe/TargetedExe.csproj
```diff
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>net6.0</TargetFramework>
-    <QscVerbosity>detailed</QscVerbosity>
+	<QscVerbosity>detailed</QscVerbosity>
```
bettinaheim, line 5:
```suggestion
    <QscVerbosity>detailed</QscVerbosity>
```

PR #1027, opened 2022-05-27
Fixed Parity() in C# RT.

### LINKED ISSUE #993: An issue about the Parity function
## Description

Hello! When executing the following testcase, the program throws an exception. It seems that the value of argument `a` should not be less than `0`. But the [document](https://docs.microsoft.com/zh-cn/qsharp/api/qsharp/microsoft.quantum.bitwise.parity) doesn't state this condition. 

## Testcase

~~~c#
namespace NISLNameSpace {
	open Microsoft.Quantum.Bitwise;

	@EntryPoint()
	operation main() : Unit {
		Message($"{Parity(-1)}");
	}
}
~~~

## Actual Behavior

~~~c#
Unhandled exception. System.OverflowException: Value was either too large or too small for a UInt64.
 ---> NISLNameSpace.main on E:\GithubProjects\QuantumFuzzing\src\Fuzzing\qsharpPattern\Program.qs:line 7
~~~

## Expected Behavior

show the range of argument `a` in the document

## Environment

`Operating system`: Windows 10

`dotnet version`: 3.1.414

`QDK`: 0.24.201332

## Command

dotnet run

kuzminrobin:
Fixes #993.

### DIFF
--- a/src/Simulation/QSharpFoundation/Bitwise/Bitwise.cs
+++ b/src/Simulation/QSharpFoundation/Bitwise/Bitwise.cs
@@ -49,11 +49,13 @@ namespace Microsoft.Quantum.Bitwise
             static long ParityFunc(long val)
             {
                 // parity function using idea described at http://graphics.stanford.edu/~seander/bithacks.html#ParityMultiply
-                ulong v = System.Convert.ToUInt64(val);
+                ulong v = unchecked((ulong)val);
                 v ^= v >> 1;
                 v ^= v >> 2;
                 v = (v & 0x1111111111111111UL) * 0x1111111111111111UL;
-                return System.Convert.ToInt64((v >> 60) & 1);
+
+                var retVal = unchecked((long)((v >> 60) & 1));
+                return retVal;
             }
 
             public Native(IOperationFactory m) : base(m) { }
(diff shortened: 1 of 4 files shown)


src/Simulation/QSharpFoundation/Bitwise/Bitwise.cs
```diff
@@ -49,11 +49,13 @@ public class Native : Parity
             static long ParityFunc(long val)
             {
                 // parity function using idea described at http://graphics.stanford.edu/~seander/bithacks.html#ParityMultiply
-                ulong v = System.Convert.ToUInt64(val);
+                ulong v = unchecked((ulong)val);
```
cesarzc, line 52: Does this line really needs to be unchecked? It seems like without any modification it could be casted to `ulong`, even using `System.Convert.ToUInt64`.

kuzminrobin: Using [System.Convert.ToUInt64()](https://docs.microsoft.com/en-us/dotnet/api/system.convert.touint64?view=net-6.0#system-convert-touint64(system-int64)) it throws (if arg is negative), see the bug in the Description of this PR.
Using explicit cast with `(ulong)` it may or may not throw depending on the [CheckForOverflowUnderflow compiler option](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/checked-and-unchecked).
With `unchecked()` I expect it to never throw.

cesarzc (approved):

PR #1034, opened 2022-06-07
Removing doc generation from tests

anpaz:


kuzminrobin (approved):

kuzminrobin:
Looks like there was an intermittent failure for `microsoft.qsharp-runtime`, I pressed "re-run".

PR #1040, opened 2022-06-14
Use record to get `with` keyword.

cgranade:
This PR modifies the `SubmissionOptions` class to be a record class, allowing other code to use the `with` keyword to modify instances of the class on a field-by-field basis. The existing `With` method can then be refactored to use the `with` keyword.

### DIFF
--- a/src/Simulation/Core/Submitters/SubmissionOptions.cs
+++ b/src/Simulation/Core/Submitters/SubmissionOptions.cs
@@ -5,32 +5,37 @@ using System.Collections.Immutable;
 
 #nullable enable
 
+namespace System.Runtime.CompilerServices
+{
+    public class IsExternalInit {}
+}
+
 namespace Microsoft.Quantum.Runtime.Submitters
 {
     /// <summary>
     /// Options for a job submitted to Azure Quantum.
     /// </summary>
-    public class SubmissionOptions
+    public record SubmissionOptions
     {
         /// <summary>
         /// A name describing the job.
         /// </summary>
-        public string FriendlyName { get; }
+        public string FriendlyName { get; init; }
 
         /// <summary>
         /// The number of times the program will be executed.
         /// </summary>
-        public int Shots { get; }
+        public int Shots { get; init; }
 
         /// <summary>
         /// Additional target-specific parameters for the job.
         /// </summary>
-        public ImmutableDictionary<string, string> InputParams { get; }
+        public ImmutableDictionary<string, string> InputParams { get; init; }
 
         /// <summary>
         /// The target capability.
         /// </summary>
-        public string TargetCapability { get; set; }
+        public string TargetCapability { get;  init; }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


kuzminrobin (approved):

src/Simulation/Core/Submitters/SubmissionOptions.cs
```diff
@@ -60,7 +65,12 @@ public class SubmissionOptions
             int? shots = null,
             ImmutableDictionary<string, string>? inputParams = null,
             string? targetCapability = null) =>
-            new SubmissionOptions(
-                friendlyName ?? FriendlyName, shots ?? Shots, inputParams ?? InputParams, targetCapability ?? TargetCapability);
+            this with
```
samarsha, line 68: I'd rather remove `With`, since it 's redundant.

src/Simulation/Core/Submitters/SubmissionOptions.cs
```diff
@@ -5,32 +5,37 @@
...
+    public class IsExternalInit {}
+}
+
 namespace Microsoft.Quantum.Runtime.Submitters
 {
     /// <summary>
     /// Options for a job submitted to Azure Quantum.
     /// </summary>
-    public class SubmissionOptions
+    public record SubmissionOptions
```
samarsha, line 18: Maybe it would be better to remove `Default` and the private constructor, and instead give each init property a default value?

src/Simulation/Core/Submitters/SubmissionOptions.cs
```diff
@@ -5,32 +5,37 @@
 
 #nullable enable
 
+namespace System.Runtime.CompilerServices
+{
+    public class IsExternalInit {}
```
samarsha, line 10: As a public type, won't this conflict if another reference uses the same workaround? Can we update Microsoft.Quantum.Runtime.Core target framework to net6.0 instead?

src/Simulation/Core/Submitters/SubmissionOptions.cs
cgranade: Sounds good, my only thought was to not break existing callers, but happy to go on and pull that method, then.

src/Simulation/Core/Submitters/SubmissionOptions.cs
cgranade: I've tested in other places, and that type seems to be special-cased to not conflict, allowing this as a suggested workaround (as per https://developercommunity.visualstudio.com/t/error-cs0518-predefined-type-systemruntimecompiler/1244809).

src/Simulation/Core/Submitters/SubmissionOptions.cs
cgranade: To unblock e2e, would it be OK to mark this and `Default` as obsolete for now?

AtariDreams:
Any updates on this?

PR #1041, opened 2022-06-15
Adjusting for Sdk updates to allow target specific diagnostics in libraries, and always passing target capability

bettinaheim:


kuzminrobin (approved):
Approving to unblock, but I didn't dive deep in to understand the changes.

cesarzc (approved):

PR #1043, opened 2022-06-17
Added @kuzminrobin's code ownership.

kuzminrobin:


.github/CODEOWNERS
```diff
@@ -1,2 +1,44 @@
-# Automatically tag in libraries maintainers on Q# API changes.
-*.qs    @microsoft/quantumlibraries
```
kuzminrobin, line 2: I see that the team `@microsoft/quantumlibraries` exists, has 4 [members](https://github.com/orgs/microsoft/teams/quantumlibraries/members) - Cassandra, Mathias, Guen, and XField. There is no Q#RT among the [team's repos](https://github.com/orgs/microsoft/teams/quantumlibraries/repositories). Hence the [error reported by GitHub](https://github.com/microsoft/qsharp-runtime/pull/1042/files).

I don't think we want those 4 members to be added as reviewers for every .qs file changed in Q#RT repo. So I'm removing the line `*.qs @microsoft/quantumlibraries` from the `.github/CODEOWNERS` file.

cgranade: This was originally done to make sure that API breaking changes were correctly identified; part of the Q# libraries are in this repo due to dependency ordering, and it is helpful for the libraries team to be able to review. I'll leave it to @msoeken, but my suggestion would be to keep this line and modify the GH team to be accurate to the current membership of the Q# libraries team.

kuzminrobin: Thank you for the explanation, @cgranade. I will restore these lines. But before that it seems reasonable to add the Q#RT repo to the [repos](https://github.com/orgs/microsoft/teams/quantumlibraries/repositories) of this team (and the write permissions, see [here](https://docs.github.com/es/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners) "When the code owner is a team, that team must be visible and it must have write permissions, even if all the individual members of the team already have write permissions...").
Seems that I have no permission to add the Q#RT repo to the team. Can anybody please do that for me? @bettinaheim, @cgranade?

kuzminrobin: Any update?

DmitryVasilevsky (approved):

PR #1052, opened 2022-07-05
Fix BinSkim issues for native simulator

swernli:
This updates the native simulator to participate in BinSkim checks by ensuring that it produces symbols even when built in release mode. Any further BinSkim fixes will be added if necessary.


swernli:
I've confirmed the e2e build now includes symbols (pdb) for the native simulator as part of the build artifacts that get consumed by later stages of the pipeline. This should unblock BinSkim scanning.

kuzminrobin (approved):

PR #1054, opened 2022-07-08
Add Sparse Sim to manifest, link against Spectre libs

swernli:
This adds the sparse state simulator to the manifest script to ensure it gets properly scanned. It also updates the SparseSimulator native binary to compile with control flow guard security flags and link against Spectre mitigated static runtime such that it can pass the manifest checks. For ease of management, all security flags and runtime linking logic has been pulled into a common CMake macro so that the full state simulator, sparse state simulator, and QIR runtime can all share the same setup.

### DIFF
--- a/src/Common/cmake/secure_dependencies.cmake
+++ b/src/Common/cmake/secure_dependencies.cmake
@@ -38,7 +38,7 @@ macro(locate_win32_spectre_static_runtime)
             NAMES vswhere
             PATHS "$ENV{ProgramFiles\(x86\)}/Microsoft Visual Studio/Installer")
         if (NOT ${vswhere})
-            message(FATAL_ERROR "Could not locate vswhere.exe - unable to source vc redistributable")
+            message(FATAL_ERROR "Could not locate vswhere - unable to search for installed vcruntime libraries.")
         endif()
         execute_process(
             COMMAND "${_vswhere_tool}" -latest -products * -requires Microsoft.VisualStudio.Component.VC.Tools.x86.x64 -find **/14.29.*/**/lib/spectre/x64 -sort
--- a/src/Common/cmake/secure_dependencies.cmake
+++ b/src/Common/cmake/secure_dependencies.cmake
@@ -1,0 +1,71 @@
+#===============================================================================
+# Configure the flags used to ensure that required security settings are enabled
+macro(configure_security_flags)
+    # Enable Control Flow Guard
+    add_link_options("LINKER:/guard:cf")
+
+    # Enable Compiler supported Spectre mitigations
+    if (NOT APPLE)
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -mspeculative-load-hardening -mretpoline")
+        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -mspeculative-load-hardening -mretpoline")
+    endif()
+endmacro()
+
+#===============================================================================
+# Set MSVC static runtime library policy
+# NOTE: Since this sets a cmake policy for ensuring Windows compilation uses static runtime,
+# this policy MUST be set before any projects are declared for it to take effect.
+macro(set_msvc_static_runtime_policy)
+    # Enforce use of static runtime (avoids target machine needing msvcrt installed).
+    # Must set policy CMP0091 before any projects are declared (see https://cmake.org/cmake/help/latest/variable/CMAKE_MSVC_RUNTIME_LIBRARY.html#cmake-msvc-runtime-library)
+    cmake_policy(SET CMP0091 NEW)
+    set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
+endmacro()
+
+#===============================================================================
+# Locate Spectre mitigated static C runtime on Windows. Sets the `SPECTRE_LIBS`
+# variable for use in later calls to `target_link_libraries`.
+# NOTE: This sets a cmake policy for ensuring Windows compilation uses static runtime.
+# this policy MUST be set before any projects are declared for it to take effect.
+macro(locate_win32_spectre_static_runtime)
+    if (WIN32)
+        # Locate the vswhere application, which will provide paths to any installed Visual Studio instances.
+        # By invoking it with "-find **/lib/spectre/x64" we will find any Spectre mitigated libaries that
+        # have been installed.
+        find_program(_vswhere_tool
+            NAMES vswhere
+            PATHS "$ENV{ProgramFiles\(x86\)}/Microsoft Visual Studio/Installer")
+        if (NOT ${vswhere})
+            message(FATAL_ERROR "Could not locate vswhere.exe - unable to source vc redistributable")
+        endif()
+        execute_process(
+            COMMAND "${_vswhere_tool}" -latest -products * -requires Microsoft.VisualStudio.Component.VC.Tools.x86.x64 -find **/14.29.*/**/lib/spectre/x64 -sort
+            OUTPUT_VARIABLE _vs_install_loc_out
+            RESULT_VARIABLE _vs_where_exitcode
+            OUTPUT_STRIP_TRAILING_WHITESPACE)
+        file(TO_CMAKE_PATH "${_vs_install_loc_out}" SPECTRE_LIB_PATH_OUT)
+        string(REGEX REPLACE "[\r\n]+" ";" SPECTRE_LIB_PATH ${SPECTRE_LIB_PATH_OUT})
+        message(INFO "*** install loc: ${SPECTRE_LIB_PATH}")
+
+        # Locate the spectre mitigated runtime libraries and fail if they can't be found. Targets in this
+        # cmake project can use the variables to explicitly link these libraries rather than using the 
+        # non-mitigated libraries that are found by default.
+        find_library(LIBCMT_SPECTRE_REL libcmt PATHS ${SPECTRE_LIB_PATH} REQUIRED)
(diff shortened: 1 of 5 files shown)


src/Common/cmake/secure_dependencies.cmake
```diff
@@ -0,0 +1,73 @@
...
+#===============================================================================
+# Set MSVC static runtime library policy
+# NOTE: Since this sets a cmake policy for ensuring Windows compilation uses static runtime,
+# this policy MUST be set before any projects are declared for it to take effect.
+macro(set_msvc_static_runtime_policy)
+    # Enforce use of static runtime (avoids target machine needing msvcrt installed).
+    # Must set policy CMP0091 before any projects are declared (see https://cmake.org/cmake/help/latest/variable/CMAKE_MSVC_RUNTIME_LIBRARY.html#cmake-msvc-runtime-library)
+    cmake_policy(SET CMP0091 NEW)
+    set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
+endmacro()
```
kuzminrobin, line 25: The fragment "for ensuring **Windows** compilation uses static runtime" makes me think that the `if (WIN32)` might be needed:
```suggestion
macro(set_msvc_static_runtime_policy)
    if (WIN32)
        # Enforce use of static runtime (avoids target machine needing msvcrt installed).
        # Must set policy CMP0091 before any projects are declared (see https://cmake.org/cmake/help/latest/variable/CMAKE_MSVC_RUNTIME_LIBRARY.html#cmake-msvc-runtime-library)
        cmake_policy(SET CMP0091 NEW)
        set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
    endif (WIN32)
endmacro()
```

src/Simulation/NativeSparseSimulator/CMakeLists.txt
```diff
@@ -1,13 +1,24 @@
...
 set(CMAKE_CXX_EXTENSIONS OFF)
 
 set(CMAKE_MACOSX_RPATH 1)
+
+locate_win32_spectre_static_runtime()
+configure_security_flags()
+
 # Main build files
 add_library(Microsoft.Quantum.SparseSimulator.Runtime SHARED factory.cpp capi.cpp)
+target_link_libraries(Microsoft.Quantum.SparseSimulator.Runtime ${SPECTRE_LIBS})
```
kuzminrobin, line 21: I assume that 
`${SPECTRE_LIBS}` expands to _nothing_ on non-Win platforms (because `SPECTRE_LIBS` is defined inside of `if (WIN32)` and on other platforms it stays undefined).

src/Common/cmake/secure_dependencies.cmake
```diff
@@ -0,0 +1,73 @@
...
+    if (WIN32)
+        # Locate the vswhere application, which will provide paths to any installed Visual Studio instances.
+        # By invoking it with "-find **/lib/spectre/x64" we will find any Spectre mitigated libaries that
+        # have been installed.
+        find_program(_vswhere_tool
+            NAMES vswhere
+            PATHS "$ENV{ProgramFiles\(x86\)}/Microsoft Visual Studio/Installer")
+        if (NOT ${vswhere})
+            message(FATAL_ERROR "Could not locate vswhere.exe - unable to source vc redistributable")
+        endif()
```
kuzminrobin, line 42: ;-) Definitely paranoid, but strictly speaking,
The code is looking for _one thing_ ("vswhere"), and if not found, then complains that some _other thing_ ("vswhere **.exe**") is not found.
What if, when looking for "vswhere", the "vswhere **.com**" or "vswhere **.cmd**" or "vswhere **.bat**" is found?

Consider looking for and complaining about the same thing, 
either "vswhere" both in L38 and L41,
or "vswhere **.exe**" both in L38 and L41.

src/Common/cmake/secure_dependencies.cmake
```diff
@@ -0,0 +1,73 @@
...
+macro(locate_win32_spectre_static_runtime)
+    if (WIN32)
+        # Locate the vswhere application, which will provide paths to any installed Visual Studio instances.
+        # By invoking it with "-find **/lib/spectre/x64" we will find any Spectre mitigated libaries that
+        # have been installed.
+        find_program(_vswhere_tool
+            NAMES vswhere
+            PATHS "$ENV{ProgramFiles\(x86\)}/Microsoft Visual Studio/Installer")
+        if (NOT ${vswhere})
+            message(FATAL_ERROR "Could not locate vswhere.exe - unable to source vc redistributable")
```
kuzminrobin, line 41: (nit)
If the users try to build locally and have no expected version of MS VC++ installed (e.g. our ReadMe, Prerequisites section, gets out-of-date and recommends to install the VS version older than needed), then the message "unable to source vc redistributable" may confuse them. Consider more detailed, more actionable message.

kuzminrobin: Disregard. This is the old/existing code.

src/Common/cmake/secure_dependencies.cmake
kuzminrobin: Disregard. This is the old/existing code.

src/Common/cmake/secure_dependencies.cmake
```diff
@@ -0,0 +1,73 @@
...
+        endif()
+        execute_process(
+            COMMAND "${_vswhere_tool}" -latest -products * -requires Microsoft.VisualStudio.Component.VC.Tools.x86.x64 -find **/14.29.*/**/lib/spectre/x64 -sort
```
kuzminrobin, line 44: IIRC, when adapting for VS2022, I had to do the following change:
```suggestion
            COMMAND "${_vswhere_tool}" -latest -products * -requires Microsoft.VisualStudio.Component.VC.Tools.x86.x64 -find **/14.*.*/**/lib/spectre/x64 -sort
```

src/Common/cmake/secure_dependencies.cmake
swernli: The CMake API docs clarify that this value is ignored on non-Windows platforms so it is safe to always set:
```
The value is ignored on non-MSVC compilers but an unsupported value will be rejected as an error when using a compiler targeting the MSVC ABI.
```

src/Simulation/NativeSparseSimulator/CMakeLists.txt
swernli: That is correct, making the line a no-op. This matches the behavior we use for the other native binaries.

src/Common/cmake/secure_dependencies.cmake
kuzminrobin: (I assume that with "The value is ignored" you mean "the  CMAKE_MSVC_RUNTIME_LIBRARY is ignored")

You write: "this value is ignored **on non-Windows platforms**".
The citation writes: "The value is ignored **on non-MSVC compilers**".
Our compiler is **clang**/**clang++** (including on Win), I assume it is non-MSVC compiler. Which makes me think that the value is ignored even on Windows.
This all sounds suspicious ;-), but unfortunately I cannot (at least quickly) provide any sensible recommendation. Leaving it to you.

kuzminrobin: Disregard, this is existing/old code.

kuzminrobin (approved):

src/Common/cmake/secure_dependencies.cmake
swernli: Unfortunately the VS2022 version binaries didn't pass the BinSkim checks. You can see that in the log from the SDL pipeline [here](https://dev.azure.com/ms-quantum/Quantum%20Program/_build/results?buildId=220094&view=logs&j=663ee715-dcc5-5b8a-8432-471131753a80&t=c27f7e01-8477-5121-dfc4-3699acde542a&l=137), and the lines in the build log [here](https://dev.azure.com/ms-quantum/Quantum%20Program/_build/results?buildId=220094&view=logs&j=01472334-6cb8-5a65-5848-8350d96b6b87&t=407b8715-ddab-5d3e-3fdb-ca4f6740b4ba&l=104) showing that the selected runtime was the `14.32.31326` with Spectre mitigations. It's not clear to me why the latest binaries don't pass the checks even though they come from the spectre path, but I do know that if the `14.16` or `14.29` versions are chose those pass the checks, and both of those are installed in the build pipeline by default (the build pipeline has MSVC v143, v142, and v141 installed in the VS2022 folder). This means building locally will require installing the Spectre mitigated v142 components, which are optional installs on VS2022.

src/Common/cmake/secure_dependencies.cmake
swernli: You are right, the wording in the docs is misleading. The page I'm looking at is the docs for [CMAKE_MSVC_RUNTIME_LIBRARY](https://cmake.org/cmake/help/latest/variable/CMAKE_MSVC_RUNTIME_LIBRARY.html#cmake-msvc-runtime-library), which at the top says:

> Select the MSVC runtime library for use by compilers targeting the MSVC ABI. This variable is used to initialize the [MSVC_RUNTIME_LIBRARY](https://cmake.org/cmake/help/latest/prop_tgt/MSVC_RUNTIME_LIBRARY.html#prop_tgt:MSVC_RUNTIME_LIBRARY) property on all targets as they are created.

In this case, while we are using clang as our compiler instead of MSVC, we are using clang to compile a binary that targets the MSVC runtime, which means CMake needs to use logic to decide which runtime to pass to the compilation and linking. So even though we call into clang and lld-link, the latter is the LLVM linker for Windows platforms that targets MSVC and as such obeys these CMake variables (for more info check out https://lld.llvm.org/windows_support.html).

PR #1057, opened 2022-07-13
Addressing potential Deserialization security issue reported

DmitryVasilevsky:
Message: Call to method `Deserialize` is converted to T, allowing malicious subtypes
Description: Doing a conversion rather than a precise type check will succeed for potentially malicious subtypes.
Alert Locations:
src/Simulation/QCTraceSimulator/Utils.cs#L123
---
This is a non-issue as the buffer is fully trusted and has just been serialized into.
However, this is an attempt to change the code to comply with the rule.


swernli (approved):

DmitryVasilevsky:
Abandoning this in favor of https://github.com/microsoft/qsharp-runtime/pull/1059

PR #1063, opened 2022-07-20
Output Recording Fixups

ScottCarda-MS:
Updates the EntryPointOperationLoader to use the QIR bit-code to get Entry Point information instead of the Q# AST. This allows the loader to get the wrapped versions of the entry point operation appropriate for QIR compilations.

This also fixes a misnamed function in the API, from `__quantum__rt__integer_record_output` to `__quantum__rt__int_record_output`.

### DIFF
--- a/src/Qir/Tools/Microsoft.Quantum.Qir.Runtime.Tools.csproj
+++ b/src/Qir/Tools/Microsoft.Quantum.Qir.Runtime.Tools.csproj
@@ -36,7 +36,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.Logging" Version="5.0.0" />
-	<PackageReference Include="Microsoft.Quantum.QirGeneration" Version="0.25.223055-beta" />
+    <PackageReference Include="Microsoft.Quantum.QirGeneration" Version="0.25.223055-beta" />
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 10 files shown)


src/Qir/Tools/Microsoft.Quantum.Qir.Runtime.Tools.csproj
```diff
@@ -36,7 +36,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.Logging" Version="5.0.0" />
-    <PackageReference Include="Microsoft.Quantum.QirGeneration" Version="0.24.217480-beta" />
+	<PackageReference Include="Microsoft.Quantum.QirGeneration" Version="0.25.223055-beta" />
```
kuzminrobin, line 39: The indentation of 4 spaces has been replaced with a single tab. You likely didn't mean to do that.

kuzminrobin (approved):
The .cpp/.hpp part looks good to me. Approving to unblock.
For the .cs part I would recommend at least one more approval.

src/Qir/Tools/Microsoft.Quantum.Qir.Runtime.Tools.csproj
ScottCarda-MS: fixed

PR #1065, opened 2022-07-23
Upgraded clang from 13 to 14[.0.6].

kuzminrobin:
Parallel [PR](https://github.com/microsoft/qdk/pull/274).
Fixes the DevOps item 42842.

The CI build of `microsoft.qsharp-runtime.e2e (Build Build Native components linux) ` is expected to fail (because the mentioned above parallel PR's changes are not in main yet). But overall `qdk.release` (0.25.227433) is expected to succeed.

### DIFF
--- a/src/Qir/Runtime/prerequisites.ps1
+++ b/src/Qir/Runtime/prerequisites.ps1
@@ -3,6 +3,8 @@
 
 #Requires -Version 6.0
 
+Write-Host "##[info] Runtime/prerequisites.ps1"
+
 if ($Env:ENABLE_QIRRUNTIME -ne "false") {
     if (($IsWindows) -or ((Test-Path Env:/AGENT_OS) -and ($Env:AGENT_OS.StartsWith("Win")))) {
         if (!(Get-Command clang        -ErrorAction SilentlyContinue) -or `
@@ -36,14 +38,16 @@ if ($Env:ENABLE_QIRRUNTIME -ne "false") {
                 sudo add-apt-repository "deb https://apt.llvm.org/focal/ llvm-toolchain-focal-14 main"
             }
             sudo apt update
-            sudo apt-get install -y ninja-build clang-14 clang-tidy-14 clang-format-14
+            sudo apt-get install -y ninja-build
+            sudo apt-get install -y clang-14 clang-tidy-14 clang-format-14
         } else {
             if ($needClang) {
                 wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key|apt-key add -
                 add-apt-repository "deb https://apt.llvm.org/focal/ llvm-toolchain-focal-14 main"
             }
             apt update
-            apt-get install -y ninja-build clang-14 clang-tidy-14 clang-format-14
+            apt-get install -y ninja-build
+            apt-get install -y clang-14 clang-tidy-14 clang-format-14
         }
     }
 }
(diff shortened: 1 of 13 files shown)


src/Simulation/Native/src/simulator/factory.cpp
```diff
@@ -63,7 +63,7 @@ MICROSOFT_QUANTUM_DECL unsigned create(unsigned maxlocal)
 {
     std::lock_guard<std::shared_mutex> lock(_mutex);
 
-    size_t emptySlot = -1;
+    size_t emptySlot = (size_t)-1;
```
DmitryVasilevsky, line 66: Doesn't matter much, but probably static_cast<>?

src/Qir/Runtime/prerequisites.ps1
```diff
@@ -3,12 +3,14 @@
...
 
+Write-Host "##[info] Runtime/prerequisites.ps1"
+
 if ($Env:ENABLE_QIRRUNTIME -ne "false") {
     if (($IsWindows) -or ((Test-Path Env:/AGENT_OS) -and ($Env:AGENT_OS.StartsWith("Win")))) {
         if (!(Get-Command clang        -ErrorAction SilentlyContinue) -or `
             !(Get-Command clang-format -ErrorAction SilentlyContinue) -or `
             (Test-Path Env:/AGENT_OS)) {
-            choco install llvm --version=13.0.0 --allow-downgrade
+            choco install llvm --version=14.0.6 --allow-downgrade
```
swernli, line 13: The readme under this same Runtime folder should be updated to say we use version 14 too.

src/Simulation/Native/prerequisites.ps1
```diff
@@ -19,21 +22,25 @@ if (($IsMacOS) -or ((Test-Path Env:AGENT_OS) -and ($Env:AGENT_OS.StartsWith("Dar
...
         if ($needClang) {
             wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key|sudo apt-key add -
-            sudo add-apt-repository "deb https://apt.llvm.org/focal/ llvm-toolchain-focal-13 main"
+            sudo add-apt-repository "deb https://apt.llvm.org/focal/ llvm-toolchain-focal-14 main"
         }
         sudo apt update
-        sudo apt-get install -y clang-13
+        sudo apt-get install -y ninja-build
+        sudo apt-get install -y libomp-14-dev
+        sudo apt-get install -y clang-14
```
swernli, line 34: These can be merged into a single apt-get install call, which is how the QIR Runtime does it.

swernli (approved):
Some minor things, but otherwise looks good!

src/Simulation/Native/src/simulator/factory.cpp
kuzminrobin: I assume that the C-style casts are equivalent to `reinterpret_cast<>()`, and at least when applied to primitive types I personally prefer the C-style casts for brevity.
If nobody has strong objections I would prefer to leave as is. Otherwise no problem on my end to change, just let me know.

src/Qir/Runtime/prerequisites.ps1
kuzminrobin: I haven't been looking at that README for probably a year :-) Thanks for the reminder. Updated.

src/Simulation/Native/prerequisites.ps1
kuzminrobin: Yes, agreed. But this break-up into multiple lines originates from my observation of one of the intermediate build failures.

There ninja and clang-14 were installed (in this order) by a single `sudo apt-get install -y` command. But the installation of clang-14 was conflicting with the version installed by that moment - clang-13 (see the [parallel PR](https://github.com/microsoft/qdk/pull/274/files)). So the installation of clang-14 was failing and the installation of ninja was not done (or was reverted). So the subsequent steps were complaining for missing ninja instead of wrong clang version.

To make the failures and failure reporting consistent for the described cases I decided to break-up into multiple lines.

kuzminrobin:
The qdk.release (0.25.227449) succeeded. Force-merging.

PR #1066, opened 2022-07-25
Commented out static linking MSVC++ libs and Spectre mitigation runtime.

kuzminrobin:
Works around the issue with linker warnings. 
Item 42842.


bettinaheim (changes requested):
Blocking the PR for now, since this will reintroduce a security alert related to spectre mitigated libraries. Let's examine whether and when we could possibly do an update to clang 14.

kuzminrobin:
The changes in this PR are not the way to go. They work around SDL.vnext failure but break the BinSkim.

PR #1069, opened 2022-08-03
Cgranade/lock nightly

cgranade:
This PR locks the version of the `nightly` Rust toolchain used to build the open systems simulator, preventing upstream changes from breaking builds until we are ready to upgrade to more recent nightly builds. In particular, the removal of `std::error::Error::backtrace` from the `#![feature(backtrace)]` feature caused a recent breakage, such that we can restore PR by locking before https://github.com/rust-lang/rust/pull/99431 landed — it would be good to update to use the new provider API instead, but for now, locking the toolchain allows us to restore the build.


src/Simulation/qdk_sim_rs/rust-toolchain
```diff
@@ -0,0 +1 @@
+nightly-2022-08-01
```
ricardo-espinoza, line 1: How often will we need to update this?

ricardo-espinoza (approved):

cgranade:
Apologies, it looks like I accidentally included #1058 in this PR as well; since that has been previously approved, I'll leave it in for now unless there's an objection.

src/Simulation/qdk_sim_rs/rust-toolchain
cgranade: Strictly speaking, we don't need to on any fixed schedule — old nightly builds persist and are valid, but will not include bugfixes from future nightlies. From that perspective, it's good practice to update as part of the regular maintenance of the project. Since we use a minimal number of unstable features (specifically, only backtrace), the cost to updating is very minimal — everything not gated on a feature attribute is still stable and subject to API stability rules.

PR #1070, opened 2022-08-03
Added unit documentation to Rx Rz and Ry operations

HalOwens:
Currently the documentation for these gates does not mention that theta is expressed in radians. Added this to each operation's summary.


tcNickolas:
There are several more gates that also take angle parameter (R and R1) - it makes sense to update those as well for uniformity.

Also, I'm not sure if it's best to add the clarification to the summary or to the input description. We try to keep our summaries short, so input description might be better. @swernli Any suggestions?

swernli:
Thanks for bringing this up and working on the doc update, @HalOwens! I agree with the recommendation of moving these to the input description, so they would look like:
```qsharp
    /// # Summary
    /// Applies a rotation about the $x$-axis by a given angle.
    ///
    /// # Description
    /// \begin{align}
    ///     R_x(\theta) \mathrel{:=}
    ///     e^{-i \theta \sigma_x / 2} = 
    ///     \begin{bmatrix}
    ///         \cos \frac{\theta}{2} & -i\sin \frac{\theta}{2}  \\\\
    ///         -i\sin \frac{\theta}{2} & \cos \frac{\theta}{2}
    ///     \end{bmatrix}.
    /// \end{align}
    ///
    /// # Input
    /// ## theta
    /// Angle in radians about which the qubit is to be rotated.
    /// ## qubit
    /// Qubit to which the gate should be applied.
    ///
    /// # Remarks
    /// Equivalent to:
    /// ```qsharp
    /// R(PauliX, theta, qubit);
    /// ```
```

In addition, the following callables should be updated as well for consistency:
https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.intrinsic.r
https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.intrinsic.r1
https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.intrinsic.r1frac
https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.intrinsic.rfrac
https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.intrinsic.exp
https://docs.microsoft.com/en-us/qsharp/api/qsharp/microsoft.quantum.intrinsic.expfrac

Thanks!

HalOwens:
Thanks, I went ahead and added that information to the functions in Intrinsic/ and Decompositions/

src/Simulation/TargetDefinitions/Decompositions/Rx.qs
```diff
@@ -4,7 +4,7 @@
 namespace Microsoft.Quantum.Intrinsic {
 
     /// # Summary
-    /// Applies a rotation about the $x$-axis by a given angle.
+    /// Applies a rotation about the $x$-axis by a given angle.
```
swernli, line 7: Looks like an errant space was leftover.
```suggestion
    /// Applies a rotation about the $x$-axis by a given angle.
```

swernli (approved):

PR #1073, opened 2022-08-12
QirFullstateExecutable to take Runner as Arg

ScottCarda-MS:
Adds a constructor the QirFullstateExecutable that takes a QirExecutableRunner as an argument.


SECURITY.md
```diff
@@ -0,0 +1,41 @@
...
+
+Microsoft takes the security of our software products and services seriously, which includes all source code repositories managed through our GitHub organizations, which include [Microsoft](https://github.com/Microsoft), [Azure](https://github.com/Azure), [DotNet](https://github.com/dotnet), [AspNet](https://github.com/aspnet), [Xamarin](https://github.com/xamarin), and [our GitHub organizations](https://opensource.microsoft.com/).
+
+If you believe you have found a security vulnerability in any Microsoft-owned repository that meets [Microsoft's definition of a security vulnerability](https://aka.ms/opensource/security/definition), please report it to us as described below.
+
+## Reporting Security Issues
+
+**Please do not report security vulnerabilities through public GitHub issues.**
+
+Instead, please report them to the Microsoft Security Response Center (MSRC) at [https://msrc.microsoft.com/create-report](https://aka.ms/opensource/security/create-report).
```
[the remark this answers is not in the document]
kuzminrobin: Applicable to more than one case.

src/Qir/Runtime/lib/QIR/QirOutputHandling.cpp
```diff
@@ -15,9 +16,61 @@ static void PrintCStr(const char* cStr)
...
 extern "C"
 {
+    // Message Records
+
+    void __quantum__rt__message_record_output(QirString* qstr) // NOLINT
+    {
+        std::stringstream strStream;
+        strStream << QOH_INFO_PREFIX << EscapeString(qstr->str);
+        PrintCStr(strStream.str().c_str());
+    }
```
kuzminrobin, line 71: What is the functionality of `EscapeString(qstr->str)` needed for? 
Why not just `strStream << QOH_INFO_PREFIX << qstr->str;`?

src/Qir/Runtime/lib/QIR/QirOutputHandling.cpp
```diff
@@ -15,9 +16,61 @@ static void PrintCStr(const char* cStr)
     WriteToCurrentStream(&msg);
 }
 
+static std::string EscapeString(std::string const& s)
+{
+    std::string out = "";
+    for (std::string::const_iterator i = s.begin(), end = s.end(); i != end; i++)
+    {
+        char c = *i;
```
kuzminrobin, line 24: Consider using the [range-based `for` loop](https://github.com/AnthonyCalandra/modern-cpp-features#range-based-for-loops), it is considered less error-prone than the classical `for` loop.
```suggestion
    for (auto c: s)
    {
```

src/Qir/Runtime/lib/QIR/QirOutputHandling.cpp
kuzminrobin: I assume the reason for `EscapeString(qstr->str)` is the fact that in the fragment 
`strStream << QOH_INFO_PREFIX << EscapeString(qstr->str);`
after the part 
`strStream << QOH_INFO_PREFIX << ` 
there can be output a sequence of chars that contains the `\n`, and after that char the output will look like a next line of output (requiring its own output prefix, like `QOH_INFO_PREFIX`). 

To disambiguate that the `EscapeString(qstr->str)` is used. And later some intermediate service will revert the change of the `EscapeString(qstr->str)` and convert the output back to the one the user expects.

src/Qir/Runtime/lib/QIR/QirOutputHandling.cpp
```diff
@@ -15,9 +16,61 @@ static void PrintCStr(const char* cStr)
...
+            case '\n':
+                out += 'n';
+                break;
+            default:
+                char const* const hexdig = "0123456789ABCDEF";
+                out += 'x';
+                unsigned char uc = (unsigned char)c;
+                out += hexdig[uc >> 4];
+                out += hexdig[uc & 0xF];
+            }
```
kuzminrobin, line 55: (Code Reliability)
Consider placing `break;` in the end of the `default:` block:
```suggestion
            default:
                char const* const hexdig = "0123456789ABCDEF";
                out += 'x';
                unsigned char uc = (unsigned char)c;
                out += hexdig[uc >> 4];
                out += hexdig[uc & 0xF];
                break;
            }
```
I believe the `default:` block can be placed in any part inside of the `case(..) { . . . }`, including in the beginning, i.e. between the lines 33 and 34. If someone decides to refactor the code and move the `default:` block to that place (and the block does not end with `break`) then the execution of the `default:` block will _fall through_ to line 35. Such a behavior will likely be unexpected and can result in a bug.

kuzminrobin (approved):

kuzminrobin:
As a general note,
looks like this PR contains changes not related to the header and description of this PR. Consider breaking up into multiple smaller PRs.

src/Qir/Runtime/lib/QIR/QirOutputHandling.cpp
ScottCarda-MS: done

src/Qir/Runtime/lib/QIR/QirOutputHandling.cpp
ScottCarda-MS: done

SECURITY.md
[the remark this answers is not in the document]
ScottCarda-MS: This change is the result of merging in main to the local branch before merging in main to the feature branch. As such, these are not my changes.

ScottCarda-MS:
Out of date

PR #1077, opened 2022-08-24
Remove unused Rust sys crate, minor clean up

swernli:
This removes the Rust sys crate that wrapped the C++ runtime since it was no longer used. In addition, it moves some rustc flags configuration out of build scripts and into a repo-wide settings file, resolves a cargo warning by moving profile settings up to the root of the workspace, and cleans up a few leftover functions listed as dllexport in the runtime header files despite having no implementation.


kuzminrobin (approved):
I'm not familiar with this code base, but at a glance the description and the changes seem reasonable to me.
Approving to unblock, But I would recommend getting Cassandra's approval.

.cargo/config.toml
```diff
@@ -0,0 +1,5 @@
+[target.'cfg(target_os = "windows")']
+rustflags = ["-Clink-args=/guard:cf /DYNAMICBASE /CETCOMPAT"]
+
+[target.x86_64-pc-windows-msvc]
+rustflags = ["-Ctarget-feature=+crt-static"]
```
idavis, line 5: @swernli why are static link crt on windows?

src/Qir/Runtime/public/QirRuntime.hpp
```diff
@@ -61,12 +61,6 @@ extern "C"
     // Utils
     // ------------------------------------------------------------------------
 
-    // Allocate a block of memory on the heap.
-    QIR_SHARED_API char* __quantum__rt__heap_alloc(uint64_t size); // NOLINT
```
idavis, line 65: This change doesn't seem to be in line with the rest of the PR.

.cargo/config.toml
cgranade, line 5: Ah, this is a neat way to configure these flags! Do we need similar for other targets as well?

cgranade (approved):
LGTM! Just a minor question about non-Windows targets.

.cargo/config.toml
```diff
@@ -0,0 +1,5 @@
+[target.'cfg(target_os = "windows")']
+rustflags = ["-Clink-args=/guard:cf /DYNAMICBASE /CETCOMPAT"]
```
idavis, line 2: Isn't `/guard:cf /DYNAMICBASE /CETCOMPAT` msvc only?

.cargo/config.toml
swernli: Is the question why do it on Windows at all or why do it only on Windows (as in, we should do it for all platforms)? I think not depending on the client machine having the C runtime installed is good, but I only thought about and saw references to doing this on Windows. Do you thing we should do it more broadly or not do it all?

swernli: Good question, and relates to Ian's question above. I had only seen examples of doing this for Windows and hadn't considered we could just do it across all platforms.

.cargo/config.toml
swernli: those flags are respected by the linker that Rust invokes, such that when passed on Windows they have the desired effect. Rust either calls `lld-link.exe` or `link.exe` which both support these flags.

src/Qir/Runtime/public/QirRuntime.hpp
swernli: Yeah, it's not, but it's an annoying little thing that I missed more than a year ago and keep forgetting to fix. I felt it fit under the theme of "clean-up" but agree that it is not Rust related where every other change is. Should I separate it out to another PR?

.cargo/config.toml
cgranade: IIRC, `+crt-static` is a no-op on platforms that don't support static linking, such that it _should_ be safe to include it on all targets.

swernli: After some discussion, we landed at simplifying these flags here.

PR #1079, opened 2022-08-26
Q#-Runtime - Fullstate Simulator V1 Output Format

ScottCarda-MS:
1. Corrects the naming of `__quantum__rt__integer_record_output` to `__quantum__rt__int_record_output`.
2. Adds `__quantum__rt__message_record_output` along with associated string-escaping logic.
3. Changes the EntryPointOperationLoader to use the LLVM bitcode instead of the AST, because the bitcode has qir-specific changes to it that the AST does not.
4. Adds a QirFullStateExecutable constructor that accepts a QirExecutableRunner as an argument. This allows the service to create its own runner logic.


kuzminrobin (approved):

swernli (approved):

src/Qir/Tools/EntryPointOperationLoader.cs
```diff
@@ -3,92 +3,74 @@
...
-            if (!AssemblyLoader.LoadReferencedAssembly(assemblyFileInfo.FullName, out var compilation))
-            {
-                throw new ArgumentException("Unable to read the Q# syntax tree from the given DLL.");
-            }
-
-            return GenerateEntryPointOperations(compilation);
+            return module.Functions
+                .Where(f =>
+                    f.Attributes.ContainsKey(LlvmBindings.Values.FunctionAttributeIndex.Function)
+                    && f.Attributes[LlvmBindings.Values.FunctionAttributeIndex.Function].Any(a => a.Name == "InteropFriendly"))
```
swernli, line 30: Actually, I think this string needs to be `"EntryPoint"`

fedonman (approved):

ScottCarda-MS:
We are no longer pursuing this.

PR #1087, opened 2022-09-14
Replace C++ QIR Runtime with Rust QIR stdlib

swernli:
This change removes the C++ implementation for the QIR Runtime and replaces it with a Rust implementation of a QIR stdlib that matches the classical functionality exposed.

Previously, the C++ QIR Runtime was compiled into three dynamic libraries, `Microsoft.Quantum.Qir.Runtime`, `Microsoft.Quantum.Qir.QSharp.Foundation`, and `Microsoft.Quantum.Qir.QSharp.Core`, and the "core" library would load the simulator dll at runtime. With this change, these dynamic libraries are no longer shipped, and we instead build a single static library, `qir_stdlib`, that is linked into the simulator dynamic library, `Microsoft.Quantum.Simulator.Runtime`. This makes the full state simulator a "QIR Backend" which means it exposes the classical and quantum functions necessary to link in and execute a QIR program.

To help orient reviews of this rather large change, here are some key landmarks in the code:

- **src/Qir/Runtime** The existing C++ code is removed in favor of a single `stdlib` folder that implements the classical functions described in the QIR spec as a static library written in Rust. This includes unit tests for those functions.
- **src/Qir/Tools** The generation logic for a main.cpp to run a given QIR program is updated to remove the support for the C++ specific runtime elements (namely QIR Execution Context and IRuntimeDriver references) in favor of just an invocation of the entry point function. The QIR libraries are removed in favor of linking directly against the full state simulator as a QIR backend.
- **src/Qir/Tests** The C++ scenario tests that compile against QIR generated from Q# are adapted to the new library implementation, requiring the removal of any tests dependent on the internal C++ representations. Build infrastructure is updated to link against the full state simulator.
- **src/Simulation/Native** The full state simulator is adapted to statically link the `qir_stdlib` as a dependency and re-export the QIR classical library support. New QIR-compliant C-API is added to allow QIR programs to link directly against the quantum operations supported by the simulator without the need of an additional shim library.
- **src/Simulation/qdk_sim_rs** Cleanup to some Rust patterns to make clippy happy with the repo (not specific to the C++ -> Rust QIR transition, just minor general linter fixes).

NOTE: e2e builds will fail until a corresponding update is made to the e2e build script logic in order to build the QIR stdlib before the simulator to satisfy dependencies.


build/steps-init.yml
```diff
@@ -28,8 +28,8 @@ steps:
 
 - script: |
     rustup install nightly
-    rustup component add rustfmt clippy
-    rustup component add rustfmt clippy --toolchain nightly
+    rustup component add rustfmt clippy llvm-tools-preview
```
swernli, line 31: Add comment as to when llvm-tools-preview dependency can be removed.

build/steps-codecheck.yml
```diff
@@ -19,7 +19,7 @@ steps:
   displayName: "Install QIR Runtime Prerequisites"
   workingDirectory: $(System.DefaultWorkingDirectory)
 
-- pwsh: src/Qir/Runtime/build-qir-runtime.ps1
+- pwsh: src/Qir/Runtime/build-qir-stdlib.ps1
   displayName: "Build QIR Runtime"
```
swernli, line 23: Update display name to say "QIR stdlib"

build/test.ps1
```diff
@@ -43,8 +43,8 @@ function Test-One {
 Test-One '../Simulation.sln'
 
 if ($Env:ENABLE_QIRRUNTIME -ne "false") {
-    $qirRuntime = (Join-Path $PSScriptRoot "../src/Qir/Runtime")
-    & "$qirRuntime/test-qir-runtime.ps1"
+    $qirstdlib = (Join-Path $PSScriptRoot "../src/Qir/Runtime")
```
swernli, line 46: Use multiple arguments to `Join-Path` instead of explicit `/` in paths.

src/Simulation/Native/src/CMakeLists.txt
```diff
@@ -34,9 +34,40 @@ else(BUILD_SHARED_LIBS)
...
+    "-L${PROJECT_BINARY_DIR}/../../../Qir/drops/bin/$ENV{BUILD_PLATFORM}/native"
+    qir_stdlib
+    Ws2_32
+    Bcrypt
+    Userenv
+  )
+  # On Windows, use a .def file to force export stdlib functions from the static library.
+  target_sources(Microsoft.Quantum.Simulator.Runtime PUBLIC "${PROJECT_BINARY_DIR}/../../../Qir/drops/include/qir_stdlib.def")
+elseif (APPLE)
+  # On MacOS, use -force_load and explicit path to pull in entire contents of stdlib static archive.
```
swernli, line 51: Inlcude links for Windows, Mac, and Linux static linking strategies (flags definitions where available)

src/Simulation/Native/src/CMakeLists.txt
```diff
@@ -34,9 +34,40 @@ else(BUILD_SHARED_LIBS)
...
+  target_link_libraries(Microsoft.Quantum.Simulator.Runtime
+    "-Wl,-force_load,${PROJECT_BINARY_DIR}/../../../Qir/drops/bin/$ENV{BUILD_PLATFORM}/native/libqir_stdlib.a"
+    pthread
+    dl
+  )
+else()
+  # On Unix systems, use --whole-archive to to pull in entire contents of stdlib  static archive.
```
swernli, line 58: clarifiy that whole archive must be turned on and then off to avoid future gotcha.

src/Simulation/Native/build-native-simulator.ps1
```diff
@@ -3,6 +3,8 @@
 
 Write-Host "##[info]Build Native simulator for $Env:BUILD_CONFIGURATION"
 
+& (Join-Path $PSScriptRoot .. .. .. build set-env.ps1)
```
swernli, line 6: Move above line 4.

src/Qir/Runtime/stdlib/include/qir_stdlib.h
```diff
@@ -0,0 +1,230 @@
...
+{
+#endif
+
+    struct QirString;
+    struct QirArray;
+    struct QirCallable;
+    struct QirTuple;
+    struct QirBigInt;
+
+    enum PauliId : int8_t
```
swernli, line 23: Possibly redundant enum definition?

src/Qir/Runtime/stdlib/src/lib.rs
```diff
@@ -0,0 +1,126 @@
...
+unsafe fn update_counts<T>(raw_rc: *const T, update: i32, is_alias: bool) {
+    let mut remaining = update;
+    while remaining != 0 {
+        let rc = Rc::from_raw(raw_rc);
+        if remaining > 0 {
+            // Create and leak new instances to increment the count on the contained item.
+            if is_alias {
+                let _ = Weak::into_raw(Rc::downgrade(&rc));
+            } else {
+                let _ = Rc::into_raw(Rc::clone(&rc));
```
swernli, line 31: Are there reasons to considuer using `Arc` for atomics support; does it have weak references?

src/Qir/Runtime/stdlib/src/bigints.rs
```diff
@@ -0,0 +1,335 @@
...
+            let bigint_7 = __quantum__rt__bigint_shiftright(bigint_6, 3);
+            assert_eq!(*bigint_7, (21).try_into().unwrap());
+            __quantum__rt__bigint_update_reference_count(bigint_0, -1);
+            __quantum__rt__bigint_update_reference_count(bigint_1, -1);
+            __quantum__rt__bigint_update_reference_count(bigint_2, -1);
+            __quantum__rt__bigint_update_reference_count(bigint_3, -1);
+            __quantum__rt__bigint_update_reference_count(bigint_4, -1);
+            __quantum__rt__bigint_update_reference_count(bigint_5, -1);
+            __quantum__rt__bigint_update_reference_count(bigint_6, -1);
+            __quantum__rt__bigint_update_reference_count(bigint_7, -1);
```
swernli, line 314: Consider releaseing in reverse allocation order.

build/build.ps1
```diff
@@ -7,8 +7,8 @@ $ErrorActionPreference = 'Stop'
 $all_ok = $True
 
 if ($Env:ENABLE_QIRRUNTIME -ne "false") {
-    $qirRuntime = (Join-Path $PSScriptRoot "../src/Qir/Runtime")
-    & "$qirRuntime/build-qir-runtime.ps1"
+    $qirstdlib = (Join-Path $PSScriptRoot "../src/Qir/Runtime")
+    & "$qirstdlib/build-qir-stdlib.ps1"
```
kuzminrobin, line 11: (Feel free to ignore)
For the variables that contain the paths I prefer the name of the var to end with "Path" (and sometimes it really helps if the name ends with either "FilePath" or "DirPath").

build/steps-init.yml
```diff
@@ -28,8 +28,8 @@ steps:
...
+    rustup component add rustfmt clippy llvm-tools-preview --toolchain nightly
```
kuzminrobin, line 32: (Just curiosity, feel free to ignore)
What will be lost if the green line 31 (L31) is removed?

build/test.ps1
```diff
@@ -43,8 +43,8 @@ function Test-One {
...
+    & "$qirstdlib/test-qir-stdlib.ps1"
```
kuzminrobin, line 47: (Nit)
```suggestion
    $qirStdlib = (Join-Path $PSScriptRoot "../src/Qir/Runtime")
    & "$qirStdlib/test-qir-stdlib.ps1"
```

build/steps-init.yml
swernli: This is a quirk of rustup and cargo component installations. Rustup (the tool that handles installing Rust and related components) can install multiple "toolchains" (see https://rust-lang.github.io/rustup/concepts/toolchains.html), and components can be installed in each of those toolchains. The first part on L31 installs the components for the default toolchain, and the second line installs those same coponents for the "nightly" toolchain which includes additional Rust features that have not yet been fully stabilized in an official release.

src/Qir/Runtime/stdlib/src/callables.rs
```diff
@@ -0,0 +1,207 @@
...
+    let index = (if *call.is_adj.borrow() { 1 } else { 0 })
+        + (if *call.is_ctl.borrow() > 0 { 2 } else { 0 });
+
+    // Collect any nested controls into a single control list.
+    let mut args_copy: *mut *const Vec<u8> = std::ptr::null_mut();
+    if !args_tup.is_null() {
+        // Copy the tuple so we can potentially edit it.
+        args_copy = __quantum__rt__tuple_copy(args_tup.cast::<*const Vec<u8>>(), true);
+
+        if *call.is_ctl.borrow() > 0 {
```
cgranade, line 58: Ah, does `is_ctl` represent the number of controls? I may suggest a more descriptive name like `n_control_qubits`, then?

src/Qir/Runtime/stdlib/src/callables.rs
```diff
@@ -0,0 +1,207 @@
...
+            // to balance the decrement that will happen in the loop and at the end of invoking the callable
+            // to ensure the original, non-owned list does not get incorrectly cleaned up.
+            __quantum__rt__array_update_reference_count(*args_copy.cast::<*const QirArray>(), 1);
+
+            let mut ctl_count = *call.is_ctl.borrow();
+            while ctl_count > 1 {
+                let ctls = *args_copy.cast::<*const QirArray>();
+                let inner_tuple = *args_copy
+                    .cast::<*const QirArray>()
+                    .wrapping_add(1)
```
cgranade, line 69: Why is this a wrapping add?

src/Qir/Runtime/stdlib/src/callables.rs
```diff
@@ -0,0 +1,207 @@
...
+                    .cast::<*mut *const Vec<u8>>();
+                let inner_ctls = *inner_tuple.cast::<*const QirArray>();
```
cgranade, line 71: Can some of this code be abstracted into `unsafe` methods on `QirArray`?

src/Qir/Runtime/stdlib/src/output_recording.rs
```diff
@@ -0,0 +1,63 @@
...
+#[no_mangle]
+pub extern "C" fn __quantum__rt__bool_record_output(val: bool) {
+    println!("RESULT\t{}", val);
+}
+
+/// # Safety
+///
+/// This function should only be called with strings created by `__quantum__rt__string_*` functions.
+#[no_mangle]
+pub unsafe extern "C" fn __quantum__rt__message_record_output(str: *const CString) {
```
swernli, line 55: This adds the message output recording consistent with the in-progress updates from https://github.com/microsoft/qsharp-runtime/pull/1079

src/Qir/Runtime/stdlib/src/lib.rs
```diff
@@ -0,0 +1,126 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+#![deny(clippy::all, clippy::pedantic)]
```
cgranade, line 3: Are the other deny attributes needed given this at the crate level?

src/Qir/Runtime/stdlib/src/lib.rs
```diff
@@ -0,0 +1,126 @@
...
+pub unsafe extern "C" fn __quantum__rt__fail(str: *const CString) {
+    __quantum__rt__message(str);
+    panic!("{}", (*str).to_str().expect("Unable to convert string"));
+}
+
+/// # Safety
+///
+/// This function should only be called with a string created by `__quantum__rt__string_*` functions.
+#[no_mangle]
+pub unsafe extern "C" fn __quantum__rt__message(str: *const CString) {
```
cgranade, line 89: Is there any way for consumers to change the stream that gets used for these messages?

src/Qir/Runtime/stdlib/src/math.rs
```diff
@@ -0,0 +1,119 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+#![deny(clippy::all, clippy::pedantic)]
+
+// TODO: transition math functions to `__quantum__rt` once compiler support is ready.
```
cgranade, line 5: Is there an issue for this TODO to help indicate when the comment can be successfully resolved?

src/Qir/Runtime/stdlib/src/strings.rs
```diff
@@ -0,0 +1,407 @@
...
+use num_bigint::BigInt;
+use std::{
+    ffi::{CStr, CString},
+    os::raw::{c_char, c_double},
+    rc::Rc,
+};
+
+/// # Safety
+///
+/// This function should only be called with a valid, null-terminated, C-style string.
```
cgranade, line 16: Feels very funny somehow that, even in quantum, we're still carrying around C-style strings. Ah, well.

src/Qir/Runtime/test-qir-stdlib.ps1
```diff
@@ -0,0 +1,26 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
cgranade, line 1:
```suggestion
# Copyright (c) Microsoft Corporation.
```

src/Qir/Tests/QIR-dynamic/CMakeLists.txt
```diff
@@ -1,29 +1,22 @@
...
   target_source_from_qir(qir-dynamic-tests ${file})
 endforeach()
 
 target_link_libraries(qir-dynamic-tests PUBLIC
-  "-L${runtime_lib_path}"
-  -lMicrosoft.Quantum.Qir.Runtime
-  -lMicrosoft.Quantum.Qir.QSharp.Foundation
-  -lMicrosoft.Quantum.Qir.QSharp.Core
+  "-L${simulator_lib_path}"
+  -lMicrosoft.Quantum.Simulator.Runtime
```
cgranade, line 15: Goodness, how many times does CMake require duplicating this declaration?

src/Qir/Tools/Executable/QirFullStateExecutable.cs
```diff
@@ -21,17 +21,15 @@ public class QirFullStateExecutable : QirExecutable
         public override string DriverFileExtension => "cpp";
 
         public override IList<string> LinkLibraries => new List<string> {
-                "Microsoft.Quantum.Qir.Runtime",
-                "Microsoft.Quantum.Qir.QSharp.Foundation",
-                "Microsoft.Quantum.Qir.QSharp.Core"
+                "Microsoft.Quantum.Simulator.Runtime"
```
cgranade, line 24: Nit, but this should likely use next-line brace style.

src/Simulation/Native/build-native-simulator.ps1
```diff
@@ -1,6 +1,8 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
```
cgranade, line 1:
```suggestion
# Copyright (c) Microsoft Corporation.
```

src/Simulation/Native/src/simulator/qir.cpp
```diff
@@ -0,0 +1,504 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+/*=============================================================================
+    QIR assumes a single global execution context.
+    To support the dispatch over the qir-bridge, the clients must register their
+    Microsoft::Quantum::IRuntimeDriver* first.
```
cgranade, line 7: Is this still accurate?

src/Simulation/Native/src/simulator/qir.hpp
```diff
@@ -0,0 +1,117 @@
...
+// is a unique type in the QIR.
+
+class QUBIT;
+typedef intptr_t QubitIdType;
+
+class RESULT;
+
+struct QirRTuple
+{
+    PauliId pauli;
```
cgranade, line 36: Is the alignment / packing of this struct deterministic across platforms? More to the point, does the Rust side assume any specific layout for these structs?

src/Simulation/qdk_sim_rs/prerequisites.ps1
```diff
@@ -22,5 +22,5 @@ if (-not (Get-Command rustup -ErrorAction SilentlyContinue)) {
 # rustfmt and clippy is available.
 rustup install nightly
 rustup toolchain install nightly
-rustup component add rustfmt clippy
-rustup component add rustfmt clippy --toolchain nightly
+rustup component add rustfmt clippy llvm-tools-preview
```
cgranade, line 25: Are these needed by `qdk_sim_rs` as well now?

src/Qir/Runtime/stdlib/include/qir_stdlib.h
```diff
@@ -0,0 +1,230 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+#include <cstdint>
+
+#ifdef _WIN32
+#define QIR_SHARED_API __declspec(dllexport)
```
kuzminrobin, line 7: I have an impression that one more condition is missing, totally something like
```cpp
#ifdef _WIN32
#ifdef EXPORT_QIR_API
#define QIR_SHARED_API __declspec(dllexport)
#else
#define QIR_SHARED_API __declspec(dllimport)
#endif
#else
#define QIR_SHARED_API
#endif
```

src/Qir/Runtime/stdlib/include/qir_stdlib.h
```diff
@@ -0,0 +1,230 @@
...
+    enum PauliId : int8_t
+    {
+        PauliId_I = 0,
+        PauliId_X = 1,
+        PauliId_Z = 2,
+        PauliId_Y = 3,
+    };
+
+    typedef void (*t_CallableEntry)(QirTuple*, QirTuple*, QirTuple*);
+    typedef void (*t_CaptureCallback)(QirTuple*, int32_t);
```
kuzminrobin, line 32: FYI: If a C file includes this header, and it is compiled with C (not C++) compiler, then I believe the compiler will complain  until we make the following change:
```suggestion
    typedef void (*t_CallableEntry)(struct QirTuple*, struct QirTuple*, struct QirTuple*);
    typedef void (*t_CaptureCallback)(struct QirTuple*, int32_t);
```
Applicable below for all the uses of identifiers
```cpp
    struct QirString;
    struct QirArray;
    struct QirCallable;
    struct QirTuple;
    struct QirBigInt;
```

src/Qir/Runtime/stdlib/include/qir_stdlib.h
```diff
@@ -0,0 +1,230 @@
...
+
+    // Returns a pointer to the malloc-allocated block.
+    QIR_SHARED_API char* __quantum__rt__memory_allocate(uint64_t size); // NOLINT
```
kuzminrobin, line 35: Nit, FYI:
I'm not sure why this has not been done earlier but it makes more sense for this function to return `void*` rather than `char*`.
```suggestion
QIR_SHARED_API void* __quantum__rt__memory_allocate(uint64_t size); // NOLINT
```

build/build.ps1
```diff
@@ -7,8 +7,8 @@ $ErrorActionPreference = 'Stop'
 $all_ok = $True
 
 if ($Env:ENABLE_QIRRUNTIME -ne "false") {
-    $qirRuntime = (Join-Path $PSScriptRoot "../src/Qir/Runtime")
-    & "$qirRuntime/build-qir-runtime.ps1"
+    $qirstdlib = (Join-Path $PSScriptRoot "../src/Qir/Runtime")
```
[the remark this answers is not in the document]
swernli: Updated!

build/build.ps1
swernli: Updated to include Path in the name!

src/Qir/Runtime/stdlib/include/qir_stdlib.h
kuzminrobin: Or we can try to introduce the type aliases for the struct names:
```c
    struct QirString;    // Original structs are retained.
    struct QirArray;
    struct QirCallable;
    struct QirTuple;
    struct QirBigInt;
    
    typedef struct QirString QirString;  // Type aliases with the same name are added.
    typedef struct QirArray QirArray;
    typedef struct QirCallable QirCallable;
    typedef struct QirTuple QirTuple;
    typedef struct QirBigInt QirBigInt;
```

build/test.ps1
swernli: Updated pattern to sue better variable name and `Join-Path` arguments.

src/Qir/Runtime/stdlib/include/qir_stdlib.h
```diff
@@ -0,0 +1,230 @@
...
+
+    // Returns a string representation of the big integer.
+    QIR_SHARED_API QirString* __quantum__rt__bigint_to_string(QirBigInt*); // NOLINT
+
+    // Creates a big integer with the specified initial value.
+    QIR_SHARED_API QirBigInt* __quantum__rt__bigint_create_i64(int64_t); // NOLINT
+
+    // Creates a big integer with the initial value specified by the i8 array. The 0-th element of the array is the
+    // highest-order byte, followed by the first element, etc.
+    QIR_SHARED_API QirBigInt* __quantum__rt__bigint_create_array(int, char*); // NOLINT
```
kuzminrobin, line 152: FYI: The `int` param is not documented in the QIR Spec. I filed an [issue](https://github.com/qir-alliance/qir-spec/issues/26) against the QIR Spec.

src/Qir/Runtime/stdlib/include/qir_stdlib.h
```diff
@@ -0,0 +1,230 @@
...
+    QIR_SHARED_API QirBigInt* __quantum__rt__bigint_subtract(QirBigInt*, QirBigInt*); // NOLINT
+
+    // Multiplies two big integers and returns their product.
+    QIR_SHARED_API QirBigInt* __quantum__rt__bigint_multiply(QirBigInt*, QirBigInt*); // NOLINT
+
+    // Divides the first big integer by the second and returns their quotient.
+    QIR_SHARED_API QirBigInt* __quantum__rt__bigint_divide(QirBigInt*, QirBigInt*); // NOLINT
+
+    // Returns the first big integer modulo the second.
+    QIR_SHARED_API QirBigInt* __quantum__rt__bigint_modulus(QirBigInt*, QirBigInt*); // NOLINT
```
kuzminrobin, line 174: FYI: I filed a QIR Spec issue [__quantum__rt__bigint_divide: The division by zero is not documented](https://github.com/qir-alliance/qir-spec/issues/27)

src/Qir/Runtime/stdlib/include/qir_stdlib.h
```diff
@@ -0,0 +1,230 @@
...
+    QIR_SHARED_API QirBigInt* __quantum__rt__bigint_bitand(QirBigInt*, QirBigInt*); // NOLINT
+
+    // Returns the bitwise-OR of two big integers.
+    QIR_SHARED_API QirBigInt* __quantum__rt__bigint_bitor(QirBigInt*, QirBigInt*); // NOLINT
+
+    // Returns the bitwise-XOR of two big integers.
+    QIR_SHARED_API QirBigInt* __quantum__rt__bigint_bitxor(QirBigInt*, QirBigInt*); // NOLINT
+
+    // Returns the bitwise complement of the big integer.
+    QIR_SHARED_API QirBigInt* __quantum__rt__bigint_bitnot(QirBigInt*); // NOLINT
```
kuzminrobin, line 189: FYI: https://github.com/qir-alliance/qir-spec/issues/28

src/Qir/Runtime/stdlib/include/qir_stdlib.h
```diff
@@ -0,0 +1,230 @@
...
+
+    // Returns the big integer arithmetically shifted left by the integer amount of bits.
+    QIR_SHARED_API QirBigInt* __quantum__rt__bigint_shiftleft(QirBigInt*, int64_t); // NOLINT
+
+    // Returns the big integer arithmetically shifted right by the integer amount of bits.
+    QIR_SHARED_API QirBigInt* __quantum__rt__bigint_shiftright(QirBigInt*, int64_t); // NOLINT
```
kuzminrobin, line 195: 1. FYI: https://github.com/qir-alliance/qir-spec/issues/29
2. Consider:
```suggestion
    // Returns the big integer arithmetically shifted left by the (positive) integer amount of bits.
    QIR_SHARED_API QirBigInt* __quantum__rt__bigint_shiftleft(QirBigInt*, int64_t); // NOLINT

    // Returns the big integer arithmetically shifted right by the (positive) integer amount of bits.
    QIR_SHARED_API QirBigInt* __quantum__rt__bigint_shiftright(QirBigInt*, int64_t); // NOLINT
```

src/Qir/Runtime/stdlib/include/qir_stdlib.h
```diff
@@ -0,0 +1,230 @@
...
+    QIR_SHARED_API QirBigInt* __quantum__rt__bigint_multiply(QirBigInt*, QirBigInt*); // NOLINT
+
+    // Divides the first big integer by the second and returns their quotient.
+    QIR_SHARED_API QirBigInt* __quantum__rt__bigint_divide(QirBigInt*, QirBigInt*); // NOLINT
+
+    // Returns the first big integer modulo the second.
+    QIR_SHARED_API QirBigInt* __quantum__rt__bigint_modulus(QirBigInt*, QirBigInt*); // NOLINT
+
+    // Returns the big integer raised to the integer power.
+    QIR_SHARED_API QirBigInt* __quantum__rt__bigint_power(QirBigInt*, int); // NOLINT
```
kuzminrobin, line 177:
```suggestion
    QIR_SHARED_API QirBigInt* __quantum__rt__bigint_power(QirBigInt*, int32_t); // NOLINT
```

src/Qir/Runtime/stdlib/src/arrays.rs
```diff
@@ -0,0 +1,335 @@
...
+    pub(crate) elem_size: usize,
+    pub(crate) data: Vec<u8>,
+}
+
+/// # Panics
+///
+/// This function panics if the passed in sizes do not fit into the usize type for the
+/// current platform.
+#[no_mangle]
+pub extern "C" fn __quantum__rt__array_create_1d(elem_size: u32, size: u64) -> *const QirArray {
```
kuzminrobin, line 19: FYI: About the second param `size: u64`: I personally prefer to call the number of elements in an array (or number of chars in a string) with the word "length" or "itemCount" or "count", and I prefer to use the word "size" to designate the _size in bytes_.
Some more details are in [Distinguish Between Size and Length](https://github.com/kuzminrobin/code_review_notes/blob/master/cpp_design_bookmarks.md#distinguish-between-size-and-length).

Consider renaming the second param:
```suggestion
pub extern "C" fn __quantum__rt__array_create_1d(elem_size: u32, itemCount: u64) -> *const QirArray {
```

build/set-env.ps1
```diff
@@ -39,10 +39,18 @@ If (-not (Test-Path -Path $Env:WHEEL_OUTDIR)) { [IO.Directory]::CreateDirectory(
 If ($Env:DOCS_OUTDIR -eq $null) { $Env:DOCS_OUTDIR =  (Join-Path $Env:DROPS_DIR "docs") }
 If (-not (Test-Path -Path $Env:DOCS_OUTDIR)) { [IO.Directory]::CreateDirectory($Env:DOCS_OUTDIR) }
 
+$env:BUILD_PLATFORM = "win-x64"
```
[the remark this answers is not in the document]
swernli: Fair question... it would default to using the folder name `win-x64` and would likely break. But it would also break in a lot of other places, such as our CMakefile.txt build infrastructure that similary assumes one of those platforms. My understanding is that anything other than those three is not part of our support statement at the moment, so lots of work would have to be done throughout to make our repo build compatible outside of those three.

src/Qir/Runtime/build-qir-stdlib.ps1
```diff
@@ -0,0 +1,61 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+#Requires -PSEdition Core
```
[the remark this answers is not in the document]
swernli: Ah, here it is! Yes, I think we should have this in all of top level scripts, including set-env.ps1, to make sure the expectations are made clear as early as possible.

src/Qir/Runtime/stdlib/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "qir-stdlib"
+version = "0.1.0"
+authors = ["Microsoft"]
+edition = "2021"
+
```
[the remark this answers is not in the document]
swernli: I agree, though I would be ok defering that until we are ready to publish it and make the addition of description and readme part of the change that adds the publishing to the CI build.

src/Qir/Runtime/stdlib/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "qir-stdlib"
+version = "0.1.0"
```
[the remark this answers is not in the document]
swernli: For now, since it is not published anywhere, the version number is not actively used. Any consumers would need to use a git depenency. I do believe that once we publish it as a crate, this should take on it's own consistent versioning scheme, but hopefully at that point it is in a separate repo such that we don't have to tie it to QDK versions.

src/Qir/Runtime/stdlib/build.rs
```diff
@@ -0,0 +1,57 @@
...
+
+fn main() -> Result<(), String> {
+    // Compile the LLVM IR bridge file. Requires the llvm-tools-preview component.
+    let out_dir = env::var_os("OUT_DIR")
+        .map(PathBuf::from)
+        .ok_or_else(|| "Environment variable OUT_DIR not defined.".to_string())?;
+
+    let llvm_tools = LlvmTools::new().map_err(|err| {
+        format!(
+            "Failed to locate llvm tools: {:?}. Is the llvm-tools-preview component installed?",
```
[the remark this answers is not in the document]
swernli: Oh, good call. Always better to have actionable error messages!

src/Qir/Runtime/stdlib/include/qir_stdlib.def
```diff
@@ -0,0 +1,76 @@
+EXPORTS
```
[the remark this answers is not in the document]
swernli: Turns out it does! Added the copyright and description.

src/Qir/Runtime/stdlib/include/qir_stdlib.h
swernli: Good call on the type aliases, I hadn't anticipated that. Added them!

src/Qir/Runtime/stdlib/include/qir_stdlib.h
```diff
@@ -0,0 +1,230 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+#include <cstdint>
+
```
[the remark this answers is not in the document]
swernli: Ah, good call out. I'm currently online including it in exactly one place, so would not have noticed.

src/Qir/Runtime/stdlib/include/qir_stdlib.h
swernli: Usually yes, if this file were also used by C/C++ code that defined these APIs. But since this header is meant for consumers of the static library who wish to re-export the defined functions (like what the simulator does) they need to be forward delcaed as `dllexport`. A consumer who is not re-exporting the APIs would be expected to forward declare them, similar to how they are handled in Rust or QIR itself.

src/Qir/Runtime/stdlib/include/qir_stdlib.h
swernli: Looks like it's not redundant, as the enum values must be declared somewhere for C to use them.

src/Qir/Runtime/stdlib/include/qir_stdlib.h
swernli: That makes me realize I didn't correctly make this a 32-bit integer type in this signature. It should be `int32_t`.

src/Qir/Runtime/stdlib/src/arrays.rs
```diff
@@ -0,0 +1,335 @@
...
+/// # Safety
+///
+/// This function should only be called with an array created by `__quantum__rt__array_*` functions.
+#[no_mangle]
+pub unsafe extern "C" fn __quantum__rt__array_copy(
+    arr: *const QirArray,
+    force: bool,
+) -> *const QirArray {
+    let rc = Rc::from_raw(arr);
+    if force || Rc::weak_count(&rc) > 0 {
```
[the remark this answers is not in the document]
swernli: True, though only if the array is accessed across multiple threads. Given our conversations about not speculatively supporting multithreaded runtime execution, I think I'm comfortable maintaining the behavior that assumed exclusive (single-threaded) access that was already true with the previous implementation.

src/Qir/Runtime/stdlib/src/arrays.rs
```diff
@@ -0,0 +1,335 @@
...
+    pub(crate) data: Vec<u8>,
+}
+
+/// # Panics
+///
+/// This function panics if the passed in sizes do not fit into the usize type for the
+/// current platform.
+#[no_mangle]
+pub extern "C" fn __quantum__rt__array_create_1d(elem_size: u32, size: u64) -> *const QirArray {
+    let elem_size = elem_size.try_into().unwrap();
```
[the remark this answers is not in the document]
swernli: Ah, that's true. I'd been working through the `unwrap()` usages in the code and replacing them with `expect()` and must not have gotten to this file. Incidentally, the use of `expect()` removes the clippy requirement for a `# Panics` docs section, even though the panic is still possible, presumably because the convention for `expect()` is that it is genuinely never expected to happen. Here that doesn't seem the case, because if we compiled and used this runtime in an environment where `usize` was smaller than `u64` it is possible runtime provided values would trigger the panic.

src/Qir/Runtime/stdlib/src/arrays.rs
```diff
@@ -0,0 +1,335 @@
...
+
+#[derive(Debug, Clone)]
+pub struct QirArray {
+    pub(crate) elem_size: usize,
+    pub(crate) data: Vec<u8>,
+}
+
+/// # Panics
+///
+/// This function panics if the passed in sizes do not fit into the usize type for the
```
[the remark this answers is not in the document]
swernli: As written, yes... and I think I need a consistent approach to that. Perhaps we could consult on that more? I'll add more thoughts in your comment on `panic = "unwind"` at the Cargo.toml file. Maybe the all up holistic fix for the panic patterns here is something that can be deffered to a follow up PR?

src/Qir/Runtime/stdlib/src/arrays.rs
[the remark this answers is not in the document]
cgranade: Makes sense, just jumped out at me, but that makes a lot of sense.

build/set-env.ps1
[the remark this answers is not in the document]
cgranade: Makes sense, as I said I'm surely being too paranoid here. Just gotten wary of fallthrough control flow...

src/Qir/Runtime/stdlib/src/arrays.rs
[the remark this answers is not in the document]
swernli: Actually, I think we've landed at a good solution: using the .cargo/config.toml settings in the qir_stdlib folder to override the panic setting and explicitly use abort. This is in keeping with the advice in the Profiles section of the Cargo Book (see https://doc.rust-lang.org/cargo/reference/profiles.html):
>Additionally, profiles can be overridden from a config definition. Specifying a profile in a config file or environment variable will override the settings from Cargo.toml.

src/Qir/Runtime/stdlib/src/arrays.rs
```diff
@@ -0,0 +1,335 @@
...
+/// This function should only be called with an array created by `__quantum__rt__array_*` functions.
+#[no_mangle]
+pub unsafe extern "C" fn __quantum__rt__array_update_alias_count(
+    arr: *const QirArray,
+    update: i32,
+) {
+    update_counts(arr, update, true);
+}
+
+#[cfg(test)]
```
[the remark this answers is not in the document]
swernli: Agreed. I'll add a note to the lib.rs to that effect.

src/Qir/Runtime/stdlib/src/lib.rs
swernli: I only just recently learned that I don't need to put it in every file :). I'll clean that up for clarity.

src/Qir/Runtime/stdlib/src/bigints.rs
```diff
@@ -0,0 +1,335 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+#![deny(clippy::all, clippy::pedantic)]
```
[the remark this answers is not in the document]
swernli: Yup, you are right and it's not needed here.

src/Qir/Runtime/stdlib/src/bigints.rs
```diff
@@ -0,0 +1,335 @@
...
+use std::{mem::ManuallyDrop, rc::Rc};
+
+#[no_mangle]
+pub extern "C" fn __quantum__rt__bigint_create_i64(input: i64) -> *const BigInt {
+    Rc::into_raw(Rc::new(input.into()))
+}
+
+/// # Safety
+///
+/// This function expects the second argument to be a well-formed C-style array of signed big-endian bytes
```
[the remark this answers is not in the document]
swernli: We landed on a different tactic for documentation, and have a top level comment on the module pointing to the QIR spec.

src/Qir/Runtime/stdlib/src/arrays.rs
```diff
@@ -0,0 +1,335 @@
...
+        let _ = Rc::into_raw(Rc::clone(&rc));
+        let _ = Rc::into_raw(rc);
+        arr
+    }
+}
+
+/// # Safety
+///
+/// This function should only be called with arrays created by `__quantum__rt__array_*` functions.
+/// # Panics
```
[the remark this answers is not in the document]
swernli: We've opted for a different commenting strategy and this has been removed.

src/Qir/Runtime/stdlib/src/arrays.rs
```diff
@@ -0,0 +1,335 @@
...
+    arr: *const QirArray,
+    force: bool,
+) -> *const QirArray {
+    let rc = Rc::from_raw(arr);
+    if force || Rc::weak_count(&rc) > 0 {
+        let copy = rc.as_ref().clone();
+        let _ = Rc::into_raw(rc);
+        Rc::into_raw(Rc::new(copy))
+    } else {
+        let _ = Rc::into_raw(Rc::clone(&rc));
```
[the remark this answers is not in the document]
swernli: Ah, nice, I didn't realize that. While some part of me likes the explicit `let _ = ` to make clear I'm ignoring them on purpose, but the code also looks cleaner without it (and clippy is happy) so I'll remove them.

src/Qir/Runtime/stdlib/src/bigints.rs
```diff
@@ -0,0 +1,335 @@
...
+        std::slice::from_raw_parts(input, size as usize),
+    )))
+}
+
+/// # Safety
+///
+/// This function should only be called with a QIR bigint created by the `__quantum__rt__bigint_*` functions.
+#[no_mangle]
+pub unsafe extern "C" fn __quantum__rt__bigint_get_data(input: *const BigInt) -> *const u8 {
+    ManuallyDrop::new((*input).to_signed_bytes_be()).as_ptr()
```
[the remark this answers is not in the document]
swernli: That's the fun part, you don't! (Insert meme here...)
On a more serious note, the bigint `to_signed_bytes_be()` does not provide a view into the data but rather copies it into a new `Vec`. The QIR API spec doesn't describe a way to clean up the data returned by this function, which would imply that the assumption is that the pointer is a view into the bigint with library-managed lifetime. I don't know if there is a better way to support that with how Rust works here, or if we need updates to the QIR spec to capture the notion that this is a managed array. Perhaps it could return an opaque QirArray instead?
@bettinaheim for her thoughts on this one.

src/Qir/Runtime/stdlib/src/bigints.rs
```diff
@@ -0,0 +1,335 @@
...
+/// # Safety
+///
+/// This function expects the second argument to be a well-formed C-style array of signed big-endian bytes
+/// with the size of that array passed as the first argument.
+#[no_mangle]
+pub unsafe extern "C" fn __quantum__rt__bigint_create_array(
+    size: u32,
+    input: *const u8,
+) -> *const BigInt {
+    Rc::into_raw(Rc::new(BigInt::from_signed_bytes_be(
```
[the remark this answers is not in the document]
swernli: It appears no, from what I can see in the code.

src/Qir/Runtime/stdlib/src/bridge-rt.ll
```diff
@@ -1,43 +1,38 @@
...
-define dllexport %String* @__quantum__rt__range_to_string(%Range %.range) {
-  %.prange = alloca %Range
-  store %Range %.range, %Range* %.prange
-  %range = bitcast %Range* %.prange to %"struct.QirRange"*
-  %str = call %"struct.QirString"* @quantum__rt__range_to_string(%"struct.QirRange"* dereferenceable(24) %range)
-  %.str = bitcast %"struct.QirString"* %str to %String*
-  ret %String* %.str
-}
\ No newline at end of file
+; Copyright (c) Microsoft Corporation.
```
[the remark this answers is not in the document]
swernli: In this case it was the path; it moved from src/Qir/Runtime/lib to src/Qir/Runtime/stdlib/src.

src/Qir/Runtime/stdlib/src/callables.rs
swernli: `.wrapping_add` calculates the pointer offset in a safe way, delaying Udefined Behavior until that pointer is dereferenced, while `.add` could itself have UB right away. It's a small difference, but one I thought was worth it at the time and now I'm less convinced by. The docs also note that `.add` is often further optimized than `.wrapping_add` which is tempting. However, I'm leaning toward leaving these as wrapping for now.

src/Qir/Runtime/stdlib/src/callables.rs
swernli: I went with `ctls_count`, good call on clarifying the name!

src/Qir/Runtime/stdlib/src/callables.rs
swernli: Perhaps, though I think the casts are relatively small and are the right way to go between different pointers. I could imagine that making sense in the future though.

src/Qir/Runtime/stdlib/src/lib.rs
swernli: Not currently, no. Since there isn't a QIR API for that, it would be something specific to the Rust implementation, which could be nice but would be an additional API to design/doc/manage.

src/Qir/Runtime/stdlib/src/lib.rs
swernli: After some discussion we landed at leaving this as an `Rc` for now and not speculatively add some parts of multithreaded support.

src/Qir/Runtime/stdlib/src/math.rs
swernli: Added as https://github.com/microsoft/qsharp-compiler/issues/1557

src/Qir/Runtime/stdlib/src/output_recording.rs
```diff
@@ -34,6 +34,8 @@ pub extern "C" fn __quantum__rt__double_record_output(val: c_double) {
     println!(
         "RESULT\t{}",
         if (val.floor() - val.ceil()).abs() < c_double::EPSILON {
```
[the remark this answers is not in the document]
swernli: Deduped!

src/Qir/Runtime/stdlib/src/strings.rs
swernli: Agreed... the legacy that cannot be escaped!

src/Qir/Tests/QIR-dynamic/CMakeLists.txt
swernli: Lots... every consumer needs to specify it :)

src/Qir/Tools/Executable/QirFullStateExecutable.cs
swernli: Adjusted!

src/Simulation/Native/src/simulator/qir.cpp
swernli: Good catch! No, that is no longer accurate. I'll update the comment.

src/Qir/Runtime/stdlib/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+authors = ["Microsoft"]
+edition = "2021"
```
kuzminrobin, line 5: I don't know what this line does but intuitively feel that the line might be meant to contain 20**22** instead of 2021.

src/Qir/Runtime/stdlib/src/bridge-rt.ll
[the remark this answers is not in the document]
kuzminrobin: (FYI) Looks like 
you moved the file _manually_ (after which the git still understood that it was a move, but shows a full diff).

If you moved the file with `git mv <old> <new>` then git wouldn't show full diff, but only the lines changed after the move (and the file history and lines history would be preserved). 

Something to keep in mind for the future (if described above is the case).

src/Simulation/Native/src/simulator/qir.hpp
swernli: That's a great question, and no, currently there is no Rust code that uses this struct. This corresponds to the QIR generated from Q# for the argument tuple. From the Rust QIR stdlib standpoint, it's just an opaque tuple that's essentially a managed memory blob. So this needs to stay consistent with the Q# API and compiler QIR generation rather than Rust. I'll add comments on these to that effect, with pointers to the corresponding Q# signatures.

src/Qir/Runtime/stdlib/src/lib.rs
```diff
@@ -0,0 +1,126 @@
...
+        } else {
+            // Create and drop instances to decrement the count on contained item.
+            if is_alias {
+                let w = Weak::into_raw(Rc::downgrade(&rc));
+
+                // Need to drop two for a net decrement, since above line increments.
+                drop(Weak::from_raw(w));
+                drop(Weak::from_raw(w));
+            } else {
+                drop(Rc::from_raw(raw_rc));
```
samarsha, line 50: Why not use `Rc::decrement_strong_count` and `Rc::increment_strong_count`?

kuzminrobin:
You may also want to update the [.github/CODEOWNERS](https://github.com/microsoft/qsharp-runtime/blob/352aba026cb5f81a5b0f2ad720aee5c969547551/.github/CODEOWNERS) (some paths get deleted).

src/Qir/Tests/Tools/TestCases/FullStateDriverGenerator/UsePauliArg.cpp
```diff
@@ -43,19 +32,9 @@ int main(int argc, char* argv[])
...
-    CLI::Option* simulationOutputFileOpt = app.add_option(
-        "--simulation-output",
-        simulationOutputFile,
-        "File where the output produced during the simulation is written");
-
     // Add a command line option for each entry-point parameter.
-    PauliId PauliArgCli;
-    PauliArgCli = PauliId::PauliId_I;
+    char PauliArgCli;
+    PauliArgCli = 0;
```
kuzminrobin, line 37: The Pauli values turn into the "magic numbers". E.g. it stops being evident that in green L37 and L20 the same value is used.
Consider creating some common C header with Pauli values enum, other common stuff, and including that header wherever needed.

src/Qir/Runtime/stdlib/src/arrays.rs
```diff
@@ -0,0 +1,308 @@
...
+    let data = vec![0_u8; elem_size * count];
+    Rc::into_raw(Rc::new(QirArray { elem_size, data }))
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn __quantum__rt__array_copy(
+    arr: *const QirArray,
+    force: bool,
+) -> *const QirArray {
+    let rc = Rc::from_raw(arr);
```
samarsha, line 30: Why not:

```suggestion
    let rc = ManuallyDrop::new(Rc::from_raw(arr));
```

Then there's no chance of forgetting `Rc::into_raw(rc)` at the end of the scope, which could accidentally free memory. (This applies to many places in the code.)

src/Simulation/Native/src/simulator/qir.cpp
```diff
@@ -0,0 +1,504 @@
+// Copyright (c) Microsoft Corporation.
```
kuzminrobin, line 1: (Nit)
If this file is a result of renaming/moving of some other file, then consider renaming/moving with `git mv ..` to preserve the history and minimize the diff.

src/Qir/Runtime/stdlib/src/arrays.rs
swernli: That's a great point, and in fact makes me realize that most of the `Rc::from_raw` usage in this file (and other) places is likely not necesssary. Functions like this copy one that need to interact with the reference counts need to see it as and `Rc` can follow the pattern you suggest, and the rest can use the pointer directly without taking ownership as an `Rc`. Thanks!

src/Simulation/Native/src/simulator/qir.hpp
```diff
@@ -0,0 +1,117 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+#pragma once
+
+#include "qir_stdlib.h"
+#include <cstdint>
```
kuzminrobin, line 7: In the other C/C++ files you include the standard headers first, then the local headers.
Here - vice versa.
Is there any specific reason for that?

src/Simulation/Native/src/simulator/qir.hpp
```diff
@@ -0,0 +1,117 @@
...
+    double angle;
+    QUBIT* target;
+};
+
+struct QirExpTuple
+{
+    QirArray* paulis;
+    double angle;
+    QirArray* targets;
+};
```
kuzminrobin, line 46: Consider converting to C-style structs (see below) and moving to inside of `extern "C"` block:
```suggestion
typedef struct 
{
    QirArray* paulis;
    double angle;
    QirArray* targets;
} QirExpTuple;
```

kuzminrobin (approved):
I'm done reviewing. Approving with suggestions.
I recommend more approvals.

src/Qir/Tools/QirTools.cs
```diff
@@ -23,7 +23,7 @@ public static class QirTools
         /// <param name="libraryDirectories">Directory where the libraries to link to are located.</param>
         /// <param name="includeDirectories">Directory where the headers needed for compilation are located.</param>
         /// <param name="executablesDirectory">Directory where the created executables are placed.</param>
-        /// <param name="debug">Enable additional debugging checks at runtime.</param>
+        /// <param name="debug">Ignored.</param>
```
samarsha, line 26: Should this be removed?

src/Qir/Runtime/stdlib/src/lib.rs
swernli: Oddly enough, the original prototype for this logic predates the stabilization of those functions by one month... Now that they are stable, I can definitely simplify this logic by relying on them!

src/Simulation/qdk_sim_rs/prerequisites.ps1
swernli: no, but they are needed by the workspace now. So if someone runs even a simple `cargo check` in a folder other than src/Simulation/qdk_sim_rs it will also try to use llvm-tools-preview component for the build.rs in the qir_stdlib. For that reason I thought it made sense to include them here too.

src/Qir/Runtime/stdlib/Cargo.toml
swernli: Surprisingly, the value "2022" fails to parse, and I believe that is because 2021 is the most recent edition. I didn't know this, but the Rust tools manage this value and new projects are automatically created with `edition = "2021"`. You can read more about this in the documentation at https://doc.rust-lang.org/edition-guide/introduction.html.

src/Simulation/Native/src/simulator/qir.hpp
swernli: Nope, the other order makes more sense, so I will adjust here.

src/Simulation/Native/src/simulator/qir.hpp
swernli: Great suggestion, I hadn't thought of that!

src/Qir/Tests/Tools/TestCases/FullStateDriverGenerator/UseBoolArg.cpp
```diff
@@ -40,16 +34,6 @@ int main(int argc, char* argv[])
     CLI::App app("QIR Standalone Entry Point");
 
     // Initialize runtime.
```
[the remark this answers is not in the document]
swernli: The section is currently unused, but there is still logic there if someone else wanted to use Driver Generator to generate a main that includes a notion of an initialize call. I could remove the comment under the assumption that any description would be added by the `RuntimeInitializer` class itself.

src/Qir/Tools/Driver/QirFullStateSimulatorInitializer.cs
```diff
@@ -7,15 +7,11 @@ namespace Microsoft.Quantum.Qir.Runtime.Tools.Driver
 {
     public class QirFullStateSimulatorInitializer : IQirRuntimeInitializer
     {
-        private readonly bool debug;
+        internal QirFullStateSimulatorInitializer() {}
```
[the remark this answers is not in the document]
swernli: I think no... this class just happens to be an example of an implementor that doesn't perform any additional setup. I'll leave it up to @cesarzc to decide if/when we might refactor these tools more thoroughly.

src/Qir/Tools/QirTools.cs
swernli: Yeah, it should. I was hedging but I think I can make sure it is no longer called with this additional param.

src/Qir/Runtime/stdlib/src/callables.rs
```diff
@@ -0,0 +1,171 @@
...
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn __quantum__rt__capture_update_alias_count(
+    callable: *const Callable,
+    update: i32,
+) {
+    let call = &*callable;
+    if !call.mem_table.is_null() && !(*(call.mem_table.wrapping_add(1))).is_null() {
+        let _val = **(call.mem_table.cast::<*mut usize>().wrapping_add(1));
```
samarsha, line 165: What does this line do?

src/Simulation/Native/src/simulator/qir.cpp
samarsha: Git rename detection doesn't use `git mv`. It works the same if you `git rm` and `git add` two files with similar contents. The rename is detected at diff time and has a configurable similarity threshold (default 50%); see [`git diff --find-renames[=<n>]`](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt--Mltngt). So if more than 50% of the file changed, it won't be seen as a rename by default, even if you used `git mv`.

src/Qir/Runtime/stdlib/src/callables.rs
swernli: Oh, great catch! This is leftover from debugging investigation in an earlier prototype and was accidentally copied into here. It should be removed.

samarsha (approved):

swernli:
Merging with passed e2e builds (see https://github.com/microsoft/qdk/pull/288). Thanks everyone for the review!

PR #1091, opened 2022-09-19
Add support for new microsoft-qc provider

xinyi-joffre:
Add support for new microsoft-qc provider.

This new provider, called Microsoft Quantum Computing will contain all future targets for Microsoft's first party quantum offerings. With targets such as:

- microsoft.estimator
- microsoft.sim.*
- microsoft.qpu.*

For now, maintain support for older provider called Microsoft.Simulator (soon to be deprecated) and older targets such as:

- microsoft.simulator.*

### DIFF
--- a/src/Azure/Azure.Quantum.Client/SubmitterFactory.cs
+++ b/src/Azure/Azure.Quantum.Client/SubmitterFactory.cs
@@ -29,8 +29,8 @@ namespace Microsoft.Azure.Quantum
                 "QirSubmitter",
                 ImmutableArray.Create("FullComputation")),
             new SubmitterInfo(
-                new Regex(@"\Amicrosoft\.((?!simulator\.)[\w-_]+\.)*[\w-_]+\z"), // does not match microsoft.simulator.*, but matches any other microsoft.*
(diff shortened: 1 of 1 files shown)


src/Azure/Azure.Quantum.Client/SubmitterFactory.cs
```diff
@@ -28,6 +28,11 @@ public static class SubmitterFactory
                 "Microsoft.Quantum.Providers.Targets.MicrosoftSimulatorSubmitter, Microsoft.Quantum.Providers.Core",
                 "QirSubmitter",
                 ImmutableArray.Create("FullComputation")),
+            new SubmitterInfo(
```
cesarzc, line 31: I would suggest separating the patterns for `microsoft.sim.*` targets and `microsoft.estimator` since their configuration is different.

xinyi-joffre: Done! I changed to be specific to microsoft.estimator(.*)

cesarzc (approved):

Issue #1081, opened 2022-08-27
Inconsistent output for QuantumSimulator vs SparseSimulator


tcNickolas:
## Describe the bug
The same code produces different output when executed on QuantumSimulator vs SparseSimulator.
It appears that the basis states are labeled in little endian for QuantumSimulator, but for SparseSimulator they are sometimes labeled in big endian.

## To Reproduce

The following code:

```qsharp
use qs = Qubit[2] {
    H(qs[0]);
    Controlled Rx([qs[0]], (1.56, qs[1]));
    DumpMachine();
    ResetAll(qs);
}
```

produces this output on QuantumSimualtor (basis state |2⟩ has zero amplitude)

![image](https://user-images.githubusercontent.com/10113024/187044626-1bf3ef43-3d54-45e9-ae37-e1f87a0b2602.png)

but this output on SparseSimulator (basis state |1⟩ has zero amplitude)

![image](https://user-images.githubusercontent.com/10113024/187044617-ca2bccb9-59e8-4baa-9c89-c17886dfd100.png)

Oddly enough, for simpler examples (such as just applying an H gate to the first qubit without following it with Controlled Ry) the output of both simulators is the same.

## Expected behavior
I expect both simulators to have consistent output (up to omitting basis states with zero amplitude).

## System information
- OS: Windows 11
- .NET Core Version: 6.0.400
- QDK version: 0.24.201332, 0.25.222597

DmitryVasilevsky:
Current numbering of qubits in the sparse simulator depends on the size of the state seen. For example, the following code
```
use qs = Qubit[10] {
    H(qs[0]);
    Message("Before:");
    DumpMachine();

    H(qs[8]);
    H(qs[8]);
    Message("After:");
    DumpMachine();

    ResetAll(qs);
}
```
Will result in the following output:
```
Before:
|0⟩      0.707107 +  0.000000 i  ==     **********           [ 0.500000 ]     --- [  0.00000 rad ]
|1⟩      0.707107 +  0.000000 i  ==     **********           [ 0.500000 ]     --- [  0.00000 rad ]
After:
|0⟩      0.707107 +  0.000000 i  ==     **********           [ 0.500000 ]     --- [  0.00000 rad ]
|256⟩    0.707107 +  0.000000 i  ==     **********           [ 0.500000 ]     --- [  0.00000 rad ]
```
I believe we want both sections to be the same. Investigating.

DmitryVasilevsky:
[...]

PR #1092, opened 2022-09-21
Sparse simulator provides little-endian labels to dump callback

DmitryVasilevsky:
Dump callback requires little-endian labels as it uses BasisStateLabelToBigInt, which assumes data is little-endian. However, labels provided by the Sparse Simulator used to be in big-endian format. This resulted in wrong naming of basis vectors in the dump. Now labels are converted to little-endian format.
This resolves issue https://github.com/microsoft/qsharp-runtime/issues/1081
This change also gets rid of defines in the sparse simulator. Use of defines is not a good practice in general. They also prevented reasonable debugging and searching (they used literal string compositions to name functions).


src/Common/cmake/secure_dependencies.cmake
```diff
@@ -37,15 +37,21 @@ macro(locate_win32_spectre_static_runtime)
...
         execute_process(
             COMMAND "${_vswhere_tool}" -latest -products * -requires Microsoft.VisualStudio.Component.VC.Tools.x86.x64 -find **/14.*.*/**/lib/spectre/x64 -sort
             OUTPUT_VARIABLE _vs_install_loc_out
             RESULT_VARIABLE _vs_where_exitcode
             OUTPUT_STRIP_TRAILING_WHITESPACE)
+        message(INFO "*** _vs_install_loc_out: ${_vs_install_loc_out}")
+
         file(TO_CMAKE_PATH "${_vs_install_loc_out}" SPECTRE_LIB_PATH_OUT)
+        message(INFO "*** SPECTRE_LIB_PATH_OUT: ${SPECTRE_LIB_PATH_OUT}")
+
```
kuzminrobin, line 54: I assume that the changes in this file will be reverted (or commented out) before merging.

src/Simulation/NativeSparseSimulator/CMakeLists.txt
```diff
@@ -30,6 +30,10 @@ if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
 	target_compile_options(Microsoft.Quantum.SparseSimulator.Runtime PUBLIC -O3 -ftree-vectorize -mavx2 -mfma)
 endif()
 
+if (MSVC)
+    add_compile_options(/bigobj)
+endif()
```
kuzminrobin, line 35: How does this relate to the issue being fixed?
Code comments wouldn't harm.

kuzminrobin (approved):
Approving with minor comments.

src/Simulation/NativeSparseSimulator/capi.cpp
```diff
@@ -52,45 +52,56 @@ extern "C"
...
 
-// Generic single-qubit gate
-#define FWDGATE1(G)                                                                                                    \
-    MICROSOFT_QUANTUM_DECL void G##_cpp(simulator_id_type sim_id, logical_qubit_id q)                                                   \
-    {                                                                                                                  \
-        getSimulator(sim_id)->G(q);                                                                  \
+    // single-qubit gates
+    MICROSOFT_QUANTUM_DECL void X_cpp(simulator_id_type sim_id, logical_qubit_id q)
+    {
+        getSimulator(sim_id)->X(q);
```
kuzminrobin, line 58: I don't insist, but these changes are hardly related to the problem being fixed.

src/Common/cmake/secure_dependencies.cmake
DmitryVasilevsky: Actually, I'm planning to keep them. They help a lot when figuring out what went wrong. I didn't have some components installed and the build was failing in a weird way - such as a function not receiving enough arguments. Turned out variables were not set earlier. By printing them out I was able to see what's wrong.

src/Simulation/NativeSparseSimulator/capi.cpp
DmitryVasilevsky: Agreed. I just fixed stuff that prevented me from debugging code efficiently. I can move them to a separate PR.

src/Simulation/NativeSparseSimulator/CMakeLists.txt
DmitryVasilevsky: Normally we build this with clang. Building with cmake/MSVC will fail without this switch.

kuzminrobin: Why would one need to build it with MSVC? To debug easier? And just for local builds (not CI)?
If yes then I would recommend to comment this out and provide extra comments explaining why one may need this.

cgranade:
Before releasing this change, please make sure to test that it doesn't modify the dumps shown by IQ# (either in Q# standalone or Q# + Python notebooks). In particular, the case shown by @tcNickolas in #1081 should be checked from a Q# standalone notebook. We may also want to consider either an e2e integration test or iqsharp unit test to make sure that the fix to #1081 does not cause a regression in the use of `%simulate_sparse` / `.simulate_sparse`.

src/Simulation/NativeSparseSimulator/CMakeLists.txt
DmitryVasilevsky: Moving to a separate PR.

DmitryVasilevsky:
Since we already forked the code for September release, I'm merging this in. We'll verify all dependencies before the next release.

PR #1101, opened 2022-10-19
Update clang

AtariDreams:
Update clang to 15.0.3


swernli:
We can't accept this change right now, as we keep to specific versions of the LLVM tool chain based on what our compiler stack uses internally for generating LLVM IR.

PR #1109, opened 2022-12-02
Implementing Library functions in Q#

swernli:
This change implements several Q# functions that were previously `body intrinsic` via existing Q# functionality. This improves compatibility with QIR generation and allows us to remove the corresponding manual C# implementation in favor of compiler generated ones. This will also improve compatibility of programs that use these functions with full QIR targets like resource estimation.


src/Simulation/QSharpFoundation/Convert/Convert.qs
```diff
@@ -94,7 +107,23 @@ namespace Microsoft.Quantum.Convert {
...
+        let len = Length(arr);
+        for i in 0..(len - 2) {
+            set val += arr[i] ? 2L ^ i | 0L;
+        }
+        if arr[len - 1] {
+            set val -= 2L ^ (len - 1);
+        }
+
+        return val;
     }
```
kuzminrobin, line 127: I'm not strong in Q# but I have an impression that if the 0-length array is passed as an arg
* it is not documented what should and will happen,
* line 122 (L122), fragment `arr[len - 1]` will access outside of the array, causing an undefined or unexpected behavior.

src/Simulation/QSharpFoundation/Convert/Convert.qs
```diff
@@ -94,7 +107,23 @@ namespace Microsoft.Quantum.Convert {
...
     /// ```
     function BoolArrayAsBigInt(a : Bool[]) : BigInt {
-        body intrinsic;
+        mutable val = 0L;
+        mutable arr = a;
+
+        if Length(arr) % 8 != 0 {
+            // Padding is needed.
+            set arr += [false, size = Length(arr) % 8];
+        }
```
kuzminrobin, line 116: Currently I have an impression that if in L113 the `Length(arr)` is 6, then L115 will add **6** more (rather than **2** more), as a result the array will have length **12** (instead of **8**). Intuitively I feel that you meant this:
```suggestion
        if Length(arr) % 8 != 0 {
            // Padding is needed.
            set arr += [false, size = 8 - (Length(arr) % 8)];
        }
```

swernli: Aha, good catch! My math was wrong there, glad you spotted it.

src/Simulation/QSharpFoundation/Convert/Convert.qs
swernli: Hmm, you're right. I thought about updating the padding logic to pad a zero length array to length 8, but then figured that's a lot of extra work to get to zero value. So I've added an if-statement around the logic that should satisfy this.

src/Simulation/QSharpFoundation/Convert/Convert.qs
```diff
@@ -94,7 +107,23 @@ namespace Microsoft.Quantum.Convert {
     /// let bi2 = BoolArrayAsBigInt([false, false, false, false, false, false, false, true]); // Not padded -> -128
     /// ```
     function BoolArrayAsBigInt(a : Bool[]) : BigInt {
-        body intrinsic;
+        mutable val = 0L;
```
kuzminrobin, line 110: How many bits does `val` have?
What does the user expect if the passed array is longer than that number of bits? And what will really happen?

swernli: `BigInt` is backed by expanding storage, so should expand to whatever size is needed. This is different than `BoolArrayAsInt` which will fail if the passed array is bigger than 64 entries (see https://github.com/microsoft/QuantumLibraries/blob/main/Standard/src/Convert/Convert.qs#L103-L104).

swernli:
@kuzminrobin Thinking about your comments, I realize there is a disconnect between `BigIntAsBoolArray` and `BoolArrayAsBigInt` where my implementation of the latter is sign respecting (to match the doc comment) while my implementation for the former is not, such that I think I have a bug here....

src/Simulation/QSharpFoundation/Convert/Convert.qs
kuzminrobin: I assume you plan to refactor `BoolArrayAsBigInt()`.

src/Simulation/QSharpFoundation/Convert/Convert.qs
```diff
@@ -63,29 +63,42 @@ namespace Microsoft.Quantum.Convert {
...
 
 
     /// # Summary
     /// Converts a given big integer to an array of Booleans.
     /// The 0 element of the array is the least significant bit of the big integer.
-    /// # Remarks
-    /// See [C# BigInteger constructor](https://docs.microsoft.com/dotnet/api/system.numerics.biginteger.-ctor?view=netframework-4.7.2#System_Numerics_BigInteger__ctor_System_Int64_) for more details.
     function BigIntAsBoolArray(a : BigInt) : Bool[] {
-        body intrinsic;
+        mutable val = a;
```
msoeken, line 85: I don't think it's critical, but this will change behaviour. So far the returned array size was always a multiple of 8 due to byte padding of the internal BigInt representation in C#. Now this will not be any longer the case, and in fact behave more closely to how `IntAsBoolArray` behaves.

src/Simulation/QSharpFoundation/Convert/Convert.qs
```diff
@@ -63,29 +63,42 @@ namespace Microsoft.Quantum.Convert {
     /// Converts a given big integer to an equivalent integer, if possible.
     /// The function returns a pair of the resulting integer and a Boolean flag
     /// which is true, if and only if the conversion was possible.
-    /// # Remarks
-    /// See [C# BigInteger constructor](https://docs.microsoft.com/dotnet/api/system.numerics.biginteger.-ctor?view=netframework-4.7.2#System_Numerics_BigInteger__ctor_System_Int64_) for more details.
     function MaybeBigIntAsInt(a : BigInt) : (Int, Bool) {
-        body intrinsic;
+        let arr = BigIntAsBoolArray(a);
```
msoeken, line 67: Would it be worth to do some range check before computing a possibly very large array just to determine not to perform the conversion?

msoeken (approved):
Looks good to me, but I left some comments.

src/Simulation/QSharpFoundation/Convert/Convert.qs
swernli: You are totally right, and it broke some of the conversion logic when it came to negative numbers. I've significantly updated the logic to now match the existing behavior and correctly handle padding and two's complement for negative numbers. I included some comments too, to clarify.

src/Simulation/QSharpFoundation/Convert/Convert.qs
swernli: Good idea; we know the limits values, so it should be cheaper to calculate and compare the BigInts instead of construct the array.

src/Simulation/QSharpFoundation/Convert/Convert.qs
```diff
@@ -63,29 +63,76 @@ namespace Microsoft.Quantum.Convert {
...
     /// The 0 element of the array is the least significant bit of the big integer.
-    /// # Remarks
-    /// See [C# BigInteger constructor](https://docs.microsoft.com/dotnet/api/system.numerics.biginteger.-ctor?view=netframework-4.7.2#System_Numerics_BigInteger__ctor_System_Int64_) for more details.
     function BigIntAsBoolArray(a : BigInt) : Bool[] {
-        body intrinsic;
+        // To use two's complement, little endian representation of the integer, we fisrt need to track if the input
+        // is a negative number. If so, flip it back to positive and start tracking a carry bit.
+        let isNegative = a < 0L;
+        mutable carry = isNegative;
+        mutable val = isNegative ? -a | a;
```
kuzminrobin, line 95: You may want to make sure (with the tests or detailed code comments) that this line, fragment `-a`, works the way you expect for the most-negative 64-bit number.

E.g. (just for explanation purposes) if we are talking about 8-bit signed integers, then the most-negative number is `-128`. If we negate it, then we get `+128` which is a 9-bit signed value and does not fit in 8-bit signed integer (we get overflow, and I'm not sure what actual value we get after negation, 0 or -128 or any other).

src/Simulation/QSharpFoundation/Convert/Convert.qs
```diff
@@ -63,29 +63,76 @@ namespace Microsoft.Quantum.Convert {
     /// Converts a given big integer to an equivalent integer, if possible.
     /// The function returns a pair of the resulting integer and a Boolean flag
     /// which is true, if and only if the conversion was possible.
-    /// # Remarks
-    /// See [C# BigInteger constructor](https://docs.microsoft.com/dotnet/api/system.numerics.biginteger.-ctor?view=netframework-4.7.2#System_Numerics_BigInteger__ctor_System_Int64_) for more details.
     function MaybeBigIntAsInt(a : BigInt) : (Int, Bool) {
-        body intrinsic;
+        if a > (1L <<< 63) - 1L or a < (-1L * (1L <<< 63)) {
+            return (0, false);
+        }
```
kuzminrobin, line 69: I'm not sure the fragment `(1L <<< 63)` gives the value you expect. I expect that it causes the overflow of the 64-bit signed type. Would be nice to have tests for the cases `a > ..` and `a < ..`.

src/Simulation/QSharpFoundation/Convert/Convert.qs
```diff
@@ -63,29 +63,76 @@ namespace Microsoft.Quantum.Convert {
...
+                    set arr += [not newBit];
+                }
+            }
+            else {
+                // For positive numbers just accumulate the calculated bits into the array.
+                set arr += [newBit];
+            }
+
+            set val /= 2L;
+        }
```
kuzminrobin, line 119: 1. I have an impression that `BigIntAsBoolArray()` does `BitInt` arithmetic. I'm not sure how `BigInt` arithmetic works/implemented in Q# (C# runtime and QIR runtime).
L95: `-a`,
L95: I assume that `val` is of type `BitInt`,
L98: compares `BitInt` to `0`,
L118: does `BigInt` division.


2. If we use `BitInt` arithmetic here in `BigIntAsBoolArray()`, then why don't we use it in `MaybeBigIntAsInt()`? Something like this:
```
    function MaybeBigIntAsInt(a : BigInt) : (Int, Bool) {
        if a > (1L <<< 63) - 1L or a < (-1L * (1L <<< 63)) {  // Same as before.
            return (0, false);
        }

        mutable val = 0;  // Difference starts.
        mutable len = 32;
        while (len) {
            set val = val << 1; // Shift left.
            set val = val + ((a &&& (1L << 31)) >> 31); // The bit 31 of `a` is added to bit 0 of `val`.
            a = a << 1;  // Shift left.
            --len;  // Decrement `len`.
        }
        return (val, true);
```

src/Simulation/QSharpFoundation/Convert/Convert.qs
swernli: Because `a` is of type `BigInt` it does exactly what I'd expect, namely it negates the number and expands the underlying storage as appropriate. `BigInt` is very flexibly that way, in that it will do the right thing in terms of expanding the width of the underlying bits as needed while preserving/updating the sign bit, but that makes it tricky to reason about it as a bit string.

The example you chose is a good one to demonstrate this. -128 is an 8 bit value, so it `BigIntAsBoolArray` returns a length 8 array: `[False,False,False,False,False,False,False,True]`. Here the most significant bit is correctly interpreted as the sign bit. As you point out, positive 128 would have the 8th bit true and need one more false bit to indicate the sign. Rather than overflow or use 9 bits, the `BigInt` expands to 16 bits, and we see that behavior when using `BigIntAsBoolArray`: 128 becomes `[False,False,False,False,False,False,False,True,False,False,False,False,False,False,False,False]`. `BigInt` negation already accounts for this.

One way to confirm the behavior here is correct for different values is to create a test project that has the conversion implementations written in this change using the current release of QDK and then run both this and existing `body intrinsic` ones side by side. I did this using the following code in an entry point:
```qsharp
        let a = -128L;
        Message($"{Microsoft.Quantum.Convert.BigIntAsBoolArray(a)}");
        Message($"{Microsoft.Quantum.Convert.MaybeBigIntAsInt(a)}");
        Message("----");
        Message($"{BigIntAsBoolArray(a)}");
        Message($"{MaybeBigIntAsInt(a)}");
```
(Where the unqualified `BigIntAsBoolArray` and `MaybeBigIntAsInt` are local copies of the code in this PR). With that I'm able to see that for several corner case values, like `0`, 64 bit signed int max and min, etc, the behavior of the new implementation is the same as the old implementation.

src/Simulation/QSharpFoundation/Convert/Convert.qs
```diff
@@ -63,29 +63,76 @@ namespace Microsoft.Quantum.Convert {
...
+        mutable val = isNegative ? -a | a;
+
+        mutable arr = [];
+        while val != 0L {
+            let newBit = val % 2L == 1L;
+            if isNegative {
+                // For negative numbers we must invert the calculated bit, so treat "true" as "0"
+                // and "false" as "1". This means when the carry bit is set, we want to record the
+                // calculated new bit and set the carry to the opposite, otherwise record the opposite
+                // of the calculate bit.
```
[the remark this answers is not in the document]
swernli: I'll admit, the explanation here of two's complement is paraphrased from the Wikepedia page: https://en.wikipedia.org/wiki/Two%27s_complement. I assumed familiarity with two's complement when writing this, and rather than explain more broadly how that works would expect someone to look it up and get a baseline understanding before reading the further explanations here.

src/Simulation/QSharpFoundation/Convert/Convert.qs
swernli: The problem I'm trying to avoid is the issue of arithmetic shift right vs logical shift right. Your logic might work even accounting for that, I'd need to double check.

src/Simulation/QSharpFoundation/Convert/Convert.qs
```diff
@@ -63,29 +63,76 @@ namespace Microsoft.Quantum.Convert {
...
+        }
+
+        // Pad to the next higher byte length (size 8) if the length is not a non-zero multiple of 8 or
+        // if the last bit does not agree with the sign bit.
+        let len = Length(arr);
+        if len == 0 or len % 8 != 0 or arr[len - 1] != isNegative {
+            set arr += [isNegative, size = 8 - (len % 8)];
+        }
+        return arr;
     }
```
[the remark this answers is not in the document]
kuzminrobin: Ah... My bad, sorry :-)

src/Simulation/QSharpFoundation/Convert/Convert.qs
kuzminrobin: If we have those tests then OK (and I assume that they succeed for QIR RT also, not only for C# RT). 

I was afraid that the fragment `(1L <<< 63)` could be calculated as a signed 64-bit integer (rather than `BigInt`), resulting in a value 0x8000_0000_0000_0000 (the lowest negative 64-bit signed value). Then the fragment `-1L * ..` could also be calculated as signed 64-bit integer (rather than `BigInt`), resulting in negating the lowest negative value, which could overflow the signed 64-bit integer. And then the overflown signed 64-bit value could be compared to BigInt.

swernli: Yeah, the interesting thing to recognize here is that bitshift of a `BigInt` is calculated through on operation that takes a `BigInt` target to be shifted and an `Int` number of bits to shift, and the calculation takes place in a `BigInt` with all the fancy automatic bit width adjustments. As a result, the numbers work as expected:
`(1L <<< 63) - 1L` becomes `9223372036854775807`
`(-1L * (1L <<< 63))` becomes `-9223372036854775808`
which are 64-bit signed int max and min, respectively.

swernli: No worries, I probably should've mentioned that earlier! Always good to double check ;)

PR #1111, opened 2022-12-08
Use qir-stdlib from the qir-runner repo

swernli:
This change removes the code for the qir-stdlib and instead compiles a wrapper library for it, allowing the source code from https://github.com/qir-alliance/qir-runner to the be authoritative implementation.


src/Qir/Runtime/stdlib/Cargo.toml
```diff
@@ -6,11 +6,7 @@ edition = "2021"
 license = "MIT"
 
 [dependencies]
-num-bigint = { version = "0.4.3", default-features = false }
-rand = "0.8.5"
-
-[build-dependencies]
-llvm-tools = "0.1.1"
+qir-stdlib = { git = "https://github.com/qir-alliance/qir-runner/", rev = "54d334c0c845ebbfd4dc9512ef05cd2cbe4238c5" }
```
kuzminrobin, line 9: Curious, how often the fragment `rev = "54d334c0c845ebbfd4dc9512ef05cd2cbe4238c5"` will need to be updated and in which cases...

swernli: That's the commit hash that gets pulled down for that dependency. Cargo offers two ways to express a git dependency: `branch` and `rev`. While `branch` is convenient for local development, it can cause problems for group projects or CI because the meaning may change; if I specify `branch = "main"` then the first time I compile it will grab the contents of the main branch BUT subsequent compilation will not automatically update it. Then a few days later, if CI runs or someone else compiles for the first time, it will pull main at that point in time, and now they are using a different, newer version of the dependency.  Using `rev` and commit hash guarantees that developers and CI will be using the same version. When we decide to get a newer version of that dependency, much like updating a version number, we can update the revision hash to point to the specific commit we want to sync to, and because that is an update to Cargo.toml the system will recompile that dependency and sync to the newest hash in the process.

For our purposes, we would only update this revision when we have a newer version of the qir-stdlib that we want to use.

PR #1117, opened 2023-01-10
Expose joint rotations publicly, update test infrastructure

### LINKED ISSUE #1000: Add two-qubit rotation gates to `Microsoft.Quantum.Intrinsic` as fully supported APIs
Currently the two qubit rotation gates `IsingXX`, `IsingYY` and `IsingZZ` are internal implementation details of the target package decompositions. However, they map to native hardware intrinsics on some platforms and can be reasonably decomposed in terms entanglers and single qubit rotations across other platforms. We should consider promoting them to full APIs and document them as part of the `Microsoft.Quantum.Intrinsic` namespace so that developers and library operations can make use them.

swernli:
This change adds the two qubit joint rotations to the public API surface for Q# as `Rxx`, `Ryy`, and `Rzz`. They are decomposed on the simulator as before in terms of `Exp`, and decomposed on hardware targets via single qubit rotations. The exception is the Type1 target package, which is updated to decompose all gate operations in terms of `Rx`, `Rz`, and `Rzz` as hardware-native intrinsics. These new Type1 decompositions are tested in a scoped test project.

In addition, the scale of existing tests is reduced to improve test performance. Instead of verifying every Q# operation up to 8 controls (which has a high simulation cost), it picks a number of controls sufficient to exercise all code paths in the decomposition. This showed local perf gains that pushed tests from taking 5 minutes end-to-end to just over 1 second end-to-end.

Resolves #1000.


src/Simulation/TargetDefinitions/Decompositions/ApplyControlledX.qs
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
kuzminrobin, line 1:
```suggestion
// Copyright (c) Microsoft Corporation.
```

src/Simulation/TargetDefinitions/Decompositions/ApplyControlledX.qs
```diff
@@ -0,0 +1,45 @@
...
+    ///     \operatorname{CNOT} \mathrel{:=}
+    ///     \begin{bmatrix}
+    ///         1 & 0 & 0 & 0 \\\\
+    ///         0 & 1 & 0 & 0 \\\\
+    ///         0 & 0 & 0 & 1 \\\\
+    ///         0 & 0 & 1 & 0
+    ///     \end{bmatrix},
+    /// \end{align}
+    ///
+    /// where rows and columns are ordered as in the quantum concepts guide.
```
kuzminrobin, line 21: It probably makes sense to make the fragment `quantum concepts` a web-link, if applicable and simple.

src/Simulation/TargetDefinitions/Decompositions/ApplyControlledZFromRzz.qs
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
kuzminrobin, line 1:
```suggestion
// Copyright (c) Microsoft Corporation.
```

src/Simulation/TargetDefinitions/Decompositions/ApplyUncontrolledRy.qs
```diff
@@ -0,0 +1,42 @@
...
+    internal operation ApplyUncontrolledRy (theta : Double, qubit : Qubit) : Unit {
+        ApplyUncontrolledH(qubit);
+        ApplyUncontrolledS(qubit);
+        ApplyUncontrolledH(qubit);
+
+        ApplyUncontrolledRz(theta, qubit);
+
+        ApplyUncontrolledH(qubit);
+        ApplyUncontrolledSAdj(qubit);
+        ApplyUncontrolledH(qubit);
```
kuzminrobin, line 40: Just curiosity:
Is there a way in Q# to get these subroutines called automatically in the same order?
(something like a mechanism of destructor calls in C++)

src/Simulation/TargetDefinitions/Decompositions/ApplyUncontrolledSAdj.qs
```diff
@@ -0,0 +1,26 @@
...
+    ///     \end{bmatrix}.
+    /// \end{align}
+    ///
+    /// # Input
+    /// ## qubit
+    /// Qubit to which the gate should be applied.
+    internal operation ApplyUncontrolledSAdj (qubit : Qubit) : Unit {
+         ApplyUncontrolledRz(-PI() / 2.0, qubit);
+    }
+}
```
kuzminrobin, line 26: (This question can be very naïve, and non-applicable)
I see that
`ApplyUncontrolledS()` and 
`ApplyUncontrolledSAdj()` are implemented in different files, rather than the same file. Such an implementation makes an impression that there can be use cases when 
`ApplyUncontrolledS()` is used (and corresponding file is "included", speaking in C++ terminology), but the 
`ApplyUncontrolledSAdj()` is not (and corresponding file is _not_ included). And viced versa.

Is that really the case? (there can be such use cases) If yes, then OK. But if not then it can make sense to merge the two files together.

src/Simulation/TargetDefinitions/Decompositions/ApplyUncontrolledTAdj.qs
```diff
@@ -0,0 +1,26 @@
...
+    ///     \end{bmatrix}.
+    /// \end{align}
+    ///
+    /// # Input
+    /// ## qubit
+    /// Qubit to which the gate should be applied.
+    internal operation ApplyUncontrolledTAdj (qubit : Qubit) : Unit {
+        ApplyUncontrolledRz(-PI() / 4.0, qubit);
+    }
+}
```
kuzminrobin, line 26: (Can be a false alarm)
When I look at the implementation of the subroutines `ApplyUncontrolledT[Adj]()` I see the correspondence between the code and the documentation, e.g. the code fragment `PI() / 4.0` corresponds to the doc-comment fragment `\pi / 4`.

But when I look at the implementation of the subroutines `ApplyUncontrolledS[Adj]()` (earlier), I don't see such a correspondence. In particular, I see the code fragment `PI() / 2.0`, but I don't see the corresponding doc-comment fragment `\pi / 2`.

src/Simulation/TargetDefinitions/Decompositions/ApplyUncontrolledX.qs
```diff
@@ -0,0 +1,28 @@
...
+    /// # Input
+    /// ## qubit
+    /// Qubit to which the gate should be applied.
+    internal operation ApplyUncontrolledX (qubit : Qubit) : Unit is Adj {
+        body (...) {
+            ApplyUncontrolledRx(PI(), qubit);
+        }
+        adjoint self;
+    }
+}
```
kuzminrobin, line 28: (nit)
A thought crawls into my mind: May be in the future, instead of using `Uncontrolled` and `Controlled` it makes sense to use \<Nothing> and `Controlled`? E.g. instead of using 
`ApplyUncontrolledX` and `ApplyControlledX` to use 
`ApplyX` and `ApplyControlledX`.

src/Simulation/TargetDefinitions/Decompositions/RxxFromExp.qs
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.Quantum.Intrinsic {
+    open Microsoft.Quantum.Targeting;
+
+    /// # Summary
+    /// Applies the two qubit Ising $XX$ rotation gate.
```
kuzminrobin, line 8: I see that in the code in multiple places (in other files) you replace the fragment `Ising` with fragment `R`, but in this line the fragment `Ising` survives. I just wanted to make sure that this is by design. Applicable to about 8 more files.

src/Simulation/TargetDefinitions/Decompositions/ApplyUncontrolledSAdj.qs
kuzminrobin: Also when I look at "src/Simulation/TargetDefinitions/Decompositions/RxxFromSinglyControlled.qs", fragment 
```qsharp
    operation Rxx (theta : Double, qubit0 : Qubit, qubit1 : Qubit) : Unit is Adj + Ctl {
        . . .
        adjoint (...) {
            Rxx(-theta, qubit0, qubit1);
        }
```
I get an impression that the adjoint version calls the non-adjoint with the opposite sign for the theta. 

The same approach can be used for `ApplyUncontrolledS[Adj]` (if merging them is applicable).

src/Simulation/TargetDefinitions/Decompositions/RzzFromExp.qs
```diff
@@ -0,0 +1,31 @@
...
+    open Microsoft.Quantum.Targeting;
+
+    /// # Summary
+    /// Applies the two qubit Ising $ZZ$ rotation gate.
+    ///
+    /// # Description
+    /// \begin{align}
+    ///     R_zz(\theta) \mathrel{:=}
+    ///     \begin{bmatrix}
+    ///         e^{-i \theta / 2} & 0 & 0 & 0 \\\\
```
kuzminrobin, line 14: (May be false alarm)
The matrix on [this page](https://www.numerade.com/ask/question/in-some-quantum-computing-platforms-the-easiest-two-qubit-gate-to-implement-is-something-called-the-ising-zz-coupling-gate-in-the-standard-basis-this-kind-of-gate-is-represented-by-matrix-of-33007/) makes me suspect that the following can be applicable 
```suggestion
    ///         e^{i \theta / 2} & 0 & 0 & 0 \\\\
```

kuzminrobin: Applicable to more than one file.

src/Simulation/TargetDefinitions/Interfaces/IIntrinsicIsingXX.cs
```diff
@@ -5,7 +5,7 @@
 
 namespace Microsoft.Quantum.Intrinsic.Interfaces
 {
-    public interface IIntrinsicIsingXX : IOperationFactory
+    public interface IIntrinsicApplyUncontrolledRxx : IOperationFactory
```
kuzminrobin, line 8: FYI: The identifier is renamed but the file is not.

kuzminrobin: When renaming, prefer `git mv` (to preserve the history) rather than other commands or renaming in IDE.

src/Simulation/TargetDefinitions/Tests/Type1/Tests.Type1.csproj
```diff
@@ -0,0 +1,53 @@
+<Project Sdk="Microsoft.Quantum.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
+    <IsPackable>false</IsPackable>
+    <QscVerbosity>detailed</QscVerbosity>
+    <QSharpDocsGeneration>false</QSharpDocsGeneration>
+    <!-- we will provide our own -->
```
kuzminrobin, line 8: Consider restructuring such that it is clear which line this comment is applicable to (the line above or to the line below or other).

kuzminrobin (approved):
Approving (with some doc-comment notes) to unblock. But I would recommend at least one more approval from those who understand quantum computing, e.g. @cesarzc.

src/Simulation/TargetDefinitions/Decompositions/ApplyControlledX.qs
swernli: Unfortunately I don't have link for this, and many of our intrinsic operations list the same text without a link.

src/Simulation/TargetDefinitions/Decompositions/ApplyUncontrolledRy.qs
swernli: Usually this pattern would make use of within-apply syntax (see [Conjugations in Q#](https://learn.microsoft.com/en-us/azure/quantum/user-guide/language/statements/conjugations)), but that requires that the called operations support the  Adjoint functor, which these purposefully do not.

src/Simulation/TargetDefinitions/Decompositions/ApplyUncontrolledX.qs
swernli: Yeah, that's a good point, and I've been debating doing a broader rename of the files too. My initial thinking was to use "Uncontrolled" to make explicit the fact that the operation did not support the Controlled functor, but now it just seems unnecessary and wordy. I think a broader rename could be a follow up PR.

src/Simulation/TargetDefinitions/Decompositions/RxxFromExp.qs
swernli: That is intentional, as the name is still descriptive for the gates in text. For example, they are described on Wikipedia as:

> The qubit-qubit Ising coupling or Heisenberg interaction gates Rxx, Ryy and Rzz are 2-qubit gates that are implemented natively in some [trapped-ion quantum computers](https://en.wikipedia.org/wiki/Trapped_ion_quantum_computer), using for example the [Mølmer–Sørensen gate procedure](https://en.wikipedia.org/wiki/M%C3%B8lmer%E2%80%93S%C3%B8rensen_gate).

from [List of quantum logic gates](https://en.wikipedia.org/wiki/List_of_quantum_logic_gates#:~:text=The%20qubit-qubit%20Ising%20coupling%20or%20Heisenberg%20interaction%20gates,be%20expressed%20in%20sinusoidal%20form%20also%2C%20for%20example.).

src/Simulation/TargetDefinitions/Decompositions/RzzFromExp.qs
swernli: The matrices have been validated with the hardware teams of our partners, exactly because these kinds of angle conventions vary and can cause problems when different in the language vs the hardware. For this, we are following the conventions and matrices that do have the negative sign in the exponent on the first row, as shown here: https://en.wikipedia.org/wiki/List_of_quantum_logic_gates#Two_qubit_interaction_gates.

src/Simulation/TargetDefinitions/Interfaces/IIntrinsicIsingXX.cs
swernli: Good catch!

src/Simulation/TargetDefinitions/Decompositions/ApplyUncontrolledTAdj.qs
swernli: This is expected, as the S gate is equivalent to TT (running T twice) and so the entry in the matrix is just imaginary `i` as seen in the description of the S gate (or phase gate or sqrt(Z) gate) in https://en.wikipedia.org/wiki/List_of_quantum_logic_gates#Clifford_qubit_gates.

src/Simulation/TargetDefinitions/Decompositions/RzzFromExp.qs
kuzminrobin: (Just FYI)
I have an impression that the implementation is correct but the doc comment line that I refer to here is still not quite right.

When I look at the [link that I provided](https://www.numerade.com/ask/question/in-some-quantum-computing-platforms-the-easiest-two-qubit-gate-to-implement-is-something-called-the-ising-zz-coupling-gate-in-the-standard-basis-this-kind-of-gate-is-represented-by-matrix-of-33007/) and the [Wikipedia link that you provided](https://en.wikipedia.org/wiki/List_of_quantum_logic_gates#Two_qubit_interaction_gates) (I assume you mean the line containing "ZZ interaction") I see that in both cases the matrix consists of the main diagonal (all other elements are 0), and
* the first-most and the last-most element in the matrix have the same sign
* and the 2 middle elements have the opposite sign.

But in your doc-comment 3 elements of the matrix have one sign, and one element has opposite sign.

In other words, in both cases of the documentation 2 elements have one sign and 2 elements have the other sign.
But in your doc-comment 3 elements have one sign and 1 element has the other sign.

If your doc-comment is still correct, and the implementation corresponds to that doc-comment, then I suspect there was a bug when "The matrices have been validated with the hardware teams of our partners".

(this logic can still be a false alarm, just FYI)

swernli: Ah, you're right! The upper left corner is correct, but the rest are mismatched for each of the copy pasted `Rzz` doc comments. I'll get that fixed!

kuzminrobin: Just in case I'll remind: Applicable to more than one file.

kuzminrobin: Oops! Looks like you have already taken care of multiple files.

swernli:
Converting to draft until we are ready to proceed with this change.

PR #1125, opened 2023-02-08
Set 1-second timeout for ManagedIdentityCredential

minestarks:
Fixes https://github.com/microsoft/Quantum/issues/762

If a credential type isn't explicitly provided, `DefaultQuantumCredential` tries all the possible credential types in order. `ManagedIdentity` is only applicable for non-Azure environments (in which case it will succeed fairly quickly), but in non-Azure environments it can hang for a long time before failing. To work around this, we set a timeout of 1 second [as recommended by the Azure SDK team](https://github.com/Azure/azure-sdk-for-net/issues/24767#issuecomment-947088247).

### DIFF
--- a/src/Azure/Azure.Quantum.Client/Authentication/DefaultQuantumCredential.cs
+++ b/src/Azure/Azure.Quantum.Client/Authentication/DefaultQuantumCredential.cs
@@ -55,7 +55,11 @@ namespace Microsoft.Azure.Quantum.Authentication
             {
                 yield return CredentialFactory.CreateCredential(CredentialType.TokenFile);
                 yield return CredentialFactory.CreateCredential(CredentialType.Environment, SubscriptionId);
-                yield return CredentialFactory.CreateCredential(CredentialType.ManagedIdentity, SubscriptionId);
+                // Create a ManagedIdentityCredential with a 1-second timeout to work around delays in non-Azure environments.
+                // In non-Azure environments, ManagedIdentityCredential can cause long delays before failing to resolve.
+                // In Azure environments, in practice, 1 second seems to be enough time for the managed identity to be resolved.
+                // (See: https://github.com/Azure/azure-sdk-for-net/issues/24767)
+                yield return new ManagedIdentityCredential(options: new TokenCredentialOptions { Retry = { NetworkTimeout = TimeSpan.FromSeconds(1), MaxRetries = 0 } });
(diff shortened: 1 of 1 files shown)


minestarks:
FYI @xinyi-joffre

kuzminrobin (approved):
Approving to unblock, but I'm not familiar with this part of the code base, I would recommend at least one more approval.

swernli (approved):

cesarzc (approved):

xinyi-joffre (approved):

src/Azure/Azure.Quantum.Client/Authentication/DefaultQuantumCredential.cs
```diff
@@ -55,7 +55,11 @@ public virtual IEnumerable<TokenCredential> Sources
             {
                 yield return CredentialFactory.CreateCredential(CredentialType.TokenFile);
                 yield return CredentialFactory.CreateCredential(CredentialType.Environment, SubscriptionId);
-                yield return CredentialFactory.CreateCredential(CredentialType.ManagedIdentity, SubscriptionId);
+                // Create a ManagedIdentityCredential with a 1-second timeout to work around delays in non-Azure environments.
+                // In non-Azure environments, ManagedIdentityCredential can cause long delays before failing to resolve.
+                // In Azure environments, in practice, 1 second seems to be enough time for the managed identity to be resolved.
+                // (See: https://github.com/Azure/azure-sdk-for-net/issues/24767)
+                yield return new ManagedIdentityCredential(options: new TokenCredentialOptions { Retry = { NetworkTimeout = TimeSpan.FromSeconds(1), MaxRetries = 0 } });
```
billti, line 62: 1 second does seem pretty aggressive. Does it expect to receive network replies or resolve loading any artifacts (other assemblies, config settings, etc.) within that time to be successful? If so, maybe something like 5 or 10 seconds might be a little more conservative, but still well within the 2 min timeout we're seeing.

(And I'm speculating, so if that's totally wrong and things will fail instantly if they're going to fail, then this is fine).

billti (approved):
Left a comment to double check. If not an issue, then looks good. Thanks!

src/Azure/Azure.Quantum.Client/Authentication/DefaultQuantumCredential.cs
billti: I should have read the opening comment on the PR before my review, not after :-)  Looks good to go then. Thanks!

PR #1130, opened 2023-02-28
Clean up a few nullable warnings and an unnecessary interpolated string

minestarks:
StackTrace.cs had several nullable warnings, and I was getting quite confused following which fields are safe to access, when. I went through and added `?` annotations to make warnings go away, and `Debug.Assert`s wherever I could confirm that a field would _definitely_ be set. This file should be warning clean now.

Additionally, simplified a `string.Format` that used an interpolated string as an argument.

Came up while working on https://github.com/microsoft/iqsharp/pull/776


src/Simulation/Simulators/CommonNativeSimulator/SimulatorBase.cs
```diff
@@ -174,6 +175,9 @@ protected void WriteStackTraceToLog(Exception exception, IEnumerable<StackFrame>
             catch (Exception e) // Dumps q# call-stack in case of exception if CallStack tracking was enabled
             {
                 this.CallStack = stackTraceCollector.CallStack;
+                // StackTraceCollector should have set CallStack to a non-null value in its OnFail handler,
+                // since an exception was just thrown by the simulator.
+                Debug.Assert(this.CallStack != null, "Expected CallStack to be set");
```
kuzminrobin, line 180: FYI: One of our team members who left our team recently, did not recommend to use `Debug.Assert()` since upon assertion failure the `Debug.Assert()` was causing something like a crash not noticeable by the test infrastructure (the test execution was not noticing the test crash). I personally got one or two confirmations to that, if I remember correctly.
I'm not yet familiar with that functionality well enough (and don't remember my experience with that well enough) to have a strong opinion. Leaving the decision to you.

kuzminrobin (approved):
Approving to unblock but I would recommend an approval from someone who knows C# better, @cesarzc, @swernli.

src/Simulation/Simulators/CommonNativeSimulator/SimulatorBase.cs
minestarks: Thanks for pointing that out. I noticed `Debug.Assert` behaving very strangely too (seemed like IQ# would silently die and the caller was left hanging) , but I wasn't sure if it was just me. Leaning towards leaving this as I can't think of a better alternative, but I'll think it through.

swernli (approved):
LGTM

PR #1138, opened 2023-03-25
Revert "Clean up a few nullable warnings and an unnecessary interpolated string"

minestarks:
Reverts microsoft/qsharp-runtime#1130

Possibly causing test failures at https://ms-quantum.visualstudio.com/Quantum%20Program/_build/results?buildId=262812&view=results


minestarks:
Tests passed on rerun so I think the failure was a fluke. Abandoning the revert

